PR #11, opened 2018-07-27
Add Travis CI

mitruly:
Configure Travis for continuous integration.

Per @pauljamescleary, this is going to be iterative but we need to have a basic `.yml` file to trigger the builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,20 @@
+language: scala
+
+scala:
+  - 2.12.6
+
+jdk:
+  - oraclejdk8
+
+cache:
+  directories:
+    - $HOME/.ivy2/cache
+    - $HOME/.sbt
+
+before_cache:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -0,0 +1,20 @@
...
+    - $HOME/.ivy2/cache
+    - $HOME/.sbt
+
+before_cache:
+  # Cleanup the cached directories to avoid unnecessary cache updates (https://www.scala-sbt.org/1.0/docs/Travis-CI-with-sbt.html)
+  - find $HOME/.ivy2/cache -name "ivydata-*.properties" -print -delete
+  - find $HOME/.sbt        -name "*.lock"               -print -delete
+
+script:
+  - sbt ++$TRAVIS_SCALA_VERSION ";validate;verify"
```
rebstar6: are we doing func tests separately?

mitruly: I neglected to put the info in the description, but this will be iterative building to configure Travis. We will have `codecov` and func tests run in this build, but probably as separate PRs.

PR #21, opened 2018-07-30
Portal docker image 

nimaeskandary:
### Set up portal docker image
* Added settings to `build.sbt for `sbt-native-packager` to create a docker image for the portal module
* Added application configuration overrides in `docker/portal` for booting up the portal image for local development
* Updated `docker/docker-compose-build.yml` to include the portal image
* Added `bin/docker-up-portal-server.sh` to start up vinyl portal, api, and their dependencies
* Fixed bug in `modules/api/src/universal/conf/application.conf` that didn't use mysql docker service name
* Added information to developer documentation for how to configure and start the portal via docker


build.sbt
```diff
@@ -130,6 +130,25 @@ lazy val apiDockerSettings = Seq(
...
+  dockerExposedVolumes := Seq("/opt/docker/lib_extra"), // mount extra libs to the classpath
+  dockerExposedVolumes := Seq("/opt/docker/conf"), // mount extra config to the classpath
+
+  // add extra libs to class path via mount
+  scriptClasspath in bashScriptDefines ~= (cp => cp :+ "/opt/docker/lib_extra/*"),
+
+  // adds config file to mount
+  bashScriptExtraDefines += """addJava "-Dconfig.file=/opt/docker/conf/application.conf"""",
+  bashScriptExtraDefines += """addJava "-Dlogback.configurationFile=/opt/docker/conf/logback.xml"""",
+  bashScriptExtraDefines += """addJava "-Djavax.net.ssl.trustStore=/opt/docker/conf/trustStore.jks"""",
```
pauljamescleary: the trust store is an interesting one.  We currently need it because we use a trust store.  Not everyone will need it though.  Thinking we should leave it out for the time being.  Ideally, we would allow the user of the docker image to pass in their own command line arguments.  Is there a way we can do that?  Would like to allow passing additional JVM arguments as an option and then document it.

docker/docker-compose-func-test.yml
```diff
@@ -48,6 +48,16 @@ services:
       - elasticmq
       - dynamodb
 
+  portal:
+    image: "hub.comcast.net/denis/portal:0.1.1-SNAPSHOT"
```
pauljamescleary: do not include hub.comcast.net.  We will be releasing open.  No namespace is required for open.

docker/portal/docker.conf
```diff
@@ -0,0 +1,114 @@
...
+# This must be changed for production, but we recommend not changing it in this file.
+#
+# See http://www.playframework.com/documentation/latest/ApplicationSecret for more details.
+play.http.secret.key = "vinyldnsportal-change-this-for-production"
+
+# The application languages
+# ~~~~~
+play.i18n.langs = [ "en" ]
+
+# Router
```
pauljamescleary: let's delete all the commented out stuff from config.  We should only provide comments for the configuration settings that we actually use.

docker/docker-compose-func-test.yml
```diff
@@ -48,6 +48,16 @@ services:
       - elasticmq
       - dynamodb
 
+  portal:
+    image: "hub.docker.com/vinyldns/portal:0.1.1-SNAPSHOT"
```
pauljamescleary: you should not need `hub.docker.com` at all, just `vinyldns/` as docker hub is the default

docker/portal/docker.conf
```diff
@@ -0,0 +1,85 @@
...
+    displayOnSidebar = true
+    displayOnLoginScreen = true
+    title = "API Documentation"
+    href = "http://vinyldns.io"
+    icon = "fa fa-file-text-o"
+  }
+]
+
+// Local.conf has files specific to your environment, for example your own LDAP settings
+include "/opt/docker/conf/local.conf"
```
pauljamescleary: probably don't need this, as we allow users to come with their own config file.

docker/docker-compose-func-test.yml
```diff
@@ -48,6 +48,18 @@ services:
       - elasticmq
       - dynamodb
 
+  portal:
```
pauljamescleary: why do we have this in func test?  That docker compose file is only for API func tests?

docker/docker-compose-build.yml
```diff
@@ -46,3 +46,15 @@ services:
...
+
+  portal:
+    image: "vinyldns/portal:0.1"
+    ports:
+      - "9001:9001"
+    container_name: "vinyldns-portal"
+    volumes:
+      - ./portal/application.conf:/opt/docker/conf/application.conf
+      - ./portal/application.ini:/opt/docker/conf/application.ini
+      # - ./portal/trustStore.jks:/opt/docker/conf/trustStore.jks  # uncomment if using a custom trustStore
```
pauljamescleary: I don't think we need this.  The "how to add a trust store" should be in the "using the portal docker image"

docker/docker-compose-func-test.yml
pauljamescleary: can we remove this from func test?  The func test is for the API only.

docker/docker-compose-build.yml
nimaeskandary: @pauljamescleary Should these be taken out completely or can they be mounted by default since it will fail silently if the files aren't there

docker/docker-compose-build.yml
```diff
@@ -46,3 +46,15 @@ services:
...
+
+  portal:
+    image: "vinyldns/portal:0.1"
+    ports:
+      - "9001:9001"
+    container_name: "vinyldns-portal"
+    volumes:
+      - ./portal/application.conf:/opt/docker/conf/application.conf
+      - ./portal/application.ini:/opt/docker/conf/application.ini
+      - ./portal/trustStore.jks:/opt/docker/conf/trustStore.jks
```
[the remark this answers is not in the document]
nimaeskandary: Hm depends, I will add something extra in the readme to make sure this is covered, but since it fails silently if the trustStore is not there to mount, and people will need to hook up ldap anyway for their installation I kept it here. Fine removing it though and making sure people know to add it

DEVELOPER_GUIDE.md
```diff
@@ -136,11 +137,43 @@ From the root directory run...
...
+up by running the following `curl -v http://localhost:9000/status`. 
+
+The portal will start up on `http://localhost:9001`
 
 To stop the local setup, run `./bin/stop-all-docker-containers.sh` from the project root.
 
-#### Validating everything
+### Portal image config
+
+`docker-compose -f ./docker/docker-compose-build.yml up -d` will start up the api on `http://localhost:9000`, and the portal
```
pauljamescleary: Why is this titled "Portal image config".  Seems like this section is "starting up the portal docker image locally."

Should be `Startng a vinyldns-portal server instance locally`

The instructions are also redundant.

PR #31, opened 2018-07-31
Update README Logo

pauljamescleary:
* Add new logo to img.  This is the svg
* Updated README to point to new logo
* Delete the old logo


README.md
```diff
@@ -1,5 +1,15 @@
 [![Join the chat at https://gitter.im/vinyldns/Lobby](https://badges.gitter.im/vinyldns/vinyldns.svg)](https://gitter.im/vinyldns/Lobby)
 
+<p align="center">
```
britneywright: can we left align the logo. Also the svg isn't looking quite right in here.
screen shot 2018-07-31 at 3 30 12 pm

pauljamescleary: oof, that is weird.  It looked ok on mine, I wonder if there is a problem with the SVG?

britneywright: do we have a PNG version. svg can get a little funky sometimes.

PR #33, opened 2018-07-31
Add portal docker image

nimaeskandary:
### Set up portal docker image

While we already had the configurations in place to start the api via a docker image, we did not for the portal. This pr adds changes needed to build and publish the portal docker image, and to run vinyldns locally with the api and portal images 

* Added settings to `build.sbt for `sbt-native-packager` to create a docker image for the portal module
* Added application configuration overrides in `docker/portal` for booting up the portal image for local development
* Updated `docker/docker-compose-build.yml` to include the portal image
* Added `bin/docker-up-portal-server.sh` to start up vinyl portal, api, and their dependencies
* Fixed bug in `modules/api/src/universal/conf/application.conf` that didn't use mysql docker service name
* Updated developer docker documentation to reflect that now both the portal and api images can be run

### How to test

Start up sbt and run `;project api; docker:publishLocal`

Run `;project portal; docker:publishLocal`

Run `bin/docker-up-vinyldns.sh`


docker/docker-compose-build.yml
```diff
@@ -46,3 +46,14 @@ services:
       - bind9
       - elasticmq
       - dynamodb
+
+  portal:
+    image: "vinyldns/portal:0.1" # the version of the portal image we want to pull
```
pauljamescleary: for my local, this is `vinyldns/portal:0.8.0-SNAPSHOT` for the API and the portal.

We should update the build.sbt, add `dockerUpdateLatest := true` for both the `portalDockerSettings` and the `apiDockerSettings.

Then, we can change the image to ... `image: "vinyldns/portal:latest"` and `image: "vinyldns/api:latest"`

DEVELOPER_GUIDE.md
```diff
@@ -126,21 +138,28 @@ variables.
 #### Ports
 vinyldns only exposes port 9000 for HTTP access to all endpoints
 
-#### Starting a vinyldns installation locally in docker
-There is a handy docker-compose file for spinning up the production docker image on your local under `docker/docker-compose-build.yml`
+### Configuration for the vinyldns-portal image
 
-From the root directory run...
+#### Volume mounts
+* `/opt/docker/lib_extra` - place here additional jar files that need to be loaded into the classpath when the application starts up.
```
pauljamescleary: add a sentence before these to indicate that, by default, the image is configured to run in a docker compose environment locally.  To run in a production environment, you would have to configure the portal appropriately.

pauljamescleary: I don't want users thing that they _have_ to do _anything_ to start things up locally.

docker/portal/application.conf
```diff
@@ -0,0 +1,82 @@
...
+
+portal.dynamo_delay = 1100
+portal.vinyldns.backend.url = "http://vinyldns-api:9000"
+portal.test_login = true
+
+# configuration for the users and groups store
+dynamo {
+  key = "akid goes here"
+  secret = "secret key goes here"
+  endpoint = "http://vinyldns-dynamodb:8000"
```
mitruly: Should this be pointing to `19000`? `reference.conf` (for API) has `19000` set as the endpoint.

modules/api/src/universal/conf/application.conf
```diff
@@ -35,7 +35,7 @@ vinyldns {
     local-mode = true
     default {
       driver = "org.mariadb.jdbc.Driver"
-      migrationUrl = "jdbc:mariadb://localhost:3306/?user=root&password=pass"
+      migrationUrl = "jdbc:mariadb://vinyldns-mysql:3306/?user=root&password=pass"
```
mitruly: Just to confirm - we don't need to change this in `reference.conf` since we're assuming that the instance would be running locally, right?

docker/portal/application.conf
pauljamescleary: no, as it is going over the network address (vinyldns-dynamodb).  19000 is the port poked through on the host.

modules/api/src/universal/conf/application.conf
pauljamescleary: this is to address an issue running migration.  Basically Rebecca broke our docker image and this fixes it.

docker/portal/application.conf
mitruly: Oh right, I noticed.

PR #34, opened 2018-07-31
Add integration tests to Travis CI

mitruly:
- Added docker compose to `.travis.yml`
- Added docker compose to `sbt` command
- Updated port for MySQL from 3306 (default) to 19002 due to issues with binding to a used port
- Updated localhost port for JDBC to 19002 in `reference.conf`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,11 +3,14 @@ language: scala
 scala:
   - 2.12.6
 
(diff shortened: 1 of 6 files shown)


mitruly:
@pauljamescleary Right, that is the better approach here. I updated and re-pushed my branch.

PR #53, opened 2018-08-02
Move everything from scalaz to cats

pauljamescleary:
* Modified JsonValidations to use cats instead of scalaz
* Modified the Json protocols accordingly
* Replaced scalaz ValidationNel with cats ValidatedNel
* Replaced all Disjunctions with Either


modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeInterfaces.scala
```diff
@@ -63,12 +62,12 @@ object BatchChangeInterfaces {
   }
 
   def toSingleValidation[E <: DomainValidationError, A](
```
mitruly: This method can be removed since we're no longer using it in the code (given the `Scalaz` → `Cats` conversion).

mitruly:
A couple of other items:

- Can we also remove `vinyldns-full-logo-light.png` since it's not tied to the conversion in this PR?

- There is a blurb that we can now remove/update from `DEVELOPER_GUIDE.MD`:
`There is presently a hybrid of somethings scalaz and other things cats.  We are migrating away from scalaz, so when building new code prefer cats if possible.`

We might still want to have some blurb about preferring `Cats` to `Scalaz`, though.

mitruly:
Looks really good. I pulled down this branch and ran the tests locally and everything passed. Changes make sense.

pauljamescleary:
@mitruly will make those changes

PR #54, opened 2018-08-03
RFC: Pluggable Repos

pauljamescleary:
RFC for making the VinylDNS database pluggable, allowing other users to bring their own database of choice.


rfcs/0001-pluggable-repos.md
```diff
@@ -0,0 +1,117 @@
...
+The design mandates that there is a `class` that has a constructor taking a `Config` in order to initialize itself.  The following is the NoOp implementation...
+
+```scala
+class NoOpCrypto(config: Config) extends CryptoAlgebra {
+  def encrypt(value: String): String = value
+  def decrypt(value: String): String = value
+}
+```
+
+The current Crypto design suffers from a few flows, namely:
```
britneywright: think you mean `flaws` not `flows` here

rfcs/0001-pluggable-repos.md
```diff
@@ -0,0 +1,93 @@
...
+
+* Create a `DatabaseProvider` trait that loads _all_ repositories.  This is necessary as it supports 1) the setting up of the database and 2) the sharing of database connection information
+* Create a `Database` trait that has all of the repositories
+* The `DatabaseProvider` will be responsible for loading the `Database` via a `Config`
+* Move all hard-coded database initialization into a default `DatabaseProvider`
+* Load the `DatabaseProvider` implementation from a config section.
+
+```scala
+trait Database { 
+  def zone: ZoneRepository
```
mitruly: Does having all of the repositories defined in the trait mean that all of the repositories have to be the same database type? If so, would it be better to structure this in a way that repositories can be linked to a database type individually or by grouping?

The motivation would be that consumers of VinylDNS may want to mix-and-match database types.

pauljamescleary: You can mix and match database types as well.  The `def zone: ZoneRepository` means that it could be a MySql repository while the `users` be a `UserRepository`.

Mixing does muddy the waters a little bit though and I think needs to be clarified.

pauljamescleary: I do like having a `trait Database` that gets loaded.  I will rework the design.  I think the changes will be:

1. The DatabaseProvider will not be configuration driven
1. The Database will be a class and not a trait
1. Have a `RepositoryProvider` for each specific repository.  That would be similar to what was implemented in the `DatabaseProvider` on this version.

This allows you to greatly mix-and-match each repository in different places.

pauljamescleary:
Closing as duplicate of https://github.com/vinyldns/rfcs/pull/3

PR #56, opened 2018-08-03
RFC: Auditing CNAMEs

kontrolldkaos:


pauljamescleary:
For the resolution check, would we simply do a DNS lookup against the CNAME to ensure that it resolves (vs an NXDomain)?

I suppose that when we load the zone, we could run that process (or kick off a background process) that would set some kind of flag or message on the record set to indicate that it was not resolved?

A few questions:

1. What would this status be called, "Unresolved"?
1. How should that materialize to the user in the user interface?
1. Would we provide a "button" in the user interface to allow people to run an "Audit"?
1. We would similarly provide an API endpoint.
1. Would we want to generate a "report" separate from somehow decorating the records to indicate they don't resolve?

For implementation, we could just use straight java to run the CNAME audit in parallel.

pauljamescleary:
Close as this is moved to the new rfcs repo

PR #64, opened 2018-08-06
Enable functional tests in Travis build

mitruly:
- `travis.yml`: Add `python` support,  removed non-existent `docker-compose` service and added functional test task.
- `func-test-api-travis.sh`: Create Travis-friendly version of `func-test-api.sh` with supported `docker-compose` parameters.
- `docker-compose-build.yml`/`docker-compose-func-test.yml`: Update MySQL port to `19001`
- `conftest.py`: Add DNS IP to logging
- `DnsJsonProtocol.scala`: Add back JSON validations prior to `scalaz` → `cats` conversion.

### DIFF
--- a/bin/func-test-api-travis.sh
+++ b/bin/func-test-api-travis.sh
@@ -33,11 +33,11 @@ fi
 cp -f $DIR/../modules/api/target/scala-2.12/vinyldns.jar $WORK_DIR/docker/api
 
 echo "Starting docker environment and running func tests..."
-docker-compose -f $WORK_DIR/docker/docker-compose-func-test.yml --project-directory $WORK_DIR/docker --log-level ERROR up --build --exit-code-from functest
+docker-compose -f $WORK_DIR/docker/docker-compose-func-test.yml --project-directory $WORK_DIR/docker up --exit-code-from functest
 test_result=$?
 
 echo "Grabbing the logs..."
-
+docker logs vinyldns-functest
 docker logs vinyldns-api > $DIR/../target/vinyldns-api.log 2>/dev/null
 docker logs vinyldns-bind9 > $DIR/../target/vinyldns-bind9.log 2>/dev/null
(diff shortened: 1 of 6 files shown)


bin/func-test-api-travis.sh
```diff
@@ -0,0 +1,52 @@
...
+
+echo "Copy all docker to the target directory so we can start up properly and the docker context is small..."
+cp -af $DIR/../docker $WORK_DIR/
+
+echo "Copy over the functional tests as well as those that are run in a container..."
+mkdir -p $WORK_DIR/functest
+rsync -av --exclude='.virtualenv' $DIR/../modules/api/functional_test $WORK_DIR/docker/functest
+
+echo "Copy the vinyldns.jar to the api docker folder so it is in context..."
+if [[ ! -f $DIR/../modules/api/target/scala-2.12/vinyldns.jar ]]; then
```
nimaeskandary: @mitruly Now that we have a process for starting up the api via docker in `bin/docker-up-vinyldns.sh`, can this be done in this script without the fat jar for consistency? The process would look something like 

1. change the api target in `docker/docker-compose-func-test.yml` to mimic the one in `docker/docker-compose-build.yml`, the tag on the image should be "latest"
1. In this script, push the api docker image to the local docker daemon in travis: `sbt ;project api;clean;docker:publishLocal`
1. Then `docker-compose -f $WORK_DIR/docker/docker-compose-func-test.yml --project-directory $WORK_DIR/docker up --exit-code-from functest
test_result=$?` should be enough to start up the api without the fat jar

nimaeskandary: Actually scratch that, don't know if that should be a larger refactoring effort? Thought ideally we would stop using fat jars once the api docker image was ironed out

mitruly: @nimaeskandary That’s a great suggestion, but we probably want to get the func tests integrated into the build ASAP.

PR #66, opened 2018-08-07
reorganize documentation site

### LINKED ISSUE #67: reorganize documentation site
part of #29 

The documentation site is no longer strictly an API documentation site. We want to contain all our guides and information about VinylDNS in it. To prep for the additional sections we should figure out the general organization of the site.

britneywright:
closes #67 

To run locally:

In one terminal tab:
`sbt ";project docs ;makeMicrosite"`

In another terminal tab:
`cd modules/docs/target/site/ && jekyll serve`

In the browser go to http://127.0.0.1:4000/


modules/docs/src/main/tut/contributing/index.md
```diff
@@ -0,0 +1,184 @@
+---
```
pauljamescleary: I don't think we should move contributing in the docs.  The documentation is meant to _span_ repositories.  The contributing guide in here is specific to the main vinyldns repository.

The ruby client, python client, go, etc may have their own guides.

modules/docs/src/main/tut/developer/index.md
```diff
@@ -0,0 +1,7 @@
+---
```
pauljamescleary: Same for the developer guide.  Developer guide is specific to `vinyldns` and does not span repositories.  It should not be in the "docs" which are really meant to be "cross repository" documentation.

pauljamescleary:
Seems like a lot of things break when testing on my local.

* Click Guides
* Click "User Guide" on the left nav
* It expands, but then the rest of the links disappear, and the top "Guides" link changes to "Main Menu"

* Click "API Documentation"
* In the left nav bar click "User Guide"
* Same issue as above

Seems like in general the "Guides" navigation is broken.

Other notes:
- I would remove `Contributing` and `Developing` all together from the main menu and docs site as they are repository / project specific.

PR #68, opened 2018-08-07
Enable code coverage and gh-pages publish on merge to master

mitruly:
- Consolidate verify step for API and portal to generate code coverage, removing deprecated `verify-api` and `verify-portal` `sbt` commands. Finish successful build by uploading to CodeCov.
- Add `47deg/Github4s` dependency and configure travis environment variable to publish Github pages on merge (see: https://47deg.github.io/sbt-microsites/docs/publish-with-travis.html#publish-using-github4s).

Notes:
- I needed to create a personal token since we need to give Travis read/write repo access. There are currently two Travis builds that run - one from the organizational `vinyldns/vinyldns` repo, and one from the Travis CI checks. Once this gets merged, The simpler CI build will get removed (as it's a duplicate).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,12 +32,17 @@ before_cache:
   
 jobs:  
   include:
-    - stage: "Verify"
-      script: sbt ++$TRAVIS_SCALA_VERSION ";validate-api; validate-core; verify-api"
+    - stage: "Verify core and API modules"
+      script: sbt ++$TRAVIS_SCALA_VERSION ";validate-api; validate-core"
       name: "Validate and verify API"
     - script: ./bin/func-test-api-travis.sh
       name: "Run API functional tests"
-    - script: sbt ++$TRAVIS_SCALA_VERSION ";validate-portal; verify-portal"
-      name: "Validate and verify portal"
+    - script: sbt ++$TRAVIS_SCALA_VERSION "validate-portal"
+      name: "Validate portal"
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -32,12 +32,14 @@ before_cache:
...
     - script: ./bin/func-test-api-travis.sh
       name: "Run API functional tests"
-    - script: sbt ++$TRAVIS_SCALA_VERSION ";validate-portal; verify-portal"
-      name: "Validate and verify portal"
+    - script: sbt ++$TRAVIS_SCALA_VERSION "validate-portal"
+      name: "Validate portal"
     - script: ./bin/func-test-portal.sh
       name: "Run portal functional tests"
+    - script: sbt ++$TRAVIS_SCALA_VERSION "verify"
+      name: "Verify and run coverage on API and portal"
```
pauljamescleary: this enabled code coverage but I think we still need to publish to `codecov` no?

mitruly: @pauljamescleary Right on. It should be added here now (see: `after_success`).

.travis.yml
```diff
@@ -32,12 +32,17 @@ before_cache:
...
+
+after_success:
+  - bash <(curl -s https://codecov.io/bash)
```
mitruly: This is all that's needed to upload the goods to CodeCov. One uploader to rule them all.

PR #72, opened 2018-08-08
Update ROADMAP document.

pauljamescleary:
Organized into large features, and added plan implementation
and availability.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -45,13 +45,11 @@ the zone will live, VinylDNS creates the delegation as well as access controls f
 1. Zone Maintenance - support the modification of zone properties, like default SOA record settings.
 
 ## Other
-There are several other features that we would like to support.  We will be opening up these for RFC, and their
-implementation can hopefully be contributed by members in the open source community, or perhaps the demand shuffles
-our priorities.  These include:
+There are several other features that we would like to support.  We will be opening up these for RFC shortly.  These include:
 
 1. DNS SEC - There is no first-class support for DNS SEC.  That feature set is being defined.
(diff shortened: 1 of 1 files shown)


ROADMAP.md
```diff
@@ -1,30 +1,57 @@
...
+1. Zone Creation - allow the creation of a sub-domain from an existing Zone.  Users choose the Server Group where
+the zone will live, VinylDNS creates the delegation as well as access controls for the new zone.
+1. Zone Maintenance - support the modification of zone properties, like default SOA record settings.
+
+## Other
+There are several other features that we would like to support.  We will be opening up these for RFC, and their
+implementation can hopefully be contributed by members in the open source community, or perhaps the demand shuffles
+our priorities.  These include:
+
+1. DNS SEC - There is no first-class support for DNS SEC.  That feature set is being defined.
```
britneywright: What is DNS SEC? DNS Security?

ROADMAP.md
```diff
@@ -1,30 +1,57 @@
...
 1. Record meta data - VinylDNS will allow the "tagging" of DNS records with arbitrary key-value pairs
 1. DNS GSLB - Integration with various GSLB vendors for common DNS GSLB configurations
```
britneywright: I would write out GSLB at least once here for those who might not be familiar with what it means.

pauljamescleary:
@britneywright unclear on how granular the `ROADMAP` would become.  Hopefully this just stays to high-level features.  I imagine that a new automation tool would be valuable to see here.

Changes to existing tools I would not expect to see in the `ROADMAP` as they are too low level.

We should have a CHANGELOG, minimally we need to be disciplined about Github releases to make sure we capture all of the changes since the last release.  This holds true for the tooling repos.

Issue #73, opened 2018-08-08
Travis should publish documentation when PRs are merged


britneywright:
We tried to implement publishing of the sbt-microsite through Travis when a PR is merged, but it doesn't seem to be working. I merged a PR with doc changes but the gh-pages branch didn't update.

https://47deg.github.io/sbt-microsites/docs/publish-with-travis.html#publish-using-github4s

PR #77, opened 2018-08-09
make publishing microsites work

britneywright:
I looked at the sbt-microsites docs and the only difference i noticed between what they suggest and what we have set up is the value for micrositeGithubToken is `getEnvVar("GITHUB_TOKEN")`. We have `sys.env.get("GITHUB_TOKEN")`. So changing to match what they have. If that doesn't work we can dig deeper. for #73

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,3 +46,6 @@ jobs:
 
 after_success:
   - bash <(curl -s https://codecov.io/bash)
+  - if ["$TRAVIS_BRANCH" = "master" -a "$TRAVIS_PULL_REQUEST" = "false"]; then
+      sbt ";project docs ;publishMicrosite";
+    fi


mitruly:
@britneywright I already tried `getEnvVar` the other day, but that wasn't recognized as a command. `47deg/sbt-microsites` lists it as a way to retrieve environment variables from Travis, but that's what `sys.env.get` should be doing.

britneywright:
So i *think* we did need to add something to the travis.yml to publish the docs. I was reading the sbt-microsites gitter and someone mentioned using Github4s and also referenced this line https://github.com/47deg/sbt-microsites/blob/master/.travis.yml#L43

.travis.yml
```diff
@@ -46,3 +46,6 @@ jobs:
 
 after_success:
   - bash <(curl -s https://codecov.io/bash)
+  - if ["$TRAVIS_BRANCH" = "master" -a "$TRAVIS_PULL_REQUEST" = "false"]; then
+      sbt ";project docs ;publishMicrosite";
```
britneywright: not sure if this needs to include the SBT_VERSION

mitruly: I don't think this will work, since `after_success` will execute on a successful build, which means that it will execute regardless of a PR merge.

britneywright: but I think that's what the "$TRAVIS_PULL_REQUEST" = "false" is for, so that it doesn't execute on PRs. The *intention* is that it applies when the master build is run and is successful.

pauljamescleary: IT does seem that variable is correct `$TRAVIS_PULL_REQUEST`.

I will leave it up to you guys on this one, beyond me.  Approving as I have faith.

PR #80, opened 2018-08-09
Implement JavaCrypto

mitruly:
Implement a symmetric cryptography mechanism that uses AES-CBC padding with support for 256-bit keys, which doesn't double-encrypt or double-decrypt.

`JavaCrypto.scala` provides the implementation and `JavaCryptoSpec.scala` tests the functionality. `generate-aes-256-hex-key.sh` is a script that utilizes `openssl` to generate a 256-bit key that can be used with the cryptography mechanism.

Note for reviewers:
- To test this, I replaced updated the `crypto` object in `application.conf` to point to the new `JavaCrypto` mechanism with the appropriate key and ran unit and integration tests.
- Usage information will be added later in the operator guide.

### DIFF
--- a/modules/core/src/main/scala/vinyldns/core/crypto/JavaCrypto.scala
+++ b/modules/core/src/main/scala/vinyldns/core/crypto/JavaCrypto.scala
@@ -1,0 +1,139 @@
+/*
+ * Copyright 2018 Comcast Cable Communications Management, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package vinyldns.core.crypto
+
+import java.nio.charset.StandardCharsets
+import java.security.{GeneralSecurityException, SecureRandom, Security}
+
+import com.typesafe.config.Config
+import javax.crypto.Cipher
+import javax.crypto.spec.{IvParameterSpec, SecretKeySpec}
+import javax.xml.bind.DatatypeConverter
+import scodec.bits.ByteVector
+
+/**
+  * Provides functions for encrypting and decrypting data using a user-defined secret stored in the configuration.
+  *
+  * Uses 256-bit AES key with CBC/PKCS5Padding algorithm.
+  */
+class JavaCrypto(cryptoConfig: Config) extends CryptoAlgebra {
+
+  private val zeroByte = 0.toByte
+  private val zeroChar = 0.toChar
+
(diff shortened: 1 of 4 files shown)


DEVELOPER_GUIDE.md
```diff
@@ -274,3 +294,12 @@ fires up _only_ the dns server with our test zones.  The DNS server is accessibl
 ### Publish the API docker image
 `bin/docker-publish-api.sh` - publishes the API docker image.  You must be logged into the repo you are publishing to
 using `docker login`, or create a file in `~/.ivy/.dockerCredentials` that has your credentials in it following the format defined in https://www.scala-sbt.org/1.x/docs/Publishing.html
+
+# Generating encryption key
```
britneywright: this should be h2 and also referenced in the table of contents at the top if it's intended to be it's own major section.

DEVELOPER_GUIDE.md
```diff
@@ -215,6 +215,26 @@ Configuration of the application is done using [Typesafe Config](https://github.
...
+ crypto {
+    type = "vinyldns.core.crypto.JavaCrypto"
+    secret = "8B06A7F3BC8A2497736F1916A123AA40E88217BE9264D8872597EF7A6E5DCE61"
+ }
+ ```
+where `secret` is a 128-, 192- or 256-bit key. `bin/generate-aes-256-hex-key.sh` is a script that utilizes `openssl`
+to conveniently generate a 256-bit key.
+
+Important notes: `JavaCrypto` uses Java Cryptography Extension (JCE) Unlimited Strength Policy Jurisdiction, which
+may require some patching for older versions of Java 8 (that aren't required for Java 9 and later).
```
pauljamescleary: should add that it requires build 171 I think on Java 8, but no patching on Java 9?

DEVELOPER_GUIDE.md
pauljamescleary: this shouldn't be in the developer guide.  This would be in the Operator Guide I am working on.  The developer should never have to touch this stuff.

I recommend we remove the section all together

DEVELOPER_GUIDE.md
pauljamescleary: This should be in the operator guide, not Developer guide.  Recommend we remove the section all together.

DEVELOPER_GUIDE.md
mitruly: @pauljamescleary That makes sense - I have removed it from the `DEVELOPER_GUIDE.md`.

modules/core/src/main/scala/vinyldns/core/crypto/JavaCrypto.scala
```diff
@@ -0,0 +1,139 @@
...
+  // Required for 256-bit keys and padding cryptography mode.
+  // TODO: Remove once Java is updated to version 9, where it is enabled by default.
+  Security.setProperty("crypto.policy", "unlimited")
+
+  /*
+   * Attempt to encrypt a string. If the string is already encrypted, do not double encrypt.
+   */
+  @throws(classOf[GeneralSecurityException])
+  def encrypt(value: String): String =
+    if (value.startsWith(encryptedPrefix)) {
```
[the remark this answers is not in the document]
mitruly: @nimaeskandary Normally, no, but this is a safeguard just in case it happens. This is similar how we don't double-decrypt since we otherwise wouldn't be able to determine how many times a string has been encrypted/decrypted.

PR #85, opened 2018-08-10
only remove vinyl containers

nimaeskandary:
The old `bin/stop-all-docker-containers.sh` killed and removed all containers running on a machine. A developer can be running containers that are not related to the vinyl project. The new `bin/remove-vinyl-containers.sh` only targets containers related to the vinyl project


DEVELOPER_GUIDE.md
```diff
@@ -260,7 +258,7 @@ The following ports and services are available:
 - sqs - 9324
 - api server (main vinyl backend app) - 9000
 
-To kill the environment, run `bin/stop-all-docker-containers.sh`
+To kill the environment, run `bin/remove-vinyl-containers.sh`
```
pauljamescleary: should we as well change line 264, since you deleted the file?

bin/remove-vinyl-containers.sh
```diff
@@ -0,0 +1,29 @@
...
+#   vinyldns/api
+#   vinyldns/portal
+#
+# This script with kill and remove containers associated
+# with these names and/or tags
+#
+# Note: this will not remove the actual images from your
+# machine, just the running containers
+
+IDS=$(docker ps -a | grep -e 'mysql:5.7' -e 'cnadiminti/dynamodb-local:2017-02-16' -e 's12v/elasticmq:0.13.8' -e 'vinyldns'  | awk '{print $1}')
```
mitruly: Should we do something similar to `vinyldns` where we don't specify a specific release? Not sure how often those images would realistically be updated, so it probably wouldn't matter much. Just hypothetically thinking of someone having an older container running when our tags are updated in the code.

nimaeskandary: Yeah was a little iffy on that, this way we'd def have to make sure we update tags in this script if we change anything which is a downside, but also makes sure that on the off chance someone it running mysql for another project, theres less of a chance we'd kill their container unintentionally. Up for debate though, can change this script to be more generic with the greps

mitruly: I'm fine with this if no one else has any issues.

rebstar6: we have names on our images....grep for `vinyldns-mysql`, `vinyldns-dynamodb`...

rebstar6: see https://github.com/vinyldns/vinyldns/blob/master/docker/docker-compose-build.yml#L5

rebstar6: sorry, stream of thought commenting - along those lines, `docker ps -a | grep -e 'vinyldns' | awk '{print $1}'` is probably sufficient since all the names have vinyldns in the container name

nimaeskandary: @rebstar6 Only for docker-compose-build, sbt dockerComposeUp doesn't

DEVELOPER_GUIDE.md
nimaeskandary: updated

bin/remove-vinyl-containers.sh
rebstar6: sad.

rebstar6:
also good with this once jenks is

Issue #24, opened 2018-07-30
Set scalac compiler flag for fatal warnings


pauljamescleary:
**Motivation**
We do not fail compile if any of our compiler flags warn.  This causes failures during PR builds that could have been resolved locally.

**Scope of change**
Add `"-Xfatal-warnings"` to our compiler options.

kevinmeredith:
Perhaps it's worthwhile to extend this issue's scope to use https://github.com/DavidGregory084/sbt-tpolecat? What do you think?

pauljamescleary:
@kevinmeredith that sounds like a good idea.  Most of our `CompilerOptions` are pulled from tpolecat's recommendations.

If you are willing, would gladly take a PR?

A few things to consider that really should be fixed, but are a little beyond simple:

1. The core module currently has a `NoOpCrypto` that takes a config that is not used that will fail.  We currently in the build filter that out.  Hopefully we can still do the same with that plugin?
2. The portal module really should have a limited set of compiler flags.  It was poorly maintained and would like to ideally overhaul the portal (scalajs? http4s?).  I would be curious how many of those flags would work.

pauljamescleary:
Let's evaluate if the plugin is suitable, i.e. there aren't _tons_ of warnings.  Otherwise, let's for the scope of this issue just flip fatal on.

pauljamescleary:
@kevinmeredith had 200 some odd errors on just normal compile, forget about test compile.

Some of those errors should probably be fixed, but will be a little bigger of a cleanup effort.

Moving forward with the `fatal` for the time being.

Thanks for the recommendation!  The plugin just dropped in and worked.

kevinmeredith:
Glad to hear and thanks, Paul! I hope to contribute to this library at some point.

PR #90, opened 2018-08-10
add scalac option for fatal warnings (#24)

michael-smith:
The `NoOpCrypto` failed due to an unused `config` in the constructor. I left it there because the scaladoc for the class mentions it specifically. To fix I just added a `val` unless there is a more desired approach.

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeInterfaces.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeInterfaces.scala
@@ -17,7 +17,7 @@
 package vinyldns.api.domain.batch
 
 import cats.data.Validated.{Invalid, Valid}
-import cats.data.{NonEmptyList, _}
+import cats.data._
 import cats.implicits._
 import vinyldns.api.domain.DomainValidationError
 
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -52,6 +52,7 @@ lazy val sharedSettings = Seq(
   startYear := Some(2018),
   licenses += ("Apache-2.0", new URL("https://www.apache.org/licenses/LICENSE-2.0.txt")),
   scalacOptions += "-target:jvm-1.8",
+  scalacOptions += "-Xfatal-warnings",
```
pauljamescleary: I would add this to https://github.com/vinyldns/vinyldns/blob/master/project/CompilerOptions.scala instead of inlined here.

One could argue that the `-target:jvm-1.8" should go there as well

pauljamescleary:
Also, if you ever want to talk offline, you can join the gitter channel.  There are usually folks hanging out in there

build.sbt
michael-smith: Ah yes, that makes sense. I latched onto the target jvm scalacOption and figured that was the right place. I'll make the change and resubmit.. Thanks.

PR #93, opened 2018-08-10
Remove DB Migration Runner

pauljamescleary:
DB Migrations (flyway) are run out-of-band.  They require you to
first `sbt assembly` and build a fat jar, and then subsequently
to run a `DbMigrationRunner`.

All of those steps are unnecessary as flyway supports concurrent
attempts to run migrations.  The specific item is in the FAQ
at https://flywaydb.org/documentation/faq#parallel

Instead, just run migrations when the application starts.

* Modified `VinylDNSJDBC` to _always_ run database migrations.  Removed
the embedded flag as that is no longer needed.
* Deleted the `migrations.conf` file as that will no longer be used.
* Cleaned up the config files.  The salient change was to remove the
`local-mode` flag, as it is no longer needed.  Found some left over
cruft along the way that is unused and cleared that up.


modules/api/src/main/scala/vinyldns/api/repository/mysql/VinylDNSJDBC.scala
```diff
@@ -68,21 +61,19 @@ class VinylDNSJDBC(config: Config) {
   }
 
   // Only run migrations if we are using embedded db as embedded requires initialization
```
nimaeskandary: Is this comment still relevant?

modules/api/src/it/resources/application.conf
```diff
@@ -3,70 +3,9 @@ akka {
...
   log-dead-letters = 0
   logger-startup-timeout = 60s
-
-  actor {
-    provider = "akka.actor.LocalActorRefProvider"
-  }
 }
 
 vinyldns {
-  dns-backends = [
```
rebstar6: @pauljamescleary where did we land on the multiple backends stream of work? we still are only at one, i assume all this stuff went in when we were working towards that.

do we have cards or something tracking that (even if low priority)?

modules/api/src/main/scala/vinyldns/api/repository/mysql/VinylDNSJDBC.scala
rebstar6: nope

modules/api/src/main/scala/vinyldns/api/repository/mysql/VinylDNSJDBC.scala
```diff
@@ -68,21 +61,19 @@ class VinylDNSJDBC(config: Config) {
...
+  val dbName = config.getString("name")
+  val placeholders = Map("dbName" -> dbName)
+  migration.setPlaceholders(placeholders.asJava)
+  migration.setSchemas(dbName)
 
-    // Runs ALL flyway migrations including SQL and scala, assumes empty databases
-    migration.migrate()
-    logger.info("migrations complete")
-  }
+  // Runs ALL flyway migrations including SQL and scala, assumes empty databases
```
rebstar6: nor is this comment. we need to make sure our migrations are fine to be run multiple times on non-empty databases if this is to work

docker/api/docker.conf
```diff
@@ -50,7 +50,6 @@ vinyldns {
   # default settings point to the setup from docker compose
   db {
```
rebstar6: docker/api/run.sh has
```
echo "Running migrations..."
java -Dconfig.resource=db-migrations.conf -cp /app/vinyldns-server.jar db.migration.MigrationRunner
```

need to remove

pauljamescleary: good catch, surprised func tests pass

modules/api/src/it/resources/application.conf
pauljamescleary: we have cards in the old repo to track.  This needs some re-thinking.  I did not port over Deepak's code when we open sourced.

PR #94, opened 2018-08-11
Log for all monitors

pauljamescleary:
The amount of VinylDNS traffic is generally rather low.  It makes
sense to log every monitored event in order to calculate error
rate, latency, and volume.  All of those are easy to generate with
this change from standard log tools like Kibana and Splunk.

* Added a `logEntry` function to `Monitor` that creates a log entry
as a string
* Updated the `Monitor` to log to info for every capture or failure

In addition, removed the `log4jReporter`.  Logging the metrics in
bulk added a lot of noise to the logs.  Plus, cannot calculate
histograms across machines, which can be done by simply logging
all events.

* Removed log4jReporter
* Removed conf entries related to log4j reporter
* Cleaned up some more config that is no longer used


modules/api/src/main/scala/vinyldns/api/route/Monitor.scala
```diff
@@ -72,6 +73,14 @@ object Monitor {
   lazy val monitors: mutable.Map[String, Monitor] = concurrent.TrieMap.empty
 
   def apply(name: String): Monitor = monitors.getOrElseUpdate(name, new Monitor(name))
+
+  def logEntry(monitorName: String, durationMillis: Long, success: Boolean): String = {
+    val sb = new StringBuilder
+    sb.append("monitor='").append(monitorName).append("'")
```
mitruly: `sb.append(s"monitor='$monitorName'")` may be easier to read.

pauljamescleary: @mitruly the reason to use string builder as opposed to string interpolation is Stringbuilder is more efficient.

PR #96, opened 2018-08-13
Allow config for dynamo region in portal

pauljamescleary:
This was hard coded to us-east-1, which breaks if trying to use
any other region for dynamodb.

* Changed `Module` which bootstraps the portal to load the region
from config
* Added a unit test to ensure that the setting is indeed present
in application-test.conf


modules/portal/test/ConfigSpec.scala
```diff
@@ -0,0 +1,37 @@
...
+      val configuration = Configuration.apply(ConfigFactory.load())
+      val dynamoAKID = configuration.get[String]("dynamo.key")
+      val dynamoSecret = configuration.get[String]("dynamo.secret")
+      val dynamoEndpoint = configuration.get[String]("dynamo.endpoint")
+      val region = configuration.get[String]("dynamo.region")
+
+      dynamoAKID must beEqualTo("akid goes here")
+      dynamoSecret must beEqualTo("secret key goes here")
+      dynamoEndpoint must beEqualTo("endpoint url goes here")
+      region must beEqualTo("us-east-1")
```
mitruly: Should we be checking for `test_datastore` as well? These values are different between `application.conf` and `application-test.conf`, but figured I'd ask just in case we're checking for existence of values.

pauljamescleary: I am unit testing values that would be used in production.  I don't think it makes sense to test for that.

PR #98, opened 2018-08-13
Suppress error thrown from scaladoc compilation 

nimaeskandary:
Adding fatal warnings in this pr https://github.com/vinyldns/vinyldns/pull/90 caused errors when compiling scaladocs. Followed this thread to resolve the error: https://issues.scala-lang.org/browse/SI-9311. This is necessary because sbt docker builds scaladocs and will error out

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -53,6 +53,7 @@ lazy val sharedSettings = Seq(
   licenses += ("Apache-2.0", new URL("https://www.apache.org/licenses/LICENSE-2.0.txt")),
   scalacOptions += "-target:jvm-1.8",
   scalacOptions ++= scalacOptionsByV(scalaVersion.value),
+  scalacOptions in (Compile, doc) += "-no-link-warnings",
   // Use wart remover to eliminate code badness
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -306,6 +306,10 @@ lazy val docSettings = Seq(
 lazy val docs = (project in file("modules/docs")).enablePlugins(MicrositesPlugin)
   .settings(docSettings)
 
+scalacOptions in (core, Compile, doc) ++= Seq(
```
pauljamescleary: we don't put things at the top level anymore.  Should this just go in `CompilerOptions`?

nimaeskandary: Tried that originally and got `bad option: '-no-link-warnings'` when compiling

pauljamescleary: You can add the following to `sharedSettings` ...

```scala
  // add no link warnings for docs
  scalacOptions in (Compile, doc) += "no-link-warnings"
```

pauljamescleary: and remove it from the top level

PR #102, opened 2018-08-14
Fix NS management

rebstar6:
2 things going on here 
1. Open up NS management to zone admins/ACL folks (remove the group check)
2. Add the approved server list check! not sure what happened to that, we were supposed to be validating NS servers on a list, but seems like weve been ignoring that

### DIFF
--- a/modules/api/functional_test/live_tests/recordsets/create_recordset_test.py
+++ b/modules/api/functional_test/live_tests/recordsets/create_recordset_test.py
@@ -1528,6 +1528,30 @@ def test_ns_create_for_admin_group_succeeds(shared_zone_test_context):
             client.wait_until_recordset_deleted(result_rs['zoneId'], result_rs['id'])
 
 
+def test_ns_create_for_unapproved_server_fails(shared_zone_test_context):
+    """
+    Tests that an ns change fails if one of the servers isnt approved
+    """
+    client = shared_zone_test_context.ok_vinyldns_client
+    zone = shared_zone_test_context.parent_zone
+
+    new_rs = {
+        'zoneId': zone['id'],
+        'name': 'someNS',
+        'type': 'NS',
+        'ttl': 38400,
+        'records': [
+            {
+                'nsdname': 'ns1.parent.com.'
+            },
+            {
+                'nsdname': 'this.is.bad.'
+            }
+        ]
+    }
+    client.create_recordset(new_rs, status=422)
+
+
 def test_ns_create_for_origin_fails(shared_zone_test_context):
     """
     Tests that an ns create for origin fails
--- a/modules/api/functional_test/live_tests/recordsets/create_recordset_test.py
+++ b/modules/api/functional_test/live_tests/recordsets/create_recordset_test.py
@@ -1499,31 +1499,9 @@ def test_dotted_host_create_fails(shared_zone_test_context):
     assert_that(error, is_('Record with name record-with.dot is a dotted host which is illegal in this zone ok.'))
 
 
-def test_ns_create_for_non_approved_group_fails(shared_zone_test_context):
+def test_ns_create_for_admin_group_succeeds(shared_zone_test_context):
     """
-    Tests that an ns change on a group whose admin group is not approved fails (only ok group is approved)
-    """
-    client = shared_zone_test_context.dummy_vinyldns_client
-    zone = shared_zone_test_context.parent_zone
-
-    new_rs = {
-        'zoneId': zone['id'],
-        'name': 'someNS',
-        'type': 'NS',
-        'ttl': 38400,
-        'records': [
-            {
-                'nsdname': 'ns1.parent.com.'
-            }
-        ]
-    }
-    error = client.create_recordset(new_rs, status=403)
-    assert_that(error, is_('Do not have permissions to manage NS recordsets, please contact vinyldns-support'))
-
-
-def test_ns_create_for_approved_group_passes(shared_zone_test_context):
-    """
-    Tests that an ns change on a group whose admin group is approved passes (only ok group is approved)
+    Tests that an ns change passes
     """
     client = shared_zone_test_context.ok_vinyldns_client
     zone = shared_zone_test_context.parent_zone
--- a/modules/api/functional_test/live_tests/recordsets/delete_recordset_test.py
+++ b/modules/api/functional_test/live_tests/recordsets/delete_recordset_test.py
@@ -536,43 +536,9 @@ def test_user_can_delete_record_via_group_acl_rule(shared_zone_test_context):
             client.wait_until_recordset_change_status(delete_result, 'Complete')
 
 
-def test_ns_delete_for_non_approved_group_fails(shared_zone_test_context):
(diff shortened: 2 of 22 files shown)


modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetValidations.scala
```diff
@@ -178,6 +180,26 @@ object RecordSetValidations {
...
+    }
+
+    badNs match {
+      case Some(unapproved) =>
+        InvalidRequest(
+          s"Nsdname ${unapproved.nsdname} is not an approved name server. " +
+            s"Contact vinyldns-support for assistance.").asLeft
+      case None => ().asRight
+    }
+  }
```
pauljamescleary: This used to be a ValidatedNel.

Also, if just doing an `Either`, seems like there is a better way

```
  /* Inspects each record in the rdata, returning back the record set itself or all ns records that are not approved */
  def containsApprovedNameServers(approvedServerList: List[Regex], nsRecordSet: RecordSet): ValidatedNel[String, RecordSet] = {
    val validations: List[ValidationNel[String, NSData]] =
      nsRecordSet.records
        .collect(case ns: NSData => ns)
        .map(isApprovedNameServer(approvedServerList, _))

    validations.sequence.map(_ => nsRecordSet)
  }
```

rebstar6: wow, sloppy Rebecca

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneRecordValidations.scala
```diff
@@ -37,10 +37,14 @@ object ZoneRecordValidations {
   def containsApprovedNameServers(
       approvedServerList: List[Regex],
       nsRecordSet: RecordSet): ValidatedNel[String, RecordSet] = {
-    val validations: List[ValidatedNel[String, NSData]] =
-      nsRecordSet.records
-        .map(_.asInstanceOf[NSData])
```
rebstar6: minor changes here to get rid of the asInstanceOf. i can revert this change if people feel im overstepping

rebstar6: considered collect, though it would be an error for sure if a non NS snuck in there, which is why i didnt.

I dont have strong feelings though, would be fine will just collecting and calling it a day

modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetValidations.scala
```diff
@@ -178,6 +180,13 @@ object RecordSetValidations {
       !isOriginRecord(recordSet.name, omitTrailingDot(zone.name))
     )
 
+  private def containsApprovedNameServers(nsRecordSet: RecordSet): Either[Throwable, Unit] =
+    ZoneRecordValidations
+      .containsApprovedNameServers(VinylDNSConfig.approvedNameServers, nsRecordSet)
+      .toEither
+      .map(_ => ())
+      .leftMap(errors => InvalidRequest(errors.head))
```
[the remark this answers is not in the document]
rebstar6: we can, it just gets wordy. i'll change it

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneRecordValidations.scala
```diff
@@ -37,10 +37,14 @@ object ZoneRecordValidations {
...
-        .map(isApprovedNameServer(approvedServerList, _))
+    val validations: List[ValidatedNel[String, NSData]] = nsRecordSet.records
+      .map {
+        case ns: NSData => ns.validNel[String]
+        case x => s"Error: $x is not NSData".invalidNel[NSData]
+      }
+      .map {
+        _.andThen(ns => isApprovedNameServer(approvedServerList, ns))
+      }
```
mitruly: One weird thing that I think is with newer versions of IntelliJ (I'm on 2018.2 CE) is that it tries to resolve parameters for `validations.sequence` and fails:
```Expression of type Nothing[List[Nothing]] doesn't conform to expected type data.ValidatedNel[String, List[RecordSet]]```

Wondering if we should include version suggestions in the `DEVELOPER_GUIDE.md` or something since IntelliJ shows this line as a hard error even though it compiles and runs.

File: modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneRecordValidations.scala (hunk 2 above)
Comment (pauljamescleary) [reply modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneRecordValidations.scala]: ok, was unsure.  I would prefer we skip any records that are not NS as opposed to error.

File: modules/api/functional_test/live_tests/recordsets/delete_recordset_test.py [hunk 5]
```diff
@@ -536,43 +536,9 @@ def test_user_can_delete_record_via_group_acl_rule(shared_zone_test_context):
             client.wait_until_recordset_change_status(delete_result, 'Complete')
 
 
-def test_ns_delete_for_non_approved_group_fails(shared_zone_test_context):
+def test_ns_delete_for_admin_group_passes(shared_zone_test_context):
```
nimaeskandary: Should we test a non admin group fails here as well? same with update

modules/api/functional_test/live_tests/recordsets/create_recordset_test.py
```diff
@@ -1499,31 +1499,9 @@ def test_dotted_host_create_fails(shared_zone_test_context):
     assert_that(error, is_('Record with name record-with.dot is a dotted host which is illegal in this zone ok.'))
 
 
-def test_ns_create_for_non_approved_group_fails(shared_zone_test_context):
+def test_ns_create_for_admin_group_succeeds(shared_zone_test_context):
```
nimaeskandary: should we test a non admin group fails

rebstar6: non admin group shouldnt fail - now NS follows the same logic as all the other record types (zone admins or people with ACL access).

can add in some ACL tests too

nimaeskandary: Ah gotcha. I see the unit tests for it so I'm good with it test wise.. especially since we test every record type for this kind of logic in our functional tests. Wonder if the func tests should be re written to do certain generic types of tests for every record type eventually

pauljamescleary:
I hate that we have test coverage % technically going down.   The `loadDns` has to be refactored and unit tested.  If we send in a `ZoneViewLoader` instead of hard-coding to `DnsZoneViewLoader` (which is kinda terrible) then we could easily unit test this.  Is this possible to do here?  Small change me-thinks?

rebstar6:
i feel like the diff coverage being 100% should be sufficient. If we delete tested code, our coverage will always go down because the relative mix of covered and not will change

i really dont think refactoring/testing ZoneViewLoader makes sense in this PR, i'm not messing with that

PR #105, opened 2018-08-14
Update poster for doc site

annelisebox:
Created custom poster for doc site. It will be the thumbnail when you share links for the doc site :)


britneywright:
I tested this locally by running the microsite, then used ngrok to create a public URL that I shared in slack and iMessage.
slack screenshot
imessage screenshot

PR #109, opened 2018-08-15
A list of requirements for Windows 10 install

djbooboo:
I wanted to make sure I didn't distract from the documentation too much for Windows 10.  If these items are added to windows, the original step by step should work for both Windows and MAC users.


README.md
```diff
@@ -57,7 +58,8 @@ VinylDNS is currently used in Production managing millions of DNS records.
 Docker images for VinylDNS live on Docker Hub at https://hub.docker.com/u/vinyldns/dashboard/. 
 To start up a local instance of VinylDNS on your machine with docker:
 
-1. Ensure that you have docker and docker-compose
+If you are running VinylDNS on a Windows 10 machine, please see the [Quickstart Requirements for Windows 10](#quickstart-requirements-for-windows-10)
```
mitruly: Instead of having a separate section later on with quickstart Windows requirements, can we have a sub-section at the start for requirements and have a Mac section with only `docker` and `docker-compose` and then the Windows requirements?

Looking to be consistent with the Developer Guide section of this README.

mwinsl001c: I get worried if quickstart becomes too busy.  Especially because Windows requirements takes 5-6 lines and we are assuming that the majority of the contributors will be on MAC.  For those Mac users, I wonder if having the Windows Requirements inline would be annoying?

mitruly: Right, I often forget that I have stuff pre-installed. Maybe it would just suffice to have those list of requirements for the quickstart and not specify the OS then. Wonder if there are any other thoughts on this.

britneywright: So Windows can follow the same Quick Start as Mac, presuming you have the required software installed beforehand? If so I would just have a line about requirements that links to the Developer guide and lists of what needs to be installed for mac and windows.

djbooboo: I think you are on the right track @britneywright .  What I was struggling with is the difference in how you obtain the dependencies on Windows vs Mac.  Mac is probably a bunch of apt-get lines, while Windows requires trips to the websites and downloaded installations.

Let me throw something else up there for review.

rebstar6:
djbooboo! gotta change your commit history to the right email so you can have the CLA appear signed

djbooboo:
Yeah @rebstar6 I am probably gonna close this pull request and resubmit.  I had 2 different accounts active during this PR ... screwed up everything.

DEVELOPER_GUIDE.md
```diff
@@ -75,6 +75,40 @@ The features that the Portal provides that are not in the API include:
...
+* scala sbt - https://www.scala-sbt.org/1.x/docs/Installing-sbt-on-Mac.html
+* Docker - https://docs.docker.com/docker-for-mac/install/
+* Wget - https://www.gnu.org/software/wget/
+
+Return to [Quickstart](README.md#quickstart)
+
+### Quickstart Requirements for Windows 10
+The following applications are required for the Quickstart of the application on Windows 10 
+
+* git ( for Windows ) - https://git-scm.com/download/win
```
britneywright: do we need to emphasize the `for Windows`? the requirements has the heading for Windows 10, so users should know what forms of the software they should be installing right? Unless any of them are potentially confusing.

README.md
```diff
@@ -57,7 +58,11 @@ VinylDNS is currently used in Production managing millions of DNS records.
 Docker images for VinylDNS live on Docker Hub at https://hub.docker.com/u/vinyldns/dashboard/. 
 To start up a local instance of VinylDNS on your machine with docker:
 
-1. Ensure that you have docker and docker-compose
+Depending on your operating system, please see the requirements:
```
britneywright: I'd say something like "Refer to the software dependencies based on your operating system:"

DEVELOPER_GUIDE.md
djbooboo: I'll take those out

README.md
britneywright: Also move this requirements part to the top of the quickstart. it's weird how it breaks into the middle of the docker part.

README.md
```diff
@@ -57,7 +58,11 @@ VinylDNS is currently used in Production managing millions of DNS records.
...
+- [Quickstart Requirements for Mac](DEVELOPER_GUIDE.md#quickstart-requirements-for-mac)
```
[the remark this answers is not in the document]
djbooboo: you guys are just having fun with me now

DEVELOPER_GUIDE.md
```diff
@@ -74,7 +74,81 @@ The features that the Portal provides that are not in the API include:
...
 ## Developer Requirements
+
+### Quickstart Requirements for Mac
+The following applications are required for the Quickstart of the application on a Mac
+
+* git - https://git-scm.com/book/en/v2/Getting-Started-Installing-Git
+* JDK 8 - http://www.oracle.com/technetwork/java/javase/downloads/jdk8-downloads-2133151.html
+* scala sbt - https://www.scala-sbt.org/1.x/docs/Installing-sbt-on-Mac.html
+* Docker - https://docs.docker.com/docker-for-mac/install/
+* Wget - https://www.gnu.org/software/wget/
```
nimaeskandary: @djbooboo Do we have to tell them where to install the dependencies? Can we just list git, jdk8, sbt, docker, and wget, then leave the note in case they are on windows, and combine these sections? Then we don't have specify for mac or windows, and also mac requirements will work for linux too

djbooboo: @nimaeskandary it is important for windows users I think.  I asked Paul what to do about Mac users.  Since we are being detailed for Windows users I wondered if we needed to be detailed for Mac users.  Personally, if someone had an issue with the install and asked me for help, the 1st thing I would ask is "where did you get these from?"

Have we certified on someone's Linux machine?

nimaeskandary: @djbooboo Paul used to develop on a Linux machine for some time, nothing major has changed since then in regards to dependencies

DEVELOPER_GUIDE.md
```diff
@@ -74,7 +74,81 @@ The features that the Portal provides that are not in the API include:
 * Creation of users - when a user logs in for the first time, VinylDNS automatically creates a user for them in the
 database with their LDAP information.
 
+## Quickstart
```
mitruly: Should the Quickstart be in `README.md`?

I typically find installation instructions for Github projects in either a README file or wiki. The quickstart section is more for just running VinylDNS and not configuring a development environment, though the requirements appear to overlap.

pauljamescleary:
Looks like Mike Winslow changed to djbooboo half way through.  @djbooboo you will need to amend author your earlier commits and change them to djbooboo; otherwise the CLA tool will complain

rebstar6:
closing for now, @djbooboo we'll have you reopen when it's not blocked by @mitruly's guide changes

PR #111, opened 2018-08-15
reduce travis jobs

britneywright:
Combining jobs to reduce the number of concurrent jobs from 5 to 2. This will allow us to run more builds in parallel.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ sudo: false
 
 jdk:
   - oraclejdk8
-  
+
 services:
   - docker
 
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -26,30 +26,21 @@ install:
...
 after_success:
-  - bash <(curl -s https://codecov.io/bash)
   - if [ "$TRAVIS_BRANCH" = "master" -a "$TRAVIS_PULL_REQUEST" = "false" ]; then
-      sbt 'project docs' 'publishMicrosite';
+      sbt ";project docs; publishMicrosite";
     fi
+
+jobs:
+  include:
+    - name: "Validate and verify API"
```
nimaeskandary: name should just be validate and verify since it runs portal and core too

PR #113, opened 2018-08-15
Update to sbt-protoc

pauljamescleary:
The sbt protobuf plugin we were using forced developers to install
protoc version 2.6.1 on their local.  That makes it difficult
to onboard new developers.

sbt-protoc is an alternative plugin.  It has a lot of features we
are presently not using.  The biggest feature it brings is to
not require developers to install protoc.

* build.sbt - use the new protoc plugin
* VinylDNSProto.proto - add syntax = proto2 to ensure compile
to compatible protobuf version
* plugins.sbt - remove the old protobuf plugin
* protoc.sbt - add the new protoc plugin

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,4 +1,3 @@
-import sbtprotobuf.{ProtobufPlugin => PB}
 import Resolvers._
 import Dependencies._
 import CompilerOptions._
@@ -169,7 +168,10 @@ lazy val portalPublishSettings = Seq(
 )
 
 lazy val pbSettings = Seq(
-  version in ProtobufConfig := "2.6.1"
+  PB.targets in Compile := Seq(
+    PB.gens.java("2.6.1") -> (sourceManaged in Compile).value
+  ),
+  PB.protocVersion := "-v261"
 )
 
 lazy val allApiSettings = Revolver.settings ++ Defaults.itSettings ++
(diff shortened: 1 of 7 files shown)


project/protoc.sbt
```diff
@@ -0,0 +1,3 @@
+addSbtPlugin("com.thesamet" % "sbt-protoc" % "0.99.18")
+
+libraryDependencies += "com.thesamet.scalapb" %% "compilerplugin" % "0.7.4"
```
thesamet: unless you plan to generate  Scala code for the protos, this can be safely removed.

build.sbt
```diff
@@ -182,14 +184,14 @@ lazy val allApiSettings = Revolver.settings ++ Defaults.itSettings ++
   scalaStyleSettings
 
 lazy val api = (project in file("modules/api"))
-  .enablePlugins(JavaAppPackaging, DockerComposePlugin, AutomateHeaderPlugin, ProtobufPlugin)
+  .enablePlugins(JavaAppPackaging, DockerComposePlugin, AutomateHeaderPlugin, ProtocPlugin)
```
thesamet: No need to enable `ProtocPlugin`, it's enabled by default and does nothing unless you specify `PB.targets`.

pauljamescleary: 👋  Hey @thesamet, thanks for dropping in and giving this a review!

nimaeskandary:
Should we also stop travis from installing protoc in this pr?

PR #117, opened 2018-08-16
Travis configuration - conditionally deploy docs

### LINKED ISSUE #73: Travis should publish documentation when PRs are merged

britneywright:
Conditionally publish docs after the test stages pass.
- Only runs on the master build.
- Deploys the docs once (instead of once per job in the same stage).

I was hoping there was a way to also move the doc related installs to that stage also, but that doesn't seem possible. closes #73

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,5 +36,6 @@ jobs:
     - name: "Run API and portal functional tests"
       script: ./bin/func-test-portal.sh && ./bin/func-test-api-travis.sh
     - stage: Deploy
+      if: branch = master AND type = push
       name: "Deploy doc site"
       script: sbt ";project docs; publishMicrosite";
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,14 +29,12 @@ before_cache:
   - find $HOME/.ivy2/cache -name "ivydata-*.properties" -print -delete
   - find $HOME/.sbt        -name "*.lock"               -print -delete
 
-after_success:
-  - if [ "$TRAVIS_BRANCH" = "master" -a "$TRAVIS_PULL_REQUEST" = "false" ]; then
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -29,14 +29,13 @@ before_cache:
...
-    fi
-
 jobs:
   include:
     - name: "Validate and verify modules"
       script: sbt ++$TRAVIS_SCALA_VERSION ";validate; verify" && bash <(curl -s https://codecov.io/bash)
     - name: "Run API and portal functional tests"
       script: ./bin/func-test-portal.sh && ./bin/func-test-api-travis.sh
+    - stage: Deploy
+      if: branch = master AND type = push
```
mitruly: I'm assuming `type = push` is also valid when merging to master here. Noticed that a new branch off of master was created, so If there are any specific changes to how docs need to be added to the codebase to publish, then we would also need to update the appropriate documentation.

Didn't see any second build stage https://travis-ci.org/vinyldns/vinyldns/builds/416999659, which is good, so looks like this would only trigger during merge.

britneywright: The branch is because the deployment can't be tested from a fork. First i tested deploying then i added the if statement. I got the if statement from here https://github.com/travis-ci/travis-ci/issues/8767#issuecomment-344751779. The fork part isn't needed since forks wouldn't be able to deploy anyway because the environment variable would fail.

britneywright: Once this is merged and i confirm it works I'll delete the travis-configuration branch.

PR #120, opened 2018-08-17
update gitter

nimaeskandary:
https://gitter.im/vinyldns/ will let people see all the gitters including https://gitter.im/vinyldns/lobby


mitruly:
We have a similar link in the microsite (see https://www.vinyldns.io/; **OPEN CHAT** button on the bottom-right).

If we are updating the link from the lobby, we should also update `micrositeGitterChannelUrl := "vinyldns/Lobby"` in `build.sbt`.

nimaeskandary:
@mitruly good catch, i was only searching for "gitter.im"

nimaeskandary:
@mitruly Yeah has to stay a specific room, will keep it as lobby then

PR #121, opened 2018-08-17
WIP: Pluggable repos

rebstar6:
V1 for this work. This PR is meant to show how the new config would be structured, and the validations/load process. The specific dynamic loading process for mysql vs dynamo is a TODO (probably followup PR).

Still need to test this one, but WIPing so people can react to the approach.


modules/api/src/main/scala/vinyldns/api/repository/mysql/MySqlDataStore.scala
```diff
@@ -0,0 +1,48 @@
...
+  GroupRepository,
+  MembershipRepository,
+  UserRepository
+}
+import vinyldns.api.domain.record.{RecordChangeRepository, RecordSetRepository}
+import vinyldns.api.domain.zone.{ZoneChangeRepository, ZoneRepository}
+import vinyldns.api.repository.DataStore
+
+// TODO placeholder for mysql loader
+class MySqlDataStore(config: Config) extends DataStore {
```
rebstar6: this and DynamoDbDataStore are just placeholders for now. In followup, need to implement them to read the config settings/repos and connect accordingly.

modules/api/src/main/scala/vinyldns/api/repository/dynamodb/DynamoDbDataStore.scala
```diff
@@ -0,0 +1,41 @@
...
+  GroupRepository,
+  MembershipRepository,
+  UserRepository
+}
+import vinyldns.api.domain.record.{RecordChangeRepository, RecordSetRepository}
+import vinyldns.api.domain.zone.{ZoneChangeRepository, ZoneRepository}
+import vinyldns.api.repository.DataStore
+
+// TODO add config parameter, load based on that
+class DynamoDbDataStore() extends DataStore {
```
rebstar6: this and MySqlDataStore are just placeholders for now. In followup, need to implement them to read the config settings/repos and connect accordingly.

modules/api/src/main/resources/reference.conf
```diff
@@ -33,24 +33,90 @@ vinyldns {
...
+    },
+    {
+      type = "vinyldns.api.repository.dynamodb.DynamoDbDataStore"
+
+      settings {
+        key = "vinyldnsTest"
+        secret = "notNeededForDynamoDbLocal"
+        endpoint = "http://127.0.0.1:19000"
+        region = "us-east-1" # note: we are always in us-east-1, but this can be overridden
+      }
```
nimaeskandary: if the repository settings override the defaults can you add `provisionedReads = 30` and `provisionedWrites = 30` to `settings`?

docker/api/docker.conf
```diff
@@ -43,6 +43,95 @@ vinyldns {
...
+    {
+      type = "vinyldns.api.repository.mysql.MySqlDataStore"
+
+      settings {
+        # JDBC Settings, these are all values in scalikejdbc-config, not our own
+        # these must be overridden to use MYSQL for production use
+        # assumes a docker or mysql instance running locally
+        ## TODO will de-nest this, keeping in "default" just to push off changes
+        name = "vinyldns"
+        default {
```
pauljamescleary: since we are changing config, no need for this `default` namespace.  That was because scalikjdbc could support different environments in the same config.  We really don't want to do that.

modules/api/src/main/scala/vinyldns/api/repository/DataStore.scala
```diff
@@ -0,0 +1,107 @@
...
+    repos match {
+      case h :: Nil => h.validNel
+      case Nil => s"Invalid config! Must have one repo of type $name".invalidNel[A]
+      case _ =>
+        s"Invalid config! May not have more than one repo of type $name"
+          .invalidNel[A]
+    }
+}
+
+case class DataAccessor(
```
pauljamescleary: I'd have to do a little research, but I reserve `case class` for "data classes".  Things like User, Person, Zone, etc.  Because you get things like `.equals` and `.toString` overrides that make sense when dealing with data types like `User`.

Here, I don't like the use of a `case` class.  I would use a normal `class` as the older, and just add `val` to the constructor parameters.

modules/api/src/main/scala/vinyldns/api/repository/DataStore.scala
```diff
@@ -0,0 +1,107 @@
...
+    for {
+      className <- IO(config.getString("type"))
+      dataStore <- IO(className match {
+        case "vinyldns.api.repository.mysql.MySqlDataStore" => new MySqlDataStore(config)
+        case _ => new DynamoDbDataStore()
+      })
+    } yield dataStore
+
+  def loadAll(configs: List[Config]): IO[DataAccessor] = {
+    val dataStores = configs.map(load).parSequence
```
pauljamescleary: we need to handle the condition where no data stores are found.

It begs the question, should `loadAll` just take the "main" config (vinyldns namespace)?  This way, we do not require callers to know how to get to `vinyldns.data-stores`

modules/api/src/main/resources/reference.conf
```diff
@@ -33,24 +33,90 @@ vinyldns {
...
+
+      repositories {
+        zoneChange {
```
pauljamescleary: I'm going to recommend that we move to "typesafe config convention" while we have the change.  We should not use camel case, but rather "dashes"

`zoneChange` becomes `zone-change` and so on.

This is "expected" / default in PureConfig as well

pauljamescleary:
Another thing I looked at is the possibility of using shapeless `HList`.  It allows you to build a strongly typed Tuple.  While I think that approach has value, it would introduce a very new shiny thing to the codebase and the disruption wouldn't exceed the value IMO.

docker/api/docker.conf
rebstar6: leaving in for this PR so i dont have to update all the refs, there will be separate PRs to implemented mysql and dynamo to be loaded properly, and those will have that update

modules/api/src/main/resources/reference.conf
rebstar6: thats fine, will also get pushed to the next PR

PR #122, opened 2018-08-17
add README to docs

### LINKED ISSUE #61: Readme for gh-pages branch
Write a readme for the gh-pages branch explaining how it's deployed to and how to run it locally, or link to instructions on how to run it locally if those are included in the main VinylDNS documentation.

britneywright:
This should add the README to the root of the gh-pages branch after it's merged into master and the doc site is published to gh-pages. It will also be viewable at https://www.vinyldns.io/readme at that time although we aren't linking to it in the documentation site itself. closes #61

### DIFF
--- a/modules/docs/src/main/tut/README.md
+++ b/modules/docs/src/main/tut/README.md
@@ -1,0 +1,28 @@
+# VinylDNS documentation site
+
+https://www.vinyldns.io/
+
+## Publication
+The VinylDNS documentation is published to the `gh-pages` branch after each successful master branch build. This is configured through Travis CI.
+
+## Documentation Structure
+- The documentation site is built with the [sbt-microsites](https://47deg.github.io/sbt-microsites/) plugin.
+- The [docs module](https://github.com/vinyldns/vinyldns/tree/master/modules/docs/src/main) contains most content for the documentation site.
+- The [Contributing Guide](https://www.vinyldns.io/contributing.html) is the [CONTRIBUTING.md](https://github.com/vinyldns/vinyldns/blob/master/CONTRIBUTING.md) file at the root of the VinylDNS project.
+- The sbt-microsite configuration is in [build.sbt](https://github.com/vinyldns/vinyldns/blob/master/build.sbt) in the root of the VinylDNS project.
+
+## Build Locally
+In the terminal enter:
+1.  `sbt`
+1. `project docs`
+1. `makeMicrosite`
+
+In a separate tab enter:
+1. `cd modules/docs/target/site`
+1. `jekyll serve`
+1. View in the browser at http://localhost:4000/
+
+Tips:
+* If you make changes to the documentation you'll need to run `makeMicrosite` again.
+You don't need to restart Jekyll.
+* If you delete files you may need to stop Jekyll and delete the target directory before running `makeMicrosite` again to see the site as expected locally.
--- a/modules/docs/src/main/tut/README.md
+++ b/modules/docs/src/main/tut/README.md
@@ -7,7 +7,12 @@ The VinylDNS documentation is published to the `gh-pages` branch after each succ
 
 ## Documentation Structure
 - The documentation site is built with the [sbt-microsites](https://47deg.github.io/sbt-microsites/) plugin.
-- The [docs module](https://github.com/vinyldns/vinyldns/tree/master/modules/docs/src/main) contains most content for the documentation site.
+- The [docs module](https://github.com/vinyldns/vinyldns/tree/master/modules/docs/src/main) contains most content for the documentation site:
+  - The text content is in the [tut](https://github.com/vinyldns/vinyldns/tree/master/modules/docs/src/main/tut/) directory
+  - The primary menu is built through setting a position value in the linked file or in [build.sbt](https://github.com/vinyldns/vinyldns/blob/master/build.sbt) if the target link is not in the docs module.
+  - The sidebar menu is maintained in the [menu.yml](https://github.com/vinyldns/vinyldns/tree/master/modules/docs/src/main/resources/microsite/data/menu.yml)
+  - Images are stored in the [img](https://github.com/vinyldns/vinyldns/tree/master/modules/docs/src/main/tut/img/) directory.
+  - Custom CSS is stored in the [custom.css](https://github.com/vinyldns/vinyldns/tree/master/modules/docs/src/main/resources/microsite/css/custom.css) file.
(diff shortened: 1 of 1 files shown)


modules/docs/src/main/tut/README.md
```diff
@@ -0,0 +1,28 @@
+# VinylDNS documentation site
+
+https://www.vinyldns.io/
+
+## Publication
+The VinylDNS documentation is published to the `gh-pages` branch after each successful master branch build. This is configured through Travis CI.
+
+## Documentation Structure
+- The documentation site is built with the [sbt-microsites](https://47deg.github.io/sbt-microsites/) plugin.
+- The [docs module](https://github.com/vinyldns/vinyldns/tree/master/modules/docs/src/main) contains most content for the documentation site.
```
mitruly: Might be useful to indicate that the `menu.yml` file is used to configure the sections and links for the microsite.

modules/docs/src/main/tut/README.md
```diff
@@ -0,0 +1,28 @@
+# VinylDNS documentation site
+
+https://www.vinyldns.io/
+
+## Publication
+The VinylDNS documentation is published to the `gh-pages` branch after each successful master branch build. This is configured through Travis CI.
+
+## Documentation Structure
+- The documentation site is built with the [sbt-microsites](https://47deg.github.io/sbt-microsites/) plugin.
```
mitruly: We can combine this sentence with the configurations info or at least group them together.

modules/docs/src/main/tut/README.md
```diff
@@ -0,0 +1,28 @@
...
+- The [docs module](https://github.com/vinyldns/vinyldns/tree/master/modules/docs/src/main) contains most content for the documentation site.
+- The [Contributing Guide](https://www.vinyldns.io/contributing.html) is the [CONTRIBUTING.md](https://github.com/vinyldns/vinyldns/blob/master/CONTRIBUTING.md) file at the root of the VinylDNS project.
+- The sbt-microsite configuration is in [build.sbt](https://github.com/vinyldns/vinyldns/blob/master/build.sbt) in the root of the VinylDNS project.
+
+## Build Locally
+In the terminal enter:
+1.  `sbt`
```
mitruly: This can just be:
1. `sbt ";project docs ;makeMicrosite"`

or:
1. `sbt`
2. `";project docs ;makeMicrosite"

I don't think we need to go into too much detail about `sbt`, but minimizing steps where possible makes sense.

modules/docs/src/main/tut/README.md
```diff
@@ -0,0 +1,28 @@
...
+1. `project docs`
+1. `makeMicrosite`
+
+In a separate tab enter:
```
mitruly: I'm alright with this, but this doesn't have to be done in a separate tab (especially if we're generating the microsite with `sbt ";project docs ;makeMicrosite"`).

I'd prefer if we had building and serving the microsite in two separate sections, and modify the tip to say that they can be done in different tabs to have Jekyll continually running.

modules/docs/src/main/tut/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Documentation Structure
+- The documentation site is built with the [sbt-microsites](https://47deg.github.io/sbt-microsites/) plugin.
+- The [docs module](https://github.com/vinyldns/vinyldns/tree/master/modules/docs/src/main) contains most content for the documentation site:
+  - The text content is in the [tut](https://github.com/vinyldns/vinyldns/tree/master/modules/docs/src/main/tut/) directory
+  - The primary menu is built through setting a position value in the linked file or in [build.sbt](https://github.com/vinyldns/vinyldns/blob/master/build.sbt) if the target link is not in the docs module.
+  - The sidebar menu is maintained in the [menu.yml](https://github.com/vinyldns/vinyldns/tree/master/modules/docs/src/main/resources/microsite/data/menu.yml)
+  - Images are stored in the [img](https://github.com/vinyldns/vinyldns/tree/master/modules/docs/src/main/tut/img/) directory.
+  - Custom CSS is stored in the [custom.css](https://github.com/vinyldns/vinyldns/tree/master/modules/docs/src/main/resources/microsite/css/custom.css) file.
```
britneywright: This link doesn't actually exist yet, but it will when the portal guide is merged in.

modules/docs/src/main/tut/README.md
britneywright: I wrote it for building while working on the docs. the one line is convenient if you only need to build the site once, but not when you have to build after every change you make.

modules/docs/src/main/tut/README.md
```diff
@@ -0,0 +1,34 @@
+# VinylDNS documentation site
+
+https://www.vinyldns.io/
+
+## Publication
+The VinylDNS documentation is published to the `gh-pages` branch after each successful master branch build. This is configured through Travis CI.
```
rebstar6: maybe link to https://github.com/vinyldns/vinyldns/tree/gh-pages on `gh-pages`

not crucial

modules/docs/src/main/tut/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Documentation Structure
+- The documentation site is built with the [sbt-microsites](https://47deg.github.io/sbt-microsites/) plugin.
+- The [docs module](https://github.com/vinyldns/vinyldns/tree/master/modules/docs/src/main) contains most content for the documentation site:
+  - The text content is in the [tut](https://github.com/vinyldns/vinyldns/tree/master/modules/docs/src/main/tut/) directory
+  - The primary menu is built through setting a position value in the linked file or in [build.sbt](https://github.com/vinyldns/vinyldns/blob/master/build.sbt) if the target link is not in the docs module.
```
rebstar6: ...or in the docSettings within build.sbt...

build.sbt has a lot, just need a little info on where for ^

rebstar6: `The primary menu is built through setting a position value in the linked file ` didnt make sense to me on just reading it. Probably just add a "see https://github.com/vinyldns/vinyldns/blob/master/modules/docs/src/main/tut/index.md for an example primary menu item"

modules/docs/src/main/tut/README.md
britneywright: since you're in the gh-pages branch to read the readme I don't think the link is necessary. If we explicitly add the Readme to the doc site then I could see the link being useful.

rebstar6: oh hah, good call

PR #132, opened 2018-08-22
Pluggable repos base classes

rebstar6:
splitting up https://github.com/vinyldns/vinyldns/pull/129 because its giant

This PR brings in the DataStoreProvider and related classes, though does not yet hook them into the application

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/repository/DataStore.scala
+++ b/modules/api/src/main/scala/vinyldns/api/repository/DataStore.scala
@@ -24,16 +24,16 @@ import vinyldns.api.domain.membership.{GroupRepository, UserRepository}
 import vinyldns.api.domain.zone.ZoneRepository
 import vinyldns.api.repository.RepositoryName.RepositoryName
 
-case class DataStore(
-    userRepository: Option[UserRepository] = None,
-    groupRepository: Option[GroupRepository] = None,
-    membershipRepository: Option[MembershipRepository] = None,
-    groupChangeRepository: Option[GroupChangeRepository] = None,
-    recordSetRepository: Option[RecordSetRepository] = None,
-    recordChangeRepository: Option[RecordChangeRepository] = None,
-    zoneChangeRepository: Option[ZoneChangeRepository] = None,
-    zoneRepository: Option[ZoneRepository] = None,
-    batchChangeRepository: Option[BatchChangeRepository] = None
+class DataStore(
+    val userRepository: Option[UserRepository] = None,
+    val groupRepository: Option[GroupRepository] = None,
+    val membershipRepository: Option[MembershipRepository] = None,
+    val groupChangeRepository: Option[GroupChangeRepository] = None,
+    val recordSetRepository: Option[RecordSetRepository] = None,
+    val recordChangeRepository: Option[RecordChangeRepository] = None,
+    val zoneChangeRepository: Option[ZoneChangeRepository] = None,
+    val zoneRepository: Option[ZoneRepository] = None,
+    val batchChangeRepository: Option[BatchChangeRepository] = None
 ) {
   def asMap: Map[RepositoryName, Repository] =
     List(
@@ -49,7 +49,7 @@ case class DataStore(
     ).flatten.toMap
 }
 
-case class DataAccessor(
+final case class DataAccessor(
     userRepository: UserRepository,
     groupRepository: GroupRepository,
     membershipRepository: MembershipRepository,
@@ -58,7 +58,6 @@ case class DataAccessor(
     recordChangeRepository: RecordChangeRepository,
     zoneChangeRepository: ZoneChangeRepository,
     zoneRepository: ZoneRepository,
-    batchChangeRepository: BatchChangeRepository
-)
+    batchChangeRepository: BatchChangeRepository)
 
 case class DataStoreStartupError(msg: String) extends Throwable(msg)
--- a/modules/api/src/main/scala/vinyldns/api/repository/DataStore.scala
+++ b/modules/api/src/main/scala/vinyldns/api/repository/DataStore.scala
@@ -47,4 +47,4 @@ case class DataAccessor(
     batchChangeRepository: BatchChangeRepository
 )
 
-case class DataStoreInitializationError(msg: String) extends Throwable(msg)
+case class DataStoreStartupError(msg: String) extends Throwable(msg)
--- a/modules/api/src/main/scala/vinyldns/api/repository/DataStore.scala
+++ b/modules/api/src/main/scala/vinyldns/api/repository/DataStore.scala
@@ -27,15 +27,15 @@ import vinyldns.api.repository.RepositoryName.RepositoryName
 import scala.reflect.ClassTag
 
 class DataStore(
-    val userRepository: Option[UserRepository] = None,
-    val groupRepository: Option[GroupRepository] = None,
-    val membershipRepository: Option[MembershipRepository] = None,
-    val groupChangeRepository: Option[GroupChangeRepository] = None,
-    val recordSetRepository: Option[RecordSetRepository] = None,
-    val recordChangeRepository: Option[RecordChangeRepository] = None,
-    val zoneChangeRepository: Option[ZoneChangeRepository] = None,
-    val zoneRepository: Option[ZoneRepository] = None,
-    val batchChangeRepository: Option[BatchChangeRepository] = None
+    userRepository: Option[UserRepository] = None,
+    groupRepository: Option[GroupRepository] = None,
+    membershipRepository: Option[MembershipRepository] = None,
+    groupChangeRepository: Option[GroupChangeRepository] = None,
+    recordSetRepository: Option[RecordSetRepository] = None,
+    recordChangeRepository: Option[RecordChangeRepository] = None,
+    zoneChangeRepository: Option[ZoneChangeRepository] = None,
+    zoneRepository: Option[ZoneRepository] = None,
+    batchChangeRepository: Option[BatchChangeRepository] = None
 ) {
 
   lazy val dataStoreMap: Map[RepositoryName, Repository] =
--- a/modules/api/src/main/scala/vinyldns/api/repository/DataStore.scala
+++ b/modules/api/src/main/scala/vinyldns/api/repository/DataStore.scala
@@ -37,15 +37,15 @@ class DataStore(
 ) {
   def asMap: Map[RepositoryName, Repository] =
     List(
-      userRepository.map(x => RepositoryName.user -> x),
-      groupRepository.map(x => RepositoryName.group -> x),
-      membershipRepository.map(x => RepositoryName.membership -> x),
-      groupChangeRepository.map(x => RepositoryName.groupChange -> x),
-      recordSetRepository.map(x => RepositoryName.recordSet -> x),
-      recordChangeRepository.map(x => RepositoryName.recordChange -> x),
-      zoneChangeRepository.map(x => RepositoryName.zoneChange -> x),
-      zoneRepository.map(x => RepositoryName.zone -> x),
-      batchChangeRepository.map(x => RepositoryName.batchChange -> x)
+      userRepository.map(RepositoryName.user -> _),
+      groupRepository.map(RepositoryName.group -> _),
+      membershipRepository.map(RepositoryName.membership -> _),
+      groupChangeRepository.map(RepositoryName.groupChange -> _),
+      recordSetRepository.map(RepositoryName.recordSet -> _),
+      recordChangeRepository.map(RepositoryName.recordChange -> _),
+      zoneChangeRepository.map(RepositoryName.zoneChange -> _),
+      zoneRepository.map(RepositoryName.zone -> _),
+      batchChangeRepository.map(RepositoryName.batchChange -> _)
     ).flatten.toMap
 }
 
--- a/modules/api/src/main/scala/vinyldns/api/repository/DataStore.scala
+++ b/modules/api/src/main/scala/vinyldns/api/repository/DataStore.scala
@@ -35,7 +35,8 @@ class DataStore(
     val zoneRepository: Option[ZoneRepository] = None,
     val batchChangeRepository: Option[BatchChangeRepository] = None
 ) {
-  def asMap: Map[RepositoryName, Repository] =
+
+  lazy val dataStoreMap: Map[RepositoryName, Repository] =
     List(
       userRepository.map(RepositoryName.user -> _),
       groupRepository.map(RepositoryName.group -> _),
@@ -47,6 +48,8 @@ class DataStore(
       zoneRepository.map(RepositoryName.zone -> _),
       batchChangeRepository.map(RepositoryName.batchChange -> _)
     ).flatten.toMap
+
+  def asMap: Map[RepositoryName, Repository] = dataStoreMap
 }
 
 final case class DataAccessor(
--- a/modules/api/src/main/scala/vinyldns/api/repository/DataStore.scala
+++ b/modules/api/src/main/scala/vinyldns/api/repository/DataStore.scala
@@ -22,6 +22,7 @@ import vinyldns.api.domain.record.{RecordChangeRepository, RecordSetRepository}
 import vinyldns.api.domain.zone.ZoneChangeRepository
 import vinyldns.api.domain.membership.{GroupRepository, UserRepository}
 import vinyldns.api.domain.zone.ZoneRepository
+import vinyldns.api.repository.RepositoryName.RepositoryName
 
 case class DataStore(
     userRepository: Option[UserRepository] = None,
@@ -33,7 +34,20 @@ case class DataStore(
     zoneChangeRepository: Option[ZoneChangeRepository] = None,
     zoneRepository: Option[ZoneRepository] = None,
     batchChangeRepository: Option[BatchChangeRepository] = None
-)
+) {
+  def asMap: Map[RepositoryName, Repository] =
+    List(
+      userRepository.map(x => RepositoryName.user -> x),
+      groupRepository.map(x => RepositoryName.group -> x),
+      membershipRepository.map(x => RepositoryName.membership -> x),
+      groupChangeRepository.map(x => RepositoryName.groupChange -> x),
+      recordSetRepository.map(x => RepositoryName.recordSet -> x),
+      recordChangeRepository.map(x => RepositoryName.recordChange -> x),
+      zoneChangeRepository.map(x => RepositoryName.zoneChange -> x),
+      zoneRepository.map(x => RepositoryName.zone -> x),
+      batchChangeRepository.map(x => RepositoryName.batchChange -> x)
+    ).flatten.toMap
+}
 
 case class DataAccessor(
     userRepository: UserRepository,
--- a/modules/api/src/main/scala/vinyldns/api/repository/DataStore.scala
+++ b/modules/api/src/main/scala/vinyldns/api/repository/DataStore.scala
@@ -24,6 +24,8 @@ import vinyldns.api.domain.membership.{GroupRepository, UserRepository}
 import vinyldns.api.domain.zone.ZoneRepository
 import vinyldns.api.repository.RepositoryName.RepositoryName
 
+import scala.reflect.ClassTag
+
 class DataStore(
     val userRepository: Option[UserRepository] = None,
     val groupRepository: Option[GroupRepository] = None,
@@ -49,7 +51,13 @@ class DataStore(
       batchChangeRepository.map(RepositoryName.batchChange -> _)
     ).flatten.toMap
 
-  def asMap: Map[RepositoryName, Repository] = dataStoreMap
+  def keys: Set[RepositoryName] = dataStoreMap.keySet
+
+  def get[A <: Repository: ClassTag](name: RepositoryName): Option[A] =
+    dataStoreMap.get(name).flatMap {
+      case a: A => Some(a)
+      case _ => None
+    }
 }
 
 final case class DataAccessor(
--- a/modules/api/src/main/scala/vinyldns/api/repository/DataStore.scala
+++ b/modules/api/src/main/scala/vinyldns/api/repository/DataStore.scala
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2018 Comcast Cable Communications Management, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package vinyldns.api.repository
+
+import vinyldns.api.domain.batch.BatchChangeRepository
+import vinyldns.api.domain.membership.{GroupChangeRepository, MembershipRepository}
+import vinyldns.api.domain.record.{RecordChangeRepository, RecordSetRepository}
+import vinyldns.api.domain.zone.ZoneChangeRepository
+import vinyldns.api.domain.membership.{GroupRepository, UserRepository}
+import vinyldns.api.domain.zone.ZoneRepository
+
+case class DataStore(
+    userRepository: Option[UserRepository] = None,
+    groupRepository: Option[GroupRepository] = None,
+    membershipRepository: Option[MembershipRepository] = None,
+    groupChangeRepository: Option[GroupChangeRepository] = None,
+    recordSetRepository: Option[RecordSetRepository] = None,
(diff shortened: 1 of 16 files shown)


modules/api/src/main/scala/vinyldns/api/repository/DataStoreLoader.scala
```diff
@@ -0,0 +1,108 @@
...
+package vinyldns.api.repository
+
+import cats.data._
+import cats.effect.IO
+import cats.implicits._
+import com.typesafe.config.{Config, ConfigException}
+
+class DataStoreLoader {
+
+  def loadAll(configs: List[DataStoreConfig]): IO[DataAccessor] =
```
rebstar6: this is the main method here - boot will eventually call `loadAll`

modules/api/src/main/scala/vinyldns/api/repository/DataStoreProvider.scala
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package vinyldns.api.repository
+
+import cats.effect.IO
+
+trait DataStoreProvider {
```
rebstar6: to be implemented once for each datastore - so we will have a `MySqlDataStoreProvider` and a `DynamoDbDataStoreProvider` implementation in a coming PR

modules/api/src/test/scala/vinyldns/api/repository/DataStoreLoaderSpec.scala
```diff
@@ -0,0 +1,252 @@
...
+  )
+
+  val goodConfig = DataStoreConfig(
+    "vinyldns.api.repository.MockDataStoreProvider",
+    placeholderConfig,
+    allEnabledReposConfig)
+
+  val baseLoader = new DataStoreLoader()
+
+  class LoaderWithOverrides(
```
rebstar6: did this class overrides thing instead of mocking. seems to work ok :shrugs:

modules/api/src/test/scala/vinyldns/api/repository/MockDataStoreProvider.scala
```diff
@@ -0,0 +1,66 @@
...
+import vinyldns.api.domain.batch.BatchChangeRepository
+import vinyldns.api.domain.membership.{
+  GroupChangeRepository,
+  GroupRepository,
+  MembershipRepository,
+  UserRepository
+}
+import vinyldns.api.domain.record.{RecordChangeRepository, RecordSetRepository}
+import vinyldns.api.domain.zone.{ZoneChangeRepository, ZoneRepository}
+
```
rebstar6: needed a few classes defined to test reading the class name from config

modules/api/src/main/scala/vinyldns/api/repository/DataStore.scala
```diff
@@ -0,0 +1,50 @@
...
+
+package vinyldns.api.repository
+
+import vinyldns.api.domain.batch.BatchChangeRepository
+import vinyldns.api.domain.membership.{GroupChangeRepository, MembershipRepository}
+import vinyldns.api.domain.record.{RecordChangeRepository, RecordSetRepository}
+import vinyldns.api.domain.zone.ZoneChangeRepository
+import vinyldns.api.domain.membership.{GroupRepository, UserRepository}
+import vinyldns.api.domain.zone.ZoneRepository
+
```
rebstar6: should I skip coverage on this file? just defines classes and an error

pauljamescleary: make all case classes `final`

modules/api/src/main/scala/vinyldns/api/repository/DataStoreConfig.scala
```diff
@@ -0,0 +1,46 @@
...
+    group: Option[Config],
+    membership: Option[Config],
+    groupChange: Option[Config],
+    recordSet: Option[Config],
+    recordChange: Option[Config],
+    zoneChange: Option[Config],
+    zone: Option[Config],
+    batchChange: Option[Config]
+) {
+  def containsActiveRepo: Boolean =
```
pauljamescleary: what is this used for?  I know we mentioned enabling things, but I thought we were no longer doing that.

I think this can go away

modules/api/src/main/scala/vinyldns/api/repository/DataStoreLoader.scala
```diff
@@ -0,0 +1,111 @@
...
+ */
+
+package vinyldns.api.repository
+
+import cats.data._
+import cats.effect.IO
+import cats.implicits._
+import com.typesafe.config.Config
+
+class DataStoreLoader {
```
pauljamescleary: why is this a class?

modules/api/src/main/scala/vinyldns/api/repository/DataStoreLoader.scala
```diff
@@ -0,0 +1,111 @@
...
+        val errorString = errors.toList.mkString(", ")
+        DataStoreInitializationError(s"Config error: $errorString")
+      }
+  }
+
+  def generateAccessor(
+      dataStores: List[DataStore]): Either[DataStoreInitializationError, DataAccessor] = {
+    // Note: headOption is fine here only because we've already validated the config has a single
+    // instance defined for each repo across datastores
+    val accessor = for {
```
pauljamescleary: We have to think through this.  Even though we validated that there was a "config" section up front, there are no guarantees based on the `DataStore` interface that we have one. The `DataStoreProvider.load` could fail to put that, as the only information we have is `Option[Repository]`.

We still need to validate after that we have all of the repositories.

modules/api/src/main/scala/vinyldns/api/repository/DataStoreLoader.scala
```diff
@@ -0,0 +1,111 @@
...
+        zoneRepo,
+        batchChangeRepo
+      )
+
+    Either.fromOption(
+      accessor,
+      DataStoreInitializationError("error pulling repositories from databases"))
+  }
+
+  private def listContainsSingleConfig(
```
pauljamescleary: avoid private def, make nested function if only used one place

modules/api/src/main/scala/vinyldns/api/repository/DataStoreConfig.scala
rebstar6: its basically a filter. i only pass through configs to load that contain at least one repo. probably not a crucial filter, but it means that if for some reason someone had mySql enabled, but none of those repos turned on, we wont even call the mySql load

modules/api/src/main/scala/vinyldns/api/repository/DataStoreLoader.scala
rebstar6: testing. i wanted to extend it.

otherwise it's an object, ill look into switching up testing today

modules/api/src/main/scala/vinyldns/api/repository/DataStoreLoader.scala
rebstar6: can also just double validate - 
- the empty list is not a problem for us, this returns an either, and will go left if the list is empty. 
- a single expected value showing up None is also not a problem. same thing, this either will return left
- the bad case here is if one of our loaders ignores the config and initializes something if should not have. Ill think about it and validate for that, since we cant guarantee it will work properly with other people's code

im team double validate on this only because its not worth even trying to setup the DBs in the case that the config is bad, so id rather fail sooner. will work on that.

modules/api/src/main/scala/vinyldns/api/repository/DataStore.scala
```diff
@@ -0,0 +1,50 @@
...
+case class DataStore(
```
pauljamescleary: Using `case class` is a bit of an issue.  `case class` forces the compiler to do a lot of work on your behalf.  The compiler will generate (among other things):

* `copy` function
* `productArity` functions
* `hashCode` function
* `equals` function
* `toString` function
* `unapply` function
* `companion object with an apply`

All that boilerplate is good if you are doing pattern matching, structural equality, want sane output with toString, or want to copy things in a bunch of places (forcing immutability).

If we have no intention of using those capabilities of case classes, use regular `class` instead with `val` in the constructor.

modules/api/src/main/scala/vinyldns/api/repository/DataStoreLoader.scala
pauljamescleary: Use `validated` instead of either in double validation.  I can imagine a scenario where someone writes their own data store provider but then upgrades to a later version of core with a new repository.  We want to know which repo has the issue.

I think there is an opportunity to extract using `Validated` to say "this list may have exactly one element"

modules/api/src/main/scala/vinyldns/api/repository/DataStore.scala
```diff
@@ -0,0 +1,63 @@
...
+      groupChangeRepository.map(x => RepositoryName.groupChange -> x),
+      recordSetRepository.map(x => RepositoryName.recordSet -> x),
+      recordChangeRepository.map(x => RepositoryName.recordChange -> x),
+      zoneChangeRepository.map(x => RepositoryName.zoneChange -> x),
+      zoneRepository.map(x => RepositoryName.zone -> x),
+      batchChangeRepository.map(x => RepositoryName.batchChange -> x)
+    ).flatten.toMap
+}
+
+final case class DataAccessor(
```
rebstar6: case class because of the mapn

modules/api/src/main/scala/vinyldns/api/repository/DataStoreConfig.scala
```diff
@@ -0,0 +1,56 @@
...
+    membership: Option[Config],
+    groupChange: Option[Config],
+    recordSet: Option[Config],
+    recordChange: Option[Config],
+    zoneChange: Option[Config],
+    zone: Option[Config],
+    batchChange: Option[Config]) {
+
+  def asMap: Map[RepositoryName, Config] =
+    List(
```
rebstar6: being able to access these values in a map lets me do `RepositoryName.values.map(v => config.get(v))`, which reduces some copy/paste elsewhere

modules/api/src/main/scala/vinyldns/api/repository/DataStore.scala
```diff
@@ -0,0 +1,63 @@
...
+    val membershipRepository: Option[MembershipRepository] = None,
+    val groupChangeRepository: Option[GroupChangeRepository] = None,
+    val recordSetRepository: Option[RecordSetRepository] = None,
+    val recordChangeRepository: Option[RecordChangeRepository] = None,
+    val zoneChangeRepository: Option[ZoneChangeRepository] = None,
+    val zoneRepository: Option[ZoneRepository] = None,
+    val batchChangeRepository: Option[BatchChangeRepository] = None
+) {
+  def asMap: Map[RepositoryName, Repository] =
+    List(
```
rebstar6: see https://github.com/vinyldns/vinyldns/pull/132/files#r212666003

modules/api/src/main/scala/vinyldns/api/repository/Repository.scala
```diff
@@ -0,0 +1,19 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package vinyldns.api.repository
+
+trait Repository
```
rebstar6: this is my favorite file

modules/api/src/main/scala/vinyldns/api/repository/Repository.scala
```diff
@@ -0,0 +1,26 @@
...
+
+object RepositoryName extends Enumeration {
```
rebstar6: urg, moving makes my comments disappear -

still an enum because the main purpose of this is to be able to do `RepositoryName.values.map(...)` -- if theres a better way with objects, ill change

modules/api/src/main/scala/vinyldns/api/repository/Repository.scala
rebstar6: alas, it has now grown

modules/api/src/main/scala/vinyldns/api/repository/DataStoreLoader.scala
```diff
@@ -0,0 +1,146 @@
...
+
+object DataStoreLoader {
+  def loadAll(configs: List[DataStoreConfig]): IO[DataAccessor] =
+    for {
+      activeConfigs <- IO.fromEither(getValidatedConfigs(configs))
+      dataStores <- activeConfigs.map(load).parSequence
+      accessor <- IO.fromEither(generateAccessor(activeConfigs, dataStores.toMap))
+    } yield accessor
+
+  def load(config: DataStoreConfig): IO[(String, DataStore)] =
```
rebstar6: `load` loads a single datastore (like mysql or dynamo)

modules/api/src/main/scala/vinyldns/api/repository/DataStoreLoader.scala
```diff
@@ -0,0 +1,146 @@
...
+    for {
+      className <- IO.pure(config.className)
+      provider <- IO(Class.forName(className).newInstance.asInstanceOf[DataStoreProvider])
+      dataStore <- provider.load(config)
+      _ <- IO.fromEither(validateLoadResponse(config, dataStore))
+    } yield (className, dataStore)
+
+  // Ensures that if a datastore is configured on, load returned it, and if configured off, load did not
```
rebstar6: urg this comment is wrong, meant to say that if a *repository* is on, it should be returned (and if off, not returned)

modules/api/src/main/scala/vinyldns/api/repository/DataStoreLoader.scala
```diff
@@ -0,0 +1,146 @@
...
+    val combinedValidations = validated.fold(().validNel)(_ |+| _)
+    combinedValidations.toEither
+      .map(_ => configs.filter(_.repositories.asMap.nonEmpty))
+      .leftMap { errors =>
+        val errorString = errors.toList.mkString(", ")
+        DataStoreStartupError(s"Config validation error: $errorString")
+      }
+  }
+
+  def generateAccessor(
```
rebstar6: a DataAccessor is the final return from loadAll - each datastore (mysql, dynamo, ...) loaded has options for each repo (user, group, ...). This function finds the datastore that *should* have each repo based on config, gets out those options, and maps into the accessor structure where nothing is an option.

modules/api/src/main/scala/vinyldns/api/repository/DataStoreConfig.scala
mitruly: I think that `Repositories` and `DataStore` are similar in that we want to construct them using optional values, but we mainly use them as a map. I was thinking that instead of having a function `asMap`, there would be a value like `configMap` inside that only gets constructed once on object creation so that we don't reconstruct the map (as I believe we are only loading these config values at boot).

One thing I was even thinking was that it might make sense to have the actual class be a wrapper around the map or a type alias, and we could have the companion object take in all of the `Option[Config]` and then create the `Map[RepositoryName, Config]` - not sure if this would be overkill, though.

modules/api/src/test/scala/vinyldns/api/repository/DataStoreLoaderSpec.scala
```diff
@@ -0,0 +1,267 @@
...
+  UserRepository
+}
+import vinyldns.api.domain.record.{RecordChangeRepository, RecordSetRepository}
+import vinyldns.api.domain.zone.{ZoneChangeRepository, ZoneRepository}
+
+import scala.collection.JavaConverters._
+
+class DataStoreLoaderSpec extends WordSpec with Matchers with MockitoSugar with EitherValues {
+
+  val placeholderConfig = ConfigFactory.parseMap(Map[String, String]().asJava)
```
mitruly: IntelliJ is warning about explicit typing here.

modules/api/src/main/scala/vinyldns/api/repository/DataStoreConfig.scala
rebstar6: yeah, i'll store the map so it's not reconstructed.

the reason for the class-ing here is it makes loading the config more simple (comes into play later - see WIP pr https://github.com/vinyldns/vinyldns/pull/129/files#diff-2e7c9ec63db3042b0dfdf256f652bdd5R33)

modules/api/src/main/scala/vinyldns/api/repository/DataStoreLoader.scala
```diff
@@ -0,0 +1,146 @@
...
+      configs: List[DataStoreConfig],
+      stringToStore: Map[String, DataStore]): Either[DataStoreStartupError, DataAccessor] = {
+
+    // maps RepositoryName to the DataStore that should hold that repo based on config
+    val reposByType = RepositoryName.values.toList.flatMap { repoName =>
```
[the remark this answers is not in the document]
rebstar6: weird, mine has a weird error last week. will try again, maybe it was something else

modules/api/src/main/scala/vinyldns/api/repository/DataStoreConfig.scala
rebstar6: on datastore the map may make more sense though, will look into it

rebstar6: oh, nvm, if its a map, everything would have to be downcasted, so cant do that in datastore either unfortunately

modules/api/src/main/scala/vinyldns/api/repository/DataStoreLoader.scala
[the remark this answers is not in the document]
rebstar6: hmm without it i get 
```
diverging implicit expansion for type scala.collection.generic.CanBuildFrom[vinyldns.api.repository.RepositoryName.ValueSet,(vinyldns.api.repository.RepositoryName.Value, vinyldns.api.repository.DataStore),That]
[error] starting with method orderingToOrdered in object Ordered
```

leaving in

modules/api/src/main/scala/vinyldns/api/repository/DataStore.scala
```diff
@@ -0,0 +1,76 @@
...
+      membershipRepository.map(RepositoryName.membership -> _),
+      groupChangeRepository.map(RepositoryName.groupChange -> _),
+      recordSetRepository.map(RepositoryName.recordSet -> _),
+      recordChangeRepository.map(RepositoryName.recordChange -> _),
+      zoneChangeRepository.map(RepositoryName.zoneChange -> _),
+      zoneRepository.map(RepositoryName.zone -> _),
+      batchChangeRepository.map(RepositoryName.batchChange -> _)
+    ).flatten.toMap
+
+  def asMap: Map[RepositoryName, Repository] = dataStoreMap
```
pauljamescleary: Unclear on why we need `asMap` as well as `lazy val dataStoreMap` here.  Can we just rename `dataStoreMap` to be `repositoryMap` instead and remove `asMap`?

modules/api/src/main/scala/vinyldns/api/repository/DataStore.scala
```diff
@@ -0,0 +1,76 @@
...
+
+  def keys: Set[RepositoryName] = dataStoreMap.keySet
+
+  def get[A](name: RepositoryName): Option[A] = {
```
pauljamescleary: 👍

modules/api/src/main/scala/vinyldns/api/repository/DataStoreLoader.scala
```diff
@@ -0,0 +1,137 @@
...
+    val combinedValidations = validated.fold(().validNel)(_ |+| _)
+    combinedValidations.toEither
+      .map(_ => configs.filter(_.repositories.nonEmpty))
+      .leftMap { errors =>
+        val errorString = errors.toList.mkString(", ")
+        DataStoreStartupError(s"Config validation error: $errorString")
+      }
+  }
+
+  def generateAccessor(configs: List[DataStoreConfig],
```
pauljamescleary: If you keep the config and datastore together, you wouldn't have to do the `className` lookup in the `Map[String, DataStore]`

Change the function to...

```scala
def generateAccessor(dataStores: List[(DataStoreConfig, DataStore)])
```

modules/api/src/main/scala/vinyldns/api/repository/DataStoreLoader.scala
```diff
@@ -0,0 +1,137 @@
...
+      } yield repo
+
+
+      repository match {
+        case Some(repo) => repo.validNel
+        case None => s"Could not pull repo $repoName from configured database".invalidNel
+      }
+    }
+
+    val accessor: ValidatedNel[String, DataAccessor] =
```
pauljamescleary: this is great

modules/api/src/main/scala/vinyldns/api/repository/Repository.scala
pauljamescleary: i'm not going to complain too loudly.  Enums are bad in general in Scala.  Enumeratum (https://github.com/lloydmeta/enumeratum) is preferable.  That said, I am ok with skipping it here and moving to Enumeratum in a separate PR.

modules/api/src/main/scala/vinyldns/api/repository/DataStoreLoader.scala
```diff
@@ -0,0 +1,144 @@
...
+      .leftMap { errors =>
+        val errorString = errors.toList.mkString(", ")
+        DataStoreStartupError(s"Config validation error: $errorString")
+      }
+  }
+
+  def generateAccessor(
+      configs: List[DataStoreConfig],
+      dataStores: Map[String, DataStore]): Either[DataStoreStartupError, DataAccessor] = {
+
```
pauljamescleary: if you pair the config with the data store it belongs to, then we can remove the `Map[String, DataStore]` which is stringly typed...

```
  def generateAccessor(
      configs: List[(DataStoreConfig, DataStore)]): Either[DataStoreStartupError, DataAccessor] = {

    def getRepoOf[A <: Repository: ClassTag](repoName: RepositoryName): ValidatedNel[String, A] = {
      val repository = for {
        (_, matchingDataStore) <- configs.find { case (c, _) => c.repositories.hasKey(repoName) }
        repo <- matchingDataStore.get[A](repoName)
      } yield repo

      repository match {
        case Some(repo) => repo.validNel
        case None => s"Could not pull repo $repoName from configured database".invalidNel
      }
    }
```

to do this, you would have to modify the `load` above as follows...

```scala
  def load(config: DataStoreConfig): IO[(DataStoreConfig, DataStore)] =
    for {
      className <- IO.pure(config.className)
      provider <- IO(Class.forName(className).newInstance.asInstanceOf[DataStoreProvider])
      dataStore <- provider.load(config)
      _ <- IO.fromEither(validateLoadResponse(config, dataStore))
    } yield (config, dataStore)
```

modules/api/src/main/scala/vinyldns/api/repository/DataStoreLoader.scala
```diff
@@ -0,0 +1,146 @@
...
+      dataStores <- activeConfigs.map(load).parSequence
+      accessor <- IO.fromEither(generateAccessor(activeConfigs, dataStores.toMap))
+    } yield accessor
+
+  def load(config: DataStoreConfig): IO[(String, DataStore)] =
+    for {
+      className <- IO.pure(config.className)
+      provider <- IO(Class.forName(className).newInstance.asInstanceOf[DataStoreProvider])
+      dataStore <- provider.load(config)
+      _ <- IO.fromEither(validateLoadResponse(config, dataStore))
```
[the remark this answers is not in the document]
pauljamescleary: I wonder if we need to even do all of this check.  Does it matter if a datastore has things that are not configured?  For sql databases, I would imagine that a datastoreprovider would just return blindly everything (simplest solution); whereas DynamoDB (because of cost) would be more discerning in what it would return.

Is it horrible to just have the data store loader "pluck" the repositories that are configured?  I think we have to validate that the datastore loaded does return all repositories configured, but do we care that they return a repo that is not configured (which I imagine would be the norm for SQL databases and even document databases)?

modules/api/src/main/scala/vinyldns/api/repository/DataStoreLoader.scala
```diff
@@ -0,0 +1,146 @@
...
+    val loadedNotConfigured = dataStoreMap.diff(configMap)
+    val configuredNotLoaded = configMap.diff(dataStoreMap)
+
+    (loadedNotConfigured.isEmpty, configuredNotLoaded.isEmpty) match {
+      case (true, true) => Right((): Unit)
+      case (false, true) =>
+        Left(
+          DataStoreStartupError(
+            s"""Loaded repos were configured off for ${config.className}: ${loadedNotConfigured
+              .mkString(", ")}"""))
```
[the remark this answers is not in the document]
rebstar6: It’s because of the mkstring. I’ll change

PR #134, opened 2018-08-23
turn on coverage in batch validations file

rebstar6:
oops.


rebstar6:
looks like a +0.01% increase in coverage. i'll take it?

PR #138, opened 2018-08-24
regenerate credentials

britneywright:


britneywright:
one thing i’m still trying to figure out: in the portal you can’t seem to regenerate your credentials multiple times in a row. You can do it once but then you need to refresh the page or go to a new page. I believe the session isn't updating with the regenerated access key on the angular side.

nimaeskandary:
@britneywright you you force refresh the page via javascript after getting a success response?

britneywright:
@nimaeskandary I was trying to do it in an "angular way" but since that didn't work I'll use plain javascript.

britneywright:
latest CodeCov report: 94.26% https://codecov.io/gh/vinyldns/vinyldns/pull/138/diff

Coverage increased by 2.99% in VinylDNS.scala

modules/portal/app/controllers/VinylDNS.scala
```diff
@@ -245,6 +245,33 @@ class VinylDNS @Inject()(
...
+        case Success(None) =>
+          throw new UnsupportedOperationException(s"Error - User account for $username not found")
+        case Failure(ex) => throw ex
+      }
+    }
+  }
+
+  def processRegenerate(oldAccount: UserAccount): Result = {
+    val account = oldAccount.regenerateCredentials()
+    userAccountAccessor.put(account)
```
pauljamescleary: `userAccountAccessor.put` is a `Try`.  This means that it could fail. `auditLogAccessor.log` can also fail. We have to handle the failure case here.

The simplest way to do that is to pull that stuff into a for comprehension

```
for {
  saved <- userAccountAccessor.put(account)
  _ <- auditLogAccessor.log( ... )
} yield {
  Logger.info(...)
  Status(200)...
}
```

modules/portal/app/controllers/VinylDNS.scala
```diff
@@ -245,6 +245,33 @@ class VinylDNS @Inject()(
     }
   }
 
+  def regenerateCreds(): Action[AnyContent] = Action.async { implicit request =>
+    withAuthenticatedUser { username =>
+      userAccountAccessor.get(username) match {
+        case Success(Some(account)) => Future(processRegenerate(account))
+        case Success(None) =>
+          throw new UnsupportedOperationException(s"Error - User account for $username not found")
```
pauljamescleary: I cannot believe we `throw` in the portal.  There is precedent so I won't ask for a change.  

Hopefully no one feels great about this and chalks it up to the portal needing a lot of love.

modules/portal/app/views/main.scala.html
```diff
@@ -112,6 +117,31 @@ <h4>Log Out?</h4>
...
+        <div class="fade in modal" id="mb-creds" data-sound="alert">
+            <div class="modal-dialog">
+                <div class="modal-content">
+                    <div class="modal-header">
+                        <button type="button" class="close" data-dismiss="modal">
+                            <span aria-hidden="true">×</span><span class="sr-only">Close</span></button>
+                        <h4>Regenerate Credentials?</h4>
+                    </div>
+                    <div class="modal-body">
+                        <p>If you regenerate your credentials you'll receive new credentials and your current credentials
```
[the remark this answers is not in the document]
pauljamescleary: Should we be a little stronger with the error message?  If they renew credentials, existing credentials will no longer be valid.  If you are using VinylDNS automation not through the portal, this will break that automation.

modules/portal/test/controllers/VinylDNSSpec.scala
```diff
@@ -124,6 +124,33 @@ class VinylDNSSpec extends Specification with Mockito {
...
 
+    ".regenerateCreds" should {
+      "change the access key and secret for the current user" in new WithApplication(app) {
+        val authenticator: LdapAuthenticator = mock[LdapAuthenticator]
+        val userAccessor: UserAccountAccessor = mock[UserAccountAccessor]
+        val config: Configuration = Configuration.load(Environment.simple())
+        val ws: WSClient = mock[WSClient]
+
+        authenticator.authenticate("frodo", "secondbreakfast").returns(Success(frodoDetails))
+        userAccessor.get(anyString).returns(Success(Some(frodoAccount)))
```
pauljamescleary: add unit tests for failure scenarios:

1. user not found
1. `userAccountAccessor.put` fails
1. `auditLogAccessor.log` fails

modules/docs/src/main/tut/faq.md
```diff
@@ -80,7 +80,7 @@ in the zone are in a list of approved name servers.  If your zone is hosted on
 name servers that are not in this list, you will not be able to use VinylDNS to manage your zone.
 
 ### 8. How do I get API credentials? <a id="8"></a>
-After logging in to the portal, click your username at the top right and select *Download API Credentials*.
+After logging in to the portal, click your username at the top right and select *Download API Credentials*. If you need new API credentials select *Regenerate Credentials*. This will invalidate your previous credentials.
```
pauljamescleary: add stronger verbiage (same as appears in the portal).  This will break automation.

pauljamescleary: probably should add the tooling piece up here

modules/portal/app/controllers/VinylDNS.scala
```diff
@@ -245,6 +245,49 @@ class VinylDNS @Inject()(
     }
   }
 
+  def regenerateCreds(): Action[AnyContent] = Action.async { implicit request =>
+    withAuthenticatedUser { username =>
+      Future
+        .fromTry {
```
pauljamescleary: make this `Future.fromTry(processRegenerate(username))` and you have my +1

britneywright:
updated @pauljamescleary!

PR #139, opened 2018-08-25
Move from Future to cats effect IO

pauljamescleary:
Most of the changes are find and replace.

The only code changes to note happened inside the JdbcRepositories.  Since IO is lazily evaluated, you need to obtain a connection _inside_ of IO.  Otherwise, you get "session already finished" errors.

We should look to use doobie instead of scalikejdbc as I believe you can do for comprehensions there.

In the tests, we heavily use `whenReady`.  The correct way to update the tests would be to do an await with timeout on `IO`.  Instead, in this PR, am using `IO.unsafeToFuture()` and leaving the tests as is.  Definitely need to open a separate PR to remove those "future based" tests.

This PR does _not_ address the Portal, as that will be a separate PR as well.

Some rules around `IO`...

* `IO.pure` should only capture already computed _values_.  For example, doing anything that could fail (including seemingly not bad things like `string.toInt` should _never_ be done inside `pure`.
* `IO(x)` is predominantly used, especially if the expression that is being lifted into `IO` could fail
* `IO.unit` nice handy shortcut to `IO.pure(())`


modules/api/src/it/scala/vinyldns/api/domain/record/RecordSetServiceIntegrationSpec.scala
```diff
@@ -186,6 +186,7 @@ class RecordSetServiceIntegrationSpec
       val originalRecord = testRecordSetService
         .getRecordSet(apexTestRecordA.id, apexTestRecordA.zoneId, auth)
         .value
+        .unsafeToFuture()
```
pauljamescleary: we will see this a lot in the unit and integration tests.  Gotta create a scalatest matcher for IO and remove `unsafeToFuture` everywhere

modules/api/src/main/scala/vinyldns/api/Interfaces.scala
```diff
@@ -16,99 +16,89 @@
...
-  def withTimeout[A](
-      theFuture: => Future[A],
-      duration: FiniteDuration,
-      error: Throwable,
-      scheduler: Scheduler)(implicit ec: ExecutionContext): Result[A] = result[A] {
-    val timeOut = after(duration = duration, using = scheduler)(Future.failed(error))
-
-    Future.firstCompletedOf(Seq(theFuture, timeOut))
+  def withTimeout[A](theIo: => IO[A], duration: FiniteDuration, error: Throwable): Result[A] = {
+    val timeOut = IO.sleep(duration) *> IO(error)
```
pauljamescleary: `*>` operator same as `.flatMap(_ => xxx)`

The race here is neat.  Timeout is on the _left_ (first position in the race), what we want to happen is on the _right_

We have to `handleError` in the event that `theIo` blows up, we want to force it to EitherT

modules/api/src/main/scala/vinyldns/api/Interfaces.scala
```diff
@@ -16,99 +16,89 @@
...
     // TODO: Can use OptionT here instead
-    def orFail(ifNone: => Throwable): Future[Either[Throwable, A]] = fut.map {
+    def orFail(ifNone: => Throwable): IO[Either[Throwable, A]] = fut.map {
       case Some(a) => Right(a)
       case None => Left(ifNone)
     }
   }
 
   /* Enhances any value to easily lift the class to a Result */
-  implicit class AnyResultImprovements[A](a: A)(implicit ec: ExecutionContext) {
```
pauljamescleary: we don't need to pass around execution contexts everywhere!!!  🎉

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -126,15 +125,17 @@ class BatchChangeService(
...
     // TODO - this implementation may be problematic with many changes. Need to perf test/later change DB
     val uniqueGets = changes.getValid.map(change => (change.zone.id, change.recordName)).toSet
 
-    val allFutures = uniqueGets.map {
+    val allIO = uniqueGets.map {
       case (zoneId, rsName) => recordSetRepository.getRecordSetsByName(zoneId, rsName)
     }
 
-    Future.sequence(allFutures).map(lst => ExistingRecordSets(lst.toList.flatten))
+    val allSeq: IO[List[List[RecordSet]]] = allIO.toList.sequence
```
pauljamescleary: same as before, but with `IO` we had to give type hints

modules/api/src/main/scala/vinyldns/api/engine/DnsConnector.scala
```diff
@@ -1,43 +0,0 @@
-/*
```
pauljamescleary: this was the `IO --> Future` Shim we can delete now that the `DnsConnection` talks `IO`

modules/api/src/main/scala/vinyldns/api/engine/ZoneSyncHandler.scala
```diff
@@ -35,68 +34,62 @@ object ZoneSyncHandler extends DnsConversions with Monitored {
...
-                vinyldnsLoader(zone, recordSetRepository).load())
-              val recordSetChanges =
-                for {
-                  dnsZoneView <- dnsView
-                  vinyldnsZoneView <- vinyldnsView
-                } yield vinyldnsZoneView.diff(dnsZoneView)
+          val dnsView = time(s"zone.sync.loadDnsView(${zoneChange.id})")(dnsLoader(zone).load())
+          val vinyldnsView = time(s"zone.sync.loadVinylDNSView(${zoneChange.id})")(
+            vinyldnsLoader(zone, recordSetRepository).load())
+          val recordSetChanges = (dnsView, vinyldnsView).parTupled.map {
```
pauljamescleary: we want to load `dnsView` and `vinyldnsView` in parallel.   Create the `IO` and use `parTupled` to wait until both are done

pauljamescleary: Note, removing the `IO.fromFuture` bit and `parTupled` change, all the rest here is whitespace.

modules/api/src/main/scala/vinyldns/api/repository/dynamodb/DynamoDBGroupRepository.scala
```diff
@@ -89,10 +88,10 @@ class DynamoDBGroupRepository(
       .withProvisionedThroughput(new ProvisionedThroughput(dynamoReads, dynamoWrites))
   )
 
-  def loadData: Future[List[Group]] = Await.ready(GroupRepository.loadTestData(this), 100.seconds)
-  loadData
+  def loadData: IO[List[Group]] = GroupRepository.loadTestData(this)
```
pauljamescleary: these all stink.  As part of the pluggable repos work, we need to come up with a better way to "load test data" than this.  This is pretty terrible.

modules/api/src/main/scala/vinyldns/api/repository/dynamodb/DynamoDBHelper.scala
```diff
@@ -46,10 +46,6 @@ trait DynamoUtils {
 /* Used to provide an exponential backoff in the event of a Provisioned Throughput Exception */
 class DynamoDBHelper(dynamoDB: AmazonDynamoDBClient, log: Logger) {
 
-  private val dynamoActorSystem = VinylDNSConfig.system
```
pauljamescleary: it feels good deleting these three lines

modules/api/src/main/scala/vinyldns/api/repository/dynamodb/DynamoDBHelper.scala
```diff
@@ -299,29 +295,7 @@ class DynamoDBHelper(dynamoDB: AmazonDynamoDBClient, log: Logger) {
         s"Unable to process all items in batch for table $tableName, resubmitting new batch with $unprocessed " +
           s"items remaining")
       val nextBatch = toBatchWriteItemRequest(result.getUnprocessedItems)
-      after(backoff, scheduler)(
-        batchWriteItem(tableName, nextBatch, retriesRemaining - 1, backoff * 2))
-    }
-  }
-
-  def lazyScanAll(request: ScanRequest)
```
pauljamescleary: this is not used anywhere in the system, so I removed it.  The wait it was built was kind of strange to begin with.

If we want a lazy scan, should use FS2 instead

modules/api/src/main/scala/vinyldns/api/repository/mysql/JdbcBatchChangeRepository.scala
```diff
@@ -99,204 +96,193 @@ class JdbcBatchChangeRepository
          | WHERE id={id}
         """.stripMargin
 
-  def save(batch: BatchChange): Future[BatchChange] =
+  def save(batch: BatchChange): IO[BatchChange] =
     monitor("repo.BatchChangeJDBC.save") {
-      DB.futureLocalTx { implicit s =>
-        saveBatchChange(batch)
+      IO {
+        DB.localTx { implicit s =>
```
pauljamescleary: there is no such thing as `DB.ioLocalTx`. 

To get the session we must do the `Tx` piece _inside_ of the `IO`

modules/api/src/main/scala/vinyldns/api/repository/dynamodb/DynamoDBHelper.scala
```diff
@@ -227,11 +223,11 @@ class DynamoDBHelper(dynamoDB: AmazonDynamoDBClient, log: Logger) {
 
       case lastEvaluatedKey if lastEvaluatedKey == null || lastEvaluatedKey.isEmpty =>
         // there is no last evaluated key, that means we are done querying
-        Future.successful(acc)
+        IO.pure(acc)
```
pauljamescleary: You can use `IO.pure` on _already completed values_.  _Never_ use it on an expression.

modules/api/src/main/scala/vinyldns/api/repository/mysql/JdbcZoneRepository.scala
```diff
@@ -114,97 +112,95 @@ class JdbcZoneRepository extends ZoneRepository with ProtobufConversions with Mo
...
-            } yield zone
+      case _ =>
+        monitor("repo.ZoneJDBC.save") {
+          IO {
+            DB.localTx { implicit s =>
+              deleteZoneAccess(zone)
+              putZone(zone)
+              putZoneAccess(zone)
+              zone
+            }
```
pauljamescleary: this was the goofy change.

Basically, you cannot `Tx` until you are _inside_ an `IO` because it is lazily evaluated.

So, we lose `futureTx` things.  

Had to take the 3 functions `deleteZoneAccess`, `putZone`, `putZoneAccess` _out_ of a future.  So, they will most definitely blow up if something goes wrong.  **We want this to happen** because we are inside `DB.localTx`, we _want_ things to blow up in this block because it will cause the entire transaction to rollback.

If any of the 3 throw an error, the transaction is rolled back and the `IO` will be a failure.

modules/api/src/main/scala/vinyldns/api/repository/mysql/JdbcZoneRepository.scala
```diff
@@ -349,30 +338,20 @@ class JdbcZoneRepository extends ZoneRepository with ProtobufConversions with Mo
...
-    Future {
-      blocking {
-        // make sure that we do a distinct, so that we don't generate unnecessary inserts
-        PUT_ZONE_ACCESS.batchByName(allAccessors.distinct: _*).apply()
-        zone
-      }
-    }
+    // make sure that we do a distinct, so that we don't generate unnecessary inserts
+    PUT_ZONE_ACCESS.batchByName(allAccessors.distinct: _*).apply()
+    zone
```
pauljamescleary: See how we remove the 3 functions from `Future`.  If they fail, caplooey.  Transaction rolls back and IO fails which is what we want.

modules/api/src/main/scala/vinyldns/api/route/BatchChangeRouting.scala
```diff
@@ -77,15 +76,16 @@ trait BatchChangeRoute extends Directives {
...
-    case Left(cle: ChangeLimitExceeded) => complete(StatusCodes.RequestEntityTooLarge, cle.message)
-    case Left(bie: BatchChangeIsEmpty) => complete(StatusCodes.UnprocessableEntity, bie.message)
-    case Left(crl: InvalidBatchChangeResponses) => complete(StatusCodes.BadRequest, crl)
-    case Left(cnf: BatchChangeNotFound) => complete(StatusCodes.NotFound, cnf.message)
-    case Left(una: UserNotAuthorizedError) => complete(StatusCodes.Forbidden, una.message)
-    case Left(uct: BatchConversionError) => complete(StatusCodes.BadRequest, uct)
-    case Left(uce: UnknownConversionError) => complete(StatusCodes.InternalServerError, uce)
-  }
+  private def execute[A](f: => EitherT[IO, BatchChangeErrorResponse, A])(rt: A => Route): Route =
+    onSuccess(f.value.unsafeToFuture()) {
```
pauljamescleary: All of our `Routes` do this now.  The `f.value` is an `IO`.  `onSuccess` _only_ works with `Future`.  Gotta get our `IO` to a future, which is what `unsafeToFuture()` does.

Ideally, there would be an akka-http analogue to `onSuccess` that could take `IO`.

modules/api/src/test/scala/vinyldns/api/AkkaTestJawn.scala
```diff
@@ -22,36 +22,44 @@ import akka.testkit.{EventFilter, TestKit}
...
+final case class TimeoutException(message: String) extends Throwable(message)
+
 trait ResultHelpers {
 
   implicit val baseTimeout: Timeout = new Timeout(2.seconds)
 
-  def await[T](f: => Future[_], duration: FiniteDuration = 1.second)(implicit tag: ClassTag[T]): T =
-    Await.ready(f, duration).mapTo[T].value.get.get
+  def await[T](f: => IO[_], duration: FiniteDuration = 1.second): T =
+    awaitResultOf[T](f.map(_.asInstanceOf[T]).attempt, duration).toOption.get
```
pauljamescleary: there is no `mapTo` in `IO`, so we have this goofy `map(_.asInstanceOf[T])` which is unsavory but this is test code.  

This function is not great because it is an `await` which _could_ fail, however it is guaranteeing a `T` out.  In other words, it is throwing.  Oof.

modules/api/src/test/scala/vinyldns/api/CatsHelpers.scala
```diff
@@ -17,34 +17,43 @@
...
 
+import scala.concurrent.ExecutionContext.Implicits.global
+
 trait CatsHelpers {
 
   implicit val baseTimeout: Timeout = new Timeout(2.seconds)
 
-  def await[T](f: => Future[T], duration: FiniteDuration = 1.second): T =
-    Await.ready(f, duration).value.get.get
+  def await[E, T](f: => IO[T], duration: FiniteDuration = 1.second): T = {
```
pauljamescleary: do we still need `CatsHelpers` now that everything is cats?

modules/api/src/main/scala/vinyldns/api/Interfaces.scala
rebstar6: i kinda hate operators that cant be googled. was one of my main beefs with the scalaz stuff. its really frustrating when you 1st start and are trying to figure out a function in the codebase.

i may rather just flatmap tbh

pauljamescleary: It is idiomatic to do `*>`, if we want to avoid symbols I can do the just as annoying `productR` which is what the `*>` operator does.  That said, I think we are ok with using operators when it is idiomatic and we are not using them heavily.

modules/api/src/it/scala/vinyldns/api/domain/zone/ZoneServiceIntegrationSpec.scala
```diff
@@ -132,18 +132,19 @@ class ZoneServiceIntegrationSpec extends DynamoDBIntegrationSpec with MockitoSug
         testZoneService
           .deleteZone(zone.id, badAuth)
           .value
-          .mapTo[Either[Throwable, ZoneCommandResult]]
-      whenReady(result, timeout) { _ =>
-        val error = leftResultOf(result)
-        error shouldBe a[NotAuthorizedError]
-      }
+      val error = leftResultOf(result)
```
rebstar6: why are we waiting in the RecordSetServiceIntegrationSpec case, but not here?

rebstar6: comparing to https://github.com/vinyldns/vinyldns/pull/139/files#diff-e0fc3af2f510b4df4fd3e25585a89ab5R309

modules/api/src/it/scala/vinyldns/api/domain/zone/ZoneServiceIntegrationSpec.scala
```diff
@@ -153,8 +154,8 @@ class ZoneServiceIntegrationSpec extends DynamoDBIntegrationSpec with MockitoSug
     }
   }
 
-  private def waitForSuccess[T](f: => Future[T]): T = {
-    val waiting = f.recover { case _ => Thread.sleep(2000); waitForSuccess(f) }
+  private def waitForSuccess[T](f: => IO[T]): T = {
```
rebstar6: probably should have thrown this in the test helpers, but fine with it here for now

modules/api/src/it/scala/vinyldns/api/repository/mysql/JdbcZoneRepositoryIntegrationSpec.scala
```diff
@@ -468,20 +469,23 @@ class JdbcZoneRepositoryIntegrationSpec
       val expectedSecondPage = sorted.drop(4).take(4)
       val expectedThirdPage = sorted.drop(8).take(4)
 
-      whenReady(saveZones(testZones), timeout) { _ =>
-        whenReady(repo.listZones(superUserAuth, offset = None, pageSize = 4), timeout) {
-          firstPage =>
-            (firstPage should contain).theSameElementsInOrderAs(expectedFirstPage)
+      whenReady(saveZones(testZones).unsafeToFuture(), timeout) { _ =>
+        whenReady(
```
rebstar6: man, these nested whenReady's are kinda yucky

modules/api/src/main/scala/vinyldns/api/Interfaces.scala
```diff
@@ -16,99 +16,89 @@
...
-      theFuture: => Future[A],
-      duration: FiniteDuration,
-      error: Throwable,
-      scheduler: Scheduler)(implicit ec: ExecutionContext): Result[A] = result[A] {
-    val timeOut = after(duration = duration, using = scheduler)(Future.failed(error))
-
-    Future.firstCompletedOf(Seq(theFuture, timeOut))
+  def withTimeout[A](theIo: => IO[A], duration: FiniteDuration, error: Throwable): Result[A] = {
+    val timeOut = IO.sleep(duration) *> IO(error)
+    EitherT(IO.race(timeOut, theIo).handleError(e => Left(e)))
```
rebstar6: its not clear to me what purpose the `handleError` here is serving. Type-wise, nothing is different between `IO.race(timeOut, theIo).handleError(e => Left(e))` and `IO.race(timeOut, theIo)`

can you explain?

pauljamescleary: This is because `IO` is _lazily evaluated_...not eagerly evaluated like `Future`.  What we want to do with `IO` is _always_ go to an `Either`.

When we construct an IO, the thing inside is not evaluated until the program is actually run, _not_ at the point it is created.  We can see this in the ammonite repl...

```
@ val ioBad = IO(throw new RuntimeException("FAIL!")) 
ioBad: IO[Nothing] = Delay(ammonite.$sess.cmd19$$$Lambda$2245/1853366079@680f2217)

@ ioBad.unsafeRunSync 
java.lang.RuntimeException: FAIL!
  ammonite.$sess.cmd19$.$anonfun$ioBad$1(cmd19.sc:1)
  cats.effect.internals.IORunLoop$.step(IORunLoop.scala:167)
  cats.effect.IO.unsafeRunTimed(IO.scala:304)
  cats.effect.IO.unsafeRunSync(IO.scala:239)
  ammonite.$sess.cmd20$.<init>(cmd20.sc:1)
  ammonite.$sess.cmd20$.<clinit>(cmd20.sc)
```

See, when we constructed the `IO`, we didn't get the error.  Contrast this with `Future` that is _eagerly_ evaluated...

```
@ val futureBad = Future(throw new RuntimeException("FAIL!")) 
futureBad: Future[Nothing] = Future(Failure(java.lang.RuntimeException: FAIL!))
```

This is why `EitherT` in cats cannot automatically force a failure to a `Left`

Here is output from the repl demonstrating this in action...

What happens when the race times out???  Well, we get our time out (exception) in a `Left`, which is what we want...

```
@ val raceTimeout = IO.race(timeout, IO { Thread.sleep(2000); "hey" }) 
raceTimeout: IO[Either[RuntimeException, String]] = Async(cats.effect.IO$$$Lambda$2096/675199364@630c579c)

@ EitherT(raceTimeout).value.unsafeRunSync 
res32: Either[RuntimeException, String] = Left(java.lang.RuntimeException: TIMEOUT!!!)
```

Ok, so what happens if the race runs successfully??? We get our value in a `Right` as expected...

```
@ val raceOk = IO.race(timeout, IO("we good")) 
raceOk: IO[Either[RuntimeException, String]] = Async(cats.effect.IO$$$Lambda$2096/675199364@6343c70)

@ EitherT(raceOk).value.unsafeRunSync 
res31: Either[RuntimeException, String] = Right("we good")
```

Now, what happens if during evaluation of our `IO`, we throw an error?  Well, that is **not good!!**.  We actually threw an error, instead of getting our failure in a `Left`

```
@ val raceIoFailure = IO.race(timeout, IO(throw new RuntimeException("fail right away!"))) 
raceIoFailure: IO[Either[RuntimeException, Nothing]] = Async(cats.effect.IO$$$Lambda$2096/675199364@262cafd5)

@ EitherT(raceIoFailure).value.unsafeRunSync 
java.lang.RuntimeException: fail right away!
  ammonite.$sess.cmd29$.$anonfun$raceIoFailure$1(cmd29.sc:1)
  cats.effect.internals.IORunLoop$.cats$effect$internals$IORunLoop$$loop(IORunLoop.scala:85)
  cats.effect.internals.IORunLoop$.startCancelable(IORunLoop.scala:42)
  cats.effect.internals.IORace$.$anonfun$simple$1(IORace.scala:76)
  cats.effect.IO$.liftedTree1$1(IO.scala:720)
  cats.effect.IO$.$anonfun$cancelable$1(IO.scala:720)
  cats.effect.IO$.$anonfun$cancelable$1$adapted(IO.scala:714)
  cats.effect.internals.IORunLoop$.cats$effect$internals$IORunLoop$$loop(IORunLoop.scala:118)
  cats.effect.internals.IORunLoop$.start(IORunLoop.scala:35)
  cats.effect.IO.unsafeRunAsync(IO.scala:257)
  cats.effect.internals.IOPlatform$.unsafeResync(IOPlatform.scala:38)
  cats.effect.IO.unsafeRunTimed(IO.scala:308)
  cats.effect.IO.unsafeRunSync(IO.scala:239)
  ammonite.$sess.cmd33$.<init>(cmd33.sc:1)
  ammonite.$sess.cmd33$.<clinit>(cmd33.sc)
```

That is why we need to handle the error, to force an unexpected error into `Left`...

```
@ 
val raceIoFailureHandle = IO.race(timeout, IO(throw new RuntimeException("fail right away"))).handleError { e => Left(e) } 
raceIoFailureHandle: IO[Either[Throwable, Nothing]] = Bind(
  Async(cats.effect.IO$$$Lambda$2096/675199364@1ef73f10),
  <function1>
)

@ EitherT(raceIoFailureHandle).value.unsafeRunSync 
res39: Either[Throwable, Nothing] = Left(java.lang.RuntimeException: fail right away)
```

See, this does not "blow up", but puts an unexpected exit of our `race` into `Left`

modules/api/src/it/scala/vinyldns/api/repository/mysql/JdbcZoneRepositoryIntegrationSpec.scala
pauljamescleary: agree

modules/api/src/it/scala/vinyldns/api/domain/zone/ZoneServiceIntegrationSpec.scala
pauljamescleary: yea, probably should be for consistency

modules/api/src/test/scala/vinyldns/api/CatsHelpers.scala
pauljamescleary: I created a follow on card to cleanup tests and test helpers.

modules/api/src/test/scala/vinyldns/api/AkkaTestJawn.scala
rebstar6: theres some gross stuff using this. tried messing with it and making the signature somethign like `def await[T](f: => IO[_ <: T], duration: FiniteDuration = 1.second): T`
and you get errors on lines like:
`await[List[Zone]](store.getRecordSets(rsOk.zoneId, rsOk.name, rsOk.typ))`

...which makes no sense

rebstar6: also, same problem, but seems to work as an alternative:
```
  def await[T: ClassTag](f: => IO[_], duration: FiniteDuration = 1.second): T = {
    val res = awaitResultOf(f.attempt, duration)

    res.toOption.get match {
      case t: T => t
      case _ => throw new RuntimeException("BOOO!!!")
    }
  }
```

PR #140, opened 2018-08-27
Test DDNS connectivity when connecting to a zone

mitruly:
When "creating" (ie. connecting to) a zone in VinylDNS, perform a DDNS delete and add for a test record set to check connectivity.
- Record set is named `vinyldns-ddns-connectivity-test` of type `TXT`. If a user is connecting to a zone with an existing record, its record data will be replaced.
- Add pre-requisite for delete messages, asserting that record to delete actually exists (so DNS response is `NXRRSET`).
- When attempting to delete the record, `RecordSetNotFound` (`NXRRSET`) and `NameNotFound` (`NXDOMAIN`) errors are suppressed.
- Update TSIG key to HMAC-SHA512 algorithm (from default of HMAC-MD5).

Updated main source files, unit and func tests as well as API documentation.

### DIFF
--- a/docker/bind9/etc/named.conf.local
+++ b/docker/bind9/etc/named.conf.local
@@ -11,6 +11,11 @@ key "vinyldns." {
   secret "nzisn+4G2ldMn0q1CV3vsg==";
 };
 
+key "vinyldns-no-updates" {
+	algorithm hmac-md5;
+	secret "BjI47RjxzSQyOktSnmbD4A==";
+};
+
 // Consider adding the 1918 zones here, if they are not used in your
 // organization
 //include "/etc/bind/zones.rfc1918";
--- a/docker/bind9/etc/named.conf.local
+++ b/docker/bind9/etc/named.conf.local
@@ -7,13 +7,13 @@
 //include "/etc/bind/zones.rfc1918";
 
 key "vinyldns." {
-    algorithm hmac-md5;
-    secret "nzisn+4G2ldMn0q1CV3vsg==";
+	algorithm hmac-sha512;
+	secret "wCZZS9lyRr77+jqfnkZ/92L9fD5ilmfrG0sslc3mgmTFsF1fRgmtJ0rj RkFITt8VHQ37wvM/nI9MAIWXYTvMqg==";
 };
 
 key "vinyldns-no-updates" {
-	algorithm hmac-md5;
-	secret "BjI47RjxzSQyOktSnmbD4A==";
+	algorithm hmac-sha512;
+	secret "1GOhWm/nwqlQop1YQ6sl96eVTjULth0E7LonKB6X4uycygaCUQRG2JPQ kHVFgp768cyUuCv4j/tvL8C+cUCkcA==";
 };
 
 // Consider adding the 1918 zones here, if they are not used in your
--- a/docker/bind9/etc/named.conf.local
+++ b/docker/bind9/etc/named.conf.local
@@ -7,8 +7,8 @@
 //include "/etc/bind/zones.rfc1918";
 
 key "vinyldns." {
-  algorithm hmac-md5;
-  secret "nzisn+4G2ldMn0q1CV3vsg==";
+    algorithm hmac-md5;
+    secret "nzisn+4G2ldMn0q1CV3vsg==";
 };
 
 key "vinyldns-no-updates" {
--- a/docker/bind9/zones/one-time.hosts
+++ b/docker/bind9/zones/one-time.hosts
@@ -12,4 +12,3 @@ test             IN  A   3.3.3.3
 test             IN  A   4.4.4.4
 @                IN  A   5.5.5.5
 already-exists   IN  A   6.6.6.6
-vinyldns-ddns-connectivity-test    IN  TXT "random existing text"
--- a/docker/bind9/zones/one-time.hosts
+++ b/docker/bind9/zones/one-time.hosts
@@ -12,4 +12,4 @@ test             IN  A   3.3.3.3
 test             IN  A   4.4.4.4
 @                IN  A   5.5.5.5
 already-exists   IN  A   6.6.6.6
-vinyldns-test    IN  TXT "random existing text"
+vinyldns-ddns-connectivity-test    IN  TXT "random existing text"
--- a/docker/bind9/zones/one-time.hosts
+++ b/docker/bind9/zones/one-time.hosts
@@ -12,3 +12,4 @@ test             IN  A   3.3.3.3
 test             IN  A   4.4.4.4
 @                IN  A   5.5.5.5
 already-exists   IN  A   6.6.6.6
+vinyldns-test    IN  TXT "random existing text"
--- a/modules/api/functional_test/conftest.py
+++ b/modules/api/functional_test/conftest.py
@@ -19,12 +19,14 @@ def pytest_addoption(parser):
                      help="The zone name that will be used for testing")
     parser.addoption("--dns-key-name", dest="dns_key_name", action="store", default="vinyldns.",
                      help="The name of the key used to sign updates for the zone")
-    parser.addoption("--dns-key", dest="dns_key", action="store", default="nzisn+4G2ldMn0q1CV3vsg==",
-                     help="The tsig key")
+    parser.addoption("--dns-key", dest="dns_key", action="store",
+                    default="wCZZS9lyRr77+jqfnkZ/92L9fD5ilmfrG0sslc3mgmTFsF1fRgmtJ0rj RkFITt8VHQ37wvM/nI9MAIWXYTvMqg==",
+                    help="The tsig key")
     parser.addoption("--dns-no-updates-key-name", dest="dns_no_updates_key_name", action="store",
                      default="vinyldns-no-updates", help="DDNS test TSIG key name")
     parser.addoption("--dns-no-updates-key", dest="dns_no_updates_key", action="store",
-                     default="BjI47RjxzSQyOktSnmbD4A==", help="DDNS test TSIG key")
+                     default="1GOhWm/nwqlQop1YQ6sl96eVTjULth0E7LonKB6X4uycygaCUQRG2JPQ kHVFgp768cyUuCv4j/tvL8C+cUCkcA==",
+                     help="DDNS test TSIG key")
 
     # optional
     parser.addoption("--basic-auth", dest="basic_auth_creds",
--- a/modules/api/functional_test/conftest.py
+++ b/modules/api/functional_test/conftest.py
@@ -21,6 +21,10 @@ def pytest_addoption(parser):
                      help="The name of the key used to sign updates for the zone")
     parser.addoption("--dns-key", dest="dns_key", action="store", default="nzisn+4G2ldMn0q1CV3vsg==",
                      help="The tsig key")
+    parser.addoption("--dns-no-updates-key-name", dest="dns_no_updates_key_name", action="store",
+                     default="vinyldns-no-updates", help="DDNS test TSIG key name")
+    parser.addoption("--dns-no-updates-key", dest="dns_no_updates_key", action="store",
+                     default="BjI47RjxzSQyOktSnmbD4A==", help="DDNS test TSIG key")
 
     # optional
     parser.addoption("--basic-auth", dest="basic_auth_creds",
--- a/modules/api/functional_test/conftest.py
+++ b/modules/api/functional_test/conftest.py
@@ -17,16 +17,6 @@ def pytest_addoption(parser):
                      help="The ip address for the dns server to use for the tests")
     parser.addoption("--dns-zone", dest="dns_zone", action="store", default="vinyldns.",
                      help="The zone name that will be used for testing")
-    parser.addoption("--dns-key-name", dest="dns_key_name", action="store", default="vinyldns.",
-                     help="The name of the key used to sign updates for the zone")
-    parser.addoption("--dns-key", dest="dns_key", action="store",
-                    default="wCZZS9lyRr77+jqfnkZ/92L9fD5ilmfrG0sslc3mgmTFsF1fRgmtJ0rj RkFITt8VHQ37wvM/nI9MAIWXYTvMqg==",
-                    help="The tsig key")
-    parser.addoption("--dns-no-updates-key-name", dest="dns_no_updates_key_name", action="store",
-                     default="vinyldns-no-updates", help="DDNS test TSIG key name")
-    parser.addoption("--dns-no-updates-key", dest="dns_no_updates_key", action="store",
-                     default="1GOhWm/nwqlQop1YQ6sl96eVTjULth0E7LonKB6X4uycygaCUQRG2JPQ kHVFgp768cyUuCv4j/tvL8C+cUCkcA==",
-                     help="DDNS test TSIG key")
 
     # optional
     parser.addoption("--basic-auth", dest="basic_auth_creds",
@@ -68,8 +58,6 @@ def pytest_configure(config):
 
     VinylDNSTestContext.configure(config.getoption("dns_ip"),
                                config.getoption("dns_zone"),
-                               config.getoption("dns_key_name"),
-                               config.getoption("dns_key"),
                                config.getoption("url"))
 
 def pytest_report_header(config):
--- a/modules/api/functional_test/live_tests/recordsets/list_recordset_changes_test.py
+++ b/modules/api/functional_test/live_tests/recordsets/list_recordset_changes_test.py
@@ -106,11 +106,11 @@ def test_list_recordset_changes_paging(zone_history_context):
     response_2 = client.list_recordset_changes(original_zone['id'], start_from=response_1['nextId'], max_items=3)
     # nextId differs local/in dev where we get exactly the last item
     # Requesting one over the total in the local in memory dynamo will force consistent behavior.
-    response_3 = client.list_recordset_changes(original_zone['id'], start_from=response_2['nextId'], max_items=12)
+    response_3 = client.list_recordset_changes(original_zone['id'], start_from=response_2['nextId'], max_items=11)
 
     check_changes_response(response_1, recordChanges=True, nextId=True, startFrom=False, maxItems=3)
     check_changes_response(response_2, recordChanges=True, nextId=True, startFrom=response_1['nextId'], maxItems=3)
-    check_changes_response(response_3, recordChanges=True, nextId=False, startFrom=response_2['nextId'], maxItems=12)
+    check_changes_response(response_3, recordChanges=True, nextId=False, startFrom=response_2['nextId'], maxItems=11)
 
     for change in response_1['recordSetChanges']:
         assert_that(change['changeType'], is_('Delete'))
@@ -126,8 +126,8 @@ def test_list_recordset_changes_exhausted(zone_history_context):
     """
     client = zone_history_context.client
     original_zone = zone_history_context.results['zone']
-    response = client.list_recordset_changes(original_zone['id'], start_from=None, max_items=18)
-    check_changes_response(response, recordChanges=True, startFrom=False, nextId=False, maxItems=18)
+    response = client.list_recordset_changes(original_zone['id'], start_from=None, max_items=17)
+    check_changes_response(response, recordChanges=True, startFrom=False, nextId=False, maxItems=17)
 
     deleteChanges = response['recordSetChanges'][0:3]
     updateChanges = response['recordSetChanges'][3:6]
--- a/modules/api/functional_test/live_tests/recordsets/list_recordset_changes_test.py
+++ b/modules/api/functional_test/live_tests/recordsets/list_recordset_changes_test.py
@@ -106,11 +106,11 @@ def test_list_recordset_changes_paging(zone_history_context):
     response_2 = client.list_recordset_changes(original_zone['id'], start_from=response_1['nextId'], max_items=3)
     # nextId differs local/in dev where we get exactly the last item
     # Requesting one over the total in the local in memory dynamo will force consistent behavior.
-    response_3 = client.list_recordset_changes(original_zone['id'], start_from=response_2['nextId'], max_items=11)
+    response_3 = client.list_recordset_changes(original_zone['id'], start_from=response_2['nextId'], max_items=12)
 
     check_changes_response(response_1, recordChanges=True, nextId=True, startFrom=False, maxItems=3)
     check_changes_response(response_2, recordChanges=True, nextId=True, startFrom=response_1['nextId'], maxItems=3)
-    check_changes_response(response_3, recordChanges=True, nextId=False, startFrom=response_2['nextId'], maxItems=11)
+    check_changes_response(response_3, recordChanges=True, nextId=False, startFrom=response_2['nextId'], maxItems=12)
 
     for change in response_1['recordSetChanges']:
         assert_that(change['changeType'], is_('Delete'))
@@ -126,8 +126,8 @@ def test_list_recordset_changes_exhausted(zone_history_context):
     """
     client = zone_history_context.client
     original_zone = zone_history_context.results['zone']
-    response = client.list_recordset_changes(original_zone['id'], start_from=None, max_items=17)
-    check_changes_response(response, recordChanges=True, startFrom=False, nextId=False, maxItems=17)
+    response = client.list_recordset_changes(original_zone['id'], start_from=None, max_items=18)
+    check_changes_response(response, recordChanges=True, startFrom=False, nextId=False, maxItems=18)
 
     deleteChanges = response['recordSetChanges'][0:3]
     updateChanges = response['recordSetChanges'][3:6]
--- a/modules/api/functional_test/live_tests/recordsets/list_recordsets_test.py
+++ b/modules/api/functional_test/live_tests/recordsets/list_recordsets_test.py
@@ -13,8 +13,7 @@ class ListRecordSetsFixture():
         self.client = shared_zone_test_context.ok_vinyldns_client
         self.new_rs = {}
         existing_records = self.client.list_recordsets(self.test_context['id'])['recordSets']
-        # If there is not already a 'TXT' record for 'vinyldns-ddns-connectivity-test', it will be added
-        assert_that(existing_records, has_length(8))
+        assert_that(existing_records, has_length(7))
         rs_template = {
             'zoneId': self.test_context['id'],
             'name': '00-test-list-recordsets-',
@@ -54,19 +53,6 @@ class ListRecordSetsFixture():
 
         for i in range(7):
             self.all_records[i + 10] = existing_records[i]
-
-        # Adding in DDNS connectivity record
-        self.all_records[17] = {
-            'zoneId': self.test_context['id'],
-            'name': 'vinyldns-ddns-connectivity-test',
-            'type': 'TXT',
-            'ttl': 86400,
-            'records': [
-                {
-                    'text': 'connection test'
-                }
-            ]
-        }
 
     def tear_down(self):
         for key in self.new_rs:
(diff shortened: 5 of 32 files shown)


modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneConnectionValidator.scala
```diff
@@ -71,6 +71,41 @@ class ZoneConnectionValidator(defaultConnection: ZoneConnection, scheduler: Sche
...
+      "vinyldns-test",
+      RecordType.TXT,
+      30,
+      RecordSetStatus.Pending,
+      DateTime.now,
+      records = List(TXTData("connection test")))
+
+    def performDdnsChecks(rsList: List[RecordSet]): Result[Unit] =
+      if (rsList.isEmpty) {
+        // Perform add then delete of test record
```
pauljamescleary: I think you should change the check up.

The logic currently follows:
1. If the test record does not exist, then add it and then delete it
2. If the test record does exist, then delete it and then add it

You will end up in two different places.   If the record does not exist (first case), it will not exist after the logic.   If the record does exist (second case), it will still exist after the logic.

Rather than doing the resolve, I would recommend:

1. Attempt to delete the record.  If you get a not found error coming back, that is ok!  Keep moving, swallowing that error
2. Create the record

I think doing an immediate delete is not what we want, but open to discussion.  DNS servers maybe eventually consistent, so that churn may wind up with some unexpected errors?  Perhaps just leave it there after step 2. 

**Note: we do not need to verify the record was created.  If we get back a NoError on the add we should be good**.

mitruly: @pauljamescleary Thanks for the comment - discussion would definitely be useful for me.
- One caveat that I found without using `resolve` is that if we are deleting a record set that matches the same name and type, we would potentially be re-adding it with different record data, TTL and other settings. I was playing around with `resolve` and I think that `NoError` is also returned in the DNS response for a delete of a non-existent record.
- It might also be confusing if we never delete the record since it would show up in the zone unless we are somehow filtering that out.

With that said, if we are fine with saying that `vinyldns-test` is off-limits to users and will be created in a zone upon connection, then I think a simple delete and add like you suggest would suffice.

docker/bind9/etc/named.conf.local
```diff
@@ -11,6 +11,11 @@ key "vinyldns." {
   secret "nzisn+4G2ldMn0q1CV3vsg==";
 };
 
+key "vinyldns-no-updates" {
```
mitruly: Generated `secret` using `ddns-confgen`. `secret` must match algorithm; otherwise, invalid TSIG key error is thrown.

docker/bind9/zones/one-time.hosts
```diff
@@ -12,3 +12,4 @@ test             IN  A   3.3.3.3
 test             IN  A   4.4.4.4
 @                IN  A   5.5.5.5
 already-exists   IN  A   6.6.6.6
+vinyldns-ddns-connectivity-test    IN  TXT "random existing text"
```
mitruly: Testing replacement of existing record matching DDNS test record with different `text`.

modules/api/functional_test/live_tests/recordsets/list_recordset_changes_test.py
```diff
@@ -106,11 +106,11 @@ def test_list_recordset_changes_paging(zone_history_context):
     response_2 = client.list_recordset_changes(original_zone['id'], start_from=response_1['nextId'], max_items=3)
     # nextId differs local/in dev where we get exactly the last item
     # Requesting one over the total in the local in memory dynamo will force consistent behavior.
-    response_3 = client.list_recordset_changes(original_zone['id'], start_from=response_2['nextId'], max_items=11)
```
mitruly: List record set change tests needed to get updated to include test DDNS record.

modules/api/functional_test/live_tests/zones/create_zone_test.py
```diff
@@ -33,9 +33,12 @@
      'records': [{u'address': u'5.5.5.5'}]},
     {'name': u'already-exists',
      'type': u'A',
-     'records': [{u'address': u'6.6.6.6'}]}]
+     'records': [{u'address': u'6.6.6.6'}]},
+    {'name': u'vinyldns-ddns-connectivity-test',
```
mitruly: Adding check to create zone test.

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneConnectionValidator.scala
```diff
@@ -71,6 +71,26 @@ class ZoneConnectionValidator(defaultConnection: ZoneConnection, scheduler: Sche
     }
   }.toResult
 
+  def testDdnsConnectivity(dnsConnection: DnsConnection, zone: Zone): Result[Unit] = {
+    val rs = RecordSet(
```
mitruly: Couldn't find a better place to store this since it is based off of `zone`, which gets passed in.

mitruly:
CodeCov doesn't appear to consistently update with subsequent commits.. This is from running `sbt verify` locally in this branch:
```
[info] Aggregation complete. Coverage was [94.20]
[info] All done. Coverage was [94.20%]
```

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneConnectionValidator.scala
```diff
@@ -71,6 +71,26 @@ class ZoneConnectionValidator(defaultConnection: ZoneConnection, scheduler: Sche
...
+      zone.id,
+      "vinyldns-ddns-connectivity-test",
+      RecordType.TXT,
+      30,
```
mitruly: Was considering changing `TTL` to the max value - would like some thoughts on this.

modules/api/functional_test/vinyldns_context.py
```diff
@@ -5,6 +5,8 @@ class VinylDNSTestContext:
     dns_rev_v6_zone_name = '1.9.e.f.c.c.7.2.9.6.d.f.ip6.arpa.'
     dns_key_name = 'vinyldns.'
     dns_key = 'nzisn+4G2ldMn0q1CV3vsg=='
+    dns_no_updates_key_name = 'vinyldns-no-updates'
+    dns_no_updates_key = 'BjI47RjxzSQyOktSnmbD4A=='
```
pauljamescleary: this will break deployment as our deployment relies on these items being on our test bind9 servers.

I think our test servers do not support 128 bit keys.  

Ideally, we would use the same key strength as our test bind9 servers for this key (this key is 128 bit).  If you can make it work, that is ideal.

Otherwise, we need to make the key value a parameter to the func tests, passed into conftest.py

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneConnectionValidator.scala
pauljamescleary: TTL should be higher for sure, a day would be sufficient.  Noone will really query for it tbh.

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneConnectionValidator.scala
```diff
@@ -71,6 +71,26 @@ class ZoneConnectionValidator(defaultConnection: ZoneConnection, scheduler: Sche
...
+      RecordSetStatus.Pending,
+      DateTime.now,
+      records = List(TXTData("connection test")))
+
+    val result = for {
+      _ <- dnsConnection.applyChange(RecordSetChange.forDelete(rs, zone))
```
pauljamescleary: applyChange will fail `Left` with `RecordSetNotFound` or `NameNotFound`.

You need to "swallow" these two cases and proceed to the add.

modules/api/src/test/scala/vinyldns/api/domain/zone/ZoneConnectionValidatorSpec.scala
```diff
@@ -138,6 +144,20 @@ class ZoneConnectionValidatorSpec
       result should be(right)
     }
 
+    "respond with a failure if DDNS connectivity fails" in {
```
pauljamescleary: need to add unit tests here:

1. delete returns a `NameNotFound`
1. delete returns a `RecordSetNotFound`
1. delete returns an unexpected error code, like `Refused` (which we see a lot)

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneConnectionValidator.scala
```diff
@@ -71,6 +71,26 @@ class ZoneConnectionValidator(defaultConnection: ZoneConnection, scheduler: Sche
...
 
+  def testDdnsConnectivity(dnsConnection: DnsConnection, zone: Zone): Result[Unit] = {
+    val rs = RecordSet(
+      zone.id,
+      "vinyldns-ddns-connectivity-test",
+      RecordType.TXT,
+      30,
+      RecordSetStatus.Pending,
+      DateTime.now,
+      records = List(TXTData("connection test")))
```
pauljamescleary: rather than connection test, let's put something more meaningful to make sure the value is updated.  Something like the current date and time

modules/api/functional_test/vinyldns_context.py
mitruly: @pauljamescleary I think you are right. I tried using a different strength key, and that failed a few integration tests with some TSIG not authorized error message. I added options to the `conftest.py`.

docker/bind9/etc/named.conf.local
```diff
@@ -7,8 +7,13 @@
 //include "/etc/bind/zones.rfc1918";
 
 key "vinyldns." {
-  algorithm hmac-md5;
-  secret "nzisn+4G2ldMn0q1CV3vsg==";
+    algorithm hmac-md5;
+    secret "H3qEuLC+7sG6exm7+F+dBwuV+pbu6cJIs0Yq3E607A4dB8yEtk5j5EeP U1HnR6jUFa2g+hYNYW+iubB8Ii90Eg==";
```
rebstar6: if were changing this here do we not need to change the default connection keys in docker.conf to be this too (?)

docker/bind9/zones/one-time.hosts
rebstar6: isnt the point of one-time that we only have one test using it?

id probably use a different zone for this check

modules/api/functional_test/live_tests/recordsets/list_recordset_changes_test.py
rebstar6: no, we shouldnt be storing the change in the database i thought (?)

this shouldnt be visible to people

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneConnectionValidator.scala
```diff
@@ -67,6 +69,31 @@ class ZoneConnectionValidator(defaultConnection: ZoneConnection)
     }
   }.toResult
 
+  def testDdnsConnectivity(dnsConnection: DnsConnection, zone: Zone): Result[Unit] = {
+    val rs = RecordSet(
+      zone.id,
+      "vinyldns-ddns-connectivity-test",
+      RecordType.TXT,
+      86400,
```
rebstar6: id think we'd want a low TTL on this guy so it doesnt hang around

modules/api/functional_test/live_tests/recordsets/list_recordset_changes_test.py
mitruly: @rebstar6 That is the approach that we initially took with this. There was some follow-up conversation at https://github.com/vinyldns/vinyldns/pull/140#discussion_r212969853 where it was suggested to keep this connectivity test record. If we're keeping the record but not updating the database (which might be what you are suggesting), then I will have to update how it is being handled.

modules/api/functional_test/live_tests/recordsets/list_recordsets_test.py
```diff
@@ -54,6 +55,19 @@ def __init__(self, shared_zone_test_context):
         for i in range(7):
             self.all_records[i + 10] = existing_records[i]
 
+        # Adding in DDNS connectivity record
+        self.all_records[17] = {
```
rebstar6: same here, if were removing this record, why are people seeing it in listRecords?

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneConnectionValidator.scala
```diff
@@ -67,6 +69,31 @@ class ZoneConnectionValidator(defaultConnection: ZoneConnection)
...
+        .applyChange(RecordSetChange.forDelete(rs, zone))
+        .map(_ => ())
+        .recover {
+          case _: RecordSetNotFound | _: NameNotFound => ()
+        }
+      _ <- dnsConnection.applyChange(RecordSetChange.forAdd(rs, zone))
+    } yield ()
+
+    result.leftMap {
+      case e => ConnectionFailed(zone, s"Unable to test DDNS connectivity to zone: ${e.getMessage}")
```
rebstar6: i dont think "Unable to test DDNS connectivity" is the right message here...Unable to apply changes in zone, or something

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneConnectionValidator.scala
mitruly: Initially this was set to the minimum TTL of `30`, but it was suggested to be bumped up to 24 hours if the record is sitting around.

modules/api/src/test/scala/vinyldns/api/domain/zone/ZoneConnectionValidatorSpec.scala
```diff
@@ -76,6 +78,12 @@ class ZoneConnectionValidatorSpec
     override val opTimeout: FiniteDuration = 10.milliseconds
     override def dnsConnection(conn: ZoneConnection): DnsConnection = testDnsConnection(conn)
     override def loadDns(zone: Zone): IO[ZoneView] = testLoadDns(zone)
+    override def testDdnsConnectivity(dnsConnection: DnsConnection, zone: Zone): Result[Unit] =
+      if (zone == testZone) {
+        ().toResult
+      } else {
+        Left(ConnectionFailed(zone, s"DDNS connection failed").asInstanceOf[Throwable]).toResult
```
rebstar6: shouldnt need asInstanceOf here

modules/api/src/test/scala/vinyldns/api/domain/zone/ZoneConnectionValidatorSpec.scala
```diff
@@ -130,11 +138,68 @@ class ZoneConnectionValidatorSpec
...
         .resolve(testZone.name, testZone.name, RecordType.SOA)
+      doReturn(NoError(mock[Message]).toResult)
+        .when(mockDnsConnection)
+        .applyChange(any[RecordSetChange])
+
+      val result = awaitResultOf(underTest.validateZoneConnections(testZone).value)
+      result should be(right)
+    }
+
+    "respond with a success if DDNS connectivity test returns NXDOMAIN on delete" in {
```
rebstar6: these unit tests dont all have to be on validateZoneConnections, it makes the mocking hard to follow. test testDdnsConnectivity individually and then just have one test confirming that if testDdnsConnectivity fails, validateZoneConnections fails

docker/bind9/etc/named.conf.local
```diff
@@ -7,8 +7,13 @@
 //include "/etc/bind/zones.rfc1918";
 
 key "vinyldns." {
-  algorithm hmac-md5;
-  secret "nzisn+4G2ldMn0q1CV3vsg==";
+	algorithm hmac-sha512;
+	secret "wCZZS9lyRr77+jqfnkZ/92L9fD5ilmfrG0sslc3mgmTFsF1fRgmtJ0rj RkFITt8VHQ37wvM/nI9MAIWXYTvMqg==";
```
mitruly: To generate this TSIG key, I used: `dnssec-keygen -a HMAC-SHA512 -b 512 -n user vinyldns.`. This *is different* from not specifying `-n user` (since it will default to `-n host`). `dnssec-keygen` can be installed through bind-utils.

`ddns-confgen` invokes `tsig-keygen`. Not sure if `ddns-confgen` can be used to generate the same exact key, but `tsig-keygen` might.

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneViewLoader.scala
```diff
@@ -35,7 +35,7 @@ object DnsZoneViewLoader extends DnsConversions {
   def dnsZoneTransfer(zone: Zone): ZoneTransferIn = {
     val conn =
       zone.transferConnection.getOrElse(VinylDNSConfig.defaultTransferConnection).decrypted()
-    val TSIGKey = new TSIG(conn.keyName, conn.key)
+    val TSIGKey = new TSIG(TSIG.HMAC_SHA512, conn.keyName, conn.key)
```
mitruly: If we are switching to HMAC-SHA512 in this PR, this is more of a stop gap since different keys can have different algorithms. Need to make this more configurable, but don't want to blow up this PR.

modules/api/functional_test/conftest.py
```diff
@@ -19,8 +19,14 @@ def pytest_addoption(parser):
                      help="The zone name that will be used for testing")
     parser.addoption("--dns-key-name", dest="dns_key_name", action="store", default="vinyldns.",
                      help="The name of the key used to sign updates for the zone")
-    parser.addoption("--dns-key", dest="dns_key", action="store", default="nzisn+4G2ldMn0q1CV3vsg==",
-                     help="The tsig key")
+    parser.addoption("--dns-key", dest="dns_key", action="store",
+                    default="wCZZS9lyRr77+jqfnkZ/92L9fD5ilmfrG0sslc3mgmTFsF1fRgmtJ0rj RkFITt8VHQ37wvM/nI9MAIWXYTvMqg==",
```
pauljamescleary: do we still need this?  in other words, can't we use the new key in our internal vinyldns bind9 servers?

If we can use the same key, we can drop this all together, right?

modules/api/src/it/scala/vinyldns/api/domain/dns/DnsConversionsIntegrationSpec.scala
```diff
@@ -34,10 +34,18 @@ class DnsConversionsIntegrationSpec
...
     "test@test.com",
     ZoneStatus.Active,
-    connection =
-      Some(ZoneConnection("vinyldns.", "vinyldns.", "nzisn+4G2ldMn0q1CV3vsg==", "127.0.0.1:19001")),
-    transferConnection =
-      Some(ZoneConnection("vinyldns.", "vinyldns.", "nzisn+4G2ldMn0q1CV3vsg==", "127.0.0.1:19001"))
+    connection = Some(
+      ZoneConnection(
+        "vinyldns.",
+        "vinyldns.",
```
pauljamescleary: I am ok with replacing the existing "test" key, but we just need to be sure that we haven't broken things elsewhere (like our internal deployment)

modules/api/src/main/scala/vinyldns/api/domain/dns/DnsConversions.scala
```diff
@@ -511,14 +512,18 @@ trait DnsConversions {
...
+        update.delete(r)
+      }
       addRecords.foreach(update.add)
     }
     Right(update)
   }
 
   def toDeleteRecordMessage(r: DNS.RRset, zoneName: String): Either[Throwable, DNS.Update] = {
     val update = new DNS.Update(zoneDnsName(zoneName))
+    update.present(r.getName, r.getType)
```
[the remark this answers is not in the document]
pauljamescleary: Not sure if we should add this condition here.  In the `RecordSetChangeHandler` during the apply, the only thing we check for is `NoError`.  In reality, getting a `NXRRSET` back during a delete needs to be properly handled depending on context.  For the most part, we respect idempotency, so deleting a record set that doesn't exist is _ok_, and we should not error our.

It is ok if we don't get `NXRRSET` back in the validation.  As long as we swallow errors appropriately.

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneConnectionValidator.scala
pauljamescleary: Yea, this should be high.  Otherwise, it will expire in caching resolvers and those resolvers will query upstream to the authoritative masters for updates to the record.

The record set always hangs around with the current implementation, we just want to avoid frequent cache refreshes.

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneViewLoader.scala
mitruly: Created https://github.com/vinyldns/vinyldns/issues/156 so wrap up configurability of TSIG key algorithms.

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneConnectionValidator.scala
```diff
@@ -71,6 +72,31 @@ class ZoneConnectionValidator(defaultConnection: ZoneConnection, scheduler: Sche
...
+      RecordType.TXT,
+      86400,
+      RecordSetStatus.Pending,
+      DateTime.now,
+      records = List(TXTData("connection test")))
+
+    val result = for {
+      _ <- dnsConnection
+        .applyChange(RecordSetChange.forDelete(rs, zone))
+        .map(_ => ())
```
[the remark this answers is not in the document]
pauljamescleary: use `.as(())` instead of `.map(_ => ())`

modules/api/functional_test/live_tests/recordsets/list_recordset_changes_test.py
rebstar6: oh, hrmm. so every zone in vinyl will have this record in it after creation (?)

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneConnectionValidator.scala
```diff
@@ -67,6 +69,31 @@ class ZoneConnectionValidator(defaultConnection: ZoneConnection)
...
+        .recover {
+          case _: RecordSetNotFound | _: NameNotFound => ()
+        }
+      _ <- dnsConnection.applyChange(RecordSetChange.forAdd(rs, zone))
```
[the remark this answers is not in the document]
mitruly: In the case of a bad key, this is failing on the delete first. @pauljamescleary might be able to provide more insight, but I think we wanted to make sure there was some update of some sort.

I suppose a record existing would also run into a bad key error first, so it might suffice.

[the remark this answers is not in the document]
pauljamescleary: It probably makes sense to test the different cases.  What happens if we use a bad key?  What happens if the key is not authorized?

Should be rather simple to test on a docker bind9.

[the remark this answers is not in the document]
mitruly: We already have separate func tests that test both of these cases.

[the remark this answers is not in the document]
pauljamescleary: I think @mitruly is, do we need to do an `add` at all here?  Can we just issue the delete?  The idea is to verify that the key is valid and has access to make changes in the zone.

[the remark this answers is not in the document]
mitruly: @pauljamescleary Right, I was discussing offline with @rebstar6 and that makes sense. I'll make the changes accordingly. 👍

modules/api/functional_test/vinyldns_context.py
```diff
@@ -1,16 +1,18 @@
+import os
+
 class VinylDNSTestContext:
     dns_ip = 'localhost'
     dns_zone_name = 'vinyldns.'
     dns_rev_v4_zone_name = '30.172.in-addr.arpa.'
     dns_rev_v6_zone_name = '1.9.e.f.c.c.7.2.9.6.d.f.ip6.arpa.'
-    dns_key_name = 'vinyldns.'
-    dns_key = 'nzisn+4G2ldMn0q1CV3vsg=='
+    dns_key_name = os.getenv('DNS_KEY_NAME', 'vinyldns.')
```
pauljamescleary: Sorry if I wasn't clear on this.  If the keys are the same in _all_ environments, then we should be able to just hard code them here, as opposed to using environment variables.  Is there a reason that you decided to use environment variables?

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneViewLoader.scala
pauljamescleary: This will probably break the other place we create TSIG keys (`DnsConnection`).

We cannot assume the key algorithm of the backend.  Until now, we have used the default constructor for TSIG, which assumes an algorithm of HMAC_MD5

That has _always_ worked for us.

I am worried that this will break everything, especially if what we are using today is HMAC_MD5

modules/api/functional_test/vinyldns_context.py
mitruly: @pauljamescleary Changed this to hard code.

mitruly:
@pauljamescleary The `TSIG` algorithm was already changed in `DnsConnection` in https://github.com/vinyldns/vinyldns/pull/140/commits/5ccc17de069c876766562c4a6a317c1e07f5454e#diff-9ccc2444aca86af06e87f0385327bafa; otherwise the tests would not pass.

modules/docs/src/main/tut/api/create-zone.md
```diff
@@ -7,7 +7,14 @@ section: "api"
 # Create Zone
 
 Connects user to an existing zone. User must be a member of the group that has access to the zone. Connection info is optional,
-if no info is provided the default VinylDNS connections will be used
+if no info is provided the default VinylDNS connections will be used. 
+
+When attempting a zone connection, VinylDNS will attempt to perform a DDNS delete of a `TXT` record with name
```
pauljamescleary: is this still happening?

modules/api/functional_test/vinyldns_context.py
mitruly: The commit that I pushed up should have addressed this: https://github.com/vinyldns/vinyldns/pull/140/files#diff-4dc0a0a54250794c10d0d0ecc4f11369

modules/docs/src/main/tut/api/create-zone.md
mitruly: Yes, we are attempting a DDNS delete of the record still. If it succeeds, any existing record match will be deleted. I updated the verbiage from delete/add/lingering record.

pauljamescleary: Right, I was looking at an older commit where it said it was adding the record, this is good now

modules/api/functional_test/live_tests/zones/create_zone_test.py
```diff
@@ -7,7 +7,7 @@
 from utils import *
 
 records_in_dns = [
-    {'name': 'one-time.',
```
rebstar6: what happened to one-time?!

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneConnectionValidator.scala
```diff
@@ -67,6 +69,30 @@ class ZoneConnectionValidator(defaultConnection: ZoneConnection)
...
+      "vinyldns-ddns-connectivity-test",
+      RecordType.TXT,
+      86400,
+      RecordSetStatus.Pending,
+      DateTime.now,
+      records = List(TXTData("connection test")))
+
+    val result = for {
+      _ <- dnsConnection
+        .applyChange(RecordSetChange.forDelete(rs, zone))
```
rebstar6: cool, i like the delete-only

modules/api/functional_test/live_tests/zones/create_zone_test.py
mitruly: With the change, it would affect all one-time tests so opted to use a different zone

PR #147, opened 2018-08-28
WIP add zone tests to VinylDNS.scala

britneywright:


britneywright:
I was going to wait until I covered all the zone methods to remove the WIP, but wanted to see what the stats would look like so far.

britneywright:
based on the tests i wrote for regenerating credentials I think i could do some cleanup on these new tests I made. this issue can be ignored for the most part until I take the WIP off.

PR #160, opened 2018-08-30
move a bunch of stuff to core

rebstar6:
this will be fun.

tomorrow ill comment on what's what


modules/api/src/main/scala/vinyldns/api/domain/auth/AuthPrincipalProvider.scala
```diff
@@ -50,8 +52,11 @@ class MembershipAuthPrincipalProvider(
 }
 
 object MembershipAuthPrincipalProvider {
-  val userRepository: UserRepository = UserRepository()
-  val membershipRepository: MembershipRepository = MembershipRepository()
+  // TODO this has to be dynamic!!!
```
rebstar6: this is actually a problem (for later). this isnt tied into boot, gonna have to think about how we want to have this pull out these repos

pauljamescleary:
When you do this move, all of hte package names should change from `vinyldns.api` to `vinyldns.core`

pauljamescleary:
Also, brings up a good question about what to do with `VinylDNSConfig`.  `core` module should have a `reference.conf` with defaults on whatever is in core.

That said, I believe eventually all of the "real" things (dynamodb, mysql) will move to _separate_ repositories.  So, then what would be config in `core`?  I don't know there is any.

Should just be repository interfaces, queue interfaces (when we BYOQ), protobufs (since that is how we store things), and our domain models (whatever is in the repo or queue interfaces).  tl;dr; I don't know if we have any defaults reference.conf in core

pauljamescleary:
Are you also going to move over the `Jdbc` repos as well?  My guess is that we move all the repos to `core` as a "step", and then likely to separate submodules as a separate PR / step.

modules/api/src/main/scala/vinyldns/api/Boot.scala
```diff
@@ -74,14 +75,16 @@ object Boot extends App {
...
+          .toRight[Throwable](DataStoreStartupError("Missing zone repository")))
+      // TODO this also will all be removed with dynamic loading
+      userRepo <- IO(DynamoDBUserRepository())
+      groupRepo <- IO(DynamoDBGroupRepository())
+      membershipRepo <- IO(DynamoDBMembershipRepository())
+      groupChangeRepo <- IO(DynamoDBGroupChangeRepository())
+      recordSetRepo <- IO(DynamoDBRecordSetRepository())
+      recordChangeRepo <- IO(DynamoDBRecordChangeRepository())
+      zoneChangeRepo <- IO(DynamoDBZoneChangeRepository())
+      _ <- TestDataLoader.loadTestData(userRepo, groupRepo, membershipRepo)
```
rebstar6: hardcoded DynamoDB loads will change once we have the accessor working as planned. moved test data loading into it's own object for now (https://github.com/vinyldns/vinyldns/pull/160/files#diff-a0e66a954e4b01189947e1f8f0caf829)

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeErrors.scala
```diff
@@ -0,0 +1,50 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package vinyldns.api.domain.batch
+
+import vinyldns.api.domain.batch.BatchChangeInterfaces.ValidatedBatch
+import vinyldns.api.domain.batch.BatchTransformations.ChangeForValidation
+
+/* Error response options */
```
rebstar6: kept most errors in api

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchTransformations.scala
```diff
@@ -0,0 +1,141 @@
...
+package vinyldns.api.domain.batch
+
+import vinyldns.api.domain.dns.DnsConversions.{getIPv4NonDelegatedZoneName, getIPv6FullReverseName}
+import vinyldns.api.domain.record.{RecordSet, RecordSetChange}
+import vinyldns.api.domain.record.RecordType.RecordType
+import vinyldns.api.domain.zone.Zone
+
+/* Helper types for intermediate transformations of the batch data */
+// ALL of these are subject to change as implementation needs
+object BatchTransformations {
```
rebstar6: we had a lot in the batch file. these are still just helper types used as we transform things through the batch service

modules/api/src/main/scala/vinyldns/api/domain/dns/DnsConversions.scala
```diff
@@ -50,15 +52,6 @@ trait DnsConversions {
     else
       DNS.Name.fromString(recordName).relativize(zoneDnsName(zoneName)).toString
 
-  def omitTrailingDot(name: String): String =
```
rebstar6: these are just string methods we use all over. pulled into DomainHelpers in core

modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetChangeGenerator.scala
```diff
@@ -21,80 +21,9 @@ import java.util.UUID
...
-    sb.append("zoneName=\"").append(zone.name).append("\"; ")
-    sb.append(recordSet.toString)
-    sb.append(" ]")
-    sb.append("singleBatchChangeIds=\"").append(singleBatchChangeIds).append("\"; ")
-    sb.toString
-  }
-}
-
-object RecordSetChange extends DnsConversions {
+object RecordSetChangeGenerator extends DnsConversions {
```
rebstar6: wanted to keep this in api because of the reliance on DNSConversions

theres now both a RecordSetGenerator and ZoneChangeGenerator with all the forAdd/forUpdate/forDelete's

modules/api/src/it/scala/vinyldns/api/domain/dns/DnsConversionsIntegrationSpec.scala
```diff
@@ -45,7 +45,7 @@ class DnsConversionsIntegrationSpec
       val testRecord = aaaa.copy(zoneId = testZone.id)
       val conn = DnsConnection(testZone.connection.get)
       val result: DnsResponse =
-        rightResultOf(conn.addRecord(RecordSetChange.forAdd(testRecord, testZone)).value)
+        rightResultOf(conn.addRecord(RecordSetChangeGenerator.forAdd(testRecord, testZone)).value)
```
pauljamescleary: Why were these functions pulled out of `RecordSetChange`?

modules/api/src/it/scala/vinyldns/api/engine/ZoneCommandHandlerIntegrationSpec.scala
```diff
@@ -119,9 +121,10 @@ class ZoneCommandHandlerIntegrationSpec extends DynamoDBIntegrationSpec with Eve
     status = RecordSetStatus.Active,
     created = DateTime.now,
     records = List(AData("1.2.3.4")))
-  private val inDbRecordChange = ChangeSet(RecordSetChange.forSyncAdd(inDbRecordSet, testZone))
+  private val inDbRecordChange = ChangeSet(
+    RecordSetChangeGenerator.forSyncAdd(inDbRecordSet, testZone))
   private val inDbZoneChange =
-    ZoneChange.forUpdate(testZone.copy(email = "new@test.com"), testZone, okUserAuth)
+    ZoneChangeGenerator.forUpdate(testZone.copy(email = "new@test.com"), testZone, okUserAuth)
```
pauljamescleary: Same here

modules/api/src/main/scala/vinyldns/api/Instrumented.scala
```diff
@@ -23,10 +23,10 @@ object VinylDNSMetrics {
 
   val metricsRegistry: MetricRegistry = new MetricRegistry
 
-  // Output all VinylDNS metrics as jmx under the "vinyldns.api" domain as milliseconds
+  // Output all VinylDNS metrics as jmx under the "vinyldns.core" domain as milliseconds
   JmxReporter
     .forRegistry(metricsRegistry)
-    .inDomain("vinyldns.api")
+    .inDomain("vinyldns.core")
```
pauljamescleary: Not sure if this will break any dashboard work, probably worth giving the SRE guys a heads up

modules/api/src/main/scala/vinyldns/api/domain/auth/AuthPrincipalProvider.scala
pauljamescleary: Right, provider should be loaded in boot, and passed into somehow in the main VinylDNSService

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchTransformations.scala
```diff
@@ -0,0 +1,148 @@
...
+
+  object ChangeForValidation {
+    def apply(zone: Zone, recordName: String, changeInput: ChangeInput): ChangeForValidation =
+      changeInput match {
+        case a: AddChangeInput => AddChangeForValidation(zone, recordName, a)
+        case d: DeleteChangeInput => DeleteChangeForValidation(zone, recordName, d)
+      }
+  }
+
+  case class AddChangeForValidation(zone: Zone, recordName: String, inputChange: AddChangeInput)
```
pauljamescleary: I assume this is copy pasta, but trait should be sealed, case classes `final`

modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetChangeGenerator.scala
pauljamescleary: Would it be easier to move `relativize` to `DomainHelpers`, then can just keep the functions where they were?  I am not a fan of the `Generator`s, mostly because the functions live in the right place, the companion object of the domain type they work on.

modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetHelpers.scala
```diff
@@ -0,0 +1,44 @@
...
+
+import vinyldns.api.domain.dns.DnsConversions
+import vinyldns.core.domain.record.RecordSet
+
+object RecordSetHelpers {
+
+  def matches(left: RecordSet, right: RecordSet, zoneName: String): Boolean = {
+
+    val isSame = for {
+      thisDnsRec <- DnsConversions.toDnsRecords(left, zoneName)
```
pauljamescleary: It is all about these DnsConversions..*sigh*

PR #166, opened 2018-09-04
WIP lock users and check user lock status

### LINKED ISSUE #142: Lock User API Endpoint
Currently, there is no way to lock a user out / expire a user in the event a user leaves the organization.

1. Only super users can lock / un-lock a user
1. Create an API endpoint `POST /users/<user id>?locked=true|false` that locks a user.  Apply the change immediately and return a `200` on success.
1. If the user attempts any action (read or write) in the API, reply with a Forbidden
1. If the user attempts to login to the portal, notify the user that their account is presently locked.

britneywright:
closes #142 

This adds the ability for super users to lock user accounts and also adds a check of the user lock status in each API call and the portal credential requests.

This doesn't actually provide the ability to change user lock status, it just creates the endpoint, so there isn't a way to test that yet. We have to decide if we want to add that functionality to the portal or one of the tools, or both. 

To test the experience of being locked out go to the UserAccount.scala file in the portal module (https://github.com/vinyldns/vinyldns/blob/d303c5ea593aa4a24499ea7aa7828a568e3d374b/modules/portal/app/models/UserAccount.scala#L34) and change false to true.

### DIFF
--- a/modules/portal/app/controllers/VinylDNS.scala
+++ b/modules/portal/app/controllers/VinylDNS.scala
@@ -61,6 +61,7 @@ object VinylDNS {
       lastName: Option[String],
       email: Option[String],
       isSuper: Boolean,
+      isLocked: Boolean,
       id: String)
   object UserInfo {
     def fromAccount(account: UserAccount): UserInfo =
@@ -70,6 +71,7 @@ object VinylDNS {
         lastName = account.lastName,
         email = account.email,
         isSuper = account.isSuper,
+        isLocked = account.isLocked,
         id = account.userId
       )
   }
@@ -114,6 +116,34 @@ class VinylDNS @Inject()(
   implicit val userInfoReads: Reads[VinylDNS.UserInfo] = Json.reads[VinylDNS.UserInfo]
   implicit val userInfoWrites: Writes[VinylDNS.UserInfo] = Json.writes[VinylDNS.UserInfo]
 
+  private def checkSuperUser(username: String): Try[UserAccount] =
+    for {
+      account <- userAccountAccessor.get(username).flatMap {
+        case Some(account) if account.isSuper => Success(account)
+        case Some(account) if !account.isSuper =>
+          Failure(new UserIsNotAnAdminException(s"User account for $username is not an admin"))
+        case None =>
+          Failure(new UserDoesNotExistException(s"User account for $username not found"))
+      }
+    } yield account
+
+  private def withUnlockedUser(username: String): Try[UserAccount] =
+    for {
+      account <- userAccountAccessor.get(username).flatMap {
+        case Some(account) if !account.isLocked => Success(account)
+        case Some(account) if account.isLocked =>
+          Failure(new UserAccountIsLockedException(s"User account for $username is locked"))
+        case None => Failure(new UserDoesNotExistException(s"User account for $username not found"))
+      }
+    } yield account
+
+  private def errorResponse(e: Throwable): Result =
+    e match {
+      case e: UserAccountIsLockedException => Forbidden(e.getMessage).withHeaders(cacheHeaders: _*)
+      case e: UserIsNotAnAdminException => Unauthorized(e.getMessage).withHeaders(cacheHeaders: _*)
+      case _ => NotFound("User was not found").withHeaders(cacheHeaders: _*)
+    }
+
   def login(): Action[AnyContent] = Action { implicit request =>
     val userForm = Form(
       tuple(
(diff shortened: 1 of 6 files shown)


modules/portal/app/controllers/VinylDNS.scala
```diff
@@ -348,7 +459,7 @@ class VinylDNS @Inject()(
               }
             case Some(user) =>
               Logger.info(s"User account for ${user.username} exists with id ${user.userId}")
-              Success(user)
+              withUnlockedUser(username)
           }
           .recoverWith {
```
britneywright: We through an error here but I want to redirect to the login with the error message. Haven't quite figured it out. I also couldn't figure out how to test for the thrown exception in the current state.

modules/portal/app/controllers/VinylDNS.scala
```diff
@@ -114,6 +116,34 @@ class VinylDNS @Inject()(
   implicit val userInfoReads: Reads[VinylDNS.UserInfo] = Json.reads[VinylDNS.UserInfo]
   implicit val userInfoWrites: Writes[VinylDNS.UserInfo] = Json.writes[VinylDNS.UserInfo]
 
+  private def checkSuperUser(username: String): Try[UserAccount] =
+    for {
+      account <- userAccountAccessor.get(username).flatMap {
+        case Some(account) if account.isSuper => Success(account)
+        case Some(account) if !account.isSuper =>
```
pauljamescleary: patterns are matched top down; therefore the guard in the second pattern is unnecessary.

modules/portal/app/controllers/VinylDNS.scala
```diff
@@ -127,90 +157,123 @@ class VinylDNS @Inject()(
...
-      })
+    withAuthenticatedUser { username =>
+      withUnlockedUser(username) match {
+        case Success(_) => {
+          val json = request.body.asJson
+          val payload = json.map(Json.stringify)
+          val vinyldnsRequest =
+            new VinylDNSRequest("POST", s"$vinyldnsServiceBackend", "groups", payload)
+          val signedRequest =
+            signRequest(vinyldnsRequest, getUserCreds(request.session.get("accessKey")))
```
pauljamescleary: we call `getUserCreds` here, which also does a user account look up, so we would be making a database call twice for the same request check.

We should make this call once, which will mean this is a little bigger.  Instead of calling `getUserCreds` here, we only need to create the AwsCredentials...

```scala
new BasicAWSCredentials(account.accessKey, account.accessSecret)
```

To get the user account, instead of `case Success(_)` up above where we drop the account, actually pull it out `case Success(account)`

modules/portal/app/controllers/VinylDNS.scala
```diff
@@ -127,90 +157,123 @@ class VinylDNS @Inject()(
...
+          executeRequest(signedRequest).map(response => {
+            Logger.info(response.body)
+            Status(response.status)(response.body)
+              .withHeaders(cacheHeaders: _*)
+          })
+        }
+        case Failure(e) => Future(errorResponse(e))
```
pauljamescleary: It is more efficient to do a `Future.successful`  here, as `Future(x)` and `Future { }` both get run on a separate execution context, where as `Future.successful` returns immediately.

modules/portal/app/controllers/VinylDNS.scala
```diff
@@ -127,90 +157,123 @@ class VinylDNS @Inject()(
...
-        signRequest(vinyldnsRequest, getUserCreds(request.session.get("accessKey")))
-      executeRequest(signedRequest).map(response => {
-        Logger.info(s"group [$id] updated with status [${response.status}]")
-        Status(response.status)(response.body)
-          .withHeaders(cacheHeaders: _*)
-      })
+    withAuthenticatedUser { username =>
+      withUnlockedUser(username) match {
+        case Success(_) => {
+          val payload = request.body.asJson.map(Json.stringify)
```
pauljamescleary: seems like a lot of duplication across these, perhaps an opportunity to create a reusable function

pauljamescleary:
We should probably do a custom action here, it would localize the code and simplify the update.

We can pair on this...

https://www.playframework.com/documentation/2.6.x/ScalaActionsComposition

britneywright:
@pauljamescleary I was reading about the custom action builders, but i got stuck on how to pass in the needed user information.

britneywright:
closing since this is out of date, new approach in #186.

PR #168, opened 2018-09-05
move required monitor classes to core

rebstar6:
part 2 of a few for externalizing the repos, they rely on this trait

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/route/VinylDNSDirectives.scala
+++ b/modules/api/src/main/scala/vinyldns/api/route/VinylDNSDirectives.scala
@@ -74,17 +74,17 @@ trait VinylDNSDirectives extends Directives {
         try {
           inner(ctx)
             .map { result =>
-              record(Monitor(name), startTime)(result)
+              record(getMonitor(name), startTime)(result)
               result
             }
             .recoverWith {
               case nf @ NonFatal(e) =>
-                record(Monitor(name), startTime)(Failure(e))
+                record(getMonitor(name), startTime)(Failure(e))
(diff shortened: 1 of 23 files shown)


modules/api/src/main/scala/vinyldns/api/route/VinylDNSDirectives.scala
```diff
@@ -72,22 +74,52 @@ trait VinylDNSDirectives extends Directives {
...
         } catch {
           case nf @ NonFatal(e) =>
-            getMonitor(name).record(startTime)(Failure(e))
+            record(getMonitor(name), startTime)(Failure(e))
             ctx.fail(nf)
         }
       }
     }
 
+  // used to record stats about an http request / response
```
rebstar6: copied over from Monitored as this is the only file using it (and its akka dependencies...)

modules/core/src/main/scala/vinyldns/core/route/Monitor.scala
```diff
@@ -88,44 +85,14 @@ object Monitor {
   * @param name The name given to the thing we are monitoring, should be unique in the system
   */
 class Monitor(val name: String) extends Instrumented {
-  override lazy val metricBaseName = MetricName("vinyldns.api")
+  override lazy val metricBaseName = MetricName("vinyldns.core")
```
pauljamescleary: should probably have the base name passed in, this way it is not tied to core

pauljamescleary: nvm, would have to update the Reporter above as well

PR #170, opened 2018-09-05
External dynamo module

rebstar6:
have not yet touched the integration tests - this will fail


modules/dynamo/src/test/scala/vinyldns/dynamo/repository/DynamoDBGroupChangeRepositorySpec.scala
```diff
@@ -102,7 +96,7 @@ class DynamoDBGroupChangeRepositorySpec
         .when(dynamoDBHelper)
         .putItem(any[PutItemRequest])
 
-      val response = await[GroupChange](underTest.save(okGroupChange))
+      val response = underTest.save(okGroupChange).unsafeRunSync()
```
rebstar6: changed all the await[]s to just unsafeRunSync

modules/dynamo/src/test/scala/vinyldns/dynamo/repository/DynamoDBGroupChangeRepositorySpec.scala
```diff
@@ -111,10 +105,8 @@ class DynamoDBGroupChangeRepositorySpec
         .when(dynamoDBHelper)
         .putItem(any[PutItemRequest])
 
-      val result = underTest.save(okGroupChange).unsafeToFuture()
-      whenReady(result.failed) { failed =>
-        failed shouldBe a[ResourceNotFoundException]
-      }
+      val result = underTest.save(okGroupChange)
+      a[ResourceNotFoundException] shouldBe thrownBy(result.unsafeRunSync())
```
rebstar6: pulled out whenReady also

modules/dynamo/src/test/scala/vinyldns/dynamo/repository/QueryHelperSpec.scala
```diff
@@ -51,6 +48,9 @@ class QueryHelperSpec
     item
   }
 
+  def await[T](f: => IO[_]): T =
```
rebstar6: this is the only place the weird type mapping of await was needed. threw it in this class, but we can make a helper somewhere?

its pretty gross, so id rather avoid using it except when absolutely necessary for casting up in a test

modules/dynamo/src/main/scala/vinyldns/dynamo/repository/DynamoDBClient.scala
```diff
@@ -14,17 +14,16 @@
  * limitations under the License.
  */
 
-package vinyldns.api.repository.dynamodb
+package vinyldns.dynamo.repository
```
rebstar6: packaging ok for all these?

modules/api/src/it/scala/vinyldns/api/domain/record/RecordSetServiceIntegrationSpec.scala
```diff
@@ -30,7 +30,7 @@ import vinyldns.core.domain.membership.{Group, User, UserRepository}
 import vinyldns.core.domain.record.RecordType._
 import vinyldns.core.domain.zone.{Zone, ZoneRepository, ZoneStatus}
 import vinyldns.api.engine.sqs.TestSqsService
```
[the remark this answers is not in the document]
pauljamescleary: This will make the build a little tricky.  You can add the docker compose plugin to the dyamodb module, create a different docker-compose.yml file perhaps inside the new dynamodb module.  Have the new dynamodb docker compose file start up on different ports as to not conflict (in the event the api docker instances are up.  Should work fine.

https://github.com/Tapad/sbt-docker-compose

In SBT, If you do not have the api aggregate the dynamodb module (which it should not), things will work better (it can depend on it, just not aggregate it).

Obviously, as you add modules, need to re-examine our command aliases at the bottom of the main build.sbt

modules/dynamo/src/main/scala/vinyldns/dynamo/repository/DynamoDBClient.scala
pauljamescleary: Yep, that would follow the rest of the modules.

I do think you need to name the module dynamodb however (and its package).  Dynamo is a well known paper, but this is dynamodb is the AWS thing we use (db being important).

docker/docker-compose-build.yml
```diff
@@ -9,7 +9,7 @@ services:
     ports:
       - "19002:3306"
 
-  dynamodb:
+  dynamodb-local:
```
rebstar6: this being names the same thing as the dynamodb module gave me issues

modules/dynamodb/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,19 @@
...
+        <encoder>
+            <pattern>%-4relative [%thread] %-5level %logger{35} - %msg%n</pattern>
+        </encoder>
+        16
+    </appender>
+
+    <root level="DEBUG">
+        <appender-ref ref="FILE"/>
+    </root>
+</configuration>
```
pauljamescleary: newline

modules/dynamodb/src/test/scala/vinyldns/dynamodb/DynamoTestConfig.scala
```diff
@@ -0,0 +1,34 @@
...
+  lazy val config: Config = ConfigFactory.load()
+  lazy val vinyldnsConfig: Config = config.getConfig("vinyldns")
+  lazy val dynamoConfig: Config = vinyldnsConfig.getConfig("dynamo")
+  lazy val zoneChangeStoreConfig: Config = vinyldnsConfig.getConfig("zoneChanges")
+  lazy val recordSetStoreConfig: Config = vinyldnsConfig.getConfig("recordSet")
+  lazy val recordChangeStoreConfig: Config = vinyldnsConfig.getConfig("recordChange")
+  lazy val usersStoreConfig: Config = vinyldnsConfig.getConfig("users")
+  lazy val groupsStoreConfig: Config = vinyldnsConfig.getConfig("groups")
+  lazy val groupChangesStoreConfig: Config = vinyldnsConfig.getConfig("groupChanges")
+  lazy val membershipStoreConfig: Config = vinyldnsConfig.getConfig("membership")
```
pauljamescleary: in the near future, the "config" will probably be moved out of the api entirely, and into whatever dynamodbloader we build right?

The only reason I mention is because there should be no difference between "test config" and "normal config loading", as the context (src/it, src/test, src/main) will dictate what config gets loaded

modules/api/src/it/scala/vinyldns/api/DynamoDBApiIntegrationSpec.scala
```diff
@@ -0,0 +1,30 @@
...
+ * limitations under the License.
+ */
+
+package vinyldns.api
+
+import com.amazonaws.services.dynamodbv2.AmazonDynamoDBClient
+import org.slf4j.LoggerFactory
+import vinyldns.dynamodb.repository.{DynamoDBHelper, DynamoDBIntegrationSpec}
+
+trait DynamoDBApiIntegrationSpec extends DynamoDBIntegrationSpec {
```
pauljamescleary: At first I was asking "why do we need this?"  Then I remember that we have integration tests for the ZoneCommandHandler so we need our repositories.

We should discuss how we move forward with integration tests once things are moved out.  The api shouldn't have a direct dependency on dynamodb (or really mysql or sqs).  So, what do those "ZoneCommandHandlerIntegrationSpecs" look like after this breakout is complete?

pauljamescleary:
A few notes, just to keep in mind as you continue this breakout / move.  These are more "end state", I know we are in transition.

1. The API shouldn't directly depend on any of the broken out submodules.  We can hook them up in revolver (reStart) in sbt, and also in our func tests when we start the docker thingie.
1. I don't think the API should have _any_ integration tests.  At the end of the day, the `ZoneCommandHandler` is the only one we do some testing on.  We maybe able to redo with stubs as a normal unit test rather than an integration test.  The DnsConversions is the only other one there, and given that it is just testing converting from DNS lib to our lib, we should be able to move it to a normal unit test as well.
1. Have to be careful about configuration in this end game as well.  I imagine the api module would have all of the mysql / dyamodb configs removed.  Besides in reference.conf / application.conf, we should be moving all those into `DataStoreLoader` things, and so each module would be responsible for inspecting the `vinyldns` config namespace and pulling out its config.

Wanted to mention these while I thought of them, this breakout is extremely valuable, but we don't want to lose some of the additional cleanup items.

pauljamescleary:
One other food for thought here, why do we have all these unit tests _and_ an integration test?  The integration test is more exhaustive, and seemingly we are testing a lot of the same things as the unit tests, including conversion and such.  I will run an experiment to just run coverage on the integration specs and see where we land, will update this comment shortly.

modules/dynamodb/src/test/scala/vinyldns/dynamodb/DynamoTestConfig.scala
rebstar6: oh yeah config will change a bunch in general once were dynamically loading

PR #172, opened 2018-09-06
add locked account check to api authentication

britneywright:
This is part of #142 for the API side. Adds the "lockStatus" attribute to the the User class and database. The attribute is a custom type `LockStatus`, the value is either LockStatus.Locked or LockStatus.Unlocked.  Also adds a check of the attribute in the authentication.

### DIFF
--- a/modules/api/functional_test/live_tests/authentication_test.py
+++ b/modules/api/functional_test/live_tests/authentication_test.py
@@ -19,3 +19,11 @@ def test_request_fails_when_user_is_not_found():
     client = VinylDNSClient(VinylDNSTestContext.vinyldns_url, 'unknownAccessKey', 'anyAccessSecretKey')
 
     client.list_batch_change_summaries(status=401)
+
+def test_request_succeeds_when_user_is_found_and_not_locked():
+    """
+    Test request success with Success (200) when user account is found and not locked
+    """
+    client = VinylDNSClient(VinylDNSTestContext.vinyldns_url, 'okAccessKey', 'okSecretKey')
+
+    client.list_batch_change_summaries(status=200)
--- a/modules/api/functional_test/live_tests/authentication_test.py
+++ b/modules/api/functional_test/live_tests/authentication_test.py
@@ -1,0 +1,27 @@
+import pytest
+import sys
+import dns.query
+import dns.tsigkeyring
+import dns.update
+
+from utils import *
+from hamcrest import *
+from vinyldns_python import VinylDNSClient
+from dns.resolver import *
+from vinyldns_context import VinylDNSTestContext
+
+
+def test_request_fails_when_user_account_is_locked():
+    """
+    Test request fails with Forbidden (403) when user account is locked
+    """
+    client = VinylDNSClient(VinylDNSTestContext.vinyldns_url, 'lockedAccessKey', 'lockedSecretKey')
+    client.list_batch_change_summaries(status=403)
+
+def test_request_fails_when_user_is_not_found():
+    """
+    Test request fails with Unauthorized (401) when user account is not found
+    """
+    client = VinylDNSClient(VinylDNSTestContext.vinyldns_url, 'unknownAccessKey', 'anyAccessSecretKey')
+
+    client.list_batch_change_summaries(status=401)
--- a/modules/api/functional_test/live_tests/authentication_test.py
+++ b/modules/api/functional_test/live_tests/authentication_test.py
@@ -1,9 +1,3 @@
-import pytest
-import sys
-import dns.query
-import dns.tsigkeyring
-import dns.update
-
 from utils import *
 from hamcrest import *
 from vinyldns_python import VinylDNSClient
--- a/modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
@@ -232,12 +232,12 @@ class MembershipService(
 
   def updateUserLockStatus(
       userId: String,
-      setLockedTo: LockStatus,
+      lockStatus: LockStatus,
       authPrincipal: AuthPrincipal): Result[User] =
     for {
       _ <- isSuperAdmin(authPrincipal).toResult
       existingUser <- getExistingUser(userId)
-      newUser = existingUser.updateUserLockStatus(setLockedTo)
+      newUser = existingUser.updateUserLockStatus(lockStatus)
       _ <- userRepo.save(newUser).toResult[User]
     } yield newUser
 }
--- a/modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
@@ -55,7 +55,7 @@ class MembershipService(
     for {
       existingGroup <- getExistingGroup(groupId)
       newGroup = existingGroup.withUpdates(name, email, description, memberIds, adminUserIds)
-      _ <- isAdmin(existingGroup, authPrincipal).toResult
+      _ <- isGroupAdmin(existingGroup, authPrincipal).toResult
       addedMembers = newGroup.memberIds.diff(existingGroup.memberIds)
       removedMembers = existingGroup.memberIds.diff(newGroup.memberIds)
       _ <- hasMembersAndAdmins(newGroup).toResult
@@ -72,7 +72,7 @@ class MembershipService(
   def deleteGroup(groupId: String, authPrincipal: AuthPrincipal): Result[Group] =
     for {
       existingGroup <- getExistingGroup(groupId)
-      _ <- isAdmin(existingGroup, authPrincipal).toResult
+      _ <- isGroupAdmin(existingGroup, authPrincipal).toResult
       _ <- groupCanBeDeleted(existingGroup)
       _ <- groupChangeRepo
         .save(GroupChange.forDelete(existingGroup, authPrincipal))
--- a/modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
@@ -19,6 +19,7 @@ package vinyldns.api.domain.membership
 import cats.implicits._
 import vinyldns.api.Interfaces._
 import vinyldns.core.domain.auth.AuthPrincipal
+import vinyldns.core.domain.membership.LockStatus.LockStatus
 import vinyldns.core.domain.zone.ZoneRepository
 import vinyldns.core.domain.membership._
 
@@ -231,7 +232,7 @@ class MembershipService(
 
   def updateUserLockStatus(
       userId: String,
-      setLockedTo: Boolean,
+      setLockedTo: LockStatus,
       authPrincipal: AuthPrincipal): Result[User] =
     for {
       _ <- isSuperAdmin(authPrincipal).toResult
--- a/modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
@@ -174,6 +174,12 @@ class MembershipService(
       .getUsers(userIds, startFrom, pageSize)
       .toResult[ListUsersResults]
 
+  def getExistingUser(userId: String): Result[User] =
+    userRepo
+      .getUser(userId)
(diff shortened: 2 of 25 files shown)


modules/api/src/main/scala/vinyldns/api/repository/TestDataLoader.scala
```diff
@@ -142,8 +167,9 @@ object TestDataLoader {
     List(okGroup1, okGroup2).map(repository.save(_)).parSequence
 
   def loadUserTestData(repository: UserRepository): IO[List[User]] =
-    (testUser :: okUser :: dummyUser :: listGroupUser :: listZonesUser :: listBatchChangeSummariesUser ::
-      listZeroBatchChangeSummariesUser :: zoneHistoryUser :: listOfDummyUsers).map { user =>
+    (testUser :: okUser :: dummyUser :: lockedUser :: superUser :: listGroupUser :: listZonesUser ::
```
pauljamescleary: gotta figure this out.  We cannot load a test user that is a super user.  Hard stop on this.

britneywright: Can I copy existing user and change the status in specific tests? Like we do in the JdbcZoneRepositoryIntegrationSpec? https://github.com/vinyldns/vinyldns/blob/d0bcb83bdd2439e1f89b346cff000f103f233b95/modules/api/src/it/scala/vinyldns/api/repository/mysql/JdbcZoneRepositoryIntegrationSpec.scala#L104

pauljamescleary: we can never have super users in the test data as they are persistent.  Since you never actually lock in a func test, you shouldn't need it.  Your func tests only test an operation on a locked user, not actually updating the user.

So it would seem, simply deleting the `superUser` from the test data would be sufficient.

modules/api/src/test/scala/vinyldns/api/domain/membership/MembershipServiceSpec.scala
```diff
@@ -85,6 +85,16 @@ class MembershipServiceSpec
...
   )
 
+  // the update will set isLocked to true
+  private val updatedToLockUserInfo = okUser.copy(
+    isLocked = true
+  )
+
+  // the update will set isLocked to false
+  private val updatedToUnlockUserInfo = lockedUser.copy(
+    isLocked = true
```
pauljamescleary: hrm, updates isLocked to true?

modules/dynamodb/src/it/scala/vinyldns/dynamodb/repository/DynamoDBUserRepositoryIntegrationSpec.scala
```diff
@@ -141,5 +141,27 @@ class DynamoDBUserRepositoryIntegrationSpec extends DynamoDBIntegrationSpec {
       result shouldBe Some(testUser)
       result.get.isSuper shouldBe false
     }
+    "returns the locked flag when true" in {
```
pauljamescleary: here, you are testing the locked flag, you would not need the test data up above for this to work (you do a save and then a get in each integration test)

britneywright: i don't need it for the unlocked status since that's the default value, but i do need it here right? i need to create a user with a locked status and then check that it saved correctly.

modules/api/src/test/scala/vinyldns/api/domain/membership/MembershipServiceSpec.scala
britneywright: hmm yup. guess i don't need that value

modules/api/functional_test/live_tests/authentication_test.py
```diff
@@ -0,0 +1,27 @@
...
+    client = VinylDNSClient(VinylDNSTestContext.vinyldns_url, 'lockedAccessKey', 'lockedSecretKey')
+    client.list_batch_change_summaries(status=403)
+
+def test_request_fails_when_user_is_not_found():
+    """
+    Test request fails with Unauthorized (401) when user account is not found
+    """
+    client = VinylDNSClient(VinylDNSTestContext.vinyldns_url, 'unknownAccessKey', 'anyAccessSecretKey')
+
+    client.list_batch_change_summaries(status=401)
```
mitruly: Missing new line.

modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipValidations.scala
```diff
@@ -33,6 +33,11 @@ object MembershipValidations {
       group.adminUserIds.contains(authPrincipal.userId) || authPrincipal.signedInUser.isSuper
     }
 
+  def isSuperAdmin(authPrincipal: AuthPrincipal): Either[Throwable, Unit] =
```
mitruly: If we're creating `isSuperAdmin`, I think it makes sense to break off `isGroupAdmin` from `isAdmin` as well.

modules/api/functional_test/live_tests/authentication_test.py
```diff
@@ -0,0 +1,27 @@
+import pytest
+import sys
+import dns.query
+import dns.tsigkeyring
+import dns.update
```
mitruly: This top group of imports don't appear to be used. Deleted them and func tests still ran and passed.

modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
```diff
@@ -222,4 +228,15 @@ class MembershipService(
         }
       }
       .toResult
+
+  def updateUserLockStatus(
+      userId: String,
+      setLockedTo: Boolean,
```
mitruly: We are already using `isLocked` on the class and "lock status" for method names. I think changing `setLockedTo` to `lockStatus` is better.

modules/api/src/main/scala/vinyldns/api/repository/TestDataLoader.scala
```diff
@@ -51,7 +51,32 @@ object TestDataLoader {
...
+    created = DateTime.now.secondOfDay().roundFloorCopy(),
+    accessKey = "lockedAccessKey",
+    secretKey = "lockedSecretKey",
+    firstName = Some("Locked"),
+    lastName = Some("User"),
+    email = Some("testlocked@test.com"),
+    isSuper = false,
+    isLocked = true
+  )
+  final val superUser = User(
```
pauljamescleary: are we using this anywhere?  I would remove this from this file entirely just to resist anyone from ever storing the super user in our persistent test data.

modules/dynamodb/src/it/scala/vinyldns/dynamodb/repository/DynamoDBUserRepositoryIntegrationSpec.scala
pauljamescleary: correct, you can do it for the integration spec

modules/api/src/main/scala/vinyldns/api/repository/TestDataLoader.scala
britneywright: I'm using it in MembershipServiceSpec.scala and MembershipRoutingSpec.scala

modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
```diff
@@ -222,4 +228,15 @@ class MembershipService(
...
+      authPrincipal: AuthPrincipal): Result[User] =
+    for {
+      _ <- isSuperAdmin(authPrincipal).toResult
+      existingUser <- getExistingUser(userId)
+      newUser = existingUser.updateUserLockStatus(setLockedTo)
+      _ <- userRepo.save(newUser).toResult[User]
```
mitruly: Looking at `updateUserLockStatus`, we are indiscriminately updating the lock status. This means that we are potentially making database update calls with the same information.

Not sure how to do a conditional step within a for-comprehension, but doing the `userRepo.save` step makes sense if any of the fields in `existingUser` and `newUser` are different.

modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
britneywright: i preferred `setLockedTo` so it's clearly indicating what it's changing to rather than what it currently is. I could easily see lockStatus being used to check what a user's current locked status is so want to avoid any confusion now or in the future.

modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
britneywright: @pauljamescleary thoughts here? Michael brings up a good point, but we don't want to fail if the user already has the desired lock status, we would just want to skip the saving essentially.

britneywright:
> Also wary about the coverage drop for `VinylDNSDirectives.scala`.

That's because I added more lines to the function. It's tested in the func tests, the new authentication_test.py, but that's not included in CodeCov. We weren't testing that explicitly previously.

modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
pauljamescleary: we could use a little type here so it is more clear.  As opposed to `Boolean` we could use...

```scala
sealed trait UserStatus
object UserStatus {
  case object Locked extends UserStatus
  case object Unlocked extends UserStatus
```

That would make the code a little easier?

modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
pauljamescleary: It is a good point; however, not consistent with the rest of the system.  Also, not sure if it is worth the added complexity since users will be locked very rarely.

modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
```diff
@@ -174,6 +175,12 @@ class MembershipService(
       .getUsers(userIds, startFrom, pageSize)
       .toResult[ListUsersResults]
 
+  def getExistingUser(userId: String): Result[User] =
```
rebstar6: maybe this is overkill, but i kinda thing this should live in a UserService separate of the MembershipService. I wont hold up the PR on that though

modules/api/src/main/scala/vinyldns/api/route/MembershipJsonProtocol.scala
```diff
@@ -39,6 +40,7 @@ object MembershipJsonProtocol {
       description: Option[String],
       members: Set[UserInfo],
       admins: Set[UserInfo])
+  final case class UpdateUserInput(id: String, lockStatus: LockStatus)
```
rebstar6: name this something specific to locking. I can see us having an update on user at some point.

`final case class UserLockingInput(id: String, lockStatus: LockStatus)` maybe?

modules/api/src/test/scala/vinyldns/api/route/VinylDNSAuthenticatorSpec.scala
```diff
@@ -88,7 +83,7 @@ class VinylDNSAuthenticatorSpec
         .authenticateReq(any[HttpRequest], any[List[String]], any[String], any[String])
 
       val result =
-        await[Either[Cause, AuthPrincipal]](underTest.apply(context, ""))
+        await[Either[VinylDNSAuthenticationError, AuthPrincipal]](underTest.apply(context, ""))
```
rebstar6: await isnt needed anymore in 90% of cases. Trying to get away from it. This is just:

`val result = underTest.apply(context, "").unsafeRunSync()`

rebstar6: can change all to ^^ and i believe then get rid of the `ResultHelpers` import

modules/api/src/main/scala/vinyldns/api/route/MembershipJsonProtocol.scala
britneywright: actually I don't even think I need this.

britneywright:
Increased VinylDNSDirectives.scala from 37.50% coverage to 46.43% coverage.

PR #185, opened 2018-09-07
Create UserChangeRepository

pauljamescleary:
Interim step to pull the repos out of the portal.  This PR does not address strongly typed secrets covered in #193 .

Following PRs will pull the user repository _and_ the user change repository into the portal.

**Changes**

* Created a UserChangeRepository trait that will eventually replace
the DynamoDBChangeLogStore in the portal
* Created a DynamoDBUserChangeRepository implementation in the
dynamodb module
* Externalized the fromItem and toItem in the DynamoDBUserRepository
as those are needed by the DynamoDBUserChangeRepository to serialize
the new and old versions of the user
* Added CryptoAlgebra to the DynamoDBUserRepository as we need to
encrypt the secret when it is stored.

Note: users do not have protobufs like the other repositories, rather it saves things as attributes and attribute maps (getM).  Rather than re-writing both the user repository AND the user change repository, we are going to stay with plain old dynamo attributes here.

Also, the **existing** UserChangeLog from the portal uses `timestamp` as the `HASH` key in the table.  This is undesireable / terribly inconsistent.  To make it more consistent, we will store the UUID for user changes.  This will force the creation of a new table when this is merged, an update to our deployment, and eventually a migration of the change log data.

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/Boot.scala
+++ b/modules/api/src/main/scala/vinyldns/api/Boot.scala
@@ -64,7 +64,7 @@ object Boot extends App {
     // that if anything fails, the app does not start!
     for {
       banner <- vinyldnsBanner()
-      crypto <- Crypto.loadCrypto(VinylDNSConfig.cryptoConfig) // load crypto
+      crypto <- IO(Crypto.instance) // load crypto
       // TODO datastore loading will not be hardcoded by type here
       mySqlDataStore <- new MySqlDataStoreProvider().load(VinylDNSConfig.mySqlConfig)
       zoneRepo <- IO.fromEither(
--- a/modules/api/src/main/scala/vinyldns/api/Boot.scala
+++ b/modules/api/src/main/scala/vinyldns/api/Boot.scala
@@ -66,7 +66,7 @@ object Boot extends App {
       banner <- vinyldnsBanner()
       crypto <- IO(Crypto.instance) // load crypto
       // TODO datastore loading will not be hardcoded by type here
-      mySqlDataStore <- new MySqlDataStoreProvider().load(VinylDNSConfig.mySqlConfig)
+      mySqlDataStore <- new MySqlDataStoreProvider().load(VinylDNSConfig.mySqlConfig, crypto)
       zoneRepo <- IO.fromEither(
         mySqlDataStore
           .get[ZoneRepository](RepositoryName.zone)
--- a/modules/api/src/main/scala/vinyldns/api/Boot.scala
+++ b/modules/api/src/main/scala/vinyldns/api/Boot.scala
@@ -64,7 +64,7 @@ object Boot extends App {
     // that if anything fails, the app does not start!
     for {
       banner <- vinyldnsBanner()
-      _ <- Crypto.loadCrypto(VinylDNSConfig.cryptoConfig) // load crypto
+      crypto <- Crypto.loadCrypto(VinylDNSConfig.cryptoConfig) // load crypto
       // TODO datastore loading will not be hardcoded by type here
       mySqlDataStore <- new MySqlDataStoreProvider().load(VinylDNSConfig.mySqlConfig)
       zoneRepo <- IO.fromEither(
@@ -78,7 +78,9 @@ object Boot extends App {
       // TODO this also will all be removed with dynamic loading
       userRepo <- DynamoDBUserRepository(
         VinylDNSConfig.usersStoreConfig,
-        VinylDNSConfig.dynamoConfig)
+        VinylDNSConfig.dynamoConfig,
+        crypto
+      )
       groupRepo <- DynamoDBGroupRepository(
         VinylDNSConfig.groupsStoreConfig,
         VinylDNSConfig.dynamoConfig)
--- a/modules/api/src/main/scala/vinyldns/api/domain/auth/AuthPrincipalProvider.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/auth/AuthPrincipalProvider.scala
@@ -18,6 +18,7 @@ package vinyldns.api.domain.auth
 
 import cats.effect._
 import vinyldns.api.VinylDNSConfig
+import vinyldns.api.crypto.Crypto
 import vinyldns.core.domain.membership.{MembershipRepository, User, UserRepository}
 import vinyldns.dynamodb.repository.{DynamoDBMembershipRepository, DynamoDBUserRepository}
 import vinyldns.core.domain.auth.AuthPrincipal
@@ -55,7 +56,10 @@ class MembershipAuthPrincipalProvider(
 object MembershipAuthPrincipalProvider {
   // TODO this has to be dynamic!!!
   val userRepository: UserRepository =
-    DynamoDBUserRepository(VinylDNSConfig.usersStoreConfig, VinylDNSConfig.dynamoConfig)
+    DynamoDBUserRepository(
+      VinylDNSConfig.usersStoreConfig,
+      VinylDNSConfig.dynamoConfig,
+      Crypto.instance)
       .unsafeRunSync()
   val membershipRepository: MembershipRepository =
     DynamoDBMembershipRepository(VinylDNSConfig.membershipStoreConfig, VinylDNSConfig.dynamoConfig)
--- a/modules/core/src/main/scala/vinyldns/core/crypto/NoOpCrypto.scala
+++ b/modules/core/src/main/scala/vinyldns/core/crypto/NoOpCrypto.scala
@@ -16,7 +16,7 @@
 
 package vinyldns.core.crypto
 
-import com.typesafe.config.Config
+import com.typesafe.config.{Config, ConfigFactory}
 
 /**
   * Provides a no op implementation of Crypto.  Does not actually do any encryption.
@@ -27,6 +27,9 @@ import com.typesafe.config.Config
   *               to dynamically load this Crypto implementation.  However, it is not used.
   */
 class NoOpCrypto(val config: Config) extends CryptoAlgebra {
+  def this() {
+    this(ConfigFactory.load())
+  }
   def encrypt(value: String): String = value
   def decrypt(value: String): String = value
 }
--- a/modules/core/src/main/scala/vinyldns/core/domain/membership/UserChange.scala
+++ b/modules/core/src/main/scala/vinyldns/core/domain/membership/UserChange.scala
@@ -32,8 +32,8 @@ object UserChangeType {
     }
 
   def fromChange[A <: UserChange](change: A): UserChangeType = change match {
-    case CreateUser(_, _, _, _) => Create
-    case UpdateUser(_, _, _, _, _) => Update
+    case UserChange.CreateUser(_, _, _, _) => Create
+    case UserChange.UpdateUser(_, _, _, _, _) => Update
   }
 }
 
@@ -47,18 +47,16 @@ sealed trait UserChange {
   def madeByUserId: String
   def created: DateTime
 }
-final case class CreateUser(id: String, newUser: User, madeByUserId: String, created: DateTime)
-    extends UserChange
-
-final case class UpdateUser(
-    id: String,
-    newUser: User,
-    madeByUserId: String,
-    created: DateTime,
-    oldUser: User)
-    extends UserChange
-
 object UserChange {
+  final case class CreateUser(id: String, newUser: User, madeByUserId: String, created: DateTime)
+      extends UserChange
+  final case class UpdateUser(
+      id: String,
+      newUser: User,
+      madeByUserId: String,
+      created: DateTime,
+      oldUser: User)
+      extends UserChange
 
   def apply(
       id: String,
--- a/modules/core/src/main/scala/vinyldns/core/domain/membership/UserChange.scala
+++ b/modules/core/src/main/scala/vinyldns/core/domain/membership/UserChange.scala
@@ -1,0 +1,57 @@
+/*
+ * Copyright 2018 Comcast Cable Communications Management, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package vinyldns.core.domain.membership
+import java.util.UUID
+
+import org.joda.time.DateTime
+import vinyldns.core.domain.auth.AuthPrincipal
+import vinyldns.core.domain.membership.UserChangeType.UserChangeType
+
+object UserChangeType extends Enumeration {
+  type UserChangeType = Value
+
+  // Note: we do not have Delete yet, Update would be for locking and unlocking users
+  val Create, Update = Value
+}
+
+final case class UserChange(
+    newUser: User,
+    changeType: UserChangeType,
+    madeByUserId: String,
+    oldUser: Option[User],
+    id: String,
+    created: DateTime)
+
+object UserChange {
+  def forAdd(user: User, authPrincipal: AuthPrincipal): UserChange =
+    UserChange(
+      user,
+      UserChangeType.Create,
+      authPrincipal.userId,
+      None,
+      UUID.randomUUID().toString,
+      DateTime.now)
+
+  def forUpdate(newUser: User, oldUser: User, authPrincipal: AuthPrincipal): UserChange =
+    UserChange(
+      newUser,
+      UserChangeType.Update,
+      authPrincipal.userId,
+      Some(oldUser),
+      UUID.randomUUID().toString,
+      DateTime.now)
+}
--- a/modules/core/src/main/scala/vinyldns/core/domain/membership/UserChange.scala
+++ b/modules/core/src/main/scala/vinyldns/core/domain/membership/UserChange.scala
@@ -15,43 +15,65 @@
  */
 
 package vinyldns.core.domain.membership
-import java.util.UUID
-
 import org.joda.time.DateTime
-import vinyldns.core.domain.auth.AuthPrincipal
-import vinyldns.core.domain.membership.UserChangeType.UserChangeType
 
-object UserChangeType extends Enumeration {
-  type UserChangeType = Value
+sealed abstract class UserChangeType(val value: String)
+object UserChangeType {
+  case object Create extends UserChangeType("create")
+  case object Update extends UserChangeType("delete")
 
-  // Note: we do not have Delete yet, Update would be for locking and unlocking users
-  val Create, Update = Value
+  case class UnknownUserChangeType(value: String) extends Throwable(s"Unknown change type $value")
+
+  def fromString(value: String): Either[UnknownUserChangeType, UserChangeType] =
+    value.toLowerCase match {
+      case UserChangeType.Create.value => Right(Create)
+      case UserChangeType.Update.value => Right(Update)
+      case other => Left(UnknownUserChangeType(other))
+    }
+
+  def fromChange[A <: UserChange](change: A): UserChangeType = change match {
+    case CreateUser(_, _, _, _) => Create
+    case UpdateUser(_, _, _, _, _) => Update
+  }
 }
 
-final case class UserChange(
-    newUser: User,
-    changeType: UserChangeType,
-    madeByUserId: String,
-    oldUser: Option[User],
+/**
+  * The nice thing about this is that we cannot have a UserChange in an invalid state.  For example,
+  * we cannot have a UserChange(ChangeType.Update, oldUser=None)
+  */
+sealed trait UserChange {
+  def id: String
+  def newUser: User
+  def madeByUserId: String
+  def created: DateTime
+}
+final case class CreateUser(id: String, newUser: User, madeByUserId: String, created: DateTime)
+    extends UserChange
+
+final case class UpdateUser(
     id: String,
-    created: DateTime)
+    newUser: User,
+    madeByUserId: String,
+    created: DateTime,
+    oldUser: User)
+    extends UserChange
 
 object UserChange {
-  def forAdd(user: User, authPrincipal: AuthPrincipal): UserChange =
-    UserChange(
-      user,
-      UserChangeType.Create,
-      authPrincipal.userId,
-      None,
-      UUID.randomUUID().toString,
-      DateTime.now)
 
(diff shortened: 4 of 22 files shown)


modules/dynamodb/src/main/scala/vinyldns/dynamodb/repository/DynamoDBUserRepository.scala
```diff
@@ -25,36 +25,75 @@ import com.amazonaws.services.dynamodbv2.model._
...
+    )
+
+  /** Moving these so they can be reused by the user change repo */
+  def toItem(crypto: CryptoAlgebra, user: User): java.util.Map[String, AttributeValue] = {
+    val item = new java.util.HashMap[String, AttributeValue]()
+    item.put(USER_ID, new AttributeValue(user.id))
+    item.put(USER_NAME, new AttributeValue(user.userName))
+    item.put(CREATED, new AttributeValue().withN(user.created.getMillis.toString))
+    item.put(ACCESS_KEY, new AttributeValue(user.accessKey))
+    item.put(SECRET_KEY, new AttributeValue(crypto.encrypt(user.secretKey)))
```
pauljamescleary: this cryptography is hap-hazard.  We really should have a tiny type like "Secret" or something that makes this more explicit.  Currently, there is no way to know that a `String` is encrypted (we make cryptography idempotent so we don;t double encrypt).

Something like 

```scala
sealed trait Secret
final case class private Encrypted(value: String) extends AnyVal
final case class private Decrypted(value: String) extends AnyVal

object Encrypted {
  def apply(str: String, crypto: CryptoAlgebra): Encrypted = new Encrypted(crypto.encrypt(str))
}
```

modules/dynamodb/src/main/scala/vinyldns/dynamodb/repository/DynamoDBUserChangeRepository.scala
```diff
@@ -0,0 +1,110 @@
...
+  private val dynamoReads = config.getLong("dynamo.provisionedReads")
+  private val dynamoWrites = config.getLong("dynamo.provisionedWrites")
+
+  dynamoDBHelper.setupTable(
+    new CreateTableRequest()
+      .withTableName(USER_CHANGE_TABLE)
+      .withAttributeDefinitions(tableAttributes: _*)
+      .withKeySchema(new KeySchemaElement(USER_CHANGE_ID, KeyType.HASH))
+      .withProvisionedThroughput(new ProvisionedThroughput(dynamoReads, dynamoWrites))
+  )
```
pauljamescleary: we have no indexes today, no need to add any.  Just use the change id as the HASH

modules/api/src/main/scala/vinyldns/api/Boot.scala
```diff
@@ -78,7 +78,9 @@ object Boot extends App {
       // TODO this also will all be removed with dynamic loading
       userRepo <- DynamoDBUserRepository(
         VinylDNSConfig.usersStoreConfig,
-        VinylDNSConfig.dynamoConfig)
+        VinylDNSConfig.dynamoConfig,
+        crypto
```
pauljamescleary: this is needed as we need to encrypt users when they are saved (in order to supplant the portal repos)

modules/api/src/main/scala/vinyldns/api/domain/auth/AuthPrincipalProvider.scala
```diff
@@ -55,7 +56,10 @@ class MembershipAuthPrincipalProvider(
 object MembershipAuthPrincipalProvider {
   // TODO this has to be dynamic!!!
   val userRepository: UserRepository =
-    DynamoDBUserRepository(VinylDNSConfig.usersStoreConfig, VinylDNSConfig.dynamoConfig)
+    DynamoDBUserRepository(
+      VinylDNSConfig.usersStoreConfig,
+      VinylDNSConfig.dynamoConfig,
+      Crypto.instance)
```
pauljamescleary: same here.  Also, this is why we load the repo twice.  We should pass in the repo to an AuthPrincipalProvider instance.  Not in this PR

modules/dynamodb/src/main/scala/vinyldns/dynamodb/repository/DynamoDBUserChangeRepository.scala
```diff
@@ -0,0 +1,132 @@
...
+    val dynamoDBHelper: DynamoDBHelper,
+    serialize: UserChange => java.util.Map[String, AttributeValue],
+    deserialize: java.util.Map[String, AttributeValue] => UserChange)
+    extends UserChangeRepository
+    with Monitored {
+  import DynamoDBUserChangeRepository._
+
+  private val logger = LoggerFactory.getLogger(classOf[DynamoDBUserChangeRepository])
+
+  def save(change: UserChange): IO[UserChange] =
```
rebstar6: is the secret also stored in the change? does this not also need crypto, or will anything private be encrypted before the repo call?

modules/dynamodb/src/main/scala/vinyldns/dynamodb/repository/DynamoDBUserRepository.scala
pauljamescleary: Created https://github.com/vinyldns/vinyldns/issues/193 as a response.

modules/dynamodb/src/main/scala/vinyldns/dynamodb/repository/DynamoDBUserChangeRepository.scala
pauljamescleary: We call the `UserRepository.toItem` to make sure it is encrypted.

I created https://github.com/vinyldns/vinyldns/issues/193 to properly handle encryption using strong types.  I think that Issue requires a little more thinking, but the stringly typed nature of these values is uncomfortable.

modules/api/src/main/scala/vinyldns/api/Boot.scala
```diff
@@ -64,7 +64,7 @@ object Boot extends App {
     // that if anything fails, the app does not start!
     for {
       banner <- vinyldnsBanner()
-      _ <- Crypto.loadCrypto(VinylDNSConfig.cryptoConfig) // load crypto
+      crypto <- Crypto.loadCrypto(VinylDNSConfig.cryptoConfig) // load crypto
```
rebstar6: hmm should we jsut be calling Crypto.instance here?

rebstar6: then can pass that in

modules/core/src/main/scala/vinyldns/core/domain/membership/UserChange.scala
```diff
@@ -0,0 +1,57 @@
...
+final case class UserChange(
+    newUser: User,
+    changeType: UserChangeType,
+    madeByUserId: String,
+    oldUser: Option[User],
+    id: String,
+    created: DateTime)
+
+object UserChange {
+  def forAdd(user: User, authPrincipal: AuthPrincipal): UserChange =
```
rebstar6: how are we feeling about the forAdd/forUpdate thing in general? that got moved into generators for zone and recordset, and i was always kinda meh on it

modules/dynamodb/src/main/scala/vinyldns/dynamodb/repository/DynamoDBUserChangeRepository.scala
```diff
@@ -0,0 +1,132 @@
...
+  val USER_NAME: String = "username"
+  val CHANGE_TYPE: String = "change_type"
+  val NEW_USER: String = "new_user"
+  val OLD_USER: String = "old_user"
+
+  val TABLE_ATTRIBUTES: Seq[AttributeDefinition] = Seq(
+    new AttributeDefinition(USER_CHANGE_ID, "S")
+  )
+
+  final case class Settings(tableName: String, provisionedReads: Long, provisionedWrites: Long)
```
rebstar6: DynamoDBRepositorySettings?

rebstar6: unused, remove it

modules/dynamodb/src/main/scala/vinyldns/dynamodb/repository/DynamoDBUserChangeRepository.scala
```diff
@@ -0,0 +1,132 @@
...
+    val deserialize: java.util.Map[String, AttributeValue] => UserChange = fromItem
+    setup.as(
+      new DynamoDBUserChangeRepository(config.tableName, dynamoDBHelper, serialize, deserialize))
+  }
+}
+
+class DynamoDBUserChangeRepository private[repository] (
+    tableName: String,
+    val dynamoDBHelper: DynamoDBHelper,
+    serialize: UserChange => java.util.Map[String, AttributeValue],
```
rebstar6: this doesnt make sense, why pass in serializers for the changeRepo, but the crypto for the userRepo? do the same for both - either just save the crypto here, or pass in to/fromItem to the userRepo as well

modules/core/src/main/scala/vinyldns/core/domain/membership/UserChange.scala
pauljamescleary: they are more or less helper functions.  Kinda dumbs down what you have to set.

I would prefer that `UserChange` were more strongly typed.  I could do that here since we are not using protobufs, but it would deviate from the rest of the system.

modules/dynamodb/src/main/scala/vinyldns/dynamodb/repository/DynamoDBUserChangeRepository.scala
pauljamescleary: Oof, yea, this shouldn't be here.

modules/dynamodb/src/main/scala/vinyldns/dynamodb/repository/DynamoDBUserChangeRepository.scala
pauljamescleary: Should do the same for both.  It was just odd when I pulled out `toItem` and `fromItem` into the companion object, and then passed the Crypto into the class, which then just passed it back to the function.

modules/dynamodb/src/main/scala/vinyldns/dynamodb/repository/DynamoDBUserChangeRepository.scala
```diff
@@ -0,0 +1,132 @@
...
+    monitor("repo.UserChange.save") {
+      logger.info(s"Saving user change ${change.id}")
+      val item = serialize(change)
+      val request = new PutItemRequest().withTableName(tableName).withItem(item)
+      dynamoDBHelper.putItem(request).as(change)
+    }
+
+  def get(changeId: String): IO[Option[UserChange]] =
+    monitor("repo.UserChange.get") {
+      val key = new HashMap[String, AttributeValue]()
```
mitruly: Should be `val key = new util.HashMap[String, AttributeValue]()` if we're prefixing to suppress warnings like we do elsewhere.

modules/api/src/main/scala/vinyldns/api/Boot.scala
mitruly: I agree with `Crypto.instance` being simpler here.

pauljamescleary: yea, goofy, ok

modules/dynamodb/src/main/scala/vinyldns/dynamodb/repository/DynamoDBUserChangeRepository.scala
```diff
@@ -0,0 +1,141 @@
...
+  // Note: This should be an Either; however pulling everything into an Either is a big refactoring
+  def fromItem(item: java.util.Map[String, AttributeValue]): IO[UserChange] =
+    for {
+      c <- IO(item.get(CHANGE_TYPE).getS)
+      changeType <- IO.fromEither(UserChangeType.fromString(c))
+      newUser <- IO(item.get(NEW_USER).getM).flatMap(m => DynamoDBUserRepository.fromItem(m))
+      oldUser <- OptionT(IO(Option(item.get(OLD_USER)).map(_.getM)))
+        .semiflatMap(DynamoDBUserRepository.fromItem)
+        .value
+      madeByUserId <- IO(item.get(MADE_BY_ID).getS)
```
pauljamescleary: `OptionT` here.  `OptionT` provides convenience functions on `IO[Option[A]]` similar to `EitherT` except the bottom type is an `Option` instead of `Either`

Let's decompose...

1. `item.get(OLD_USER)` could be a `null`, so we do `Option(item.get(OLD_USER))` in the event that the value is not there (and it won't be for Create User)
1. `.map(i => Option(i.getM))` will grab the map out of the item _if it is there_, otherwise, it will be an option
1. `.semiflatMap` - one of my favorite convenience helpers, will only be called if the bottom value is `Some`, if it is `None` in the event the attribute is not there, this is skipped!!
1. `.value` to pull the `IO[Option[A]]` out of `OptionT`

modules/dynamodb/src/main/scala/vinyldns/dynamodb/repository/DynamoDBUserRepository.scala
```diff
@@ -83,13 +86,49 @@ object DynamoDBUserRepository {
...
+      id = item.get(USER_ID).getS,
+      userName = item.get(USER_NAME).getS,
+      created = new DateTime(item.get(CREATED).getN.toLong),
+      accessKey = item.get(ACCESS_KEY).getS,
+      secretKey = item.get(SECRET_KEY).getS,
+      firstName = if (item.get(FIRST_NAME) == null) None else Option(item.get(FIRST_NAME).getS),
+      lastName = if (item.get(LAST_NAME) == null) None else Option(item.get(LAST_NAME).getS),
+      email = if (item.get(EMAIL) == null) None else Option(item.get(EMAIL).getS),
+      isSuper = if (item.get(IS_SUPER) == null) false else item.get(IS_SUPER).getBOOL
+    )
```
pauljamescleary: the question of course is, why make this an `IO` at all.  It is for consistency with the user change repository.

This is admittedly taking the easy way out, as I am allowing things like `item.get(USER_NAME).getS` to throw a null pointer exception.

modules/dynamodb/src/main/scala/vinyldns/dynamodb/repository/DynamoDBUserRepository.scala
```diff
@@ -104,7 +143,11 @@ class DynamoDBUserRepository private[repository] (
       key.put(USER_ID, new AttributeValue(userId))
       val request = new GetItemRequest().withTableName(userTableName).withKey(key)
 
-      dynamoDBHelper.getItem(request).map(result => Option(result.getItem).map(fromItem))
+      OptionT
+        .liftF(dynamoDBHelper.getItem(request))
+        .subflatMap(r => Option(r.getItem))
+        .semiflatMap(item => deserialize(item))
+        .value
```
pauljamescleary: needed these changes as `deserialize` is an `IO` (same as above)

modules/dynamodb/src/main/scala/vinyldns/dynamodb/repository/DynamoDBUserRepository.scala
```diff
@@ -129,13 +172,13 @@ class DynamoDBUserRepository private[repository] (
...
-    def parseUsers(result: BatchGetItemResult): List[User] = {
+    def parseUsers(result: BatchGetItemResult): IO[List[User]] = {
       val userAttributes = result.getResponses.asScala.get(userTableName)
       userAttributes match {
         case None =>
-          List()
+          IO.pure(List())
         case Some(items) =>
-          items.asScala.toList.map(fromItem)
+          items.asScala.toList.map(fromItem).sequence
```
pauljamescleary: needed these changes as `deserialize` is an `IO` (same as above)

modules/dynamodb/src/main/scala/vinyldns/dynamodb/repository/DynamoDBUserRepository.scala
```diff
@@ -159,14 +202,20 @@ class DynamoDBUserRepository private[repository] (
...
-      // run the batches in parallel
       val batchGetIo = batchGets.map(dynamoDBHelper.batchGetItem)
 
-      val allBatches: IO[List[BatchGetItemResult]] = batchGetIo.sequence
-
-      val allUsers = allBatches.map { batchGetItemResults =>
-        batchGetItemResults.flatMap(parseUsers)
-      }
+      // run the batches in parallel
+      val allBatches: IO[List[BatchGetItemResult]] = batchGetIo.parSequence
```
pauljamescleary: woot!  we weren't doing `parSequence` before, now we are actually getting these in parallel!

modules/dynamodb/src/main/scala/vinyldns/dynamodb/repository/DynamoDBUserRepository.scala
```diff
@@ -159,14 +202,20 @@ class DynamoDBUserRepository private[repository] (
...
+
+      val allUsers = for {
+        batches <- allBatches
+        x <- batches.foldLeft(IO(List.empty[User])) { (acc, cur) =>
+          for {
+            u <- parseUsers(cur)
+            a <- acc
+          } yield u ++ a
+        }
+      } yield x
```
pauljamescleary: needed these changes as `deserialize` is an `IO` (same as above)

Could possibly introduce `Nested` here like `OptionT`

modules/dynamodb/src/main/scala/vinyldns/dynamodb/repository/DynamoDBUserRepository.scala
```diff
@@ -194,50 +243,15 @@ class DynamoDBUserRepository private[repository] (
         .withExpressionAttributeValues(expressionAttributeValues)
         .withKeyConditionExpression(keyConditionExpression)
 
-      dynamoDBHelper.query(queryRequest).map { results =>
-        results.getItems.asScala.headOption.map(fromItem)
+      dynamoDBHelper.query(queryRequest).flatMap { results =>
+        results.getItems.asScala.headOption.map(deserialize).sequence
```
pauljamescleary: fun facts, `Option` counts as a "list of at most one".  You can call `.sequence` on `Option` treating it like a `List` with all the cats things.  Fun!

modules/api/src/main/scala/vinyldns/api/domain/auth/AuthPrincipalProvider.scala
rebstar6: yeah dont worry about this, i have it covered in the final dynamic loading PR

modules/dynamodb/src/main/scala/vinyldns/dynamodb/repository/DynamoDBUserRepository.scala
```diff
@@ -83,13 +86,49 @@ object DynamoDBUserRepository {
...
+  def fromItem(item: java.util.Map[String, AttributeValue]): IO[User] = IO {
+    User(
+      id = item.get(USER_ID).getS,
+      userName = item.get(USER_NAME).getS,
+      created = new DateTime(item.get(CREATED).getN.toLong),
+      accessKey = item.get(ACCESS_KEY).getS,
+      secretKey = item.get(SECRET_KEY).getS,
+      firstName = if (item.get(FIRST_NAME) == null) None else Option(item.get(FIRST_NAME).getS),
+      lastName = if (item.get(LAST_NAME) == null) None else Option(item.get(LAST_NAME).getS),
+      email = if (item.get(EMAIL) == null) None else Option(item.get(EMAIL).getS),
```
rebstar6: doesn option handle this?

`Option(item.get(EMAIL)).map(_.getS)` ?

rebstar6: yeah it does, `def apply[A](x: A): Option[A] = if (x == null) None else Some(x)`

id change all ^ to Option then map

modules/dynamodb/src/main/scala/vinyldns/dynamodb/repository/DynamoDBUserRepository.scala
```diff
@@ -159,14 +202,20 @@ class DynamoDBUserRepository private[repository] (
...
-
-      val allUsers = allBatches.map { batchGetItemResults =>
-        batchGetItemResults.flatMap(parseUsers)
-      }
+      // run the batches in parallel
+      val allBatches: IO[List[BatchGetItemResult]] = batchGetIo.parSequence
+
+      val allUsers = for {
+        batches <- allBatches
+        x <- batches.foldLeft(IO(List.empty[User])) { (acc, cur) =>
```
rebstar6: can you use real variable names here? at least call x "users"

modules/dynamodb/src/main/scala/vinyldns/dynamodb/repository/DynamoDBUserRepository.scala
pauljamescleary: Yea, `Option(null).map(_.getX)` would be much preferable, will update

modules/core/src/main/scala/vinyldns/core/crypto/NoOpCrypto.scala
```diff
@@ -27,6 +27,9 @@ import com.typesafe.config.Config
   *               to dynamically load this Crypto implementation.  However, it is not used.
   */
 class NoOpCrypto(val config: Config) extends CryptoAlgebra {
+  def this() {
+    this(ConfigFactory.load())
+  }
```
pauljamescleary: added this because it was annoying in tests to do `new NoOpCrypto(ConfigFactory.load())`

modules/dynamodb/src/main/scala/vinyldns/dynamodb/repository/DynamoDBUserRepository.scala
```diff
@@ -83,13 +86,49 @@ object DynamoDBUserRepository {
...
+      id = item.get(USER_ID).getS,
+      userName = item.get(USER_NAME).getS,
+      created = new DateTime(item.get(CREATED).getN.toLong),
+      accessKey = item.get(ACCESS_KEY).getS,
+      secretKey = item.get(SECRET_KEY).getS,
+      firstName = Option(item.get(FIRST_NAME)).map(_.getS),
+      lastName = Option(item.get(LAST_NAME)).map(_.getS),
+      email = Option(item.get(EMAIL)).map(_.getS),
+      isSuper = if (item.get(IS_SUPER) == null) false else item.get(IS_SUPER).getBOOL
+    )
```
pauljamescleary: You may ask "why do isSuper differently"?  Well, following the pattern `Option(item.get(IS_SUPER)).map(_.getBOOL).getOrElse(false)` **did not compile** 😱 

It couldn't figure out the java Boolean -> scala Boolean conversion, so it inferred `Any`.

rebstar6: what a pain. ok.

PR #186, opened 2018-09-07
Portal lock users take 2

britneywright:
Based on @pauljamescleary PR I added the check of the user's "isLocked" attribute in the custom action builders. There's two action builders now, one for the API and one for the Frontend. It might be overkill to have the custom action on the api calls in VinylDNS.scala with it handled in the views and the API itself will check the user account when that PR is merged. I didn't update the tests yet with that in mind.

Overall I like the custom action builder approach. reduced a lot of redundancy in the code.


modules/portal/app/controllers/VinylDNS.scala
```diff
@@ -364,193 +329,165 @@ class VinylDNS @Inject()(
...
-          .withHeaders(cacheHeaders: _*)
-      })
-    }
+  def addZone(): Action[AnyContent] = userAction.async { implicit request =>
+    val json = request.body.asJson
+    val payload = json.map(Json.stringify)
+    val vinyldnsRequest =
+      new VinylDNSRequest("POST", s"$vinyldnsServiceBackend", "zones", payload)
+    val signedRequest =
+      signRequest(vinyldnsRequest, getUserCreds(request.session.get("accessKey")))
```
pauljamescleary: we shouldn't need to call `getUserCreds` any more, as the `implicit request` provided should be a `UserAccountRequest` and have the `account` already available.  Eliminates a call.

Also, there is a pattern to all of these, namely:

1. Sign request
2. Execute
3. Send response back out with cache headers.

We can change the `executeRequest` to be 

```scala
def executeRequest(req: VinylDNSRequest, user: UserAccount)
```

and do the signing in there, as well as the `.map(response => Status(response.status)...withHeaders` bits all in there.

modules/portal/app/models/UserAccount.scala
```diff
@@ -30,7 +32,8 @@ case class UserAccount(
     created: DateTime,
     accessKey: String,
     accessSecret: String,
-    isSuper: Boolean = false) {
+    isSuper: Boolean = false,
+    isLocked: Boolean = false) {
 
   private def generateKey: String = RandomStringUtils.randomAlphanumeric(20)
```
mitruly: There's an unused import on line 90.

modules/portal/app/controllers/UserAccountAction.scala
```diff
@@ -0,0 +1,99 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package controllers
+import controllers.VinylDNS.Alerts
```
mitruly: Both of these imports are unused:
```
import controllers.VinylDNS.Alerts
import javax.inject.Inject
```

mitruly:
I think I'm good on this once the build passes.

modules/portal/app/controllers/VinylDNS.scala
```diff
@@ -233,32 +211,24 @@ class VinylDNS @Inject()(
...
-    Logger.info(s"Serving credentials for file $fileName")
-    withAuthenticatedUser { username =>
-      userAccountAccessor.get(username) match {
-        case Success(Some(account)) => Future(processCsv(username, account))
-        case Success(None) =>
-          throw new UnsupportedOperationException(s"Error - User account for $username not found")
-        case Failure(ex) => throw ex
-      }
-    }
+  def serveCredsFile(fileName: String): Action[AnyContent] = frontendAction.async {
```
britneywright: I had to use the frontendAction here otherwise the portal renders the csv path as a blank screen with just the text "Account is Locked", now it redirects to the login screen if a user is locked and tries to download credentials.

modules/portal/app/controllers/UserAccountAction.scala
```diff
@@ -0,0 +1,98 @@
...
+              Future.successful(
+                Redirect("/login")
+                  .flashing(VinylDNS.Alerts.error(s"Account locked"))
+                  .withNewSession)
+            } else {
+              block(new UserAccountRequest(un, acc, request))
+            }
+        }
+    }
+}
```
pauljamescleary: there is duplication here that could be refactored.  I am ok with leaving it alone for the time being as it is unclear how the two actions may evolve.

modules/portal/app/controllers/VinylDNS.scala
```diff
@@ -111,6 +103,18 @@ class VinylDNS @Inject()(
     ("Pragma", "no-cache"),
     ("Expires", "0"))
 
+  // Need this guy for user actions, brings the session username and user account into the Action
+  private val userAction = Action.andThen(new ApiAction(userAccountAccessor.get))
+  private val frontendAction = Action.andThen(new FrontendAction(userAccountAccessor.get))
```
pauljamescleary: why is the frontendAction needed here?  I see it is only used on `serveCredsFile`.

modules/portal/app/controllers/VinylDNS.scala
pauljamescleary: answered my question, thanks.

modules/portal/app/controllers/VinylDNS.scala
```diff
@@ -233,32 +211,24 @@ class VinylDNS @Inject()(
...
-            .withSession("username" -> response.username, "accessKey" -> response.accessKey)
-        })
-        .recover {
-          case _: UserDoesNotExistException =>
-            NotFound(s"User $username was not found").withHeaders(cacheHeaders: _*)
-        }
-    }
+  def regenerateCreds(): Action[AnyContent] = userAction.async { implicit request =>
+    Future
+      .fromTry(processRegenerate(request.userName))
```
pauljamescleary: `processRegenerate` takes a `userName`, but then does an account lookup.  However, we already have the user account on the `request`.  We can pass the `request.account` into the `processRegenerate` instead, and then skip the account lookup there.

modules/portal/app/controllers/VinylDNS.scala
britneywright: just noting this was answered for anyone else reviewing. userAction doesn't redirect, frontendAction does, and in this case we want to redirect.

modules/portal/app/controllers/UserAccountAction.scala
```diff
@@ -0,0 +1,113 @@
...
+  * If the user is not in session, redirect to the login screen
+  * If the user is in session, but the account is not found, redirect to the login screen with a different message
+  * If the user is locked out, redirect to login screen
+  * Otherwise, load the account into a custom UserAccountRequest and pass into the action
+  */
+class ApiAction(userLookup: String => IO[Option[User]])(
+    implicit val executionContext: ExecutionContext)
+    extends ActionFunction[Request, UserRequest] {
+
+  private val cacheHeaders = Seq(
```
mitruly: These cache headers are the exact same as in the `VinylDNS` singleton and `FrontendAction`. One way to reduce code duplication is to pull these into a shared `trait`. `protected` should make this `cacheHeaders` only accessible from classes/objects which extend it.

modules/portal/app/controllers/UserAccountAction.scala
```diff
@@ -0,0 +1,113 @@
...
+    extends WrappedRequest(request)
+
+/**
+  * Custom action that will take a request and do an account lookup for the user
+  * If the user is not in session, redirect to the login screen
+  * If the user is in session, but the account is not found, redirect to the login screen with a different message
+  * If the user is locked out, redirect to login screen
+  * Otherwise, load the account into a custom UserAccountRequest and pass into the action
+  */
+class ApiAction(userLookup: String => IO[Option[User]])(
```
rebstar6: not crazy about the class breakdown all in the same file.

Unless theres a compelling reason not to, id say split this out each into its own file in an actions package

modules/portal/app/controllers/UserAccountAction.scala
```diff
@@ -0,0 +1,113 @@
...
+      case Some(un) =>
+        // user name in session, let's get it from the repo
+        userLookup(un).unsafeToFuture().flatMap {
+          case None =>
+            // Odd case, but let's handle with a different error message
+            Future.successful(
+              NotFound(s"Unable to find user account for user name '$un'")
+                .withHeaders(cacheHeaders: _*))
+
+          case Some(user) =>
```
rebstar6: de-nest the if here:

```
        userLookup(un).unsafeToFuture().flatMap {
          case None =>
            // Odd case, but let's handle with a different error message
            Future.successful(
              NotFound(s"Unable to find user account for user name '$un'")
                .withHeaders(cacheHeaders: _*))

          case Some(user) if user.lockStatus == LockStatus.Locked =>
            Future.successful(Forbidden(s"Account is locked.").withHeaders(cacheHeaders: _*))
          case Some(user) =>
            block(new UserRequest(un, user, request))
        }
```

modules/portal/app/controllers/UserAccountAction.scala
```diff
@@ -0,0 +1,113 @@
...
+        // user name in session, let's get it from the repo
+        userLookup(un).unsafeToFuture().flatMap {
+          case None =>
+            // Odd case, but let's redirect to login with a different error message
+            Future.successful(
+              Redirect("/login")
+                .flashing(VinylDNS.Alerts.error(s"Unable to find user account for user name '$un'"))
+                .withHeaders(cacheHeaders: _*))
+
+          case Some(user) =>
```
rebstar6: same nested if comment here

modules/portal/app/controllers/UserAccountAction.scala
```diff
@@ -0,0 +1,113 @@
...
+  private val cacheHeaders = Seq(
+    ("Cache-Control", "no-cache, no-store, must-revalidate"),
+    ("Pragma", "no-cache"),
+    ("Expires", "0"))
+
+  def invokeBlock[A](request: Request[A], block: UserRequest[A] => Future[Result]): Future[Result] =
+    // if the user name is not in session, redirect to the login screen
+    request.session.get("username") match {
+      case None =>
+        Future(
```
rebstar6: the others are Future.successful - any reason for this not to be?

modules/portal/app/controllers/UserAccountAction.scala
rebstar6: to this point, you actually can share the bulk of this code with something like:

```trait UserAccountAction extends ActionFunction[Request, UserRequest]  {
  
  implicit val executionContext: ExecutionContext
  val userLookup: String => IO[Option[User]]

  protected val cacheHeaders = Seq(
    ("Cache-Control", "no-cache, no-store, must-revalidate"),
    ("Pragma", "no-cache"),
    ("Expires", "0"))

  val notLoggedIn: Result

  def userNotFound(un: String): Result

  val locked: Result

  def invokeBlock[A](request: Request[A], block: UserRequest[A] => Future[Result]): Future[Result] =
  // if the user name is not in session, reject
    request.session.get("username") match {
      case None => Future(notLoggedIn)
      case Some(un) =>
        // user name in session, let's get it from the repo
        userLookup(un).unsafeToFuture().flatMap {
          case None =>
            // Odd case, but let's handle with a different error message
            Future.successful(userNotFound(un))
          case Some(user) if user.lockStatus == LockStatus.Locked => Future.successful(locked)
          case Some(user) => block(new UserRequest(un, user, request))
        }
    }
}
```

rebstar6: then your api action is just something like 
```
class ApiAction(val userLookup: String => IO[Option[User]])(
    implicit val executionContext: ExecutionContext)
    extends UserAccountAction {

  val notLoggedIn: Result =
    Unauthorized("You are not logged in. Please login to continue.").withHeaders(cacheHeaders: _*)

  def userNotFound(un: String): Result =
    NotFound(s"Unable to find user account for user name '$un'").withHeaders(cacheHeaders: _*)

  val locked: Result = Forbidden(s"Account is locked.").withHeaders(cacheHeaders: _*)
}
```

rebstar6: only thing im not sure on there is how that implicit EC gets passed around - talk to @pauljamescleary about that

modules/portal/app/controllers/VinylDNS.scala
```diff
@@ -348,308 +302,253 @@ class VinylDNS @Inject()(
...
-        signRequest(vinyldnsRequest, getUserCreds(request.session.get("accessKey")))
-      executeRequest(signedRequest).map(response => {
+        new VinylDNSRequest("PUT", s"$vinyldnsServiceBackend", s"users/$userId/lock")
+      executeRequest(vinyldnsRequest, request.user).map(response => {
         Logger.info(response.body)
         Status(response.status)(response.body)
           .withHeaders(cacheHeaders: _*)
       })
+    } else {
+      Future(Forbidden("Request restricted to super users only."))
```
rebstar6: when we are just returning known statuses like this, should it instead be in Future.successful?

pauljamescleary: Probably right.  The danger of Future.successful is if the expression fails, you throw an exception, or possibly return a Future.successful(exception).

If the expression is not known to fail, Future.successful is preferable.

modules/portal/app/actions/ApiAction.scala
```diff
@@ -0,0 +1,63 @@
...
+  * If the user is in session, but the account is not found, return Not Found message
+  * If the user is locked out, return Forbidden message
+  * Otherwise, load the account into a custom UserAccountRequest and pass into the action
+  */
+class ApiAction(userLookup: String => IO[Option[User]])(
+    implicit val executionContext: ExecutionContext)
+    extends ActionFunction[Request, UserRequest]
+    with CacheHeader {
+
+  def invokeBlock[A](request: Request[A], block: UserRequest[A] => Future[Result]): Future[Result] =
```
rebstar6: you can still move the basic logic here into the trait (CacheHeader, but with this id probably call it VinylDnsAction or something), and just override what it does at each step:

```
trait CacheHeader extends ActionFunction[Request, UserRequest] {
  val cacheHeaders = Seq(
    ("Cache-Control", "no-cache, no-store, must-revalidate"),
    ("Pragma", "no-cache"),
    ("Expires", "0"))
  
  val userLookup: String => IO[Option[User]]
  implicit val executionContext: ExecutionContext
  def notLoggedInResult: Future[Result]
  def cantFindAccountResult(un: String): Future[Result]
  def lockedUserResult: Future[Result]

  def invokeBlock[A](request: Request[A], block: UserRequest[A] => Future[Result]): Future[Result] =
  // if the user name is not in session, reject
    request.session.get("username") match {
      case None => notLoggedInResult

      case Some(un) =>
        // user name in session, let's get it from the repo
        userLookup(un).unsafeToFuture().flatMap {
          // Odd case, but let's handle with a different error message
          case None => cantFindAccountResult(un)

          case Some(user) if user.lockStatus == LockStatus.Locked => lockedUserResult

          case Some(user) =>
            block(new UserRequest(un, user, request))
        }
    }
}
```

```
class ApiAction(val userLookup: String => IO[Option[User]])(
    implicit val executionContext: ExecutionContext)
    extends CacheHeader {

  def notLoggedInResult: Future[Result] = Future.successful(
    Unauthorized("You are not logged in. Please login to continue.")
      .withHeaders(cacheHeaders: _*)
  )

  def cantFindAccountResult(un: String): Future[Result] = Future.successful(
    NotFound(s"Unable to find user account for user name '$un'")
      .withHeaders(cacheHeaders: _*))

  def lockedUserResult: Future[Result] = Future.successful(
    Forbidden(s"Account is locked.")
      .withHeaders(cacheHeaders: _*))
}
```

modules/portal/app/actions/ApiAction.scala
```diff
@@ -0,0 +1,63 @@
...
+    // if the user name is not in session, reject
+    request.session.get("username") match {
+      case None =>
+        Future(
```
rebstar6: make the Futuring consistent - doesnt make sense that this is a Future() but the others are Future.successful.

I believe all the already-determined results can be Future.successful

modules/portal/app/actions/ApiAction.scala
britneywright: I can make a separate trait. CacheHeader is just for the variable that's used in the controllers and actions. I can make a separate trait for the commonalities in the actions.

build.sbt
```diff
@@ -307,6 +307,7 @@ lazy val portal = (project in file("modules/portal")).enablePlugins(PlayScala, A
     libraryDependencies ++= portalDependencies,
     routesGenerator := InjectedRoutesGenerator,
     coverageMinimum := 75,
+    coverageFailOnMinimum := false,
```
rebstar6: why have a minimum if we dont fail on it?

where are we on coverage now in the portal? id rather lower the minimum if we have to now then just not fail when we hit it

rebstar6: actually, were above 75 as is, this PR is bringing it down. needs tests it would seem :P

britneywright:
I'm pretty sure the reason the code coverage dropped so much for the VinylDNS controller in the portal is due to how we refactored the functions. Since we're using custom actions now we no longer needed the withAuthenticatedUser function, and that wrapped most of the logic in the functions. So if a function wasn't tested (like all of the zone functions) it was only counting as one statement miss, but now there are many more statements in each function. From running the coverage reports locally I see there are 182 statements in master, but in my branch there are now 655. 

I've tested everything related to what my PR adds, even for functions that weren't previously tested at all. I've added 55 tests, that increased the coverage in my branch from 46.59% to 59.61%. Adding more tests seems outside the scope of this PR though. I can branch off this PR for that.

rebstar6:
@britneywright tricky, hmm

We definitely wanna keep coverageFailOnMinimum on, because without it it kinda defeats the point of having any kinda coverage at all. what im inclined to do is, if there are full untested files that you think we should add testing for in a followup, $COVERAGE-OFF$ them, and then write the card to add the tests and turn on coverage for that file.

Open to other suggestions though

modules/portal/app/actions/FrontendAction.scala
```diff
@@ -0,0 +1,57 @@
...
+  * If the user is locked out, redirect to login screen
+  * Otherwise, load the account into a custom UserAccountRequest and pass into the action
+  */
+class FrontendAction(val userLookup: String => IO[Option[User]])(
+    implicit val executionContext: ExecutionContext)
+    extends VinylDnsAction
+    with CacheHeader {
+
+  def notLoggedInResult: Future[Result] =
+    Future(
```
rebstar6: one last thing - should all of these be `Future` or `Future.successful`?

rebstar6:
other than https://github.com/vinyldns/vinyldns/pull/186/files#r221731723, im good here. please make sure to make followup cards to test and bring back coverage on the skipped items

PR #188, opened 2018-09-10
remove zone change repo getpendings functions

nimaeskandary:
This pr removes ZoneChangeRepository.getPending and ZoneChangeRepository.getAllPendingZoneIds


modules/api/src/test/scala/vinyldns/api/protobuf/ProtobufConversionsSpec.scala
```diff
@@ -67,7 +67,7 @@ class ProtobufConversionsSpec
     zone,
     "system",
     ZoneChangeType.Update,
-    ZoneChangeStatus.Complete,
+    ZoneChangeStatus.Synced,
```
nimaeskandary: wanted to cut back on the usage of ZoneChangeStatus.Complete, with these changes it will not be used at all unless a `fromPB` is converting a blob that has it

nimaeskandary:
@pauljamescleary updated

modules/dynamodb/src/main/scala/vinyldns/dynamodb/repository/DynamoDBZoneChangeRepository.scala
```diff
@@ -161,63 +155,6 @@ class DynamoDBZoneChangeRepository(config: Config, dynamoDBHelper: DynamoDBHelpe
...
-    val expressionAttributeNames = new HashMap[String, String]
-    expressionAttributeNames.put("#status_attribute", STATUS)
-    expressionAttributeNames.put("#zone_id_attribute", ZONE_ID)
-
-    val keyConditionExpression: String =
-      "#status_attribute = :status and #zone_id_attribute = :zone_id"
-
-    val queryRequest = new QueryRequest()
-      .withTableName(zoneChangeTable)
-      .withIndexName(ZONE_ID_STATUS_INDEX_NAME)
```
pauljamescleary: do we need this index still?

modules/dynamodb/src/main/scala/vinyldns/dynamodb/repository/DynamoDBZoneChangeRepository.scala
```diff
@@ -161,63 +155,6 @@ class DynamoDBZoneChangeRepository(config: Config, dynamoDBHelper: DynamoDBHelpe
...
-
-  def getAllPendingZoneIds: IO[List[String]] = {
-    val expressionAttributeValues = new HashMap[String, AttributeValue]
-    val expressionAttributeNames = new HashMap[String, String]
-    expressionAttributeNames.put("#status_attribute", STATUS)
-    expressionAttributeValues.put(":status", new AttributeValue("Pending"))
-    val keyConditionExpression: String = "#status_attribute = :status"
-    val queryRequest = new QueryRequest()
-      .withTableName(zoneChangeTable)
-      .withIndexName(STATUS_INDEX_NAME)
```
pauljamescleary: I don't think we need this status either.

Note: IF we no longer need these indexes, then we should remove them from the `CreateTable`.

pauljamescleary:
**Note: When you remove the indexes from the create table section, please update the docs as well, the operator guide has AWS dynamodb bits that list out the indexes**

PR #190, opened 2018-09-10
Update TSIG key algorithm

mitruly:
Update TSIG key algorithm from HMAC-MD5 to HMAC-SHA512.


mitruly:
@pauljamescleary That's a good point. I will close this and reference it in https://github.com/vinyldns/vinyldns/pull/190 for later pick up.

PR #195, opened 2018-09-12
Docker Improvements

NorseGaud:
Why I believe these changes are necessary: If I was someone not familiar with VinylDNS and was using the docker setup to play around, I am in a state of learning. I would want easier control over ENV values and explicit documentation. This allows me to then take the docker examples and set up my own production instance based on them without much headache and wasted time trying to infer things.

I have made multiple changes:

* Use of .env files (in docker/ and bin/), allowing customization for users.
* docker-compose files have been updated to use docker/.env values.
* Sourcing of .env in docker bash scripts.
* New example for disabling host filter in application.conf.
* README explaining it all.

I have tested all of these changes and it's working great.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,40 +46,40 @@ Docker images for VinylDNS live on Docker Hub at https://hub.docker.com/u/vinyld
 To start up a local instance of VinylDNS on your machine with docker:
 
 1. Ensure that you have [docker](https://docs.docker.com/install/) and [docker-compose](https://docs.docker.com/compose/install/)
-2. Clone the repo: `git clone https://github.com/vinyldns/vinyldns.git`
-3. Navigate to repo: `cd vinyldns`
-4. Run `bin/docker-up-vinyldns.sh`. This will start up the api at `localhost:9000` and the portal at `localhost:9001` along with their 
+1. Clone the repo: `git clone https://github.com/vinyldns/vinyldns.git`
+1. Navigate to repo: `cd vinyldns`
+1. Run `bin/docker-up-vinyldns.sh`. This will start up the api at `localhost:9000` and the portal at `localhost:9001` along with their 
 dependencies, ping the API on `http://localhost:9000/ping` and the portal on `http://localhost:9001`, and notify you if either failed to start.
-5. To stop the local setup, run `./bin/remove-vinyl-containers.sh` from the project root.
+1. To stop the local setup, run `./bin/remove-vinyl-containers.sh` from the project root.
 
 Things to try after VinylDNS is running:
 
 1. View the portal at <http://localhost:9001> in a web browser
-2. Login with the credentials ***testuser*** and ***testpassword***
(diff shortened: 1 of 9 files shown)


rebstar6:
and apparently accidentally closing 😬

README.md
```diff
@@ -46,38 +46,40 @@ Docker images for VinylDNS live on Docker Hub at https://hub.docker.com/u/vinyld
 To start up a local instance of VinylDNS on your machine with docker:
 
 1. Ensure that you have [docker](https://docs.docker.com/install/) and [docker-compose](https://docs.docker.com/compose/install/)
-1. Clone the repo: `git clone https://github.com/vinyldns/vinyldns.git`
-1. Navigate to repo: `cd vinyldns`
-1. Run `bin/docker-up-vinyldns.sh`. This will start up the api at `localhost:9000` and the portal at `localhost:9001` along with their
+2. Clone the repo: `git clone https://github.com/vinyldns/vinyldns.git`
```
nimaeskandary: Any inherent problem with auto-numbering?

NorseGaud: Nope, I can switch it back

nimaeskandary: Yeah wasn't sure myself which is best practice, but we do use auto numbering in general in other markdown files

docker/docker-compose-build.yml
```diff
@@ -36,11 +39,11 @@ services:
...
   api:
     image: "vinyldns/api:0.8.0"
-    environment:
-      - REST_PORT=9000
+    env_file:
+      .env
     container_name: "vinyldns-api"
     ports:
-      - "9000:9000"
+      - "${REST_PORT}:${REST_PORT}"
```
mitruly: I'm thinking that we should make a distinction for the repeated value of `REST_PORT` (where the port syntax is `"HOST_PORT:CONTAINER_PORT"`. We use the same number for the host and container pair when assigning API and portal ports because they are available, but it is possible other adopters may want to assign another host port, so ``9000:${REST_PORT}` or `9000:${API_PORT}` might be better.

We probably can also bring in the DynamoDB, Bind9, MySQL, Elastic MQ, etc. port info since those are shared between docker compose files. There is also a new `modules/dynamodb/docker/docker-compose.yml` file that can use the DynamoDB container port as well.

docker/portal/application.conf
```diff
@@ -80,3 +80,6 @@ links = [
     icon = "fa fa-file-text-o"
   }
 ]
+
+# Uncomment if using a different hostname from localhost
+# play.filters.disabled += play.filters.hosts.AllowedHostsFilter
```
pauljamescleary: fun facts, this isn't the official docker image, just for local testing.  Let's just remove the comment, leave this uncommented and the default.

NorseGaud: Sounds good!

NorseGaud:
Rebased. Ready to go.

PR #206, opened 2018-09-14
Replace the repos in the portal with dynamodb and core

pauljamescleary:
* Remove all data stores from the portal
* Use the user and user change repository from core and dynamodb
* Remove the UserAccount type, use core User instead
* Remove the UserChangeLog types, use core UserChange instead


modules/core/src/main/scala/vinyldns/core/domain/membership/User.scala
```diff
@@ -30,4 +31,11 @@ case class User(
...
     isSuper: Boolean = false
-)
+) {
+  def regenerateCredentials(): User =
+    copy(accessKey = User.generateKey, secretKey = User.generateKey)
+}
+
+object User {
+  def generateKey: String = RandomStringUtils.randomAlphanumeric(20)
+}
```
pauljamescleary: moving this stuff in from the portal as we delete the `UserAccountStore` there

modules/core/src/main/scala/vinyldns/core/domain/membership/UserRepository.scala
```diff
@@ -31,5 +31,7 @@ trait UserRepository extends Repository {
 
   def getUserByAccessKey(accessKey: String): IO[Option[User]]
 
+  def getUserByName(userName: String): IO[Option[User]]
+
```
pauljamescleary: need for the portal

modules/portal/app/Module.scala
```diff
@@ -14,29 +14,39 @@
...
     val crypto = CryptoAlgebra.load(configuration.underlying.getConfig("crypto")).unsafeRunSync()
+
+    val dynamoConfig = configuration.get[Configuration]("dynamo")
+    val dynamoSettings = DynamoDBDataStoreSettings(
+      key = dynamoConfig.get[String]("key"),
+      secret = dynamoConfig.get[String]("secret"),
+      endpoint = dynamoConfig.get[String]("endpoint"),
+      region = dynamoConfig.get[String]("region")
+    )
+
```
pauljamescleary: Rather than using pureconfig here, I stay with the existing configuration for the user repo, user change repo, and dynamo settings.  This will keep us from having to change any portal config in order to use this change.

When we "enable dynamic loading" in the portal, we will need to change the portal config

modules/portal/app/controllers/UserAccountAccessor.scala
```diff
@@ -16,24 +16,15 @@
...
-    user: String,
-    timeStamp: DateTime,
-    changeType: UserChangeType,
-    newUser: UserAccount,
-    oldUser: UserAccount)
+import vinyldns.core.domain.membership.{User, UserChange, UserChangeRepository, UserRepository}
 
 @Singleton
-class UserAccountAccessor @Inject()(store: UserAccountStore) {
+class UserAccountAccessor @Inject()(users: UserRepository, changes: UserChangeRepository) {
```
pauljamescleary: Before, we would save users and save changes in two separate repositories.  I am doing both in here now, so we need both repositories from core

modules/portal/app/controllers/VinylDNS.scala
```diff
@@ -63,14 +66,14 @@ object VinylDNS {
...
-        lastName = account.lastName,
-        email = account.email,
-        isSuper = account.isSuper,
-        id = account.userId
+        userName = user.userName,
+        firstName = user.firstName,
+        lastName = user.lastName,
+        email = user.email,
+        isSuper = user.isSuper,
+        id = user.id
```
pauljamescleary: this is just using `User` instead of `UserAccount` which was deleted as part of this PR

modules/portal/app/controllers/VinylDNS.scala
```diff
@@ -236,12 +233,15 @@ class VinylDNS @Inject()(
...
-      }
+      userAccountAccessor
+        .get(username)
+        .flatMap {
+          case Some(account) => IO(processCsv(username, account))
+          case None =>
+            IO.raiseError(
+              new UnsupportedOperationException(s"Error - User account for $username not found"))
+        }
+        .unsafeToFuture()
```
pauljamescleary: the `UserAccountAccessor` now is in `IO` instead of `Try`, some subtle adjustments there, at the end we need a `Future`

modules/portal/conf/application-test.conf
```diff
@@ -7,23 +7,23 @@ portal.vinyldns.backend.url = "http://not.real.com"
 dynamo {
   key = "akid goes here"
   secret = "secret key goes here"
-  endpoint = "endpoint url goes here"
+  endpoint = "http://127.0.0.1:19000"
```
pauljamescleary: this is kinda big, several tests require dynamodb since we removed the "in memory" stores in this PR.  We could keep the in memory repos, and rewrite them to match the core UserRepository and UserChangeRepository.  But that is a big change by itself.

Removing the in memory repositories, and making this endpoint point to dynamodb works nicely, it does require one to make sure that dynamodb is up before running tests in the portal, which makes me sad.  The truth of the matter is that so many of our portal "unit" tests are actually "integration" tests.

modules/portal/test/controllers/UserAccountAccessorSpec.scala
```diff
@@ -16,157 +16,71 @@
...
-        Some("Baggins"),
-        Some("fbaggins@hobbitmail.me"),
-        DateTime.now(),
-        "key",
-        "secret")
-      mockStore.storeUser(any[UserAccount]).returns(Success(user))
-
-      val underTest = new UserAccountAccessor(mockStore)
-      underTest.put(user) must beASuccessfulTry(user)
+      mockRepo.save(any[User]).returns(IO.pure(user))
```
pauljamescleary: look at all this duplication. delete delete delete delete

project/Dependencies.scala
```diff
@@ -32,6 +32,7 @@ object Dependencies {
     "dnsjava"                   %  "dnsjava"                        % "2.1.7",
     "org.mariadb.jdbc"          %  "mariadb-java-client"            % "2.2.3",
     "org.apache.commons"        %  "commons-lang3"                  % "3.4",
+    "org.apache.commons"        %  "commons-text"                   % "1.4",
```
pauljamescleary: commons text supposedly is the update for secure random string generation (the existing way we generate keys and secrets in the portal was deprecated).  Pulling this in removed that deprecation warning.

modules/portal/test/controllers/VinylDNSSpec.scala
```diff
@@ -81,11 +82,11 @@ class VinylDNSSpec extends Specification with Mockito {
         val ws: WSClient = mock[WSClient]
 
         authenticator.authenticate("frodo", "secondbreakfast").returns(Success(frodoDetails))
-        userAccessor.get(anyString).returns(Success(Some(frodoAccount)))
-        userAccessor.get("frodo").returns(Success(Some(frodoAccount)))
+        userAccessor.get(anyString).returns(IO.pure(Some(frodoUser)))
+        userAccessor.get("frodo").returns(IO.pure(Some(frodoUser)))
```
pauljamescleary: pretty sure every change in this file is either:

1. because we deleted the auditLog
2. because we are using IO
3. because we are using User instead of UserAccount

I made not substantive changes to this otherwise

modules/portal/app/controllers/VinylDNS.scala
```diff
@@ -93,7 +94,6 @@ class VinylDNS @Inject()(
     configuration: Configuration,
     authenticator: Authenticator,
     userAccountAccessor: UserAccountAccessor,
-    auditLogAccessor: ChangeLogStore,
```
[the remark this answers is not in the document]
britneywright: ah I see it now. nice.

modules/portal/conf/application-test.conf
```diff
@@ -7,23 +7,23 @@ portal.vinyldns.backend.url = "http://not.real.com"
...
   region = "us-east-1" # note: we are always in us-east-1, but this can be overridden
-  test_datastore = true
+  test_datastore = false
```
pauljamescleary: I left the `test_datastore` and `dummy` things in the config; as it will all be blown up when we do dynamic loading in the portal.

modules/portal/app/controllers/UserAccountAccessor.scala
```diff
@@ -42,16 +33,26 @@ class UserAccountAccessor @Inject()(store: UserAccountStore) {
     * @return Success(Some(user account)) on success, Success(None) if the user does not exist and Failure when there
     *         was an error.
     */
-  def get(identifier: String): Try[Option[UserAccount]] =
-    store.getUserById(identifier) match {
-      case Success(None) => store.getUserByName(identifier)
-      case Success(Some(user)) => Success(Some(user))
-      case Failure(ex) => Failure(ex)
+  def get(identifier: String): IO[Option[User]] =
+    users.getUser(identifier).flatMap {
```
rebstar6: in what case is it calling `getUser` vs `getUserByName`? just wondering if we really need both

modules/portal/app/controllers/UserAccountAccessor.scala
```diff
@@ -42,16 +33,26 @@ class UserAccountAccessor @Inject()(store: UserAccountStore) {
...
-  def put(user: UserAccount): Try[UserAccount] =
-    store.storeUser(user)
-
-  def getUserByKey(key: String): Try[Option[UserAccount]] =
-    store.getUserByKey(key)
+  def create(user: User): IO[User] =
+    for {
+      _ <- users.save(user)
+      _ <- changes.save(
+        UserChange.CreateUser(UUID.randomUUID().toString, user, "system", DateTime.now))
```
rebstar6: UserChange.CreateUser/UpdateUser should take care of making the change ID (rather than the caller always having to pass in UUID.randomUUID)

modules/portal/app/controllers/UserAccountAccessor.scala
```diff
@@ -16,24 +16,15 @@
...
-    userId: String,
-    user: String,
-    timeStamp: DateTime,
-    changeType: UserChangeType,
-    newUser: UserAccount,
-    oldUser: UserAccount)
+import vinyldns.core.domain.membership.{User, UserChange, UserChangeRepository, UserRepository}
 
 @Singleton
-class UserAccountAccessor @Inject()(store: UserAccountStore) {
```
rebstar6: `UserAccountAccessor` now feels like the wrong name for this class, but I can deal with it

modules/portal/app/controllers/UserAccountAccessor.scala
pauljamescleary: Good question, this was copy pasta.  `getUser` assumes the user iid (UUID).  `getUserByName` hits the user name index.

Most likely, we _only_ use the `getUserByName`; unclear on why we added the fallback in the `UserAccountAccessor`.

modules/portal/app/controllers/UserAccountAccessor.scala
pauljamescleary: I can create an apply

modules/dynamodb/src/main/scala/vinyldns/dynamodb/repository/DynamoDBUserRepository.scala
```diff
@@ -161,6 +161,30 @@ class DynamoDBUserRepository private[repository] (
...
+    val attributeValues = new util.HashMap[String, AttributeValue]()
+    attributeValues.put(":uname", new AttributeValue().withS(username))
+    val request = new QueryRequest()
+      .withTableName(userTableName)
+      .withKeyConditionExpression("#uname = :uname")
+      .withExpressionAttributeNames(attributeNames)
+      .withExpressionAttributeValues(attributeValues)
+      .withIndexName(USER_NAME_INDEX_NAME)
+
+    // the question is what to do with duplicate usernames, in the portal we just log loudly, staying the same here
```
rebstar6: bummer. we really should have used username as our ID

modules/portal/app/controllers/VinylDNS.scala
```diff
@@ -44,8 +48,6 @@ object VinylDNS {
     private val MSG = "alertMessage"
     def error(msg: String): Flash = Flash(Map(TYPE -> "danger", MSG -> msg))
     def warning(msg: String): Flash = Flash(Map(TYPE -> "warning", MSG -> msg))
-    def info(msg: String): Flash = Flash(Map(TYPE -> "info", MSG -> msg))
```
rebstar6: interesting. what are these used for as opposed to the utilityService alert stuff in the angular? probably more of a @britneywright question

pauljamescleary: Used to send messages of different severities to the client.  Can see them in `Redirect` and `Flash`  Those removed were unused.

modules/portal/app/controllers/VinylDNS.scala
```diff
@@ -236,12 +239,15 @@ class VinylDNS @Inject()(
...
     Logger.info(s"Serving credentials for file $fileName")
     withAuthenticatedUser { username =>
-      userAccountAccessor.get(username) match {
-        case Success(Some(account)) => Future(processCsv(username, account))
-        case Success(None) =>
-          throw new UnsupportedOperationException(s"Error - User account for $username not found")
-        case Failure(ex) => throw ex
-      }
+      userAccountAccessor
+        .get(username)
```
rebstar6: back to the userAccountAccessor.get thing, in the case that we know the item we are looking for is a userName, we should just be searching for that username. having 2 gets every time someone does anything is double the gets we need...

rebstar6: if need be, untangling that can be a followup. i imagine all we need to confirm is that both testuser and a real login work, that *may* be the difference (?)

modules/portal/app/controllers/VinylDNS.scala
britneywright: these alerts appear on the login page, when we return "Authentication Failed" and soon the account locked message there, since it doesn't touch the angular. We probably don't even need all these different functions, i think we only user error.

modules/portal/app/controllers/VinylDNS.scala
rebstar6: even worse one of them is a query! will be larger untangle, it would be nice throughout to have withAuthenticatedUser pass around both the userName AND ID and work off the ID. then we dont have to query all over

britneywright: my portal PR has the action that will replace withauthenticated user and start passing in the whole user instead of just the username. so we can eventually refactor userAccountAcessor.get out from all of the functions in VinylDNS.scala e.g. https://github.com/vinyldns/vinyldns/pull/186/files#diff-324ddb235c45bfd196384b7ba1aa3a26R557

britneywright: I started to change it for the other functions but i thought that might be getting too out of scope.

modules/portal/app/controllers/VinylDNS.scala
```diff
@@ -261,62 +267,50 @@ class VinylDNS @Inject()(
...
-              userAccount <- existingAccount match {
-                case Some(user) => Try(VinylDNS.UserInfo.fromAccount(user))
-                case None =>
-                  createNewUser(userDetails).map(VinylDNS.UserInfo.fromAccount)
-              }
-            } yield userAccount
-          }
+        {
+          for {
+            userDetails <- IO.fromEither(authenticator.lookup(username).toEither)
```
rebstar6: i know this is copy/pasta, but this method is silly. we only need to be LDAPing here if ``userAccountAccessor.get(username)` returns nothing.

modules/portal/conf/application-test.conf
rebstar6: you mean api docker up? why that vs making the portal have it's own dynamo in docker?

modules/portal/conf/application-test.conf
```diff
@@ -7,23 +7,23 @@ portal.vinyldns.backend.url = "http://not.real.com"
...
 }
 
 users {
-  dummy = true
-  tablename = "userAccounts"
+  dummy = false
+  tablename = "users"
```
rebstar6: this can be called anything. ive been tending to make them not the same as what they would be running on purpose ("users-test" or something)

project/Dependencies.scala
rebstar6: you dont need to add here and in the portal if its in core. i believe everyone pulls in the stuff in "coreDependencies"

modules/portal/conf/application-test.conf
pauljamescleary: we could do that as well, or keep the in memory repositories.  Not sure I love having all these dynamodbs on their own ports with only slightly different configurations and duplicating files everywhere.  I know there is a better way (using dependencies of tasks / targets in sbt).

project/Dependencies.scala
pauljamescleary: hmm, I will check

modules/portal/app/controllers/VinylDNS.scala
pauljamescleary: This is actually called from the client, like when we add a user to a group.  So the user passed in is _not_ the username being looked up.

So, this method does a few things:

1. If the user is in ldap, proceed; otherwise fail out
1. If the user is in ldap, but is NOT in our system yet, create a new user for them

You could argue that the order could be changed, but I don't see any benefit to doing so.

modules/portal/conf/application-test.conf
pauljamescleary: for this to align with the API when doing local testing (reStart on the api and `run` in the portal) the names of the tables have to align.

pauljamescleary: nevermind, right, this is application-test.conf, only used in test

project/Dependencies.scala
pauljamescleary: Ok, seems like we do not need it in both places, I will remove from the portal

modules/portal/conf/application-test.conf
pauljamescleary: Having a test user account accessor might be ok for running unit tests.  I would prefer that approach to having yet another dynamodb docker instance tbh

build.sbt
```diff
@@ -73,7 +73,7 @@ lazy val testSettings = Seq(
   parallelExecution in Test := false,
   parallelExecution in IntegrationTest := false,
   fork in IntegrationTest := false,
-  testOptions in Test += Tests.Argument("-oD"),
+  testOptions in Test += Tests.Argument("-oDNCXEHPQRMIK"),
```
pauljamescleary: surpresses a lot of output when running unit and integration tests.  This is only for scalatest; not for the portal tests unfortunately.

modules/portal/app/modules/VinylDNSModule.scala
```diff
@@ -0,0 +1,108 @@
...
+import vinyldns.core.domain.membership.{UserChangeRepository, UserRepository}
+import vinyldns.dynamodb.repository.{
+  DynamoDBDataStoreSettings,
+  DynamoDBRepositorySettings,
+  DynamoDBUserChangeRepository,
+  DynamoDBUserRepository
+}
+
+class VinylDNSModule(environment: Environment, configuration: Configuration)
+    extends AbstractModule {
```
pauljamescleary: I moved this.  It used to be in a file in root called `Module`.  However, we cannot override that for test purposes.

Created this instead under a `modules` package.

modules/portal/conf/application-test.conf
```diff
@@ -60,3 +60,5 @@ links = [
     icon = ""
   }
 ]
+
+play.modules.enabled += "modules.VinylDNSModule"
```
pauljamescleary: this has to be added to config to load our module since it is no longer this "magical" file named `Module`.  Need to update our internal deploy repo with this as well.

modules/portal/test/controllers/LdapAuthenticatorSpec.scala
```diff
@@ -63,24 +63,22 @@ class LdapAuthenticatorSpec extends Specification with Mockito {
     Mocks(contextCreator, context, searchResults, searchNext, byDomainAuthenticator, mockAttributes)
   }
 
-  val app = GuiceApplicationBuilder().build()
-  val testApp = GuiceApplicationBuilder().configure(Map("portal.test_login" -> true)).build()
-  val settings = new Settings(app.configuration)
```
pauljamescleary: turns out we don't need to build our app or testApp for the unit tests in here.  I changed the unit tests to work without them.

modules/portal/test/controllers/TestApplicationData.scala
```diff
@@ -0,0 +1,170 @@
...
+
+  val mockAuth: Authenticator = mock[Authenticator]
+  val mockUserRepo: UserRepository = mock[UserRepository]
+  val mockUserChangeRepo: UserChangeRepository = mock[UserChangeRepository]
+
+  mockAuth.authenticate("frodo", "secondbreakfast").returns(Success(frodoDetails))
+  mockUserRepo.getUser(anyString).returns(IO.pure(Some(frodoUser)))
+  mockUserChangeRepo.save(any[UserChange]).returns(IO.pure(newFrodoLog))
+
+  def app: Application =
```
pauljamescleary: this is the gist of the change I made to allow for mocks instead of relying on dynamodb for unit tests.

We have to explicitly disable our module (otherwise the GuiceApplicationBuilder thing will complain and it will actually invoke our `VinylDNSModule` to load it which is sad.

Just bind our "dependencies" to mocks that the application needs to load.  If you look at `VinylDNSModule`, you will see that these are the only bindings we load.

modules/portal/conf/application-test.conf
pauljamescleary: I modified the test setup to build the app using mocks.  No longer need dynamodb to be running silently anymore \o/

build.sbt
rebstar6: wow this looks insane

pauljamescleary: http://www.scalatest.org/user_guide/using_the_runner

modules/portal/test/controllers/TestApplicationData.scala
```diff
@@ -0,0 +1,170 @@
...
+    Some("Gamgee"))
+
+  val frodoJson: String =
+    s"""{
+       |"name": "${frodoUser.userName}"
+       |}
+     """.stripMargin
+
+  val hobbitGroupId = "uuid-12345-abcdef"
+  val hobbitGroup: JsValue = Json.parse(s"""{
```
mitruly: Not sure if this is enabled for the portal, but it's probably better to have the `Json.parse(...)` on a new line since a few of these lines exceed 120 characters.

modules/portal/app/controllers/VinylDNS.scala
rebstar6: yeah i guess, it does feel like if the user is already in our system there is no need to check LDAP though...

rebstar6: not worth changing now, i imagine its more of a pain than just flipping the order, i think the ldap check handles casing issues for us

modules/portal/test/controllers/TestApplicationData.scala
rebstar6: autoformatting is the worst with config things.

not a problem, just saying ugh

project/Dependencies.scala
rebstar6: i dont think you need in the api either if its in core, right?

build.sbt
rebstar6: oh i believe it.

modules/portal/test/controllers/TestApplicationData.scala
```diff
@@ -0,0 +1,170 @@
...
+    None,
+    UserChangeType.Create
+  ).toOption.get
+
+  val serviceAccountDetails =
+    UserDetails("CN=frodo,OU=hobbits,DC=middle,DC=earth", "service", None, None, None)
+  val serviceAccount =
+    User("service", "key", "secret", None, None, None, DateTime.now, "service-uuid")
+
+  val frodoJsonString: String = s"""{
```
rebstar6: with some of this it may just be easier to make the json directly:

```  val frodoJsonDetails: JsValue = Json.obj(
    "userName" -> frodoUser.userName,
    "firstName" -> frodoUser.firstName,
    "lastName" -> frodoUser.lastName,
    "email" -> frodoUser.email,
    "created" -> frodoUser.created.toString,
    "id" -> frodoUser.id
  )``` works i believe

PR #207, opened 2018-09-14
Generic loader

rebstar6:
Code is done i believe, just need to add more tests

### DIFF
--- a/modules/core/src/main/scala/vinyldns/core/repository/DataStore.scala
+++ b/modules/core/src/main/scala/vinyldns/core/repository/DataStore.scala
@@ -24,6 +24,33 @@ import vinyldns.core.repository.RepositoryName.RepositoryName
 
 import scala.reflect.ClassTag
 
+object DataStore {
+  def apply(
+      userRepository: Option[UserRepository] = None,
+      groupRepository: Option[GroupRepository] = None,
+      membershipRepository: Option[MembershipRepository] = None,
+      groupChangeRepository: Option[GroupChangeRepository] = None,
+      recordSetRepository: Option[RecordSetRepository] = None,
+      recordChangeRepository: Option[RecordChangeRepository] = None,
+      zoneChangeRepository: Option[ZoneChangeRepository] = None,
+      zoneRepository: Option[ZoneRepository] = None,
+      batchChangeRepository: Option[BatchChangeRepository] = None,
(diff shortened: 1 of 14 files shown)


modules/dynamodb/src/main/scala/vinyldns/dynamodb/repository/DynamoDBDataStoreProvider.scala
```diff
@@ -105,8 +100,11 @@ class DynamoDBDataStoreProvider extends DataStoreProvider {
         zoneChange,
         DynamoDBZoneChangeRepository.apply(_, dynamoConfig)),
       IO.pure[Option[ZoneRepository]](None),
-      IO.pure[Option[BatchChangeRepository]](None)
-    ).parMapN { new DataStore(_, _, _, _, _, _, _, _, _) }
+      IO.pure[Option[BatchChangeRepository]](None),
+      initializeSingleRepo[UserChangeRepository](
+        userChange,
+        DynamoDBUserChangeRepository.apply(_, dynamoConfig, crypto))
+    ).parMapN { new DataStore(_, _, _, _, _, _, _, _, _, _) }
```
pauljamescleary: wonder if this becomes easier if you add a companion object with a `def apply` on it, would it allow you to do `.parMapN(DataStore.apply)` ??  not sure, but would be prettier than all these underscores

rebstar6: updated

modules/core/src/main/scala/vinyldns/core/repository/DataStore.scala
```diff
@@ -17,18 +17,40 @@
...
-  UserRepository
-}
+import vinyldns.core.domain.membership._
 import vinyldns.core.domain.record.{RecordChangeRepository, RecordSetRepository}
 import vinyldns.core.domain.zone.{ZoneChangeRepository, ZoneRepository}
 import vinyldns.core.repository.RepositoryName.RepositoryName
 
 import scala.reflect.ClassTag
 
+object DataStore {
```
mitruly: I don't think that this companion object is needed if we convert `class DataStore` → `case class DataStore` since we aren't changing any of the parameters. One of the advantages of case classes is a default `apply` constructor that uses the same parameters for the class.

PR #208, opened 2018-09-17
MySqlZoneChangeRepository

nimaeskandary:
implements a MySql version of the zone change repository. While similar to zone and batch change mysql repos, there is a difference in how paging is done. 

Instead of using `offset` sql statements, it uses string comparisons to determine if the items for the next page are `less than` the start from key, following an approach more similar to https://www.eversql.com/faster-pagination-in-mysql-why-order-by-with-limit-and-offset-is-slow/. 

The paging key is the millisecond time, appended by the uuid. The reasoning being if two timestamps are the same, it will then sort by uuid to return consistent pages if created times are the same for multiple records


modules/api/src/main/scala/vinyldns/api/repository/mysql/MySqlZoneChangeRepository.scala
```diff
@@ -0,0 +1,125 @@
...
+      |  WHERE zc.zone_id = {zoneId} AND
+      |        (
+      |          CASE WHEN {startFrom} != '' AND zc.created_time_change_id <= {startFrom}
+      |               THEN 1
+      |               WHEN {startFrom} = ''
+      |               THEN 1
+      |               ELSE 0
+      |          END
+      |        ) = 1
+      |  ORDER BY zc.created_time_change_id DESC
```
pauljamescleary: since you order by created time, should index on there as well

modules/api/src/main/scala/vinyldns/api/repository/mysql/MySqlZoneChangeRepository.scala
```diff
@@ -0,0 +1,125 @@
...
+      |  VALUES ({change_id}, {zone_id}, {data}, {created_time_change_id})
+      """.stripMargin
+
+  private final val LIST_ZONES_CHANGES =
+    sql"""
+      |SELECT zc.data
+      |  FROM zone_change zc
+      |  WHERE zc.zone_id = {zoneId} AND
+      |        (
+      |          CASE WHEN {startFrom} != '' AND zc.created_time_change_id <= {startFrom}
```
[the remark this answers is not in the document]
pauljamescleary: why do we need this, why not just `where createdtime_change_id < x`?

Also, looks like `createdtime_change_id` is a `VARCHAR(52)`, that is not terribly efficient for sorting in the index.  Is there anyway we can just use a number instead?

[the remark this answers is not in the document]
pauljamescleary: zone changes are unlike record set changes where we have a bunch of changes come in on the same timestamp.  I think the additional overhead / checking here for zone changes maybe overdone to address a problem that likely does not exist (for zone changes).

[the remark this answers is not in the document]
nimaeskandary: @pauljamescleary What are your thoughts on not having a created field in the table at all and using an auto-incremented field?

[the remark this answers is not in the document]
pauljamescleary: @nimaeskandary Obviously, it increases contention, but there is simply not a lot of churn on zones so it is probably fine.  I wouldn't recommend that as a solution for record changes.

I also think we should have a created / timestamp column on there anyway.  It goes against YAGNI here, but since we are storing in protobuf, it would require a migration if we ever wanted to query against that field in the future.

modules/api/src/it/scala/vinyldns/api/repository/mysql/MySqlZoneChangeRepositoryIntegrationSpec.scala
```diff
@@ -0,0 +1,202 @@
...
+        repo.listZoneChanges(zones(1).id, startFrom = None, maxItems = 1).unsafeToFuture(),
+        timeout) { retrieved =>
+        retrieved.items.size should equal(1)
+        retrieved.items should equal(expectedChanges)
+        retrieved.nextId should equal(expectedNext)
+        retrieved.startFrom should equal(None)
+      }
+    }
+
+    "get zone changes using a startFrom and maxItems" in {
```
britneywright: how many changes are in this zone?

nimaeskandary: three, i updated the test to be more straightforward in what its supposed to do

modules/api/src/main/scala/vinyldns/api/repository/mysql/MySqlZoneChangeRepository.scala
```diff
@@ -0,0 +1,106 @@
...
+      |SELECT zc.data
+      |  FROM zone_change zc
+      |  WHERE zc.zone_id = {zoneId} AND zc.created_timestamp <= {startFrom}
+      |  ORDER BY zc.created_timestamp DESC
+      |  LIMIT {maxItems}
+    """.stripMargin
+
+  override def save(zoneChange: ZoneChange): IO[ZoneChange] =
+    monitor("repo.ZoneChangeMySql.save") {
+      logger.info(s"Saving zone change ${zoneChange.id}")
```
pauljamescleary: logger.info should probably be inside the `IO`

modules/api/src/main/scala/vinyldns/api/repository/mysql/MySqlZoneChangeRepository.scala
```diff
@@ -0,0 +1,106 @@
...
+      }
+    }
+
+  override def listZoneChanges(
+      zoneId: String,
+      startFrom: Option[String],
+      maxItems: Int): IO[ListZoneChangesResults] =
+    // sorted from most recent, startFrom is an offset from the most recent change
+    monitor("repo.ZoneChangeMySql.listZoneChanges") {
+      logger.info(s"Getting zone changes for zone $zoneId")
```
pauljamescleary: same here

modules/api/src/main/scala/vinyldns/api/repository/mysql/MySqlZoneChangeRepository.scala
```diff
@@ -0,0 +1,106 @@
...
+        }
+      }
+    }
+
+  override def listZoneChanges(
+      zoneId: String,
+      startFrom: Option[String],
+      maxItems: Int): IO[ListZoneChangesResults] =
+    // sorted from most recent, startFrom is an offset from the most recent change
+    monitor("repo.ZoneChangeMySql.listZoneChanges") {
```
pauljamescleary: The monitor should be `repo.ZoneChange`.  Don't add the `MySql` suffix

modules/api/src/main/scala/vinyldns/api/repository/mysql/MySqlZoneChangeRepository.scala
```diff
@@ -0,0 +1,106 @@
...
+    sql"""
+      |SELECT zc.data
+      |  FROM zone_change zc
+      |  WHERE zc.zone_id = {zoneId} AND zc.created_timestamp <= {startFrom}
+      |  ORDER BY zc.created_timestamp DESC
+      |  LIMIT {maxItems}
+    """.stripMargin
+
+  override def save(zoneChange: ZoneChange): IO[ZoneChange] =
+    monitor("repo.ZoneChangeMySql.save") {
```
pauljamescleary: make the monitor `repo.ZoneChange.save`

modules/api/src/main/scala/vinyldns/api/repository/mysql/MySqlZoneChangeRepository.scala
```diff
@@ -0,0 +1,106 @@
...
+    extends ZoneChangeRepository
+    with ProtobufConversions
+    with Monitored {
+  private final val logger = LoggerFactory.getLogger(classOf[MySqlZoneChangeRepository])
+
+  private final val PUT_ZONE_CHANGE =
+    sql"""
+      |INSERT INTO zone_change (change_id, zone_id, data, created_timestamp)
+      |  VALUES ({change_id}, {zone_id}, {data}, {created_timestamp})
+      """.stripMargin
```
pauljamescleary: This does not have `put` semantics as it will fail if it already exists.  Need to use `REPLACE INTO` I believe

modules/api/src/it/scala/vinyldns/api/repository/mysql/MySqlZoneChangeRepositoryIntegrationSpec.scala
```diff
@@ -0,0 +1,211 @@
...
+  "MySqlZoneChangeRepository" should {
+    "successfully save a change" in {
+      val change = changes(1)
+      whenReady(repo.save(change).unsafeToFuture(), timeout) { saved =>
+        saved should equal(change)
+      }
+      whenReady(repo.listZoneChanges(change.zone.id).unsafeToFuture(), timeout) { retrieved =>
+        retrieved.items should equal(List(change))
+      }
+    }
```
pauljamescleary: add an "update zone change" test here

modules/api/src/main/scala/vinyldns/api/repository/mysql/JdbcBatchChangeRepository.scala
```diff
@@ -43,7 +43,7 @@ class JdbcBatchChangeRepository
 
   private final val PUT_BATCH_CHANGE =
     sql"""
-         |INSERT INTO batch_change(id, user_id, user_name, created_time, comments)
+         |REPLACE INTO batch_change(id, user_id, user_name, created_time, comments)
```
nimaeskandary: reverting this, wrong file

modules/api/src/it/scala/vinyldns/api/repository/mysql/MySqlZoneChangeRepositoryIntegrationSpec.scala
```diff
@@ -0,0 +1,234 @@
...
+      s.executeUpdate("DELETE FROM zone_change")
+    }
+    super.afterAll()
+  }
+
+  "MySqlZoneChangeRepository" should {
+    "save a change" in {
+      val change = changes(1)
+      // save the change
+      whenReady(repo.save(change).unsafeToFuture(), timeout) { saved =>
```
pauljamescleary: we shouldn't start with `whenReady`, sorry I missed this on the last PR.

Since everything is in `IO`, we can simply do `repo.save(change).unsafeRunSync()` for all the tests.

If you need to test failure, can do `repo.save(change).attempt.unsafeRunSync()` which gives back an `Either[Throwable, A]`

nimaeskandary: Oo that's cool

PR #211, opened 2018-09-18
Quiet build output

pauljamescleary:
* set the traceLevel in the build to -1, which means no exceptions
will be output when running
* set the akka.logLevel in a few files to `OFF` so we don't
log anything during tests

This quieted a lot of exceptions.


modules/portal/conf/logback-test.xml
```diff
@@ -0,0 +1,21 @@
+<configuration>
```
pauljamescleary: this basically just turns all logging off for running unit tests so we don't see strange logs and exceptions happen.

I left the appender in so people can toggle the settings locally if they need to.

PR #212, opened 2018-09-18
Fix for null pointer in user deserialization

pauljamescleary:
The fromItem function was not checking for the possibility of
null when retrieving the LOCK_STATUS string.

Added that check, defaulting to null.  Added a unit test.


nimaeskandary:
Tested it out, 👍

PR #213, opened 2018-09-18
allow for multiple ptr records in portal

nimaeskandary:
Allows for multiple ptr records in a ptr record set through the portal. Changes the portals record set create modal, and the view for ptr records to view a list rather than just one record.


nimaeskandary:
screen shot 2018-09-18 at 1 40 32 pm

nimaeskandary:
screen shot 2018-09-18 at 1 40 45 pm

modules/portal/app/views/zones/zoneTabs/manageRecords.scala.html
```diff
@@ -171,7 +171,18 @@ <h3 class="panel-title">Records {{ getRecordPageTitle() }}</h3>
                                 </li>
                             </ul>
                         </span>
-                        <span ng-if="record.type == 'PTR'">{{ record.ptrRecordData }}</span>
+                        <span ng-if="record.type == 'PTR'">
+                            <ul class="table-cell-list">
+                                <li ng-repeat="item in record.ptrRecordData track by $index"
+                                    ng-if="!record.onlyFour || $index <= 3">{{item}}</li>
```
mitruly: Would be nice to have a configurable value and not hard code checks, but that would be more of maintenance clean up outside of this card's scope.

PR #220, opened 2018-09-19
Fix authentication error in the portal

pauljamescleary:
The root cause for the authentication error is that the portal
was not decrypting the user secret key before signing requests.

This is solved via the following:

1. Update VinylDNS controller to decrypt user secret when needed
1. Make sure that the `encrypt-user-secrets` feature flag is `on`
in the API reference.conf.  This was why in testing locally we
did not hit the same issue that we saw in the development environment.
Because the flag was false, test users secrets were not encrypted.

* `portal application.conf` - set the crypto to match the API
* `Dependencies.scala` - eliminate some duplication of dependencies
* `api reference.conf` - set the encrypt-user-secrets flag to true
* `TestApplicationData.scala` - modify the mock play app to have a
CryptoAlgebra binding
* `VinylDNS` - add secret decryption in getUserCreds and processCSV
* `VinylDNSModule` - add binding for CryptoAlgebra for dependency
injection.

### DIFF
--- a/modules/api/src/main/resources/reference.conf
+++ b/modules/api/src/main/resources/reference.conf
@@ -133,5 +133,5 @@ vinyldns {
   batch-change-limit = 20
 
   # whether user secrets are expected to be encrypted or not
-  encrypt-user-secrets = false
+  encrypt-user-secrets = true
 }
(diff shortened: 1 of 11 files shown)


modules/api/src/main/resources/reference.conf
```diff
@@ -133,5 +133,5 @@ vinyldns {
   batch-change-limit = 20
 
   # whether user secrets are expected to be encrypted or not
-  encrypt-user-secrets = false
+  encrypt-user-secrets = true
```
rebstar6: holdover. should remove in followup (or now, i think it's just used once or twice)

pauljamescleary: oof.  OOOOOOFFFF!!!

We should kill this flag since this was just during the encryption migration

pauljamescleary: Killed this in #221

PR #224, opened 2018-09-20
handle CNAME wildcard records

### LINKED ISSUE #91: Wildcard records cause others to fail
In our validations, for an add, we skip the initial existence check if there is a wildcard that matches the record, as it will resolve to something in that case (ex. `A *.test.com` exists, trying to add `A new.test.com`). However, we do not skip for CNAME wildcards even though they cause other records to resolve (ex. `CNAME *.test.com` exists, trying to add `A new.test.com`)

**VinylDNS Version**
0.8.0

**To Reproduce**
1. add a CNAME wildcard (*.zone.name.com)
2. try to add A record (some-a.zone.name.com)

**Expected behavior**
The A record should be added

britneywright:
closes #91


modules/api/src/it/scala/vinyldns/api/domain/record/RecordSetServiceIntegrationSpec.scala
```diff
@@ -304,6 +315,19 @@ class RecordSetServiceIntegrationSpec
...
+
+    "add record if CNAME wildcard exists" in {
+      val newRecord = subTestRecordA.copy(name = "any-name")
+      val result =
+        testRecordSetService
+          .addRecordSet(newRecord, auth)
+          .value
+          .unsafeToFuture()
+          .mapTo[Either[Throwable, RecordSetChange]]
+      whenReady(result, timeout) { out =>
```
pauljamescleary: We are doing away with the `whenReady` stuff.

Should just do `.value.unsafeRunSync()`

modules/api/functional_test/live_tests/recordsets/create_recordset_test.py
```diff
@@ -1465,8 +1465,54 @@ def test_create_record_with_existing_wildcard_succeeds(shared_zone_test_context)
 
         test_create = client.create_recordset(test_rs, status=202)
         test_rs = client.wait_until_recordset_change_status(test_create, 'Complete')['recordSet']
-    except:
```
britneywright: I noticed this was making the wildcard tests pass regardless of what records or status i put in. So removed it.

modules/api/functional_test/live_tests/recordsets/create_recordset_test.py
```diff
@@ -1465,8 +1465,54 @@ def test_create_record_with_existing_wildcard_succeeds(shared_zone_test_context)
...
-        pass
+    finally:
+        try:
+            delete_result = client.delete_recordset(wildcard_rs['zoneId'], wildcard_rs['id'], status=202)
```
rebstar6: more a conceptual question than something specific to this PR - if our cleanup fails, should we really be allowing it to just pass through? i wanna say it should just fail, that could be why we get bad data lingering sometimes

modules/api/functional_test/live_tests/recordsets/create_recordset_test.py
```diff
@@ -1465,8 +1465,54 @@ def test_create_record_with_existing_wildcard_succeeds(shared_zone_test_context)
...
+                pass
+
+
+def test_create_record_with_existing_cname_wildcard_succeed(shared_zone_test_context):
+    """
+    Test that creating a record when a CNAME wildcard record already exists succeeds
+    """
+    client = shared_zone_test_context.ok_vinyldns_client
+
+    wildcard_rs = {
```
rebstar6: for cleanliness, we have a function `def get_recordset_json(zone, rname, type, rdata_list, ttl=200):` that makes this less json-y

so something like:
```
get_recordset_json(shared_zone_test_context, '*', 'CNAME', [{'cname': 'cname2.'}])
```

modules/api/src/it/scala/vinyldns/api/domain/record/RecordSetServiceIntegrationSpec.scala
```diff
@@ -84,6 +84,15 @@ class RecordSetServiceIntegrationSpec
...
     List(AAAAData("fd69:27cc:fe91::60")))
+  private val wildcardTestRecordCNAME = RecordSet(
+    zone.id,
+    "*",
+    CNAME,
+    38400,
+    RecordSetStatus.Active,
+    DateTime.now,
+    None,
+    List(AData("10.1.1.1")))
```
rebstar6: CnameData not AData

modules/api/functional_test/live_tests/recordsets/create_recordset_test.py
pauljamescleary: @rebstar6 assuming we are using the shared zone test context, that thing is always torn down _before_ func tests are run.  That pattern was put in place because sometimes these things fail to clean up.

Where we need to be ultra careful is if we are not working with that test fixture.

modules/api/src/it/scala/vinyldns/api/domain/record/RecordSetServiceIntegrationSpec.scala
```diff
@@ -304,6 +315,17 @@ class RecordSetServiceIntegrationSpec
...
+
+    "add record if CNAME wildcard exists" in {
+      val newRecord = subTestRecordA.copy(name = "any-name")
+      val result =
+        testRecordSetService
+          .addRecordSet(newRecord, auth)
+          .value
+          .unsafeRunSync()
+
+      rightValue(result).asInstanceOf[RecordSetChange].recordSet.name shouldBe "any-name"
```
pauljamescleary: result is an either.  Should use `EitherMatchers` or `EitherValues`, and not use `rightValue`.  Ultimately, we want to delete `ResultHelpers`.  I tried doing this but it was a big cleanup.

modules/api/src/it/scala/vinyldns/api/engine/RecordSetChangeHandlerIntegrationSpec.scala
```diff
@@ -0,0 +1,128 @@
...
+
+    zoneRepo.save(zone).unsafeRunSync()
+
+    // Seeding records in DB
+    val records = List(
+      wildcardTestRecordCNAME,
+      testRecordA
+    )
+
+    records.map(record => recordSetRepo.putRecordSet(record).unsafeRunSync())
```
britneywright: I tried to do `records.map(record => recordSetRepo.putRecordSet(record)).parSequence.unsafeRunSync()` here but it wasn't recognizing parSequence

pauljamescleary: `parSequence` needs to have the right imports.  I usually add `import cats._, cats.data._, cats.implicits._` at the top and then "Organize Imports" once everything compiles

mitruly: FWIW, I imported only `cats.implicits._` and it worked for me.

modules/api/src/it/scala/vinyldns/api/engine/RecordSetChangeHandlerIntegrationSpec.scala
```diff
@@ -0,0 +1,128 @@
...
+      recordSetRepo,
+      mock[RecordChangeRepository],
+      mock[UserRepository],
+      TestSqsService,
+      AccessValidations)
+  }
+
+  def tearDown(): Unit = ()
+
+  "RecordSetChangeHandler" should {
```
rebstar6: waiiit a sec, why is this not just in the existing RecordSetServiceIntegrationSpec?

pauljamescleary: I would argue this shouldn't be an integration spec at all.  There is already an existing unit test in `RecordSetChangeHandlerSpec` called `"bypass the validate and verify steps if a wildcard record exists"`

Should be an additional unit test there for CNAME.

pauljamescleary:
I wrote up the unit test, this just needs to be added to `RecordSetChangeHandlerSpec`.  I ran it locally and it worked fine.

Due to a rebase, posting to your PR was messed up...


```scala
"bypass the validate and verify steps if a wildcard CNAME exists" in {
      // Return empty as the wildcard record matching the type
      doReturn(IO.pure(List.empty))
        .when(mockRsRepo)
        .getRecordSets(rsChange.recordSet.zoneId, "*", rsChange.recordSet.typ)

      // Return a wildcard matching CNAME
      doReturn(IO.pure(List(rsChange.recordSet)))
        .when(mockRsRepo)
        .getRecordSets(rsChange.recordSet.zoneId, "*", RecordType.CNAME)

      // The second return is for verify
      doReturn(Interfaces.result(List()))
        .doReturn(Interfaces.result(List(rs)))
        .when(mockConn)
        .resolve(rs.name, rsChange.zone.name, rs.typ)

      doReturn(Interfaces.result(Right(NoError(mockDnsMessage))))
        .when(mockConn)
        .applyChange(rsChange)
      doReturn(IO.pure(cs)).when(mockChangeRepo).save(any[ChangeSet])
      doReturn(IO.pure(cs)).when(mockRsRepo).apply(any[ChangeSet])

      val test = underTest.apply(mockConn, rsChange)
      val res = test.unsafeRunSync()

      res.status shouldBe RecordSetChangeStatus.Complete

      verify(mockRsRepo).apply(rsRepoCaptor.capture())
      verify(mockChangeRepo).save(changeRepoCaptor.capture())

      val appliedCs = rsRepoCaptor.getValue
      appliedCs.status shouldBe ChangeSetStatus.Complete
      appliedCs.changes.head.status shouldBe RecordSetChangeStatus.Complete
      appliedCs.changes.head.recordSet.status shouldBe RecordSetStatus.Active

      // Our change should be successful
      val savedCs = changeRepoCaptor.getValue
      savedCs.status shouldBe ChangeSetStatus.Complete
      savedCs.changes.head.status shouldBe RecordSetChangeStatus.Complete

      // make sure the record was applied
      verify(mockConn).applyChange(rsChange)

      // make sure we never called resolve, as we skip validate step and verify
      verify(mockConn, never).resolve(rs.name, rsChange.zone.name, rs.typ)

      val batchChangeUpdates = batchRepo.getBatchChange(batchChange.id).unsafeRunSync()
      val updatedSingleChanges = completeCreateAAAASingleChanges.map { ch =>
        ch.copy(
          status = SingleChangeStatus.Complete,
          recordChangeId = Some(rsChange.id),
          recordSetId = Some(rsChange.recordSet.id))
      }
      val scExpected = notUpdatedChange :: updatedSingleChanges
      batchChangeUpdates.get.changes shouldBe scExpected
    }
```

PR #229, opened 2018-09-20
Load configured repos in the portal

rebstar6:
🎆 ?


modules/portal/app/modules/VinylDNSModule.scala
```diff
@@ -32,78 +32,44 @@ package modules
...
-      key = dynamoConfig.get[String]("key"),
-      secret = dynamoConfig.get[String]("secret"),
-      endpoint = dynamoConfig.get[String]("endpoint"),
-      region = dynamoConfig.get[String]("region")
-    )
+    val startApp = for {
+      cryptoConf <- settings.cryptoConfig
+      crypto <- CryptoAlgebra.load(cryptoConf)
+      repoConfigs <- settings.dataStoreConfigs
+      repositories <- DataStoreLoader.loadAll[PortalDataAccessor](repoConfigs, crypto, PortalDataAccessorProvider)
```
rebstar6: this was delightfully seamless. go me.

modules/portal/app/controllers/Settings.scala
```diff
@@ -36,6 +40,13 @@ class Settings(private val config: Configuration) {
 
   val portalTestLogin: Boolean = config.getOptional[Boolean]("portal.test_login").getOrElse(false)
 
+  val dataStoreConfigs: IO[List[DataStoreConfig]] =
+    IO(config.underlying.getStringList("data-stores").asScala.toList).flatMap { lst =>
```
pauljamescleary: this looks unpleasant, isn't there something in pureconfig where we can do a loadConfig on data-stores? and just flatmap through it?

rebstar6: data stores is a list of strings that are basically keys into the data store configs. we use "mysql" and "dynamodb" as said keys, but really if someone wanted to put
```
data-stores = ["sweetkey"]

sweetkey {
   class-name/settings/repos...
}
```
it would also work

rebstar6: actually, i *think* i can cut out some of the java nonsense with 
```
  val dataStoreConfigs: IO[List[DataStoreConfig]] =
    loadConfigF[IO, List[String]](config.underlying, "data-stores").flatMap { lst =>
      lst.map(loadConfigF[IO, DataStoreConfig](config.underlying, _)).parSequence
    }
```

will test it out

PR #230, opened 2018-09-21
Parallelizing the build

pauljamescleary:
Using the `all` command in sbt allows us to run certain tasks
in parallel.  Not everything can be done in parallel, so we have to use
judgement here.  Some things like dockerCompose cannot be done in parallel.

With the adjustments, local `;validate;verify` went from 12 minutes to 9
minutes, a savings of 3 full minutes!

**build.sbt**
Most changes are here.

Allow parallelExecution by default.  The reason is that test suites will be run
in parallel.  The play framework by default turns this off.  Some of the changes
made in here around the `InMemoryBatchChangeRepository` were necessary when I
flipped this on.

We cannot run in parallel by default for IntegrationTest.  The reason is that
several of the api integration tests use the same zone repo, so they wind up
stomping on each other.

Added a `killDocker` task that is much faster to run than `dockerComposeStop`

Enabled parallelExecution in IntegrationTest in dynamodb.  The integration tests
do not conflict with each other here.

Changed the command aliases to use parallel `all`

**InMemoryBatchCHangeRepository**
This used to be a singleton, which prevents running unit tests in parallel.
Made this a class, and updated unit tests to use the class instead.

**logback-test.xml**
We were still logging in odd ways in places.  Removed this to turn logging off
by default for tests.

**.jvmopts**
Kept running out of metaspace.  Increased the memory needed to help slow that
down.


mitruly:
Timed this (with bash `time`) and got:
```
real	4m4.919s
user	14m46.894s
sys	0m47.839s
```

In comparison, `master` branch shows:
```
real	5m44.509s
user	16m4.061s
sys	0m52.005s
```

build.sbt
```diff
@@ -70,7 +70,7 @@ lazy val sharedSettings = Seq(
 )
 
 lazy val testSettings = Seq(
-  parallelExecution in Test := false,
+  parallelExecution in Test := true,
   parallelExecution in IntegrationTest := false,
```
mitruly: Heads up for reviewers: this has to be `true` because there are some conflicts when running API integration tests in parallel. (Tested this locally.)

PR #239, opened 2018-09-25
Increase MySQL retry count/sleep duration for API startup script

mitruly:
There are transient failures in Docker where max retries (currently set to 30) are exceeded for MySQL, and then functional tests fail. Example build: https://travis-ci.org/vinyldns/vinyldns/jobs/431178316.

- Increase retries from 30 to 40
- Increase timeout duration of API start from 0.5 seconds to 1 second (to also match func test start)

### DIFF
--- a/docker/api/run.sh
+++ b/docker/api/run.sh
@@ -23,7 +23,7 @@ export JDBC_MIGRATION_URL="jdbc:mariadb://${MYSQL_ADDRESS}:${MYSQL_PORT}/?user=$
 # wait until mysql is ready...
 echo 'Waiting for MYSQL to be ready...'
 DATA=""
-RETRY=30
+RETRY=60
 while [ $RETRY -gt 0 ]
 do
     DATA=$(nc -vzw1 vinyldns-mysql 3306)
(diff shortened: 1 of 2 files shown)


rebstar6:
@mitruly did you compare the mysql output in that build to a working one? im curious whats taking it so long

mitruly:
@rebstar6 Yes, I compared the output builds between passing and failing merges to master and it seemed that ones that failed had exceeded the MySQL timeout (not to be confused with the longer retry counts on the vinyldns-api).

docker/api/run.sh
```diff
@@ -23,7 +23,7 @@ export JDBC_MIGRATION_URL="jdbc:mariadb://${MYSQL_ADDRESS}:${MYSQL_PORT}/?user=$
 # wait until mysql is ready...
 echo 'Waiting for MYSQL to be ready...'
 DATA=""
-RETRY=30
+RETRY=60
 while [ $RETRY -gt 0 ]
```
rebstar6: at the same time, vinyldns-functest is waiting on the api to startup. if you increase this, i think you also have to increase the func test wait time.

i also wonder if we should just increase the time between checks.

docker/functest/run.sh
```diff
@@ -4,6 +4,7 @@ VINYLDNS_URL="http://vinyldns-api:9000"
 echo "Waiting for API to be ready at ${VINYLDNS_URL} ..."
 DATA=""
 RETRY=60
+SLEEP_DURATION=1
```
rebstar6: this file needs to have more time than docker/api/run.sh (higher RETRY). assume they both start at the same time, this needs to wait for the api, which really means waiting for both mysql AND the api. it was intentional that this had a higher retry

Issue #200, opened 2018-09-12
Create SQS Implementation of Message Queue


pauljamescleary:
**Scope**

1. Create a new SQS sub-module
1. Create the Message Queue "interface" in the core sub-module
1. Create an implementation of the Message Queue interface for SQS
1. Use integration tests / docker compose (like we do for the dynamodb module) to spin up SQS in docker to verify functionality

**Notes**

1. The message queue implementation is responsible for the serialization of the zone command
1. The message queue implementation is _not_ responsible for retries (although optionally it may choose to do so).
1. The existing `EngineCommandBus` can likely be replaced with the new `MessageQueue`.

**Do not be afraid to alter the design of the MessageQueue algebra.  SQS may dictate subtle changes than the proposed design.**

**Proposed Design**

```scala
// for ultimately acknowledging or re-queue of the message, this is queue implementation independent
trait MessageHandle
abstract class Message[A <: ZoneCommand] {
  def handle: MessageHandle 

  // A vinyldns deserialized zone command
  def command: A
}

// main consumer
trait MessageQueue {

  // receives a single message, the command should already be deserialized
  def receive(): IO[Message]

  // receives a batch of messages
  def receiveBatch(): IO[List[Message]]

  // puts the message back on the queue with the intention of having it re-processed again
  def requeue(message: Message): IO[Unit]

  // removes a message from the queue, indicating completion or the message should never be processed
  def remove(message: Message): IO[Unit]

  // alters the visibility timeout for a message on the queue.  
  def changeMessageTimeout(message: Message, duration: FiniteDuration): IO[Unit]

  // sends a zone command to the queue, the queue will need to serialize it (if necessary) first
  def send(command: ZoneCommand): IO[Unit]

  // sends a batch of messages to the queue, the queue should serialize those and return List[IO[Message]]
  // where each IO could individually fail.  Note: retry semantics is not a requirement of the queue implementation
  def sendBatch(messages: List[ZoneCommand]): List[IO[Message]]
}

// follow data store loading on how to load the message queue settings using pure config, and how the 
// implementation loads and initializes itself.
trait MessageQueueProvider {
  def load(settings: MessageQueueSettings): IO[MessageQueue]
}
```

PR #242, opened 2018-10-01
Implement trait for MessageQueue in core module

mitruly:
Create implementation of `MessageQueue` in core module.

Part of #200.


modules/core/src/main/scala/vinyldns/core/queue/MessageQueueConfig.scala
```diff
@@ -0,0 +1,9 @@
+package vinyldns.core.queue
+
+final case class MessageQueueConfig(
```
rebstar6: this should be the main config, which is just 
```
final case class MessageQueueConfig(
className: String
settings: Config
}
```

the stuff you have in here is SQS-specific I believe

modules/core/src/main/scala/vinyldns/core/queue/MessageQueueProvider.scala
```diff
@@ -0,0 +1,8 @@
+package vinyldns.core.queue
+import cats.effect.IO
+
+// follow data store loading on how to load the message queue settings using pure config, and how the
+// implementation loads and initializes itself.
+trait MessageQueueProvider[A] {
+  def load(settings: MessageQueueConfig): IO[MessageQueue[A]]
```
rebstar6: is MessageQueue typed? doesnt appear so

modules/core/src/main/scala/vinyldns/core/queue/MessageQueue.scala
```diff
@@ -0,0 +1,40 @@
...
+trait MessageHandle
+abstract class Message[A <: ZoneCommand] {
+  def handle: MessageHandle
+
+  // A vinyldns deserialized zone command
+  def command: A
+}
+
+// main consumer
+trait MessageQueue {
```
rebstar6: to the last comment, if the idea is to have `Message[A]`, Id assume you want `trait MessageQueue[A <: ZoneCommand]` where each of these functions return `IO[Message[A]]`

rebstar6: err actually not sure, is each MessageQueue tied to a type of ZoneCommand, or it can return many types?

pauljamescleary: MessageQueue can represent different commands.  The `Message` is the only thing that needs a type parameter.

modules/core/src/main/scala/vinyldns/core/queue/MessageQueue.scala
```diff
@@ -0,0 +1,40 @@
...
+
+  // receives a single message, the command should already be deserialized
+  def receive(): IO[Message[ZoneCommand]]
```
pauljamescleary: @rebstar6 is right, this return type is a little funny.  `IO[Message[_]]` perhaps?

Maybe we need kind projector to help us here

modules/core/src/main/scala/vinyldns/core/queue/MessageQueue.scala
rebstar6: but then wont the type always be set at ZoneCommand?

modules/core/src/main/scala/vinyldns/core/queue/MessageQueueConfig.scala
pauljamescleary: @mitruly leave the config stuff out for this PR.  We only need to establish a MessageQueue and an SQSMessageQueue implementation.

modules/core/src/main/scala/vinyldns/core/queue/MessageQueue.scala
rebstar6: like, wouldnt you then need to have receive instead be
```
 def receive[A <: ZoneCommand](): IO[Message[A]]
````

Comment (pauljamescleary) [reply modules/core/src/main/scala/vinyldns/core/queue/MessageQueue.scala]: I think we have to play with the type parameters here.  In theory, you shouldn't know what type of zone command you will receive.  Will do some tinkering and post back an update.

Comment (rebstar6) [reply modules/core/src/main/scala/vinyldns/core/queue/MessageQueue.scala]: when does it get split into it's type? if its not til later, you can just kill the typing on Message id think and case match on it Message.command later

Comment (mitruly) [reply modules/core/src/main/scala/vinyldns/core/queue/MessageQueue.scala]: Would it be simpler to define a trait as `Message` and have `RecordSetChangeMessage`/`ZoneChangeMessage` extend it? We can then avoid typing of `Message`..

Comment (pauljamescleary) [reply modules/core/src/main/scala/vinyldns/core/queue/MessageQueue.scala]: 1. I did not use `Message` since it name collides with a billion other `Message` things, importantly for us the `Message` in the sqs client.  Changed to `CommandMessage`
1. You could argue we don't need the sub types at all if CommandMessage is already polymorphic on the command itself.
1. Use `CommandMessage[_]` to omit the actual type.

Comment (pauljamescleary) [reply modules/core/src/main/scala/vinyldns/core/queue/MessageQueue.scala]: In SQS, the `receiptHandle` is different than the `messageId`.  So, it does not make a lot of sense to hold some kind of `handle` on any `send` operations.

The errors that can come through on any operation are an `Exception`, which we cannot do much about.  So `IO` makes sense.

It may make sense to just have the `send` all work on `Unit`

Comment (mitruly) [reply modules/core/src/main/scala/vinyldns/core/queue/MessageQueue.scala]: @pauljamescleary That's a good point. I updated the `send` and `sendBatch` commands to return `IO[Unit]`.

File: modules/core/src/main/scala/vinyldns/core/queue/MessageQueue.scala [hunk 5]
```diff
@@ -0,0 +1,54 @@
...
+
+  // alters the visibility timeout for a message on the queue.
+  def changeMessageTimeout(message: CommandMessage[_], duration: FiniteDuration): IO[Unit]
+
+  // sends a zone command to the queue, the queue will need to serialize it (if necessary) first
+  def send[A <: ZoneCommand](command: A): IO[Unit]
+
+  // sends a batch of messages to the queue, the queue should serialize those and return List[IO[Message]]
+  // where each IO could individually fail.  Note: retry semantics is not a requirement of the queue implementation
+  def sendBatch(messages: List[ZoneCommand]): IO[Unit]
```
rebstar6: `[A <: ZoneCommand]` here too?

rebstar6: fwiw, i believe this will only be called with RecordSetChanges, so may not even need it to be generic

modules/core/src/main/scala/vinyldns/core/queue/MessageQueue.scala
```diff
@@ -0,0 +1,54 @@
...
+package vinyldns.core.queue
+import cats.effect.IO
+import vinyldns.core.domain.record.RecordSetChange
+import vinyldns.core.domain.zone.{ZoneChange, ZoneCommand}
+
+import scala.concurrent.duration.FiniteDuration
+
+// for ultimately acknowledging or re-queue of the message, this is queue implementation independent
+trait MessageHandle
+sealed abstract class CommandMessage[A <: ZoneCommand](val handle: MessageHandle, val command: A) {}
```
rebstar6: whats the difference between having this sealed abstract class and:
```
trait CommandMessage[A <: ZoneCommand]{
  val handle: MessageHandle
  val command: A
}

final case class RecordChangeMessage(handle: MessageHandle, command: RecordSetChange) extends CommandMessage[RecordSetChange]
```

mitruly: I read that the only difference is that you can't extend multiple abstract classes (see: https://stackoverflow.com/questions/41078797/difference-between-using-sealed-trait-and-sealed-abstract-class-as-a-base-class). Not sure if there is a preference either way.

modules/core/src/main/scala/vinyldns/core/queue/MessageQueue.scala
pauljamescleary: I simplified the interface a little:

1. there is no need to have send vs. sendBatch, you can send a `List` of 1
1. we do not need multiple receives, as we can return a list of messages, and a list of messages can have one.  We can discuss whether or not we want to add a `count` parameter to the `receive`

Can't say I love the `send` yet?

`IO[NonEmptyList[Either[SendError, SendOk]]]`

mitruly:
Also removed `MessageQueueProvider` and `MessageQueueConfig` per @pauljamescleary's comment on implementing `MessageQueueTrait` and the corresponding SQS implementation, leaving the config for a later PR.

modules/core/src/main/scala/vinyldns/core/queue/MessageQueue.scala
rebstar6: that said, you could overload directly in the trait with 
```
def send(messages: ZoneCommand): IO[Either[SendError, SendOk]] = {
   // call NEL send, return .head
}
```
and then no implementation is needed in each queue type

rebstar6: you also probably dont need SendOK, could just be `Either[SendError, ZoneCommand]` in theory

pauljamescleary: agree @rebstar6, may make sense to have a `send` vs. a `sendBatch` still.  I haven't found another return type yet that I love.  The requirement / pain in the butt is that we want to know which commands failed vs. which ones succeeded from a caller's perspective.

We could do something simpler like...

```scala
case class SendBatchResult(successes: List[ZoneCommand], failures: List[ZoneCommand]
```

And call it a day.

mitruly: On a slightly related note, if we're going with the approach of using `NonEmptyList[A]` for commands, thinking about doing the same for `remove` since SQS supports batch removal as well.

pauljamescleary: @mitruly good point, but I imagine we would have the same `send` problem.  Namely, need to communicate successes and failures somehow.

For our existing `ZoneCommandHandler`, we do not batch.  The only reason to batch is to reduce traffic to the queue.  I don't think we have that requirement yet.  We _do_ however have the requirement to send a batch, and to know the specific results of reach item in the batch

PR #243, opened 2018-10-01
remove pending code from record set changes

nimaeskandary:
Seems the only places status was being used was when the zone sync handler was saving a changeset, it would make it `complete`

This removes status related code from changesets and from the repository. I could change this pr to only remove the repository functions, and keep status in the changeset object. If so, would have the question on whether to keep status in `fromItem` and `toItem` should do on a changeset,


nimaeskandary:
Will make internal cards to update our prod and dev dynamo tables, but the exact changes depend on how the discussion goes on keeping status in the change set model

modules/api/src/main/scala/vinyldns/api/engine/ZoneSyncHandler.scala
```diff
@@ -71,7 +71,7 @@ object ZoneSyncHandler extends DnsConversions with Monitored {
             } else {
               logger.info(
                 s"Zone sync for change $zoneChange found ${changes.size} changes to be saved")
-              val changeSet = ChangeSet(changes).copy(status = ChangeSetStatus.Applied)
+              val changeSet = ChangeSet(changes).copy()
```
pauljamescleary: probably don't need `copy` if you aren't changing the values.  

Also, the reason we move it to `applied` is to indicate for a zone sync that the change set is complete.  We do not want to process it.  Why remove this?  I think it should stay in.

modules/dynamodb/src/main/scala/vinyldns/dynamodb/repository/DynamoDBRecordChangeRepository.scala
```diff
@@ -151,74 +133,6 @@ class DynamoDBRecordChangeRepository private[repository] (
...
-      dynamoDBHelper.queryAll(queryRequest).map { queryResults =>
-        queryResults.flatMap { queryresult =>
-          queryresult.getItems.asScala.toList.map { item =>
-            item.get(ZONE_ID).getS()
-          }
-        }.distinct
-      }
-    }
-
   def getChanges(zoneId: String): IO[List[ChangeSet]] =
```
rebstar6: we dont use getChanges either it seems, we also have listRecordSetChanges...

modules/api/src/main/scala/vinyldns/api/engine/ZoneSyncHandler.scala
rebstar6: empty copy

modules/api/src/test/scala/vinyldns/api/engine/ZoneSyncHandlerSpec.scala
```diff
@@ -139,7 +139,7 @@ class ZoneSyncHandlerSpec
 
   private val testRecordSetChange = RecordSetChangeGenerator.forSyncAdd(testRecord2, testZone)
   private val testChangeSet =
-    ChangeSet.apply(testRecordSetChange).copy(status = ChangeSetStatus.Applied)
+    ChangeSet.apply(testRecordSetChange).copy()
```
rebstar6: empty copy

PR #247, opened 2018-10-02
main queue loader

rebstar6:


modules/core/src/main/scala/vinyldns/core/queue/MessageQueueLoader.scala
```diff
@@ -0,0 +1,34 @@
...
+import org.slf4j.LoggerFactory
+
+object MessageQueueLoader {
+
+  private val logger = LoggerFactory.getLogger("MessageQueueLoader")
+
+  def load(config: MessageQueueConfig): IO[MessageQueue] = {
+    logger.error(s"Attempting to load queue ${config.className}")
+    for {
+      className <- IO.pure(config.className)
```
pauljamescleary: probably don;t need the `IO.pure` here, just do in the second `IO(Class.forName(config.className)...`

modules/core/src/test/scala/vinyldns/core/queue/MockMessageQueueProvider.scala
```diff
@@ -0,0 +1,40 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package vinyldns.core.queue
+
+import cats.effect.IO
+import org.scalatest.mockito.MockitoSugar
+
+object MockMessageQueueProvider extends MockitoSugar {
```
pauljamescleary: curious as to why have these in a separate file, if they are only used by the unit test

modules/core/src/main/scala/vinyldns/core/queue/MessageQueueLoader.scala
rebstar6: hah yeah, for some reason we do that in DataStoreLoader too. will clean up in both

modules/core/src/test/scala/vinyldns/core/queue/MockMessageQueueProvider.scala
rebstar6: shrugs, can move in

modules/core/src/main/scala/vinyldns/core/queue/MessageQueueLoader.scala
```diff
@@ -0,0 +1,32 @@
...
+import cats.effect.IO
+import org.slf4j.LoggerFactory
+
+object MessageQueueLoader {
+
+  private val logger = LoggerFactory.getLogger("MessageQueueLoader")
+
+  def load(config: MessageQueueConfig): IO[MessageQueue] =
+    for {
+      _ <- IO(logger.error(s"Attempting to load queue ${config.className}"))
```
nimaeskandary: should this be an logger.error? believe there's a logger.info as well

rebstar6: yeah, our log level doesnt show info, anything startup related i error to guarantee it will print

modules/core/src/test/scala/vinyldns/core/queue/MessageQueueLoaderSpec.scala
```diff
@@ -0,0 +1,82 @@
...
+ */
+
+package vinyldns.core.queue
+
+import cats.effect.IO
+import cats.scalatest.{EitherMatchers, EitherValues, ValidatedMatchers}
+import com.typesafe.config.{Config, ConfigFactory}
+import org.scalatest.mockito.MockitoSugar
+import org.scalatest.{Matchers, WordSpec}
+import vinyldns.core.crypto.{CryptoAlgebra, NoOpCrypto}
```
mitruly: The tests don't seem to require the crypto stuff.

pauljamescleary: Right, are these required?

Also, we should fail to compile with unused imports.  Perhaps something isn't quite right with sbt and core?

rebstar6: its that i had a crypto defined, bad copy/pasta.

modules/core/src/test/scala/vinyldns/core/queue/MessageQueueLoaderSpec.scala
```diff
@@ -0,0 +1,80 @@
...
+
+}
+
+class MessageQueueLoaderSpec
+    extends WordSpec
+    with Matchers
+    with MockitoSugar
+    with EitherValues
+    with EitherMatchers
+    with ValidatedMatchers {
```
pauljamescleary: do we need `EitherValues`, `EitherMatchers`, `ValidatedMatchers`, or `MockitoSugar` here?

modules/core/src/test/scala/vinyldns/core/queue/MessageQueueLoaderSpec.scala
```diff
@@ -0,0 +1,76 @@
...
+
+  def load(config: MessageQueueConfig): IO[MessageQueue] =
+    IO.raiseError(new RuntimeException("boo"))
+
+}
+
+class MessageQueueLoaderSpec
+    extends WordSpec
+    with Matchers
+    with MockitoSugar {
```
pauljamescleary: are you using MockitoSugar, (things like `mock[A]` or `spy[B]`)

PR #248, opened 2018-10-02
Create pull_request_template.md

britneywright:
Changes in this pull request:

- Add a pull request template

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,5 @@
+Fixes #.
+
+Changes in this pull request:
+-
+-


britneywright:
It's so much though 😩. maybe we should simplify that. Why do we need to list impacted files if you can see that in the PR.

rebstar6:
yeah, i think anything we put in here we're gonna have to be pretty good about actually doing ourselves...the less overhead the better i think

britneywright:
my main goal is to get us back in the habit of referencing the issue primarily and putting in some type of description. The template is intended as more of a reminder than a strict set of requirements.

pauljamescleary:
I like it @britneywright

pauljamescleary:
Am I the only one who uses `*` instead of `-` for bulleted lists?  Curious.

britneywright:
yes 😁 
just kidding, but I definitely use `-` instead `*`

PR #249, opened 2018-10-02
Add new command handler for new queue algebra

pauljamescleary:
The new command handler largely mimics the existing `ZoneCommandHandler`.  There
are some subtle differences:

1. No longer rely on the ApiDataAccessor, instead, explicitly pass in the repos
required.  This better practice and simplifies testing.
1. No longer need this intermediate `Request` types.  The new `CommandMessage`
takes care of pairing up the `MessageHandle` with the message for us
1. No longer deserialize in the command handler.  The `MessageQueue`
implementation is expected to deserialize for us
1. Always call `requeue` on failure (as opposed to nothing).  While SQS does not
need a requeue, other message queue implementations may need one


modules/api/src/main/scala/vinyldns/api/backend/CommandHandler.scala
```diff
@@ -0,0 +1,210 @@
...
+        mq.remove(msg)
+      case RetryMessage(msg) =>
+        // Nothing to do here, the message will be retried after visibility timeout
+        logger.error(s"Message failed, retrying; $msg")
+        mq.requeue(msg)
+    }.map(_ => ())
+
+  def run(
+      mq: MessageQueue,
+      count: MessageCount,
```
rebstar6: just wording - count feels too generic of a word for what this represents (to the point of being confusing). Nothing way better comes to mind immediately - something like amountPerPull but less meh sounding

rebstar6: `amountPerPull: PositiveInt` (would need to rebrand MessageCount, but it gets the point across better I think)

rebstar6: pollingInterval/pollingCount maybe?

modules/api/src/main/scala/vinyldns/api/backend/CommandHandler.scala
```diff
@@ -0,0 +1,210 @@
...
+
+    // concurrently run 4 message batches, so we can have 40 messages max running concurrently
+    def flow(): Stream[IO, Unit] =
+      messageSource
+        .join(4)
+        .observe(increaseTimeoutForZoneSyncs)
+        .through(changeRequestProcessor)
+        .to(updateQueue)
+        .handleErrorWith { error =>
+          logger.error("Encountered unexpected error in main flow", error)
```
rebstar6: more context on this error - "Encountered unexpected error in CommandProcessor main flow"...or something

modules/api/src/main/scala/vinyldns/api/backend/CommandHandler.scala
```diff
@@ -0,0 +1,210 @@
...
+      // every delay duration, we poll
+      scheduler
+        .fixedDelay[IO](pollingInterval)
+        .evalMap[Chunk[CommandMessage]] { _ =>
+          // get the messages from the queue, transform them to a Chunk of messages
+          mq.receive(count).map(msgs => Chunk(msgs: _*))
+        }
+        .map {
+          // Need to produce a stream of streams so we can run "batches" in parallel
+          Stream.chunk(_).covary[IO]
```
rebstar6: hmm, can you explain this briefly at some point. looks like were going List to Chunk to Stream?

modules/api/src/main/scala/vinyldns/api/backend/CommandHandler.scala
```diff
@@ -0,0 +1,210 @@
...
+      mq: MessageQueue,
+      count: MessageCount,
+      pollingInterval: FiniteDuration,
+      pauseSignal: Signal[IO, Boolean])(implicit scheduler: Scheduler): Stream[IO, Unit] = {
+
+    // Polls queue for message batches, connected to the signal which is toggled in the status endpoint
+    val messageSource = startPolling(mq, count, pollingInterval).pauseWhen(pauseSignal)
+
+    // Increase timeouts for zone syncs as they can take 10s of minutes
+    val increaseTimeoutForZoneSyncs = changeVisibilityTimeoutForZoneSyncs(mq)
```
rebstar6: while we're in here, can we rename this too? its now for sync AND create (because create does a sync). Dont want people thinking its just for explicit sync

modules/api/src/main/scala/vinyldns/api/backend/CommandHandler.scala
```diff
@@ -0,0 +1,210 @@
...
+            } yield syncChange
+
+          outcomeOf(message)(doSync)
+
+        case zoneChange: ZoneChange =>
+          outcomeOf(message)(zoneChangeProcessor(zoneChange))
+
+        case rcr: RecordSetChange =>
+          val dnsConn =
+            DnsConnection(rcr.zone.connection.getOrElse(VinylDNSConfig.defaultZoneConnection))
```
rebstar6: i know this is old news, but while were in here I wouldnt hate if the default zone connection was passed in rather than read from this config object

modules/api/src/main/scala/vinyldns/api/backend/CommandHandler.scala
```diff
@@ -0,0 +1,210 @@
...
+    val changeRequestProcessor =
+      processChangeRequests(zoneChangeHandler, recordChangeHandler, zoneSyncHandler)
+
+    // Delete messages from message queue when complete
+    val updateQueue = messageSink(mq)
+
+    // concurrently run 4 message batches, so we can have 40 messages max running concurrently
+    def flow(): Stream[IO, Unit] =
+      messageSource
+        .join(4)
```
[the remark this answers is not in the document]
pauljamescleary: Good idea, but I want to postpone any changes until we "rethink" processing.

As it is now, we have 6 nodes x 40 concurrent messages per node so a theoretical max throughput of 240 concurrent messages being processed.  It is somewhat non-deterministic, and I would like bigger changes, but for the time being sticking with what has worked for the last 16 months

modules/api/src/main/scala/vinyldns/api/backend/CommandHandler.scala
pauljamescleary: I can rename it, perhaps `msgsPerPoll`

modules/api/src/main/scala/vinyldns/api/backend/CommandHandler.scala
```diff
@@ -0,0 +1,210 @@
...
+
+        case rcr: RecordSetChange =>
+          val dnsConn =
+            DnsConnection(rcr.zone.connection.getOrElse(VinylDNSConfig.defaultZoneConnection))
+          outcomeOf(message)(recordChangeProcessor(dnsConn, rcr))
+      }
+    }
+
+  def outcomeOf[A](message: CommandMessage)(p: => IO[A]): IO[MessageOutcome] =
+    IO.pure(logger.info(s"Running change request $message"))
```
rebstar6: how much info is gonna be in each of these log statements?

modules/api/src/main/scala/vinyldns/api/backend/CommandHandler.scala
pauljamescleary: We need to produce a "stream of streams", so that we can run multiple streams in parallel.

When I reviewed this earlier this year with Michael Pilquist, his recommendation on doing `List -> Stream` was via chunking.

modules/api/src/main/scala/vinyldns/api/backend/CommandHandler.scala
pauljamescleary: Good idea

modules/api/src/main/scala/vinyldns/api/backend/CommandHandler.scala
pauljamescleary: can do

modules/api/src/main/scala/vinyldns/api/backend/CommandHandler.scala
pauljamescleary: The log statements aren't great, I was mostly pulling through what we already had.  I think before we just logged the message id.  This will log the `MessageHandle` assuming a suitable `toString` along with the change itself.  We have `toString` overrides for those.

modules/api/src/main/scala/vinyldns/api/backend/CommandHandler.scala
pauljamescleary: Actually, in the log output, you always get the class name, so the log entry would look something like ...

`vinyldns.api.backend.CommandHandler - Encountered unexpected error in main flow`

And the exception would log all of the line numbers

rebstar6: ah, right, cool

PR #255, opened 2018-10-03
Implement message queue in AWS SQS

### LINKED ISSUE #200: Create SQS Implementation of Message Queue

mitruly:
Implement SQS `MessageQueue` interface in separate module.
- Extract current core SQS implementation into its own module
- Implement integration tests
- Configure module for docker

Closes #200.

----
### Note to reviewers:
- Lots of the changed files were from moving `ZoneCommand.scala` from `vinyldns.core.domain.zone` to `vinyldns.core.domain`.
- AWS SQS does not really offer a clean solution to keep track of current message visibility timeout. There are basically two approaches:
  - Use a `RetryPolicy` to define retry condition, exponential backoff and max delay for config when building the AWS service (easier)
  - Add timeout to `CommandMessage`/`MessageHandle` and manually remove and re-add `ZoneCommand`s for re-queue


build.sbt
```diff
@@ -293,6 +294,22 @@ lazy val mysql = (project in file("modules/mysql"))
     organization := "io.vinyldns",
   ).dependsOn(core % "compile->compile;test->test")
 
+lazy val sqs = (project in file("modules/sqs"))
+  .enablePlugins(DockerComposePlugin, AutomateHeaderPlugin)
+  .configs(IntegrationTest)
+  .settings(sharedSettings)
+  .settings(headerSettings(IntegrationTest))
+  .settings(inConfig(IntegrationTest)(scalafmtConfigSettings))
+  .settings(name := "sqs")
```
pauljamescleary: probably should just be down in the last `.settings` setting the `name`

modules/sqs/docker/custom.conf
```diff
@@ -0,0 +1,22 @@
+node-address {
+  protocol = http
+  host = "localhost"
+  host = ${?APP_HOST}
+  port = 19005
```
mitruly: Assigning `19005` to SQS docker host port. `queueUrl` is defined by: `<host>:<port>/queue/<queue_name from queues>`.

modules/core/src/main/scala/vinyldns/core/queue/MessageQueue.scala
```diff
@@ -64,5 +64,8 @@ trait MessageQueue {
 
   // sends a single message, exceptions will be raised via IO
   def send[A <: ZoneCommand](command: A): IO[Unit]
+
+  // shutdown queue client
+  def shutdown(): Unit
```
pauljamescleary: I thought we were not doing this?  I know there is a `shutdown` in the aws `SqsClient`, but based on its documentation it is not necessary

modules/sqs/docker/custom.conf
```diff
@@ -0,0 +1,22 @@
...
+  context-path = ""
+}
+
+rest-sqs {
+  enabled = true
+  bind-port = 9324
```
pauljamescleary: shouldn't this be the same as the `node-address` port?

modules/sqs/src/it/scala/vinyldns/sqs/queue/SqsMessageQueueSpec.scala
```diff
@@ -0,0 +1,113 @@
...
+    "change message visibility timeouts" in {
+      queue.send(makeTestAddChange(rsOk)).unsafeRunSync()
+
+      val result = queue.receive(MessageCount(2).right.value).unsafeRunSync()
+      result.length shouldBe 1
+
+      noException should be thrownBy
+        queue.changeMessageTimeout(CommandMessage(result(0).handle, makeTestAddChange(rsOk)),
+          FiniteDuration(5, SECONDS)).unsafeRunSync()
+    }
```
mitruly: Looking into adding test for `shutdown`, which is currently throwing `RejectedRequestException`.

modules/sqs/src/it/scala/vinyldns/sqs/queue/SqsMessageQueueSpec.scala
```diff
@@ -0,0 +1,113 @@
...
+
+  // Re-create queue before tests
+  override protected def beforeAll(): Unit = {
+    queue.client.createQueue("sqs")
+  }
+
+  override protected def afterEach(): Unit = {
+    // Remove items from queue after each test
+    val result = queue.receive(MessageCount(10).right.value).unsafeRunSync()
+    result.foreach(queue.remove)
```
pauljamescleary: probably should loop continue receiving until they are all removed, right?

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueue.scala
```diff
@@ -0,0 +1,158 @@
...
+
+  def remove(message: CommandMessage): IO[Unit] =
+    monitored("sqs.removeMessage")(
+      sqsAsync[DeleteMessageRequest, DeleteMessageResult](
+        new DeleteMessageRequest(queueUrl, message.handle.asInstanceOf[ReceiptHandle].value),
+        client.deleteMessageAsync)).map(_ => ())
+
+  // AWS SQS has no explicit requeue mechanism; use changeMessageTimeout instead
+  def requeue(message: CommandMessage): IO[Unit] =
+    changeMessageTimeout(message, new FiniteDuration(120, SECONDS))
```
mitruly: Could potentially pattern match on re-queueing. Currently setting to two minutes.

modules/sqs/src/it/scala/vinyldns/sqs/queue/SqsMessageQueueSpec.scala
```diff
@@ -0,0 +1,113 @@
...
+
+  val rsAddChange: RecordSetChange = makeTestAddChange(rsOk)
+
+  "SqsMessageQueueSpec" should {
+    "receive a single message from the queue" in {
+      queue.send(rsAddChange).unsafeRunSync()
+
+      val result = queue.receive(MessageCount(1).right.value).unsafeRunSync()
+      result.length shouldBe 1
+      result.head.command shouldBe a[RecordSetChange]
```
pauljamescleary: Don't use `.head`

Should just check that the change is in there.

`result.map(_.command) should contain theSameElementsAs List(rsAddChange)`

build.sbt
```diff
@@ -464,7 +483,8 @@ addCommandAlias("validate", "; root/clean; " +
 addCommandAlias("verify", "; project root; killDocker; " +
   "project api; dockerComposeUp; project dynamodb; dockerComposeUp; project mysql; dockerComposeUp; " +
```
rebstar6: presumably we need to dockerComposeUp sqs as well?

modules/sqs/src/it/scala/vinyldns/sqs/queue/SqsMessageQueueSpec.scala
```diff
@@ -0,0 +1,113 @@
...
+import org.scalatest._
+import vinyldns.core.TestRecordSetData._
+import vinyldns.core.TestZoneData._
+import vinyldns.core.domain.record.RecordSetChange
+import vinyldns.core.domain.zone.ZoneCommand
+import vinyldns.core.queue.{CommandMessage, MessageCount}
+
+import scala.concurrent.duration.FiniteDuration
+
+class SqsMessageQueueSpec extends WordSpec
```
rebstar6: if this is gonna live in IT, call it SqsMessageQueueIntegrationSpec

modules/sqs/src/it/scala/vinyldns/sqs/queue/SqsMessageQueueSpec.scala
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    "receive up to MessageCount messages from the queue" in {
+      for (_ <- 1 to 6) {
+        queue.send(rsAddChange).unsafeRunSync()
+        queue.send(zoneChangePending).unsafeRunSync()
+      }
+
+      val result = queue.receive(MessageCount(4).right.value).unsafeRunSync()
+      result.length shouldBe 4
```
pauljamescleary: Try to use matchers as opposed to individual values.  The result of this failing would be `3 is not equal to 4`.  Whereas, you will get better failure information if you used a list matcher from scalatest...

`List() should have length(4)`

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueue.scala
```diff
@@ -0,0 +1,158 @@
...
+
+  def send[A <: ZoneCommand](command: A): IO[Unit] =
+    monitored("sqs.sendMessage")(
+      sqsAsync[SendMessageRequest, SendMessageResult](
+        toSendMessageRequest(command)
+          .withQueueUrl(queueUrl),
+        client.sendMessageAsync)).map(_ => ())
+
+  def send[A <: ZoneCommand](messages: NonEmptyList[A]): IO[SendBatchResult] = {
+    val idLookup = messages.toList.map(UUID.randomUUID().toString -> _)
```
mitruly: Not sure if there is a more elegant way to do this. Essentially we need unique IDs for each send request (since a message can be re-queued), and also need to use it as a lookup as `SendMessageBatchResult` returns message info as an `md5`.

modules/sqs/src/it/scala/vinyldns/sqs/queue/SqsMessageQueueSpec.scala
```diff
@@ -0,0 +1,113 @@
...
+      result.foreach(_.command shouldBe a[ZoneCommand])
+    }
+
+    "return Unit when attempting to remove item from empty queue" in {
+      val result = queue.receive(MessageCount(1).right.value).unsafeRunSync()
+      result.length shouldBe 0
```
pauljamescleary: I don't think you need to do this receive first.

Possibly add a helper function in the test class that says "drain queue", or beforeEach just drain the queue so you can assert that it is empty in each test

modules/sqs/src/it/scala/vinyldns/sqs/queue/SqsMessageQueueSpec.scala
```diff
@@ -0,0 +1,113 @@
...
+
+      noException should be thrownBy
+        queue.remove(CommandMessage(ReceiptHandle("does-not-exist"), makeTestAddChange(rsOk))).unsafeRunSync()
```
pauljamescleary: I don't think an exception test is what we want here.  Just `remove` is unit.

Can acheive that through `queue.remove(xxx).attempt.unsafeRunSync should beRight(())` - need `EitherMatchers`

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueue.scala
```diff
@@ -0,0 +1,158 @@
...
+import scala.concurrent.duration.FiniteDuration
+
+// Unique identifier corresponding to action of receiving the message, not the message itself. Only this info
+// is required for actions like removal and changing message timeout in AWS SQS
+final case class ReceiptHandle(value: String) extends MessageHandle
+
+case class SqsMessageQueue(queueUrl: String, client: AmazonSQSAsync)
+    extends MessageQueue
+    with ProtobufConversions {
+
```
mitruly: SQS seems to have a supported max of 10 items in a batch. We could either handle the error ourselves or let the AWS library.

modules/core/src/main/scala/vinyldns/core/queue/MessageQueue.scala
rebstar6: did we ever settle on if we even need shutdown?

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsConverters.scala
```diff
@@ -0,0 +1,155 @@
...
+import vinyldns.proto.VinylDNSProto
+
+import scala.collection.JavaConverters._
+import scala.util.Try
+
+object SqsConverters extends ProtobufConversions {
+
+  private val logger = LoggerFactory.getLogger("vinyldns.sqs.queue.SqsConverters")
+
+  sealed trait SqsMessageType {
```
pauljamescleary: this can be simplified using `sealed abstract class SqsMessageType(val name: String)`

And then 

`case object SqsRecordSetChangeMessage extends SqsMessageType("SqsRecordSetChangeMessage")`

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsConverters.scala
```diff
@@ -0,0 +1,155 @@
...
+      messages: NonEmptyList[A],
+      idLookup: List[(String, A)]): SendMessageBatchRequest = {
+    val messageTypeBytesTuple = messages.map {
+      case recordSetChange: RecordSetChange =>
+        (SqsRecordSetChangeMessage.name, toPB(recordSetChange).toByteArray)
+      case zoneChange: ZoneChange => (SqsZoneChangeMessage.name, toPB(zoneChange).toByteArray)
+    }
+
+    val messageBatchRequestEntryList = messageTypeBytesTuple.zipWithIndex
+      .map { tuple =>
```
pauljamescleary: rather than have `tuple` as a parameter, deconstruct the tuple like `.map { case (foo, bar) => ` so this is easier to read.

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsConverters.scala
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  def fromMessage[A <: ZoneCommand](
+      sendMessageBatchResultEntry: SendMessageBatchResultEntry,
+      idLookup: Map[String, A]): ZoneCommand = {
+    logger.info(
+      s"Received message with attributes ${sendMessageBatchResultEntry.getMD5OfMessageAttributes}")
+
+    val zoneCommand = idLookup.getOrElse(sendMessageBatchResultEntry.getId, None)
+    zoneCommand match {
```
pauljamescleary: if you `getOrElse(xx, None)`, your pattern does not account for the `None` case

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsConverters.scala
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  def fromMessage[A <: ZoneCommand](
+      batchResultErrorEntry: BatchResultErrorEntry): (Exception, ZoneCommand) = {
+    logger.info(
+      s"Received message with attributes ${batchResultErrorEntry.getMessage}, " +
+        s"${batchResultErrorEntry.getCode}")
+
+    val messageBytes = Base64.getDecoder.decode(batchResultErrorEntry.getMessage)
+    throw new Exception(Base64.getDecoder.decode(messageBytes).toString)
```
pauljamescleary: this cannot be right.  We want to return a tuple (Exception, ZoneCommand), not throw

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueue.scala
```diff
@@ -0,0 +1,158 @@
...
+import vinyldns.core.queue._
+import vinyldns.core.route.Monitor
+import vinyldns.sqs.queue.SqsConverters._
+
+import scala.collection.JavaConverters._
+import scala.concurrent.duration.FiniteDuration
+
+// Unique identifier corresponding to action of receiving the message, not the message itself. Only this info
+// is required for actions like removal and changing message timeout in AWS SQS
+final case class ReceiptHandle(value: String) extends MessageHandle
```
pauljamescleary: I would argue `SqsMessageHandle` is a better name; this way when it is logged out we know what it is

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueue.scala
pauljamescleary: we should loop through everything passed in on the batch.  Each message queue maybe different on the max batch size.

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueue.scala
pauljamescleary: 2 minutes is too long.  We want this to retry.

An interesting trick is if we can get the number of retries off of the message.  then we could exponential backoff (adjust the message timeout to some ceiling).  For example, if 0 then set the timeout to 2 seconds, then 4, then 8, then 16, then 32, then 64, keep 64 as a ceiling until retries are exhausted

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueue.scala
pauljamescleary: fun facts.  Each change has it's own id.  Why not just use the change id on the change as the unique id on the message?

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueue.scala
```diff
@@ -0,0 +1,158 @@
...
+    monitored("sqs.changeMessageTimeout")(
+      sqsAsync[ChangeMessageVisibilityRequest, ChangeMessageVisibilityResult](
+        new ChangeMessageVisibilityRequest()
+          .withReceiptHandle(message.handle.asInstanceOf[ReceiptHandle].value)
+          .withVisibilityTimeout(duration.toSeconds.toInt) // 1800 seconds == 30 minutes
+          .withQueueUrl(queueUrl),
+        client.changeMessageVisibilityAsync
+      )).map(_ => ())
+
+  def shutdown(): Unit = client.shutdown()
```
pauljamescleary: we can remove this

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueue.scala
mitruly: Hmm, that should work. I'll try that out.

modules/sqs/docker/custom.conf
mitruly: I get the following error when updating the `node-address`:
```
com.amazonaws.SdkClientException: Unable to execute HTTP request: The target server failed to respond
```

pauljamescleary: guess if it works :)

modules/sqs/src/it/scala/vinyldns/sqs/queue/SqsMessageQueueIntegrationSpec.scala
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  val rsAddChange: RecordSetChange = makeTestAddChange(rsOk)
+
+  "SqsMessageQueueSpec" should {
+    "receive a single message from the queue" in {
+      queue.send(rsAddChange).unsafeRunSync()
+
+      val result = queue.receive(MessageCount(1).right.value).unsafeRunSync()
+      result.length shouldBe 1
```
pauljamescleary: no need for the length check if you are doing `contain theSameElementsAs`

modules/sqs/src/it/scala/vinyldns/sqs/queue/SqsMessageQueueIntegrationSpec.scala
```diff
@@ -0,0 +1,125 @@
...
+
+    "receive up to MessageCount messages from the queue" in {
+      for (_ <- 1 to 6) {
+        queue.send(rsAddChange).unsafeRunSync()
+        queue.send(zoneChangePending).unsafeRunSync()
+      }
+
+      val result = queue.receive(MessageCount(4).right.value).unsafeRunSync()
+      result should have length 4
+      result.foreach(_.command shouldBe a[ZoneCommand])
```
pauljamescleary: check out http://www.scalatest.org/user_guide/using_matchers for lots of great matchers

Also, I think we can do `contains theSameElementsAs` here as well.  We know that we should have 2 addChange and 2 zoneChangePending

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsConverters.scala
```diff
@@ -0,0 +1,157 @@
...
+        s"${batchResultErrorEntry.getCode}")
+
+    val messageBytes = Base64.getDecoder.decode(batchResultErrorEntry.getMessage)
+    val zoneCommand = idLookup.getOrElse(batchResultErrorEntry.getId, None)
+    zoneCommand match {
+      case recordSetChange: RecordSetChange =>
+        (new Exception(Base64.getDecoder.decode(messageBytes).toString), recordSetChange)
+      case zoneChange: ZoneChange =>
+        (new Exception(Base64.getDecoder.decode(messageBytes).toString), zoneChange)
+      case None => throw new Exception("Could not find send batch error zone command")
```
pauljamescleary: never ever `throw`, possibility of failure needs to be explicit in the return type of the function if there is no acceptable handling of the situation.

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueue.scala
pauljamescleary: If you have to, can add an `def id: String` to the `ZoneCommand` trait

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsConverters.scala
mitruly: Yeah, this is a weird situation since the `Map` is used to simply tie in the `id` value, which we know exists. `get` should always succeed here, but wasn't sure how to handle this properly. I don't think modifying the the return type signature would be the route to go.

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessage.scala
```diff
@@ -0,0 +1,60 @@
...
+  sealed abstract class SqsMessageError(message: String) extends Exception(message)
+  final case class InvalidSqsMessageContents(messageId: String)
+      extends SqsMessageError(s"Unable to parse SQS Message with ID '$messageId'")
+  final case class EmptySqsMessageContents(messageId: String)
+      extends SqsMessageError(s"No message body found for SQS message with ID '$messageId'")
+
+  final case class InvalidMessageType(className: String)
+      extends Exception(s"Invalid command message type '$className'")
+
+  def cast(msg: CommandMessage): Either[InvalidMessageType, SqsMessage] = msg match {
```
pauljamescleary: ah, doh.  Take a look at my PR.  Updated `CommandMessage` to have a `def id: MessageId` field.

That was after comments from @rebstar6 

Make the same adjustments, and then you can delete the `cast` here, and calls in the message queue itself

modules/sqs/src/it/scala/vinyldns/sqs/queue/SqsMessageQueueIntegrationSpec.scala
```diff
@@ -0,0 +1,129 @@
...
+
+  // Delete queue after tests since stuff can potentially linger
+  override protected def afterAll(): Unit = {
+    queue.client.deleteQueue(queue.queueUrl)
+  }
+
+  val rsAddChange: RecordSetChange = makeTestAddChange(rsOk)
+
+  "SqsMessageQueueIntegrationSpec" should {
+    "receive a single message from the queue" in {
```
pauljamescleary: should add a test that a malformed message gets dropped / removed from the queue.

You would have to manually do an sqs.send as opposed to using the MessageQueue abstraction to insert some garbage data.

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageType.scala
```diff
@@ -0,0 +1,62 @@
...
+  val messageAttribute: (String, MessageAttributeValue) =
+    "message-type" -> new MessageAttributeValue()
+      .withStringValue(name)
+      .withDataType("String")
+}
+
+object SqsMessageType {
+  case object SqsRecordSetChangeMessage extends SqsMessageType("SqsRecordSetChangeMessage")
+  case object SqsZoneChangeMessage extends SqsMessageType("SqsZoneChangeMessage")
+  sealed abstract class SqsMessageTypeError(message: String) extends Exception(message)
```
pauljamescleary: bah, should probably make it extend `Throwable` instead for consistency in our codebase

modules/core/src/main/scala/vinyldns/core/domain/ZoneCommand.scala
```diff
@@ -14,28 +14,36 @@
...
 import org.joda.time.DateTime
-import vinyldns.core.domain.zone.{Zone, ZoneCommand, ZoneCommandResult}
+import vinyldns.core.domain.RecordSetChangeStatus.RecordSetChangeStatus
+import vinyldns.core.domain.RecordSetChangeType.RecordSetChangeType
+import vinyldns.core.domain.ZoneChangeStatus.ZoneChangeStatus
+import vinyldns.core.domain.ZoneChangeType.ZoneChangeType
+import vinyldns.core.domain.record.{RecordSet, RecordSetStatus}
+import vinyldns.core.domain.zone.Zone
+
+sealed trait ZoneCommand {
```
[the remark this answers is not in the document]
mitruly: Would like to get some discussion with @pauljamescleary on this item. I'm a bit indifferent to this and can easily revert this.

modules/sqs/src/it/scala/vinyldns/sqs/queue/SqsMessageQueueIntegrationSpec.scala
```diff
@@ -0,0 +1,128 @@
...
+
+    "return Unit when attempting to remove item from empty queue" in {
+      val result = queue.receive(MessageCount(1).right.value).unsafeRunSync()
+      result.length shouldBe 0
+
+      noException should be thrownBy
+        queue.remove(CommandMessage(ReceiptHandle("does-not-exist"), makeTestAddChange(rsOk))).unsafeRunSync()
+    }
+
+    "return Unit when attempting to remove item from queue" in {
```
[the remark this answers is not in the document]
mitruly: I couldn't find this so I think this might be a stale commit 😬

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueue.scala
```diff
@@ -0,0 +1,232 @@
...
+import vinyldns.core.domain.zone.{ZoneChange, ZoneCommand}
+import vinyldns.core.protobuf.ProtobufConversions
+import vinyldns.core.queue._
+import vinyldns.core.route.Monitored
+import vinyldns.sqs.queue.SqsMessageType.{SqsRecordSetChangeMessage, SqsZoneChangeMessage}
+
+import scala.collection.JavaConverters._
+import scala.concurrent.duration.FiniteDuration
+
+class SqsMessageQueue(val queueUrl: String, val client: AmazonSQSAsync)
```
rebstar6: should we be passing in the queueUrl or name? kinda feels like name, as theres a `getQueueUrl(name: String)` function on the AmazonSQSAsync type

rebstar6: it just smells a little weird - what if someone puts in something totally different for queueUrl than where the client is

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueue.scala
```diff
@@ -0,0 +1,232 @@
...
+  def receive(count: MessageCount): IO[List[SqsMessage]] =
+    monitor("queue.SQS.receiveMessageBatch") {
+      sqsAsync[ReceiveMessageRequest, ReceiveMessageResult](
+        new ReceiveMessageRequest()
+          .withMaxNumberOfMessages(count.value)
+          .withMessageAttributeNames(".*")
+          .withWaitTimeSeconds(1)
+          .withQueueUrl(queueUrl),
+        client.receiveMessageAsync
+      ).flatMap(batchResult => parseBatch(batchResult.getMessages.asScala.toList))
```
rebstar6: can we rename this not to use the word 'batch' - it took me more than a second to realize we werent talking about batch as in batchChanges

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueue.scala
```diff
@@ -0,0 +1,232 @@
...
+
+  def remove(message: CommandMessage): IO[Unit] =
+    monitor("queue.SQS.removeMessage") {
+      IO(delete(message.id.value))
+    }.as(())
+
+  // AWS SQS has no explicit requeue mechanism; It can be configured to
+  // natively apply an exponential back-off retry mechanism
+  // (see: https://docs.aws.amazon.com/general/latest/gr/api-retries.html).
+  def requeue(message: CommandMessage): IO[Unit] = IO.unit
```
rebstar6: we had discussed changeMessageTimeout back low (like, not 20 mins for a sync or create) at some point - still thinking about building that in?

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueue.scala
```diff
@@ -0,0 +1,232 @@
...
+
+  def send[A <: ZoneCommand](command: A): IO[Unit] =
+    monitor("queue.SQS.sendMessage")(
+      sqsAsync[SendMessageRequest, SendMessageResult](
+        toSendMessageRequest(command)
+          .withQueueUrl(queueUrl),
+        client.sendMessageAsync)).as(())
+
+  def sendBatch[A <: ZoneCommand](cmds: NonEmptyList[A]): IO[SendBatchResult] =
```
rebstar6: hmm i guess we also use batch here. this one never bothered me because i know this is only ever called from the batch interface...

i can live without this change, but still wouldnt mind it i think https://github.com/vinyldns/vinyldns/pull/255/files#r223863488

modules/sqs/src/it/scala/vinyldns/sqs/queue/SqsMessageQueueIntegrationSpec.scala
```diff
@@ -0,0 +1,152 @@
...
+
+      val messages = NonEmptyList.fromListUnsafe(commands)
+
+      val result = queue.sendBatch(messages).unsafeRunSync()
+
+      result.successes should contain theSameElementsAs commands
+      result.failures shouldBe empty
+    }
+
+    "change message visibility timeouts" in {
```
rebstar6: can test that it gets requeued - drop to 1 second and sleep, then take again. the item should have become available

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueue.scala
pauljamescleary: @rebstar6 in theory, when the `SqsMessageQueueProvider` loads up, it should detect the presence of the queues.  You could argue that it should create them if they are not present, but we do not do that today.

`sqs.listQueues()` and `sqs.getQueueUrl(name)` both return essentially the same thing, the `queueUrl` as a `String`, ugh.

That string is required to complete all the sqs requests, as they all need `withQueueUrl`.

Could make it strongly typed as opposed to a string as an optimization.

I think the validity of the queueUrl should be left to the provider (or perhaps we hide the constructor and add it on the companion object).

rebstar6: yeah, i guess this is more a point on how we config this -passing in the queue url from config seems wrong. the queue name is really what should be configured, and then in apply, we should be doing some kind of `createIfNotExists` on the client, and pulling the queue url from there. 

right now it looks like we also assume the queue exists already at that url

rebstar6: were saying the same thing on create. we can wait on that but i think the loader should do it

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueue.scala
pauljamescleary: Definitely changing it to something shorter would be ideal, even if we do not backoff (something short say 10 seconds).

The retry you mention here @mitruly has to do with the aws sdk retries, not message retries.  

The recommendation is especially pertinent for zone syncs, where we change the visibility timeout to 20 minutes, oi!  If we  do nothing, those will continue to cycle for 2000 minutes before exhausting SQS retries, woa!

What we want to do is reduce that back down to a few seconds so the retry can happen almost immediately.  By doing a change message timeout to 10 seconds, we change that retry timeframe from 2000 minutes down to 1000 seconds

modules/sqs/src/it/scala/vinyldns/sqs/queue/SqsMessageQueueIntegrationSpec.scala
rebstar6: oh yeah, 0, even better :)

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueue.scala
pauljamescleary: `receiveResult` should be fine

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueue.scala
```diff
@@ -0,0 +1,232 @@
...
+    monitor("queue.SQS.sendMessageBatch")(
+      sqsAsync[SendMessageBatchRequest, SendMessageBatchResult](
+        toSendMessageBatchRequest(cmds)
+          .withQueueUrl(queueUrl),
```
pauljamescleary: this will be a problem I think if we have a limit on the number of messages we can send.  From the documentation...

```
The maximum allowed individual message size and the maximum total payload size (the sum of the individual lengths of all of the batched messages) are both 256 KB (262,144 bytes).
```

So, we need to convert each message to bytes, calculate the total size in number of bytes, and then group on that sending multiple batches.

I don't know if the message attribute contributes to the message size, assume "no" and then see what happens in the test.

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueue.scala
```diff
@@ -0,0 +1,232 @@
...
+        client.sendMessageBatchAsync))
+      .map { batchResult =>
+        toSendBatchResult(batchResult, cmds)
+      }
+
+  def changeMessageTimeout(message: CommandMessage, duration: FiniteDuration): IO[Unit] =
+    monitor("queue.SQS.changeMessageTimeout") {
+      IO {
+        sqsAsync[ChangeMessageVisibilityRequest, ChangeMessageVisibilityResult](
```
pauljamescleary: need to do a validation on the duration.  Doesn't seem like it will explicitly error out, or what would happen if the message timeout is exceeded.

```
Values values: 0 to 43200. Maximum: 12 hours
```

I would just add a small function on the companion object `def validateMessageTimeout(duration: FiniteDuration): Either[InvalidMessageTimeout, FiniteDuration] = ???`

Then, in here just `IO.fromEither(validateMessageTimeout(duration).map { validDuration => sqsAsync... }`

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueue.scala
mitruly: Updated the send batch logic and updated the code to do a grouping based off of the max message body size.

modules/sqs/src/it/scala/vinyldns/sqs/queue/SqsMessageQueueIntegrationSpec.scala
```diff
@@ -0,0 +1,193 @@
...
+      val messages = NonEmptyList.fromListUnsafe(commands)
+
+      val result = queue.sendBatch(messages).unsafeRunSync()
+
+      result.successes should contain theSameElementsAs commands
+      result.failures shouldBe empty
+    }
+
+    "send a message batch request with more than 256KB successfully" in {
+      val recordSet = aaaa.copy(name = "a"*100000, zoneId = "b"*10000)
```
mitruly: This test had previously failed before the chunking approach due to a broken pipe error.

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueue.scala
```diff
@@ -0,0 +1,265 @@
...
+      .map { lst: List[Either[Throwable, SqsMessage]] =>
+        lst.collect {
+          case Right(message) => message
+        }
+      }
+
+  // If we cannot parse the message, remove it from the queue
+  def parse(message: Message): IO[Either[Throwable, SqsMessage]] =
+    // This is tricky, we need to attempt to parse the message.  If we cannot, delete it; otherwise return ok
+    IO.fromEither(SqsMessage.parseSqsMessage(message)).attempt.flatMap {
```
pauljamescleary: we probably don't need the `fromEither ... attempt` here.

Can we just do `IO(SqsMessage.parse...).flatMap` instead?

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueue.scala
```diff
@@ -0,0 +1,265 @@
...
+  def sendBatch[A <: ZoneCommand](cmds: NonEmptyList[A]): IO[SendBatchResult] =
+    monitor("queue.SQS.sendMessageBatch") {
+      toSendMessageBatchRequest(cmds)
+        .map { sendRequest =>
+          sqsAsync[SendMessageBatchRequest, SendMessageBatchResult](
+            sendRequest
+              .withQueueUrl(queueUrl),
+            client.sendMessageBatchAsync)
+        }
+        .parSequence
```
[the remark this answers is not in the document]
mitruly: I believe that they are. We have two integration tests that check for matching starting/ending `ZoneCommand`s which pass. I also tested with print statements and `toSendBatchResult` accumulates all of the `SendBatchRequestResult`s.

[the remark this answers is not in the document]
pauljamescleary: Ok, I see it

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueue.scala
```diff
@@ -0,0 +1,267 @@
...
+
+    messageTypeBytesTuple match {
+      case (messageType, messageBytes) =>
+        new SendMessageRequest()
+          .withMessageBody(Base64.getEncoder.encodeToString(messageBytes))
+          .withMessageAttributes(Map(messageType.messageAttribute).asJava)
+    }
+  }
+
+  def toSendMessageBatchRequest[A <: ZoneCommand](
```
pauljamescleary: Can you add a few unit tests for this.  I did the math on some sketch paper and it appears to work out.

At least one unit test that builds x messages, with x bytes to predetermine the number of "batch requests", and make sure the result matches what is expected and that the sum total size of each individual "batch request" is less than the MAX_BATCH_SIZE?

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueue.scala
```diff
@@ -0,0 +1,267 @@
...
+            )))
+        .withEndpointConfiguration(new EndpointConfiguration(serviceEndpoint, signingRegion))
+        .withCredentials(new AWSStaticCredentialsProvider(
+          new BasicAWSCredentials(accessKey, secretKey)))
+        .build()
+
+    new SqsMessageQueue(queueUrl, client)
+  }
+
+  def validateMessageTimeout(
```
pauljamescleary: add a few unit tests for this as well

1. < 0
2. > max
3. correct value

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueue.scala
```diff
@@ -0,0 +1,267 @@
...
+          .withQueueUrl(queueUrl),
+        client.receiveMessageAsync
+      ).flatMap(receiveResult => parseBatch(receiveResult.getMessages.asScala.toList))
+    }
+
+  def parseBatch(messages: List[Message]): IO[List[SqsMessage]] =
+    // attempt to parse each message that arrives, failures will be removed and not returned
+    messages
+      .map(parse)
+      .sequence
```
rebstar6: sequence or parSequence here?

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueue.scala
```diff
@@ -0,0 +1,267 @@
...
+            .withMessageBody(msgBody)
+            .withId(cmd.id)
+            .withMessageAttributes(Map(SqsMessageType.fromCommand(cmd).messageAttribute).asJava)
+      }
+      .toList
+
+    // Group entries into batches
+    val maxMessageSize = entries.map(_.getMessageBody.getBytes().length).max
+    entries
+      .grouped(math.floor(MAXIMUM_BATCH_SIZE.toDouble / maxMessageSize.toDouble).toInt)
```
rebstar6: dont think you need to floor, pretty sure toInt does that already

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueue.scala
```diff
@@ -0,0 +1,267 @@
...
+        new SendMessageBatchRequest().withEntries(groupedEntries.asJava)
+      }
+      .toList
+  }
+
+  def toSendBatchResult[A <: ZoneCommand](
+      sendResultList: List[SendMessageBatchResult],
+      cmds: NonEmptyList[A]): SendBatchResult = {
+    val successfulIds = sendResultList.flatMap(_.getSuccessful.asScala.map(_.getId))
+    val successes = cmds.toList.filter(cmd => successfulIds.contains(cmd.id))
```
rebstar6: when you collect or filter on NEL, it should turn to List (so no need to toList here).

also, the zipping in the failures makes me nervous. maybe:

```
    val successfulIds = sendResultList.flatMap(_.getSuccessful.asScala.map(_.getId))
    val successes = cmds.filter(cmd => successfulIds.contains(cmd.id))

    val failed = sendResultList.flatMap(_.getFailed.asScala.map { f =>
      (f.getId, f.getMessage)
    }).toMap
    
    val failures = cmds
       .map(cmd => (failed.get(cmd.id), cmd))
       .collect {
         case (Some(msg), cmd) => (new Exception(msg), cmd)
       }

    SendBatchResult(successes, failures)

```

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageType.scala
```diff
@@ -0,0 +1,62 @@
...
+      case SqsZoneChangeMessage.name => Right(SqsZoneChangeMessage)
+      case invalid => Left(InvalidMessageTypeValue(invalid))
+    }
+
+  def fromMessage(sqsMessage: Message): Either[SqsMessageTypeError, SqsMessageType] =
+    // getMessageAttributes guarantees a map, but it could be empty
+    // the message-type maybe present, but doesn't have a string value
+    // the message-type could have a string value, but not a valid value
+    for {
+      messageTypeAttr <- Option(sqsMessage.getMessageAttributes.get("message-type"))
```
rebstar6: i think the `.get("message-type")` can throw, which i dont believe the Option will stop

```
    for {
      messageType <- Either.catchNonFatal(sqsMessage.getMessageAttributes.get("message-type"))
      messageTypeAttr <- Either.fromOption (
        Option(messageType.getStringValue),
        MessageTypeNotFound
      )
      messageType <- fromString(messageTypeAttr)
    } yield messageType
```

rebstar6: import cats.implicits._ for ^

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessage.scala
```diff
@@ -0,0 +1,55 @@
...
+  final case class EmptySqsMessageContents(messageId: String)
+      extends SqsMessageError(s"No message body found for SQS message with ID '$messageId'")
+
+  final case class InvalidMessageType(className: String)
+      extends Throwable(s"Invalid command message type '$className'")
+
+  def parseSqsMessage(message: Message): Either[Throwable, SqsMessage] =
+    for {
+      messageType <- SqsMessageType.fromMessage(message)
+      messageBytes <- Option(message.getBody)
```
rebstar6: there's `Either.fromOption` for this, no need to map

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageType.scala
pauljamescleary: @rebstar6 can it though?   Curious in what way.

Looking at the code `getMessageAttributes` returns an empty map if not set.

In java, `.get` will return null if the key does not exist.  From the javadoc...

```
     * Returns the value to which the specified key is mapped,
     * or {@code null} if this map contains no mapping for the key.
```

pauljamescleary: Guess _technically_ because we are on the Map interface...

```
     * @throws ClassCastException if the key is of an inappropriate type for
     *         this map
     * (<a href="{@docRoot}/java/util/Collection.html#optional-restrictions">optional</a>)
     * @throws NullPointerException if the specified key is null and this map
     *         does not permit null keys
```

rebstar6: yeah ^^ is what i saw

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageType.scala
```diff
@@ -48,15 +51,16 @@ object SqsMessageType {
...
     // the message-type could have a string value, but not a valid value
-    for {
-      messageTypeAttr <- Option(sqsMessage.getMessageAttributes.get("message-type"))
-        .flatMap(attr => Option(attr.getStringValue))
-        .map(Right(_))
-        .getOrElse(Left(MessageTypeNotFound))
-      messageType <- fromString(messageTypeAttr)
-    } yield messageType
+    val messageType = for {
+      messageTypeAttr <- Either.fromTry(
```
mitruly: Tried doing:
```
for {
      messageTypeAttr <- Either.fromOption(
        Option(sqsMessage.getMessageAttributes.get("message-type").getStringValue),
        MessageTypeNotFound
      )
      messageType <- fromString(messageTypeAttr)
    } yield messageType
```
but this still resulted in an uncaught `NullPointerException` since the `sqsMessage.getMessageAttributes.get("message-type")` is hitting an error before casting to an `Option`.

pauljamescleary: right, get("message-type") could be null

I don't think we want this...

```scala
      messageTypeAttr <- Either.fromTry(
        Try(sqsMessage.getMessageAttributes.get("message-type").getStringValue))
```

As we kinda just blow up with a null pointer exception.  Would be nice to know the message type attribute was not there.

This should do the trick, gotta wrap goth the `get` and the `getStringValue` in an Option using a flatMap.

```scala
Either.fromOption (
  Option(sqsMessage.getMessageAttributes.get("message-type").flatMap(mt => Option(mt.getStringValue)
```

PR #256, opened 2018-10-03
Disable coverage for VinylDNSService

### LINKED ISSUE #250: Disable code coverage for `VinylDNSService`
**Motivation**
The file tracks negatively from a code coverage perspective.  However, much of the file is bootstrapping the application, which is very difficult to test.

**Scope of change**
Really small change, just adding two comments to disable code coverage.  When the change is complete, codecov should comment that we impacted code coverage positively.  Codecov stats for that file should look better.

* Add `// $COVERAGE-OFF$` right before the `class VinylDNSService`
* Add `// $COVERAGE-ON$` right after that same class in the same file

davidmargolin:
resolves #250


pauljamescleary:
I mean, +.35% code coverage for skipping coverage on untestable code.  This is great stuff @davidmargolin

PR #259, opened 2018-10-04
MySQL Message Queue

pauljamescleary:
Create a MySQL Message Queue implementation.

* Created a `MySqlMessageQueue` in the mysql sub module
* Created a `MySqlMessage` that implements `CommandMessage` from core
* Created a `MessageType` enum to determine which type of command is on the message
* Created a `MySqlMessageQueueIntegrationSpec` which exercises 100% code coverage on the queue, including hidden behavior in the SQL

### DIFF
--- a/modules/api/src/test/scala/vinyldns/api/backend/CommandHandlerSpec.scala
+++ b/modules/api/src/test/scala/vinyldns/api/backend/CommandHandlerSpec.scala
@@ -32,7 +32,7 @@ import vinyldns.api.domain.dns.DnsConnection
 import vinyldns.core.domain.batch.BatchChangeRepository
 import vinyldns.core.domain.record.{RecordChangeRepository, RecordSetChange, RecordSetRepository}
 import vinyldns.core.domain.zone._
-import vinyldns.core.queue.{CommandMessage, MessageCount, MessageHandle, MessageQueue}
+import vinyldns.core.queue.{CommandMessage, MessageCount, MessageQueue}
 
 import scala.concurrent.duration._
 import scala.concurrent.ExecutionContext.Implicits.global
@@ -46,12 +46,12 @@ class CommandHandlerSpec
     with EitherValues
     with EitherMatchers {
 
-  private case class TestHandle(value: String) extends MessageHandle
+  private case class TestCommandMessage(command: ZoneCommand, value: String) extends CommandMessage
   private val mq = mock[MessageQueue]
   implicit val sched: Scheduler =
     Scheduler.fromScheduledExecutorService(Executors.newScheduledThreadPool(2))
   private val messages = for { i <- 0 to 10 } yield
-    CommandMessage(TestHandle(i.toString), pendingCreateAAAA)
+    TestCommandMessage(pendingCreateAAAA, i.toString)
   private val count = MessageCount(10).right.value
 
   private val mockZoneChangeProcessor = mock[ZoneChange => IO[ZoneChange]]
@@ -96,7 +96,7 @@ class CommandHandlerSpec
     "update the timeout for zone syncs" in {
       doReturn(IO.unit).when(mq).changeMessageTimeout(any[CommandMessage], any[FiniteDuration])
       val sync = zoneCreate.copy(changeType = ZoneChangeType.Sync)
-      val msg = CommandMessage(TestHandle("foo"), sync)
+      val msg = TestCommandMessage(sync, "foo")
       Stream
         .emit(msg)
         .covary[IO]
@@ -109,7 +109,7 @@ class CommandHandlerSpec
     }
     "update the timeout for zone creates" in {
       doReturn(IO.unit).when(mq).changeMessageTimeout(any[CommandMessage], any[FiniteDuration])
-      val msg = CommandMessage(TestHandle("foo"), zoneCreate)
+      val msg = TestCommandMessage(zoneCreate, "foo")
       Stream
         .emit(msg)
         .covary[IO]
@@ -122,7 +122,7 @@ class CommandHandlerSpec
     }
     "not update the timeout for zone deletes" in {
       val del = zoneCreate.copy(changeType = ZoneChangeType.Delete)
-      val msg = CommandMessage(TestHandle("foo"), del)
+      val msg = TestCommandMessage(del, "foo")
       Stream
         .emit(msg)
         .covary[IO]
@@ -134,7 +134,7 @@ class CommandHandlerSpec
       verifyZeroInteractions(mq)
     }
     "not update the timeout for record changes" in {
-      val msg = CommandMessage(TestHandle("foo"), pendingCreateAAAA)
+      val msg = TestCommandMessage(pendingCreateAAAA, "foo")
       Stream
         .emit(msg)
         .covary[IO]
@@ -149,13 +149,13 @@ class CommandHandlerSpec
 
   "determining outcome" should {
     "generate a retry for failures" in {
-      val msg = CommandMessage(TestHandle("foo"), pendingCreateAAAA)
+      val msg = TestCommandMessage(pendingCreateAAAA, "foo")
       val result =
         CommandHandler.outcomeOf(msg)(IO.raiseError(new RuntimeException("fail"))).unsafeRunSync()
       result shouldBe RetryMessage(msg)
     }
     "generate delete for successes" in {
-      val msg = CommandMessage(TestHandle("foo"), pendingCreateAAAA)
+      val msg = TestCommandMessage(pendingCreateAAAA, "foo")
       val result = CommandHandler.outcomeOf(msg)(IO.unit).unsafeRunSync()
       result shouldBe DeleteMessage(msg)
     }
@@ -163,13 +163,13 @@ class CommandHandlerSpec
 
   "message sink" should {
     "retry messages" in {
-      val msg = RetryMessage(CommandMessage(TestHandle("foo"), pendingCreateAAAA))
+      val msg = RetryMessage(TestCommandMessage(pendingCreateAAAA, "foo"))
       doReturn(IO.unit).when(mq).requeue(msg.message)
       Stream.emit(msg).covary[IO].to(CommandHandler.messageSink(mq)).compile.drain.unsafeRunSync()
       verify(mq).requeue(msg.message)
     }
     "remove messages" in {
-      val msg = DeleteMessage(CommandMessage(TestHandle("foo"), pendingCreateAAAA))
+      val msg = DeleteMessage(TestCommandMessage(pendingCreateAAAA, "foo"))
       doReturn(IO.unit).when(mq).remove(msg.message)
       Stream.emit(msg).covary[IO].to(CommandHandler.messageSink(mq)).compile.drain.unsafeRunSync()
       verify(mq).remove(msg.message)
@@ -178,7 +178,7 @@ class CommandHandlerSpec
 
   "processing change requests" should {
     "handle record changes" in {
-      val change = CommandMessage(TestHandle("foo"), pendingCreateAAAA)
+      val change = TestCommandMessage(pendingCreateAAAA, "foo")
       doReturn(IO.pure(change))
         .when(mockRecordChangeProcessor)
         .apply(any[DnsConnection], any[RecordSetChange])
@@ -199,7 +199,7 @@ class CommandHandlerSpec
           mockZoneSyncProcessor,
           default
         )
-      val change = CommandMessage(TestHandle("foo"), noConnChange)
+      val change = TestCommandMessage(noConnChange, "foo")
       doReturn(IO.pure(change))
         .when(mockRecordChangeProcessor)
         .apply(any[DnsConnection], any[RecordSetChange])
@@ -211,7 +211,7 @@ class CommandHandlerSpec
       resolver.getAddress.getHostName shouldBe default.primaryServer
     }
     "handle zone creates" in {
-      val change = CommandMessage(TestHandle("foo"), zoneCreate)
+      val change = TestCommandMessage(zoneCreate, "foo")
       doReturn(IO.pure(zoneCreate))
         .doReturn(IO.pure(change))
         .when(mockZoneChangeProcessor)
@@ -224,7 +224,7 @@ class CommandHandlerSpec
     }
     "handle zone syncs" in {
       val sync = zoneCreate.copy(changeType = ZoneChangeType.Sync)
-      val change = CommandMessage(TestHandle("foo"), sync)
+      val change = TestCommandMessage(sync, "foo")
       doReturn(IO.pure(sync)).doReturn(IO.pure(change)).when(mockZoneChangeProcessor).apply(sync)
       doReturn(IO.pure(sync)).when(mockZoneSyncProcessor).apply(sync)
       Stream.emit(change).covary[IO].through(processor).compile.drain.unsafeRunSync()
@@ -234,7 +234,7 @@ class CommandHandlerSpec
     }
     "handle zone deletes" in {
       val del = zoneCreate.copy(changeType = ZoneChangeType.Delete)
-      val change = CommandMessage(TestHandle("foo"), del)
+      val change = TestCommandMessage(del, "foo")
       doReturn(IO.pure(del)).doReturn(IO.pure(change)).when(mockZoneChangeProcessor).apply(del)
       Stream.emit(change).covary[IO].through(processor).compile.drain.unsafeRunSync()
       verify(mockZoneChangeProcessor).apply(del)
@@ -246,7 +246,7 @@ class CommandHandlerSpec
   "main flow" should {
     "process successfully" in {
       val stop = fs2.async.signalOf[IO, Boolean](false).unsafeRunSync()
-      val cmd = CommandMessage(TestHandle("foo"), pendingCreateAAAA)
+      val cmd = TestCommandMessage(pendingCreateAAAA, "foo")
 
       // stage pulling from the message queue
       doReturn(IO.pure(List(cmd))).when(mq).receive(count)
@@ -284,7 +284,7 @@ class CommandHandlerSpec
     }
     "continue processing on unexpected failure" in {
       val stop = fs2.async.signalOf[IO, Boolean](false).unsafeRunSync()
-      val cmd = CommandMessage(TestHandle("foo"), pendingCreateAAAA)
+      val cmd = TestCommandMessage(pendingCreateAAAA, "foo")
 
       // stage pulling from the message queue, make sure we always return our command
       doReturn(IO.pure(List(cmd)))
@@ -332,7 +332,7 @@ class CommandHandlerSpec
     "process a zone update change through the flow" in {
       // testing the run method, which does nothing more than simplify construction of the main flow
       val stop = fs2.async.signalOf[IO, Boolean](false).unsafeRunSync()
-      val cmd = CommandMessage(TestHandle("foo"), zoneUpdate)
+      val cmd = TestCommandMessage(zoneUpdate, "foo")
 
       val zoneRepo = mock[ZoneRepository]
       val zoneChangeRepo = mock[ZoneChangeRepository]
--- a/modules/api/src/test/scala/vinyldns/api/backend/CommandHandlerSpec.scala
+++ b/modules/api/src/test/scala/vinyldns/api/backend/CommandHandlerSpec.scala
@@ -32,7 +32,7 @@ import vinyldns.api.domain.dns.DnsConnection
 import vinyldns.core.domain.batch.BatchChangeRepository
 import vinyldns.core.domain.record.{RecordChangeRepository, RecordSetChange, RecordSetRepository}
 import vinyldns.core.domain.zone._
-import vinyldns.core.queue.{CommandMessage, MessageCount, MessageQueue}
+import vinyldns.core.queue.{CommandMessage, MessageCount, MessageId, MessageQueue}
 
 import scala.concurrent.duration._
 import scala.concurrent.ExecutionContext.Implicits.global
@@ -46,7 +46,11 @@ class CommandHandlerSpec
     with EitherValues
     with EitherMatchers {
 
-  private case class TestCommandMessage(command: ZoneCommand, value: String) extends CommandMessage
+  private case class TestCommandMessage(command: ZoneCommand, value: String)
+      extends CommandMessage {
+    def id: MessageId = MessageId(value)
+  }
+
   private val mq = mock[MessageQueue]
   implicit val sched: Scheduler =
     Scheduler.fromScheduledExecutorService(Executors.newScheduledThreadPool(2))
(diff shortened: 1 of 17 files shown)


modules/mysql/src/main/scala/vinyldns/mysql/queue/MySQLMessageQueue.scala
```diff
@@ -0,0 +1,298 @@
...
+
+  private val logger = LoggerFactory.getLogger(classOf[MySQLMessageQueue])
+
+  private val INSERT_MESSAGE =
+    sql"""
+      |INSERT INTO message_queue(id, message_type, in_flight, data, created, updated, timeout_seconds, attempts)
+      |     VALUES ({id}, {messageType}, {inFlight}, {data}, {created}, {updated}, {timeoutSeconds}, {attempts})
+      |ON DUPLICATE KEY UPDATE updated={updated}
+    """.stripMargin
+
```
pauljamescleary: This is important for error handling.

The `ON DUPLICATE KEY` makes this idempotent.  We could in theory use `REPLACE INTO`, but that does a `DELETE` then an `INSERT`, whereas `ON DUPLICATE KEY` does an update.

Should not do `INSERT IGNORE INTO` as that will cause **all** errors (even valid SQL errors) to be ignored.  We do not want that.

In a nutshell, this enforces an "all or nothing" approach to sending messages.  If there are valid SQL errors, the entire batch of messages will fail.  If there are duplicate keys (for idempotentcy), but no SQL errors, the entire batch will **succeed**

modules/mysql/src/main/scala/vinyldns/mysql/queue/MySQLMessageQueue.scala
```diff
@@ -0,0 +1,298 @@
...
+  private val REQUEUE_MESSAGE =
+    sql"""
+      |UPDATE message_queue
+      |   SET in_flight = 0, updated = NOW()
+      | WHERE id = ?
+    """.stripMargin
+
```
pauljamescleary: `Requeue` clears the `in_flight` flag, allowing this message to be picked up on a different receive.

Note: we do not update the `attempts` here, that is updated when we actually claim the messages

modules/mysql/src/main/scala/vinyldns/mysql/queue/MySQLMessageQueue.scala
```diff
@@ -0,0 +1,298 @@
...
+  private val FETCH_UNCLAIMED =
+    sql"""
+         |SELECT id, message_type, data, attempts, timeout_seconds
+         |  FROM message_queue
+         | WHERE in_flight = 0
+         |    OR updated < DATE_SUB(NOW(),INTERVAL timeout_seconds SECOND)
+         | LIMIT ?
+    """.stripMargin
```
pauljamescleary: fetching unclaimed is interesting, so follow the logic here:

1. If the in_flight status is 0, we can take it
1. If in_flight is 1, however the the message visibility has expired, we can claim it.

That second statement is like SQS today.  In SQS, in theory 2 separate processes can process the same message at the same time.  For example, if the message timeout was 5 seconds, and a process had it queued but did not process it for say 10 seconds without changing the timeout OR requeue.  Large GC pauses can obviously force this to happen.

modules/mysql/src/main/scala/vinyldns/mysql/queue/MySQLMessageQueue.scala
```diff
@@ -0,0 +1,298 @@
...
+
+  // Important!  Make sure to update the updated timestamp to current, and increment the attempts!
+  private val CLAIM_MESSAGES =
+    sql"""
+         |UPDATE message_queue
+         |   SET in_flight=1, updated=NOW(), attempts=attempts+1
+         | WHERE id in (?)
+    """.stripMargin
+
```
pauljamescleary: when we claim a message, we make sure to mark the in flight status so noone else will pull it.

We also set the `updated` timestamp to reset message visibility

Finally, we update the number of attempts so we do not retry indefinitely on a message that maybe malformed and will just never succeed.

modules/core/src/main/scala/vinyldns/core/queue/MessageQueue.scala
```diff
@@ -22,12 +22,9 @@ import vinyldns.core.domain.zone.ZoneCommand
...
 // $COVERAGE-OFF$
-
-// Message handle is implementation specific.  For example, in SQS, this may be the `Message` itself
-trait MessageHandle
-
-// Represents a command encoded in a message on a queue
-final case class CommandMessage(handle: MessageHandle, command: ZoneCommand)
+trait CommandMessage {
+  def command: ZoneCommand
+}
```
pauljamescleary: Subtle change in the interface.  Since we do not use `MessageHandle` anywhere in any of the function signatures, it is confusing and unnecessary.  Just use the `CommandMessage` trait instead.  It makes message queue implementations simpler.

modules/api/src/test/scala/vinyldns/api/backend/CommandHandlerSpec.scala
```diff
@@ -32,7 +32,7 @@ import vinyldns.api.domain.dns.DnsConnection
 import vinyldns.core.domain.batch.BatchChangeRepository
 import vinyldns.core.domain.record.{RecordChangeRepository, RecordSetChange, RecordSetRepository}
 import vinyldns.core.domain.zone._
-import vinyldns.core.queue.{CommandMessage, MessageCount, MessageHandle, MessageQueue}
+import vinyldns.core.queue.{CommandMessage, MessageCount, MessageQueue}
```
pauljamescleary: changes in this file are needed to accomodate the small refactor of the `CommandMessage` type

modules/mysql/src/main/scala/vinyldns/mysql/queue/MySQLMessageQueue.scala
```diff
@@ -0,0 +1,298 @@
...
+/* MySQL Command Message implementation */
+final case class MySQLMessage(
+    id: MessageId,
+    attempts: Int,
+    timeout: FiniteDuration,
+    command: ZoneCommand)
+    extends CommandMessage
+object MySQLMessage {
+  final case class UnsupportedCommandMessage(msg: String) extends Throwable(msg)
```
rebstar6: this error can probably live in core, i imagine the other queues will have a similar casting structure

modules/mysql/src/main/scala/vinyldns/mysql/queue/MySQLMessageQueue.scala
```diff
@@ -0,0 +1,298 @@
...
+
+  def fromInt(i: Int): Either[InvalidMessageType, MessageType] = i match {
+    case 1 => Right(RecordChangeMessageType)
+    case 2 => Right(ZoneChangeMessageType)
+    case _ => Left(InvalidMessageType(i))
+  }
+}
+
+/* MySQL Command Message implementation */
+final case class MySQLMessage(
```
rebstar6: split MySQLMessage into its own file I think. I feel like we have a lot shoved in here

modules/mysql/src/main/resources/db/migration/V3.1__MessageQueue.sql
```diff
@@ -0,0 +1,21 @@
+CREATE SCHEMA IF NOT EXISTS ${dbName};
```
rebstar6: just wondering, any reason we did 3.1 vs 4 here?

modules/mysql/src/main/scala/vinyldns/mysql/queue/MySQLMessageQueue.scala
```diff
@@ -0,0 +1,259 @@
...
+import vinyldns.proto.VinylDNSProto
+
+import scala.concurrent.duration._
+
+object MySQLMessageQueue {
+  final case class InvalidMessageHandle(msg: String) extends Throwable(msg)
+  final case class MessageAttemptsExceeded(msg: String) extends Throwable(msg)
+  final case class InvalidMessageTimeout(timeout: Int)
+      extends Throwable(s"Invalid message timeout $timeout")
+  final case class MessageId(value: String) extends AnyVal
```
rebstar6: that's the AnyVal extension give us?

modules/mysql/src/main/scala/vinyldns/mysql/queue/MySQLMessageQueue.scala
rebstar6: not certain that we need to change updated here, but i guess it doesnt matter either way

modules/mysql/src/main/scala/vinyldns/mysql/queue/MySQLMessageQueue.scala
```diff
@@ -0,0 +1,259 @@
...
+  private val CHANGE_TIMEOUT =
+    sql"""
+      |UPDATE message_queue
+      |   SET timeout_seconds = ?
+      | WHERE id = ?
+    """.stripMargin
+
+  private val FETCH_UNCLAIMED =
+    sql"""
+         |SELECT id, message_type, data, attempts, timeout_seconds
```
rebstar6: indenting is inconsistent from here down (with the 1st 3 sqls)

modules/mysql/src/main/scala/vinyldns/mysql/queue/MySQLMessageQueue.scala
```diff
@@ -0,0 +1,259 @@
...
+    *
+    * 2. If there are any errors stemming from step 1, delete them
+    *
+    * 3. For all the good messages, mark them as in-flight, and increment the attempts
+    */
+  def receive(count: MessageCount): IO[List[CommandMessage]] =
+    monitor("queue.JDBC.receive") {
+      IO {
+        // Need a max count so the user doesn't try to take 1MM messages, 10 is sufficient
+        val limit = Math.min(10, count.value)
```
rebstar6: 10 feels low, no?

Also, if were gonna say MessageCount has a max, AND we already have a positive restriction on the type, why not just build that into the type?

modules/mysql/src/main/scala/vinyldns/mysql/queue/MySQLMessageQueue.scala
```diff
@@ -0,0 +1,259 @@
...
+          // Mark the ones we got successfully
+          claimMessages(successes.map(_.id))
+          successes
+        }
+      }
+    }
+
+  def requeue(message: CommandMessage): IO[Unit] =
+    monitor("queue.JDBC.requeue") {
+      IO.fromEither(MySQLMessage.cast(message)).map { mysql =>
```
rebstar6: do we need all this casting everywhere, or should we just store ID on the MessageQueue type? Looks like SQS Message type also has a `private String messageId`

modules/mysql/src/main/scala/vinyldns/mysql/queue/MySQLMessageQueue.scala
```diff
@@ -0,0 +1,259 @@
...
+          }
+        }
+        .as(())
+    }
+
+  def send[A <: ZoneCommand](messages: NonEmptyList[A]): IO[SendBatchResult] =
+    monitor("queue.JDBC.sendBatch") {
+      IO {
+        DB.localTx { implicit s =>
+          // Note, we do replace into, so these really cannot fail, they all succeed or all fail
```
rebstar6: we do INSERT INTO?

modules/mysql/src/main/scala/vinyldns/mysql/queue/MySQLMessageQueue.scala
```diff
@@ -0,0 +1,259 @@
...
+  def insertParams(cmd: ZoneCommand): Seq[(Symbol, Any)] = {
+    val ts = DateTime.now
+    Seq(
+      'id -> cmd.id,
+      'messageType -> MessageType.fromCommand(cmd).value,
+      'inFlight -> 0,
+      'attempts -> 0,
+      'data -> getBytes(cmd),
+      'created -> ts,
+      'updated -> ts,
```
rebstar6: feels like updated should be none to start, no?

modules/mysql/src/main/scala/vinyldns/mysql/queue/MySQLMessageQueue.scala
rebstar6: interesting. updated does kinda feel like a weird field to be changing here though. in general, i think this will all work having reviewed the whole file, but some of the treatment of the updated field in general seems a little confusing

rebstar6: more clearly, treating updated strictly as 'time last pulled from queue' is what we want - by messing with it on creation, im worried we're complicating things

modules/mysql/src/it/scala/vinyldns/mysql/repository/MySQLMessageQueueIntegrationSpec.scala
```diff
@@ -0,0 +1,313 @@
...
+  attempts: Int)
+
+final case class InvalidMessage(command: ZoneCommand) extends CommandMessage
+
+class MySQLMessageQueueIntegrationSpec extends WordSpec with Matchers
+  with BeforeAndAfterEach with EitherMatchers with EitherValues with BeforeAndAfterAll with ProtobufConversions {
+
+  private val underTest = new MySQLMessageQueue()
+
+  private val okZone: Zone = Zone("ok.zone.recordsets.", "test@test.com", adminGroupId = "group")
```
rebstar6: fyi, you can import:

vinyldns.core.TestZoneData.okZone
vinyldns.core.TestRecordSetData.aaaa

private val rsChange = vinyldns.core.TestRecordSetData.makeTestAddChange(aaaa, okZone)
private val zoneChange = vinyldns.core.TestZoneData.zoneChangePending (slightly different change, but still should work)

modules/mysql/src/it/scala/vinyldns/mysql/repository/MySQLMessageQueueIntegrationSpec.scala
```diff
@@ -0,0 +1,313 @@
...
+      insert(rsChange.id, RecordChangeMessageType.value, true, rsChangeBytes, initialTs, initialTs, 30, 1)
+      val msgs = underTest.receive(MessageCount(1).right.value).unsafeRunSync()
+      msgs should have length 1
+    }
+  }
+
+  "changeMessageTimeout" should {
+    "update the message time out" in {
+      underTest.send(rsChange).unsafeRunSync()
+      underTest.changeMessageTimeout(testMessage, 100.seconds).unsafeRunSync()
```
rebstar6: dont like sleeping in a test, but is it worth changing one to 1 second and testing that it gets pulled a 2nd time?

modules/mysql/src/main/scala/vinyldns/mysql/queue/MySQLMessageQueue.scala
pauljamescleary: I modified to update created  Problem is we have to update something ON DUPLICATE KEY.  Better than updated, less confusing I suppose

modules/mysql/src/main/scala/vinyldns/mysql/queue/MySQLMessageQueue.scala
pauljamescleary: I pulled it out, and updated the integration spec

modules/mysql/src/main/resources/db/migration/V3.1__MessageQueue.sql
pauljamescleary: grr, we should not have been updating "major" version in flyway, realistically we should be on V1.4

modules/mysql/src/main/scala/vinyldns/mysql/queue/MySQLMessageQueue.scala
pauljamescleary: probably don't need it.

modules/mysql/src/main/scala/vinyldns/mysql/queue/MySQLMessageQueue.scala
pauljamescleary: yea, that is annoying, will fix assuming that scalafmt doesn't "fix it back"

modules/mysql/src/main/scala/vinyldns/mysql/queue/MySQLMessageQueue.scala
pauljamescleary: `MessageCount` is in core, I suppose we could put that restriction in core.

modules/mysql/src/main/scala/vinyldns/mysql/queue/MySQLMessageQueue.scala
pauljamescleary: Yea, I thought the same thing too.  I just don't know what other commands may come about.  I would be ok with putting it on the `CommandMessage` though, that would simplify things.  Right now it is really annoying to do this casting

modules/mysql/src/main/scala/vinyldns/mysql/queue/MySQLMessageQueue.scala
pauljamescleary: Yep, INSERT INTO with duplicate key handling

modules/mysql/src/main/scala/vinyldns/mysql/queue/MySQLMessageQueue.scala
pauljamescleary: updated really should be "NOW", I may modify that INSERT statement

modules/mysql/src/it/scala/vinyldns/mysql/repository/MySQLMessageQueueIntegrationSpec.scala
pauljamescleary: Ahhhh, `TestZoneData`!  I saw `VinylDNSTestData` and it was in the API.  Ok, will do

modules/mysql/src/it/scala/vinyldns/mysql/repository/MySQLMessageQueueIntegrationSpec.scala
pauljamescleary: that's a good idea

modules/mysql/src/main/scala/vinyldns/mysql/queue/MySQLMessageQueue.scala
rebstar6: comment says `replace into`

modules/mysql/src/main/scala/vinyldns/mysql/queue/MySQLMessageQueue.scala
```diff
@@ -0,0 +1,259 @@
...
+      IO {
+        // Need a max count so the user doesn't try to take 1MM messages, 10 is sufficient
+        val limit = Math.min(10, count.value)
+
+        DB.localTx { implicit s =>
+          // get unclaimed messages, note these could fail during retrieval
+          val claimed = fetchUnclaimed(limit)
+
+          // Errors could not be deserialized, have an invalid type, or exceeded retries
+          val errors = claimed.collect {
```
mitruly: Was playing around with this a bit, and `claimed.separate` might make things easier to work with since it will separate a collection of `Either` into `Left`s and `Right`s.

modules/mysql/src/it/scala/vinyldns/mysql/queue/MySqlMessageQueueIntegrationSpec.scala
```diff
@@ -0,0 +1,300 @@
...
+      val r = underTest.receive(MessageCount(1).right.value).unsafeRunSync()
+
+      // second receive should not re-deliver
+      underTest.receive(MessageCount(1).right.value).unsafeRunSync() shouldBe empty
+
+      underTest.changeMessageTimeout(testMessage, 1.seconds).unsafeRunSync()
+      Thread.sleep(2000)
+      val r2 = underTest.receive(MessageCount(1).right.value).unsafeRunSync()
+      r2 should have length 1
+      r.headOption shouldBe r2.headOption`
```
mitruly: A wild ` appeared!

rebstar6: CATCH IT!!!

modules/mysql/src/main/scala/vinyldns/mysql/queue/MySqlMessageQueue.scala
```diff
@@ -0,0 +1,253 @@
...
+      IO {
+        // Need a max count so the user doesn't try to take 1MM messages, 10 is sufficient
+        val limit = Math.min(10, count.value)
+
+        DB.localTx { implicit s =>
+          // get unclaimed messages, note these could fail during retrieval
+          val claimed = fetchUnclaimed(limit)
+
+          // Errors could not be deserialized, have an invalid type, or exceeded retries
+          val errors = claimed.collect {
```
mitruly: Was playing around with this a bit, and `claimed.separate` might make things easier to work with since it will separate a collection of `Either` into `Left`s and `Right`s.

pauljamescleary: It is a good idea, but then I still need to pull out the id, doing something unsavory like `.map(_._.2)`.  Going to keep it like this, but thanks for that info!

modules/mysql/src/main/scala/vinyldns/mysql/queue/MySqlMessageQueue.scala
```diff
@@ -0,0 +1,253 @@
...
+    *
+    * 2. If there are any errors stemming from step 1, delete them
+    *
+    * 3. For all the good messages, mark them as in-flight, and increment the attempts
+    */
+  def receive(count: MessageCount): IO[List[CommandMessage]] =
+    monitor("queue.JDBC.receive") {
+      IO {
+        // Need a max count so the user doesn't try to take 1MM messages, 10 is sufficient
+        val limit = Math.min(10, count.value)
```
rebstar6: leaving this here? im still team embed this in the MessageCount type, but wont hold it up for that

pauljamescleary: The "max" seems to be queue specific.  There is a limit of 10 in SQS, but could be 100, 200, 1000 in other cases like RabbitMQ or redismq.  Even here we could receive more than 10 at a time.  Truth is, I am unclear of what the limit should be, 10 will always be fine, but 1,000,000 maybe too many.  Would need to do some benchmarking to figure out the appropriate limit.  Even then, the limit could be different based on your MySql setup.

We could loop until we satisfy the count as an alternative, but we may never return if there is a long gap (hours, days) between messages.  So that is not a good option.

Another alternative would be to encode the error in the return type, something like "MaxMessageCountExceeded".  `EitherT[IO, MaxMessageCountExceeded, List[CommandMessage]]`.  It certainly makes working with the queue more complicated.

**tl;dr** I think leaving this in is an ok tradeoff for simplicity purposes.

PR #267, opened 2018-10-05
exclude filter for local.conf

### LINKED ISSUE #209: Exclude modules/portal/conf/local.conf from packaging
**Describe the bug**
modules/portal/conf/local.conf is described in DEVELOPERS.md as somewhere you can store local test settings, like settings and creds for an ldap env. This file should not make it into a docker image if a developer publishes from their machine 

**To Reproduce**
1. have a `modules/portal/conf/local.conf`
1. start up sbt
1. `project portal`
1. `docker:stage`
1. check out `modules/portal/target/docker/stage/opt/docker/conf`, you will see your `local.conf` which would be bad 

**Additional context**
Some googling brings up potential solutions, for example https://stackoverflow.com/questions/24375624/how-to-exclude-files-under-conf-folder-for-distribution

nimaeskandary:
Fixes #209 .

Changes in this pull request:
- `modules/portal/conf/local.conf` is excluded from packaging so it doesn't show up in docker images
- `modules/portal/conf/local.conf` is excluded from `packageBin` so it doesn't show up in jars of the portal

To test: 
- have a `modules/portal/conf/local.conf` that overrides application.conf settings
- ensure that running the portal locally still has settings from local.conf
- run `sbt ";project portal;clean;docker:stage"`
- run `ls modules/portal/target/docker/stage/opt/docker/conf` and ensure `local.conf` is not there
- run `mkdir modules/portal/target/scala-2.12/test` and `cd modules/portal/target/scala-2.12/test && unzip ../portal_2.12-0.8.0-SNAPSHOT.jar && ls` and ensure `local.conf` is not in the jar

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -171,7 +171,11 @@ lazy val apiPublishSettings = Seq(
 lazy val portalPublishSettings = Seq(
   publishArtifact := false,
   publishLocal := (publishLocal in Docker).value,
-  publish := (publish in Docker).value
+  publish := (publish in Docker).value,
+  // for sbt-native-packager (docker) to exclude local.conf
+  mappings in Universal ~= (_.filterNot(_._1.name.equals("local.conf"))),
+  // for local.conf to be excluded in jars
+  mappings in (Compile, packageBin) ~= { _.filter(!_._1.getName.equals("local.conf")) }
 )
 
 lazy val pbSettings = Seq(
--- a/build.sbt
+++ b/build.sbt
@@ -173,9 +173,13 @@ lazy val portalPublishSettings = Seq(
   publishLocal := (publishLocal in Docker).value,
   publish := (publish in Docker).value,
   // for sbt-native-packager (docker) to exclude local.conf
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -171,7 +171,11 @@ lazy val apiPublishSettings = Seq(
 lazy val portalPublishSettings = Seq(
   publishArtifact := false,
   publishLocal := (publishLocal in Docker).value,
-  publish := (publish in Docker).value
+  publish := (publish in Docker).value,
+  // for sbt-native-packager (docker) to exclude local.conf
+  mappings in Universal ~= (_.filterNot(_._1.name.equals("local.conf"))),
```
mitruly: Something like the following might be easier to read and achieves the same results:
```
  mappings in Universal ~= (_.filterNot {
    case (file, _) => file.getName.equals("local.conf")
  }),
  mappings in (Compile, packageBin) ~= (_.filterNot {
    case (file, _) => file.getName.equals("local.conf")
  })
```
Not sure if they can be combined somehow.

Issue #268, opened 2018-10-07
Docs typo: SSHFP record model docs cite a `typ` attribute


mdb:
**Describe the bug**

It appears the record set model docs for SSHFP records cite a `typ` attribute. I believe this is a typo and should be `type`.

**VinylDNS Version**

`master` documentation

**To Reproduce**
1. Go to the [record set model docs](https://www.vinyldns.io/api/recordset-model.html)
2. View the [record set data info associated with SSHFP record sets](https://www.vinyldns.io/api/recordset-model.html#record-data-information-)
3. Note that it cites a `typ` attribute

**Expected behavior**
I believe the attribute should be `type`.

**Additional context**
This was pointed out to me [here](https://github.com/vinyldns/vinyldns-cli/pull/19) by @tenpaiyomi.

pauljamescleary:
Is this closed?  @mdb, @nimaeskandary ?

mdb:
@pauljamescleary Yes, I believe this is fixed, as PR #269 has been merged.

PR #269, opened 2018-10-07
fix typo re: SSHFP record model type attribute

mdb:
The record set model docs for SSHFP records cite a `typ` attribute. This is a typo and should be `type`.

Fixes issue #268

Changes in this pull request:
- correct SSHFP record set model documentation typo to read `type` rather than `typ`
- remove some whitespace


mdb:
@mitruly Good catch. I've pushed an additional PR fixing the other `typ`s.

PR #278, opened 2018-10-10
Zone saves

### LINKED ISSUE #252: Handle dead-locks in zone saves gracefully
**Describe the bug**
Presently, we can encounter a MySQL dead-lock condition when saving the zone.  This can happen when multiple concurrent updates are made to the same zone at the same time.  

Instead of gracefully handling the situation, the zone change unexpectedly fails in the ZoneCommandHandler, forcing a retry 30 minutes later.

For functional tests, this represents a problem, as the zone is saved long after the functional tests complete, leaving a straggling zone behind.

**Solution**
Handle the deadlock detection in the `MySQLZoneRepository.save`.  Wrap the entire `save` in a retry block, that will retry with a back off similar to what is done in the `DynamoDBHelper`.

The backoff in the `DynamoDBHelper` should be sufficient.  Still rather short (total retry time under 1 second).  

`private class RetryStateHolder(var retries: Int = 10, var backoff: FiniteDuration = 1.millis)`

nimaeskandary:
Fixes #252  .

Changes in this pull request:
- Retry zone save and delete with backoff in MySqlZoneRepository
- Used this example from cats docs https://typelevel.org/cats-effect/datatypes/io.html#example-retrying-with-exponential-backoff, changed slightly for testing


modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneRepository.scala
```diff
@@ -356,4 +344,35 @@ class MySqlZoneRepository extends ZoneRepository with ProtobufConversions with M
...
+          putZoneAccess(zone)
+          zone
+        }
+      }
+    }
+
+  def retryWithBackoff[A](f: A => IO[A], a: A, delay: FiniteDuration, maxRetries: Int): IO[A] = {
+    implicit val timer: Timer[IO] = IO.timer(ExecutionContext.global)
+    f(a).handleErrorWith { error =>
+      if (maxRetries > 1)
```
nimaeskandary: This should be a zero, will change

rebstar6:
looks good to me minus your 1 comment

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneRepository.scala
pauljamescleary: shouldn't have the timer inside the `def`.  Instead, make it a class member `private implicit val timer...`

PR #282, opened 2018-10-10
ignore patch failures

rebstar6:
this should report patch coverage, but always be green


rebstar6:
(see https://docs.codecov.io/docs/commit-status#section-target for ref)

PR #283, opened 2018-10-10
Common MySQL connector

rebstar6:
made MySqlConnector with shared methods (will be used by the queue provider)

### DIFF
--- a/docker/api/docker.conf
+++ b/docker/api/docker.conf
@@ -60,7 +60,7 @@ vinyldns {
       user = ${?JDBC_USER}
       password = "pass"
       password = ${?JDBC_PASSWORD}
-      pool-max-size = 20
+      maximum-pool-size = 20
       connection-timeout-millis = 1000
       max-life-time = 600000
     }
--- a/docker/api/docker.conf
+++ b/docker/api/docker.conf
@@ -60,9 +60,12 @@ vinyldns {
       user = ${?JDBC_USER}
       password = "pass"
       password = ${?JDBC_PASSWORD}
-      maximum-pool-size = 20
+      # see https://github.com/brettwooldridge/HikariCP
       connection-timeout-millis = 1000
-      max-life-time = 600000
+      idle-timeout = 10000
+      max-lifetime = 600000
+      maximum-pool-size = 20
+      minimum-idle = 20
     }
     # Repositories that use this data store are listed here
     repositories {
--- a/docker/portal/application.conf
+++ b/docker/portal/application.conf
@@ -19,25 +19,29 @@ portal.vinyldns.backend.url = "http://vinyldns-api:9000"
 portal.test_login = true
 
 # configuration for the users and groups store
-dynamo {
-  key = "akid goes here"
-  secret = "secret key goes here"
-  endpoint = "http://vinyldns-dynamodb:8000"
-  test_datastore = false
-}
+data-stores = ["dynamodb"]
+dynamodb {
+  class-name = "vinyldns.dynamodb.repository.DynamoDBDataStoreProvider"
 
-users {
-  dummy = false
-  tablename = "users"
-  provisionedReadThroughput = 100
-  provisionedWriteThroughput = 100
-}
+  settings {
+    key = "akid goes here"
+    secret = "secret key goes here"
+    endpoint = "http://vinyldns-dynamodb:8000"
+    region = "us-east-1" # note: we are always in us-east-1, but this can be overridden
+  }
 
-changelog {
-  dummy = false
-  tablename = "usersAndGroupChanges"
-  provisionedReadThroughput = 100
-  provisionedWriteThroughput = 100
+  repositories {
+    user {
+      table-name = "users"
+      provisioned-reads = 100
+      provisioned-writes = 100
+    }
+    user-change {
+      table-name = "userChange"
+      provisioned-reads = 100
+      provisioned-writes = 100
+    }
+  }
 }
 
 LDAP {
--- a/modules/api/src/main/resources/reference.conf
+++ b/modules/api/src/main/resources/reference.conf
@@ -48,9 +48,10 @@ vinyldns {
       url = "jdbc:mariadb://localhost:19002/vinyldns?user=root&password=pass"
       user = "root"
       password = "pass"
-      pool-max-size = 20
+      maximum-pool-size = 5
       connection-timeout-millis = 1000
       max-life-time = 600000
+      register-mbeans = true
     }
 
     repositories {
--- a/modules/api/src/main/resources/reference.conf
+++ b/modules/api/src/main/resources/reference.conf
@@ -48,9 +48,13 @@ vinyldns {
       url = "jdbc:mariadb://localhost:19002/vinyldns?user=root&password=pass"
       user = "root"
       password = "pass"
-      maximum-pool-size = 5
+
+      # see https://github.com/brettwooldridge/HikariCP
       connection-timeout-millis = 1000
-      max-life-time = 600000
+      idle-timeout = 10000
+      max-lifetime = 600000
+      maximum-pool-size = 5
+      minimum-idle = 5
       register-mbeans = true
     }
 
--- a/modules/mysql/src/it/resources/application.conf
+++ b/modules/mysql/src/it/resources/application.conf
@@ -11,10 +11,17 @@ mysql {
     url = "jdbc:mariadb://localhost:19004/vinyldns?user=root&password=pass"
     user = "root"
     password = "pass"
-    # TODO: Research optimal values for settings
-    pool-max-size = 20
+
+    maximum-pool-size = 5
     connection-timeout-millis = 1000
     max-life-time = 10000
+    minimum-idle = 1
+
+    my-sql-properties = {
+      cachePrepStmts=true
+      prepStmtCacheSize=250
+      prepStmtCacheSqlLimit=2048
+    }
   }
 
   repositories {
--- a/modules/mysql/src/it/resources/application.conf
+++ b/modules/mysql/src/it/resources/application.conf
@@ -15,7 +15,7 @@ mysql {
     # see https://github.com/brettwooldridge/HikariCP
     connection-timeout-millis = 1000
     idle-timeout = 10000
-    max-lifetime = 10000
+    max-lifetime = 30000
     maximum-pool-size = 5
     minimum-idle = 0
 
--- a/modules/mysql/src/it/resources/application.conf
+++ b/modules/mysql/src/it/resources/application.conf
@@ -17,7 +17,7 @@ mysql {
     idle-timeout = 10000
     max-lifetime = 10000
     maximum-pool-size = 5
-    minimum-idle = 1
+    minimum-idle = 0
 
     my-sql-properties = {
       cachePrepStmts=true
--- a/modules/mysql/src/it/resources/application.conf
+++ b/modules/mysql/src/it/resources/application.conf
@@ -12,9 +12,11 @@ mysql {
     user = "root"
     password = "pass"
 
-    maximum-pool-size = 5
+    # see https://github.com/brettwooldridge/HikariCP
     connection-timeout-millis = 1000
-    max-life-time = 10000
+    idle-timeout = 10000
+    max-lifetime = 10000
+    maximum-pool-size = 5
     minimum-idle = 1
 
     my-sql-properties = {
--- a/modules/mysql/src/it/resources/logback.xml
+++ b/modules/mysql/src/it/resources/logback.xml
@@ -1,0 +1,16 @@
+<configuration>
+    <!-- Test configuration, log to console so we can get the docker logs -->
+    <appender name="CONSOLE" class="ch.qos.logback.core.ConsoleAppender">
+        <encoder class="ch.qos.logback.classic.encoder.PatternLayoutEncoder">
+            <pattern>%d [test] %-5p | \(%logger{4}:%line\) | %msg %n %nopexception </pattern>
+        </encoder>
+    </appender>
+
+    <logger name="org.flywaydb" level="INFO">
+        <appender-ref ref="CONSOLE"/>
+    </logger>
+
+    <root level="ERROR">
+        <appender-ref ref="CONSOLE"/>
+    </root>
+</configuration>
--- a/modules/mysql/src/main/scala/vinyldns/mysql/MySqlConnectionConfig.scala
+++ b/modules/mysql/src/main/scala/vinyldns/mysql/MySqlConnectionConfig.scala
@@ -34,5 +34,5 @@ final case class MySqlConnectionConfig(
     registerMbeans: Boolean = false,
     // MySql performance settings
     // see https://github.com/brettwooldridge/HikariCP/wiki/MySQL-Configuration
-    mySqlProperties: Map[String, String] = Map(),
+    mySqlProperties: Map[String, AnyRef] = Map(),
 )
--- a/modules/mysql/src/main/scala/vinyldns/mysql/MySqlConnectionConfig.scala
+++ b/modules/mysql/src/main/scala/vinyldns/mysql/MySqlConnectionConfig.scala
@@ -24,13 +24,13 @@ final case class MySqlConnectionConfig(
     user: String,
     password: String,
     migrationSchemaTable: Option[String],
-    // Optional hikari settings
+    // Optional settings, will use Hikari defaults if unset
     // see https://github.com/brettwooldridge/HikariCP#frequently-used
     connectionTimeoutMillis: Option[Long],
     idleTimeout: Option[Long],
-    minimumIdle: Option[Int],
+    maxLifetime: Option[Long],
     maximumPoolSize: Option[Int],
-    maxLifeTime: Option[Long],
+    minimumIdle: Option[Int],
     registerMbeans: Boolean = false,
     // MySql performance settings, should be stored in a .properties file
     // see https://github.com/brettwooldridge/HikariCP/wiki/MySQL-Configuration
--- a/modules/mysql/src/main/scala/vinyldns/mysql/MySqlConnectionConfig.scala
+++ b/modules/mysql/src/main/scala/vinyldns/mysql/MySqlConnectionConfig.scala
@@ -16,6 +16,18 @@
 
 package vinyldns.mysql
 
+import com.typesafe.config.ConfigObject
+import pureconfig.ConfigReader
+
+object MySqlConnectionConfig {
+
+  import scala.collection.JavaConverters._
+
+  val mySqlPropertiesReader: ConfigReader[Map[String, AnyRef]] = {
+    ConfigReader[Option[ConfigObject]].map(_.map(_.unwrapped().asScala.toMap).getOrElse(Map()))
+  }
+}
+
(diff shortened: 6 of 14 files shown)


modules/mysql/src/main/scala/vinyldns/mysql/MySqlConnector.scala
```diff
@@ -0,0 +1,88 @@
...
+  def runDBMigrations(config: MySqlConnectionConfig): IO[Unit] = {
+    val migrationConnectionSettings = DataSourceSettings(
+      config.driver,
+      config.migrationUrl,
+      config.user,
+      config.password,
+      3,
+      30000,
+      1800000,
+      false)
```
rebstar6: when we run migrations, we were not setting these last 3 values. I threw in the default, but this may be a good time to think about what flyway connectionTimeoutMillis/maxLifeTime/setRegisterMbeans should be

modules/mysql/src/main/scala/vinyldns/mysql/MySqlConnector.scala
```diff
@@ -0,0 +1,88 @@
...
+
+    IO.fromEither(getDataSource(migrationConnectionSettings)).map { migrationDataSource =>
```
rebstar6: functionally the same as the old migration method, just using shared `getDataSource` method

modules/mysql/src/main/scala/vinyldns/mysql/MySqlConnector.scala
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  def getDataSource(settings: DataSourceSettings): Either[Throwable, DataSource] =
+    Try {
+      val ds = new HikariDataSource()
+      ds.setDriverClassName(settings.driver)
+      ds.setJdbcUrl(settings.url)
+      ds.setUsername(settings.user)
+      ds.setPassword(settings.password)
+      ds.setConnectionTimeout(settings.connectionTimeoutMillis)
```
rebstar6: some of these can technically throw

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlDataStoreProvider.scala
```diff
@@ -65,65 +62,28 @@ class MySqlDataStoreProvider extends DataStoreProvider {
...
-      migration.setTable(tableName)
-    }
+    IO.fromEither(getDataSource(dbConnectionSettings)).map { dataSource =>
+      logger.error("configuring connection pool")
 
-    val placeholders = Map("dbName" -> settings.name)
-    migration.setPlaceholders(placeholders.asJava)
-    migration.setSchemas(settings.name)
+      // pulled out of DBs.setupAll since we're no longer using the db. structure for config
+      DBs.loadGlobalSettings()
```
rebstar6: this is new.

if you look at DBs.setupAll, it calls loadGlobalSettings (logging level stuff, we still want that), and then for each dbName defined in the structure `db.default...` in config, it will add a connection pool. we're already adding the hikari connection pool specifically, AND we dont have things configured in that style, so the setupAll is a little misleading

modules/mysql/src/main/scala/vinyldns/mysql/MySqlConnector.scala
```diff
@@ -0,0 +1,88 @@
...
+      migration.setSchemas(config.name)
+
+      // Runs flyway migrations
+      migration.migrate()
+      logger.info("migrations complete")
+    }
+  }
+
+  def getDataSource(settings: DataSourceSettings): Either[Throwable, DataSource] =
+    Try {
```
pauljamescleary:
```scala
import cats._, cats.data._, cats.implicits._

// then you can do this magic
Either.catchNonFatal
```

modules/mysql/src/main/scala/vinyldns/mysql/MySqlConnector.scala
```diff
@@ -0,0 +1,88 @@
...
+
+  final case class DataSourceSettings(
+      driver: String,
+      url: String,
+      user: String,
+      password: String,
+      poolMaxSize: Int,
+      connectionTimeoutMillis: Long,
+      maxLifeTime: Long,
+      setRegisterMbeans: Boolean)
```
pauljamescleary: I think it is a good idea to explore the additional settings.  There are a lot of settings we should have configured but currently do not set that will help us tune.  The "new" pulsar-portal had all of those set.  Minimally, we should pull those over and re-verify that they are what we want.

modules/mysql/src/main/scala/vinyldns/mysql/MySqlConnector.scala
rebstar6: mm got it

modules/mysql/src/main/scala/vinyldns/mysql/MySqlConnector.scala
rebstar6: all im seeing in that project is
db.default.poolMaxSize=5
db.default.connectionTimeoutMillis=5000

we dropped a lot of the other settings

modules/mysql/src/main/scala/vinyldns/mysql/queue/MySqlMessageQueueProvider.scala
```diff
@@ -0,0 +1,64 @@
...
+import cats.effect.IO
+import org.slf4j.LoggerFactory
+import pureconfig.module.catseffect.loadConfigF
+import scalikejdbc.{ConnectionPool, DataSourceConnectionPool}
+import scalikejdbc.config.DBs
+import vinyldns.core.queue.{MessageQueue, MessageQueueConfig, MessageQueueProvider}
+import vinyldns.mysql.MySqlConnectionConfig
+import vinyldns.mysql.MySqlConnector._
+
+class MySqlMessageQueueProvider extends MessageQueueProvider {
```
rebstar6: just threw this in here too so you can see the reason i refactored as i did. will need tests, though note its *very* similar to the db Provider

modules/mysql/src/main/scala/vinyldns/mysql/queue/MySqlMessageQueueProvider.scala
```diff
@@ -0,0 +1,64 @@
...
+
+    IO.fromEither(getDataSource(queueConnectionSettings)).map { dataSource =>
+      logger.error("configuring connection pool for queue")
+
+      // note this is being called 2x in the case you use the mysql datastores and
+      // loader. That should be ok
+      DBs.loadGlobalSettings()
+
+      // Configure the connection pool
+      ConnectionPool.add("queue", new DataSourceConnectionPool(dataSource))
```
rebstar6: i believe this is what we want. will need to change the MySqlMessageQueue to use the "queue" connection pool rather than the default

mitruly:
Just one build issue in `MySqlDataStoreProviderSpec`:
```
a[pureconfig.error.ConfigReaderException[MySqlConnectionConfig]] should be thrownBy underTest
```

docker/api/docker.conf
```diff
@@ -60,7 +60,7 @@ vinyldns {
       user = ${?JDBC_USER}
       password = "pass"
       password = ${?JDBC_PASSWORD}
-      pool-max-size = 20
+      maximum-pool-size = 20
```
rebstar6: see https://github.com/brettwooldridge/HikariCP#frequently-used, consistent naming.

modules/api/src/main/resources/reference.conf
```diff
@@ -48,9 +48,10 @@ vinyldns {
       url = "jdbc:mariadb://localhost:19002/vinyldns?user=root&password=pass"
       user = "root"
       password = "pass"
-      pool-max-size = 20
+      maximum-pool-size = 20
       connection-timeout-millis = 1000
       max-life-time = 600000
+      register-mbeans = true
```
rebstar6: we hardcoded true for this before, so setting in ref.conf true for now

rebstar6: eventually all these settings should be excluded from ref.conf. optional, if missing will not set anything on the datasource

modules/mysql/src/main/scala/vinyldns/mysql/queue/MySqlMessageQueueProvider.scala
pauljamescleary: We should probably pass in the pool to each "repo" / "queue" if we are going this route.  Not sure what the right thing is to pass in.  Probably either the `ConnectionPoolContext` or `ConnectionPool` from scalikejdbc.

modules/api/src/main/resources/reference.conf
pauljamescleary: For the settings, I found the ones we used, was not in the portal but in old pulsar...

```
# these database settings are recommended here https://github.com/brettwooldridge/HikariCP/wiki/MySQL-Configuration
# as well as here http://assets.en.oreilly.com/1/event/21/Connector_J%20Performance%20Gems%20Presentation.pdf
# and here https://www.facebook.com/notes/mysql-at-facebook/jdbc-connection-overheads/10152027251270933/

# removes additional overhead for new connections
db.default.cacheServerConfiguration=true

# sets the number of prepared statements that will be cached, we don't have many so we can be conservative here
db.default.prepStmtCacheSize=50

# sets the max length of a prepared SQL statement.  We will use 2048 but may not need that
db.default.prepStmtCacheSqlLimit=2048

# make sure we cache prepared statements, and cache on server if availalbe, this makes things snappy!
db.default.cachePrepStmts=true
db.default.useServerPrepStmts=true

db.default.useLocalSessionState=true
db.default.useLocalTransactionState=true

# cache metadata for columns and tables
db.default.cacheResultSetMetadata=true
```

rebstar6: ill check for these in hikari and add them as optional too

modules/mysql/src/main/scala/vinyldns/mysql/queue/MySqlMessageQueueProvider.scala
rebstar6: per http://scalikejdbc.org/documentation/connection-pool.html, what i think we wanna do is change the queue functions to `NamedDB('queue).localTx(...)` rather than `DB...`

in that case, there shouldnt be any need to pass it in (that said, i havent tried this yet :P)

rebstar6: hmm that said we may need the ConnectionPoolContext. its implicit, so never quite sure if its there or not

mitruly:
Oof, -0.2%!

modules/mysql/src/main/scala/vinyldns/mysql/MySqlConnectionConfig.scala
```diff
@@ -14,16 +14,25 @@
...
+    // see https://github.com/brettwooldridge/HikariCP#frequently-used
+    connectionTimeoutMillis: Option[Long],
+    idleTimeout: Option[Long],
+    minimumIdle: Option[Int],
+    maximumPoolSize: Option[Int],
+    maxLifeTime: Option[Long],
+    registerMbeans: Boolean = false,
+    // MySql performance settings, should be stored in a .properties file
+    // see https://github.com/brettwooldridge/HikariCP/wiki/MySQL-Configuration
+    mySqlProperties: Map[String, String] = Map(),
```
rebstar6: @pauljamescleary anything at https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-reference-configuration-properties.html can be set, so it's pretty free-form at this point.

As an alternative to this map, we can ask for a path to a .properties file and load from there.

Note - I havent yet checked, but if in config, these may have to be in camel case vs snake, which will be weird

modules/api/src/main/resources/reference.conf
rebstar6: see https://github.com/vinyldns/vinyldns/pull/283/files#diff-ba4331e5191a431fd5a6fe18b463cdadR37

modules/mysql/src/it/resources/application.conf
```diff
@@ -11,10 +11,17 @@ mysql {
     url = "jdbc:mariadb://localhost:19004/vinyldns?user=root&password=pass"
     user = "root"
     password = "pass"
-    # TODO: Research optimal values for settings
-    pool-max-size = 20
+
+    maximum-pool-size = 5
```
rebstar6: integration specs probably dont need 20 connections...

modules/mysql/src/main/scala/vinyldns/mysql/MySqlConnectionConfig.scala
pauljamescleary: Not a fan of `Map[String, String]` as we lose a lot of safety.  I don't think we have to support _every_ my sql property.  When I was doing my research, there was manageable number that made sense.

I would say, either leave those out for the time being, or just pull in the properties I mentioned in my earlier comment.  As I benchmark, I may find others worthwhile adding and can post a separate PR

modules/mysql/src/main/scala/vinyldns/mysql/MySqlConnectionConfig.scala
```diff
@@ -14,16 +14,25 @@
...
+    migrationSchemaTable: Option[String],
+    // Optional hikari settings
+    // see https://github.com/brettwooldridge/HikariCP#frequently-used
+    connectionTimeoutMillis: Option[Long],
+    idleTimeout: Option[Long],
+    minimumIdle: Option[Int],
+    maximumPoolSize: Option[Int],
+    maxLifeTime: Option[Long],
+    registerMbeans: Boolean = false,
+    // MySql performance settings, should be stored in a .properties file
```
pauljamescleary: remove this comment, we cannot do a .properties file without changing the docker image.  Would be good just to leave it as a nested config entry like you have it.

modules/mysql/src/main/scala/vinyldns/mysql/MySqlConnector.scala
```diff
@@ -0,0 +1,83 @@
...
+      dsConfig.setPassword(settings.password)
+
+      settings.connectionTimeoutMillis.foreach(dsConfig.setConnectionTimeout)
+      settings.idleTimeout.foreach(dsConfig.setIdleTimeout)
+      settings.minimumIdle.foreach(dsConfig.setMinimumIdle)
+      settings.maximumPoolSize.foreach(dsConfig.setMaximumPoolSize)
+      settings.maxLifeTime.foreach(dsConfig.setMaxLifetime)
+      dsConfig.setRegisterMbeans(settings.registerMbeans)
+
+      settings.mySqlProperties.foreach {
```
pauljamescleary: It is interesting, this actually works.  In your application.conf, you have `prepStmtCacheSqlLimit=2048`, which is a number not a string.

Going to make sure there is a unit test for this.

modules/mysql/src/main/scala/vinyldns/mysql/MySqlConnectionConfig.scala
rebstar6: oh oops, yeah meant to change that :)

docker/portal/application.conf
```diff
@@ -19,25 +19,29 @@ portal.vinyldns.backend.url = "http://vinyldns-api:9000"
 portal.test_login = true
 
 # configuration for the users and groups store
-dynamo {
```
rebstar6: this had legacy stuff, updating

modules/mysql/src/main/scala/vinyldns/mysql/MySqlDataSourceSettings.scala
```diff
@@ -0,0 +1,73 @@
...
+
+  override def toString: String = {
+    val sb = new StringBuilder
+    def addNamedField(name: String, field: Any): Unit =
+      sb.append(name).append("=\"").append(field).append("\"; ")
+
+    sb.append("MySqlDataSourceSettings: [")
+    addNamedField("driver", driver)
+    addNamedField("user", user)
+    val maskedUrl = url.split('?').headOption.getOrElse("")
```
rebstar6: toString override to clear the query string and remove password from printout

modules/mysql/src/main/scala/vinyldns/mysql/MySqlConnectionConfig.scala
```diff
@@ -14,16 +14,25 @@
...
+    // see https://github.com/brettwooldridge/HikariCP#frequently-used
+    connectionTimeoutMillis: Option[Long],
+    idleTimeout: Option[Long],
+    maxLifetime: Option[Long],
+    maximumPoolSize: Option[Int],
+    minimumIdle: Option[Int],
+    registerMbeans: Boolean = false,
+    // MySql performance settings
+    // see https://github.com/brettwooldridge/HikariCP/wiki/MySQL-Configuration
+    mySqlProperties: Map[String, String] = Map(),
```
pauljamescleary: Make this `Map[String, Any]` or `Map[String, AnyRef]` to align with its usage ... `public void addDataSourceProperty(String propertyName, Object value)`

modules/mysql/src/main/scala/vinyldns/mysql/MySqlConnector.scala
```diff
@@ -0,0 +1,85 @@
...
+
+      // Runs flyway migrations
+      migration.migrate()
+      logger.info("migrations complete")
+    }
+  }
+
+  def getDataSource(settings: MySqlDataSourceSettings): IO[HikariDataSource] = IO {
+
+    logger.error(s"Initializing data source with settings: $settings")
```
pauljamescleary: Wouldn't this log loudly the username and password?  Should make sure we are not logging the password, either as a URL or `settings.password`

pauljamescleary: nvm, see the override on the tostring

modules/mysql/src/test/resources/application.conf
```diff
@@ -11,10 +11,19 @@ mysql {
...
     user = "some-user"
     password = "some-pass"
-    # TODO: Research optimal values for settings
-    pool-max-size = 20
+
+    # see https://github.com/brettwooldridge/HikariCP
     connection-timeout-millis = 1000
-    max-life-time = 600000
+    idle-timeout = 10000
+    max-lifetime = 10000
```
pauljamescleary: please update this as it is discarded.  I see the message...

```
mysqlDbPool - maxLifetime is less than 30000ms, setting to default 1800000ms.
```

The minimum value should be 30000.

Should be updated elsewhere as well.

modules/mysql/src/it/resources/logback.xml
```diff
@@ -0,0 +1,16 @@
+<configuration>
+    <!-- Test configuration, log to console so we can get the docker logs -->
```
rebstar6: hiding a bunch of logs in the integration spec. right now theyre printing like crazy

modules/mysql/src/main/scala/vinyldns/mysql/MySqlConnectionConfig.scala
pauljamescleary: Is this going to be a `Map[String, AnyRef]`?

PR #284, opened 2018-10-10
WIP MySQL Record Set Repository

### LINKED ISSUE #177: MySQL Record Set Repository

1. Create MySQLRecordSetRepository that implements functionality of DynamoDBRecordSetRepository

pauljamescleary:
Fixes #177

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -55,7 +55,6 @@ lazy val sharedSettings = Seq(
   scalacOptions in (Compile, doc) += "-no-link-warnings",
   // Use wart remover to eliminate code badness
   wartremoverErrors ++= Seq(
-    Wart.ArrayEquals,
     Wart.EitherProjectionPartial,
     Wart.IsInstanceOf,
     Wart.JavaConversions,
--- a/build.sbt
+++ b/build.sbt
@@ -298,6 +298,7 @@ lazy val mysql = (project in file("modules/mysql"))
   .settings(scalaStyleCompile ++ scalaStyleTest)
   .settings(
     organization := "io.vinyldns",
+    fork in IntegrationTest := true,
   ).dependsOn(core % "compile->compile;test->test")
 
 val preparePortal = TaskKey[Unit]("preparePortal", "Runs NPM to prepare portal for start")
--- a/modules/mysql/src/it/resources/application.conf
+++ b/modules/mysql/src/it/resources/application.conf
@@ -7,8 +7,8 @@ mysql {
     # assumes a docker or mysql instance running locally
     name = "vinyldns"
     driver = "org.mariadb.jdbc.Driver"
-    migration-url = "jdbc:mariadb://localhost:19004/?user=root&password=pass"
-    url = "jdbc:mariadb://localhost:19004/vinyldns?user=root&password=pass"
+    migration-url = "jdbc:mariadb://localhost:19004/?user=root&password=pass&rewriteBatchedStatements=true"
+    url = "jdbc:mariadb://localhost:19004/vinyldns?user=root&password=pass&rewriteBatchedStatements=true"
     user = "root"
     password = "pass"
     # TODO: Research optimal values for settings
@@ -20,5 +20,15 @@ mysql {
   repositories {
     batch-change {}
     zone {}
+    record-set {}
   }
 }
+
+scalikejdbc.global.loggingSQLAndTime.enabled=true
+scalikejdbc.global.loggingSQLAndTime.logLevel=error
+scalikejdbc.global.loggingSQLAndTime.warningEnabled=true
+scalikejdbc.global.loggingSQLAndTime.warningThresholdMillis=1000
+scalikejdbc.global.loggingSQLAndTime.warningLogLevel=warn
+scalikejdbc.global.loggingSQLAndTime.singleLineMode=false
+scalikejdbc.global.loggingSQLAndTime.printUnprocessedStackTrace=false
+scalikejdbc.global.loggingSQLAndTime.stackTraceDepth=10
--- a/modules/mysql/src/it/resources/logback.xml
+++ b/modules/mysql/src/it/resources/logback.xml
@@ -1,0 +1,24 @@
+<configuration>
+    <!-- Test configuration, log to console so we can get the docker logs -->
+    <appender name="CONSOLE" class="ch.qos.logback.core.ConsoleAppender">
+        <encoder class="ch.qos.logback.classic.encoder.PatternLayoutEncoder">
+            <pattern>%d [test] %-5p | \(%logger{4}:%line\) | %msg %n</pattern>
+        </encoder>
+    </appender>
+
+    <logger name="org.flywaydb" level="DEBUG">
+        <appender-ref ref="CONSOLE"/>
+    </logger>
+
+    <logger name="org.flywaydb.core.internal.dbsupport.SqlScript" level="DEBUG">
+        <appender-ref ref="CONSOLE"/>
+    </logger>
+
+    <logger name="org.flywaydb.core.internal.command.DbMigrate" level="DEBUG">
+        <appender-ref ref="CONSOLE"/>
+    </logger>
+
+    <root level="INFO">
+        <appender-ref ref="CONSOLE"/>
+    </root>
+</configuration>
--- a/modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlIntegrationSpec.scala
+++ b/modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlIntegrationSpec.scala
@@ -20,6 +20,7 @@ import com.typesafe.config.{Config, ConfigFactory}
 import vinyldns.core.domain.batch.BatchChangeRepository
 import vinyldns.core.domain.zone.{ZoneChangeRepository, ZoneRepository}
 import vinyldns.core.crypto.NoOpCrypto
+import vinyldns.core.domain.record.RecordSetRepository
 import vinyldns.core.repository.{DataStore, DataStoreConfig, RepositoryName}
 
 trait MySqlIntegrationSpec {
@@ -36,6 +37,8 @@ trait MySqlIntegrationSpec {
     instance.get[ZoneRepository](RepositoryName.zone).get
   lazy val zoneChangeRepository: ZoneChangeRepository =
     instance.get[ZoneChangeRepository](RepositoryName.zoneChange).get
+  lazy val recordSetRepository: RecordSetRepository =
+    instance.get[RecordSetRepository](RepositoryName.recordSet).get
 }
 
 object TestMySqlInstance extends MySqlIntegrationSpec {
--- a/modules/mysql/src/main/resources/db/migration/V3.2__RecordSet.sql
+++ b/modules/mysql/src/main/resources/db/migration/V3.2__RecordSet.sql
@@ -6,8 +6,10 @@ CREATE TABLE recordset (
   id CHAR(36) NOT NULL,
   zone_id CHAR(36) NOT NULL,
   name VARCHAR(256) NOT NULL,
+  fqdn VARCHAR(256) NOT NULL,
   type TINYINT NOT NULL,
   data BLOB NOT NULL,
   PRIMARY KEY (id),
-  INDEX zone_id_name_index (zone_id, name, type)
+  INDEX zone_id_name_index (zone_id, name, type),
+  INDEX fqdn_index (fqdn, type)
 );
--- a/modules/mysql/src/main/resources/db/migration/V3.2__RecordSet.sql
+++ b/modules/mysql/src/main/resources/db/migration/V3.2__RecordSet.sql
@@ -1,0 +1,13 @@
+CREATE SCHEMA IF NOT EXISTS ${dbName};
+
+USE ${dbName};
+
+CREATE TABLE recordset (
+  id CHAR(36) NOT NULL,
+  zone_id CHAR(36) NOT NULL,
+  name VARCHAR(256) NOT NULL,
+  type TINYINT NOT NULL,
+  data BLOB NOT NULL,
+  PRIMARY KEY (id),
+  INDEX zone_id_name_index (zone_id, name, type)
+);
--- a/modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
+++ b/modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
@@ -73,8 +73,13 @@ class MySqlRecordSetRepository extends RecordSetRepository with Monitored {
     *
     * This method takes a sequence of records, and appropriately builds the batch prepared statement
     * using underlying JDBC things
+    *
+    * TODO: get clarification that we need to do this, instead of using scalikejdbc
     */
   private def insert(records: Seq[InsertRecord], conn: Connection): Seq[Int] = {
+    // Important!  We must do INSERT IGNORE here as we cannot do ON DUPLICATE KEY UPDATE
+    // with this mysql bulk insert.  To maintain idempotency, we must handle the possibility
+    // of the same insert happening multiple times
     val ps = conn.prepareStatement(
       "INSERT IGNORE INTO recordset (id, zone_id, name, type, data) VALUES (?, ?, ?, ?, ?)")
     records.foreach { r =>
@@ -281,6 +286,7 @@ object MySqlRecordSetRepository extends ProtobufConversions {
 
   def toRecordSet(rs: WrappedResultSet): RecordSet =
     fromPB(VinylDNSProto.RecordSet.parseFrom(rs.bytes(1)))
+  
   def toRecordType(i: Int): Either[InvalidRecordType, RecordType] =
     recordTypeLookup.get(i).map(t => Right(t)).getOrElse(Left(InvalidRecordType(i)))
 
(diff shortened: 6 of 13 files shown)


build.sbt
```diff
@@ -55,7 +55,6 @@ lazy val sharedSettings = Seq(
   scalacOptions in (Compile, doc) += "-no-link-warnings",
   // Use wart remover to eliminate code badness
   wartremoverErrors ++= Seq(
-    Wart.ArrayEquals,
```
pauljamescleary: ugh, kept complaining when I put a `Array[Byte]` in a `case class`

modules/mysql/src/it/resources/application.conf
```diff
@@ -7,8 +7,8 @@ mysql {
     # assumes a docker or mysql instance running locally
     name = "vinyldns"
     driver = "org.mariadb.jdbc.Driver"
-    migration-url = "jdbc:mariadb://localhost:19004/?user=root&password=pass"
-    url = "jdbc:mariadb://localhost:19004/vinyldns?user=root&password=pass"
+    migration-url = "jdbc:mariadb://localhost:19004/?user=root&password=pass&rewriteBatchedStatements=true"
+    url = "jdbc:mariadb://localhost:19004/vinyldns?user=root&password=pass&rewriteBatchedStatements=true"
```
pauljamescleary: this `rewriteBatchedStatements` gets _mad_ performance boosts...

https://www.journaldev.com/2494/jdbc-batch-insert-update-mysql-oracle

modules/mysql/src/it/resources/application.conf
```diff
@@ -20,5 +20,15 @@ mysql {
...
 }
+
+scalikejdbc.global.loggingSQLAndTime.enabled=true
+scalikejdbc.global.loggingSQLAndTime.logLevel=error
+scalikejdbc.global.loggingSQLAndTime.warningEnabled=true
+scalikejdbc.global.loggingSQLAndTime.warningThresholdMillis=1000
+scalikejdbc.global.loggingSQLAndTime.warningLogLevel=warn
+scalikejdbc.global.loggingSQLAndTime.singleLineMode=false
+scalikejdbc.global.loggingSQLAndTime.printUnprocessedStackTrace=false
+scalikejdbc.global.loggingSQLAndTime.stackTraceDepth=10
```
pauljamescleary: these are actually useful for inspecting the SQL.  Can delete them too.

modules/mysql/src/main/resources/db/migration/V3.2__RecordSet.sql
```diff
@@ -0,0 +1,13 @@
...
+USE ${dbName};
+
+CREATE TABLE recordset (
+  id CHAR(36) NOT NULL,
+  zone_id CHAR(36) NOT NULL,
+  name VARCHAR(256) NOT NULL,
+  type TINYINT NOT NULL,
+  data BLOB NOT NULL,
+  PRIMARY KEY (id),
+  INDEX zone_id_name_index (zone_id, name, type)
```
pauljamescleary: gotta rethink the indexes.

Also, I would like to add an `fqdn` field as well, assuming we have a reliable way to build the FQDN from the record set.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
```diff
@@ -0,0 +1,289 @@
...
+      "INSERT IGNORE INTO recordset (id, zone_id, name, type, data) VALUES (?, ?, ?, ?, ?)")
+    records.foreach { r =>
+      ps.setString(1, r.id)
+      ps.setString(2, r.zoneId)
+      ps.setString(3, r.recordName)
+      ps.setInt(4, r.recordType)
+      ps.setBlob(5, new MariaDbBlob(r.data))
+      ps.addBatch()
+    }
+    ps.executeBatch().toSeq
```
pauljamescleary: I am still looking into if scalikejdbc is doing the right thing on `batch`.

When I looked at the SQL scalikejdbc was logging, it didn't look like it was.

Would prefer to use scalikejdbc for consistency purposes, but may not have a coice.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
```diff
@@ -0,0 +1,289 @@
...
+            println(r)
+          }
+          updates.grouped(1000).foreach { group =>
+            UPDATE_RECORDSET.batch(group: _*).apply()
+          }
+          deletes.grouped(1000).foreach { group =>
+            DELETE_RECORDSET.batch(group: _*).apply()
+          }
+        }
+      }.as(changeSet)
```
pauljamescleary: pretty sure these all have to be the same transaction.  This is one of those items to tune with benchmarking, as we have to see how this works with 10MM inserts.

Except for the initial load of a zone, for most syncs (even for large zones), the number of changes is going to be relatively small

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
```diff
@@ -0,0 +1,289 @@
...
+          val opts = (startFrom.as("AND name > {name}") ++
+            recordNameFilter.as("AND name LIKE '%{filter}'") ++
+            Some(" ORDER BY name ASC ") ++
+            maxItems.as("LIMIT {maxItems}")).toList.mkString(" ")
+
+          val params = (Some('zoneId -> zoneId) ++
+            startFrom.map(n => 'name -> n) ++
+            recordNameFilter.map(f => 'filter -> f) ++
+            maxItems.map(m => 'maxItems -> m)).toSeq
+
```
pauljamescleary: gotta dynamically build the query:

* if startFrom is not there, we assume the first page of results 
* if maxItems is not there, we return all records in the zone (for zone syncs)
* if recordNameFilter is not there, we do not apply a filter

Some big concerns with very large zones, as loading 4MM records will eat up a ton of memory.  Benchmarking will flush that problem out a bit.  We probably need to rebuild the zone syncing to use FS2 in order to enforce constant memory usage and just "pull" record sets from our database.  Hopefully, we do not need to do that right now.

modules/mysql/src/it/resources/application.conf
pauljamescleary: Also a good read, as we can increase the max packet size and get more throughput per write.  Maybe valuable to have this be a configuration setting...

There is also another important thing to take into account when dealing with batch inserts on MySQL. The property ‘max_allowed_packet’ defines the maximum size of a single network packet so, this parameter will limit the number of inserts included in a batch. This parameter is defined on the MySQL server side.

```
It is important to notice that this limit is more restrictive than the Batch size configured in Virtual DataPort for the data source. This way, the number of inserts per batch will be the minimum between the batch insert size defined in VDP and the value of the property ‘max_allowed_packet’ defined in MySQL.
```

https://community.denodo.com/kb/view/document/MySQL%20configuration%20for%20batch%20inserts?category=Performance+%26+Optimization

modules/mysql/src/main/resources/db/migration/V3.2__RecordSet.sql
rebstar6: we do, we need to pull the zones to do so though. it can be determined by the combo of record name and zone name, but were not currently storing the zone name on the record

also of note, we likely have a bunch of abandoned records from our func tests (this may hurt our stats. sad.). assuming this is done in some kinda migration, any records with a zoneId that no longer corresponds to an existing zone can be deleted.

modules/mysql/src/it/resources/logback.xml
```diff
@@ -0,0 +1,24 @@
...
+    </logger>
+
+    <logger name="org.flywaydb.core.internal.command.DbMigrate" level="DEBUG">
+        <appender-ref ref="CONSOLE"/>
+    </logger>
+
+    <root level="INFO">
+        <appender-ref ref="CONSOLE"/>
+    </root>
+</configuration>
```
pauljamescleary: will roll this back to ERROR before merging.

modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlIntegrationSpec.scala
```diff
@@ -36,6 +37,8 @@ trait MySqlIntegrationSpec {
     instance.get[ZoneRepository](RepositoryName.zone).get
   lazy val zoneChangeRepository: ZoneChangeRepository =
     instance.get[ZoneChangeRepository](RepositoryName.zoneChange).get
+  lazy val recordSetRepository: RecordSetRepository =
+    instance.get[RecordSetRepository](RepositoryName.recordSet).get
```
pauljamescleary: adds the record set repo to our test instance

modules/mysql/src/main/resources/db/migration/V3.2__RecordSet.sql
pauljamescleary: The fqdn maybe tricky for reverse zones.  I found some PTR stuff in the batch conversions.  I think a module like `FQDN` where we can generate the FQDN across record types would be valuable.

@rebstar6 perhaps a separate PR for that?  We can always add the column later (not like it is being used anytime soon) when we add that FQDN module

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
```diff
@@ -0,0 +1,302 @@
...
+  private val COUNT_RECORDSETS_IN_ZONE =
+    sql"""
+         |SELECT count(*)
+         |  FROM recordset
+         | WHERE zone_id = {zoneId}
+    """.stripMargin
+
+  //private val INSERT_RECORDSET =
+  //  sql"INSERT INTO recordset (id, zone_id, name, type, data) VALUES (?, ?, ?, ?, ?)"
```
pauljamescleary: will remove this comment.  Still haven't heard back from the scalikejdbc folks.  May need to inspect on my own if they are doing the right thing or not.

modules/mysql/src/main/resources/db/migration/V3.2__RecordSet.sql
pauljamescleary: IP6, you have some functions up there.  Actually, maybe in `DnsConversions`...

```scala
def getIPv4NonDelegatedZoneName(ip: String): Option[String] = getIPv4FullReverseName(ip).map {
    reverseName =>
      reverseName.split('.').drop(1).mkString(".") + "."
  }

  def getIPv6FullReverseName(ip: String): Option[String] =
    Try(DNS.ReverseMap.fromAddress(ip, DNS.Address.IPv6)).toOption.map(_.toString)

  def recordDnsName(recordName: String, zoneName: String): DNS.Name =
    if (omitTrailingDot(recordName) == omitTrailingDot(zoneName))
      zoneDnsName(zoneName)
    else
      DNS.Name.fromString(recordName, zoneDnsName(zoneName))

  def zoneDnsName(origin: String): DNS.Name =
    if (origin.endsWith("."))
      DNS.Name.fromString(origin)
    else
      DNS.Name.fromString(s"$origin.")
```

modules/mysql/src/main/resources/db/migration/V3.2__RecordSet.sql
```diff
@@ -0,0 +1,13 @@
+CREATE SCHEMA IF NOT EXISTS ${dbName};
+
+USE ${dbName};
+
+CREATE TABLE recordset (
+  id CHAR(36) NOT NULL,
+  zone_id CHAR(36) NOT NULL,
+  name VARCHAR(256) NOT NULL,
+  type TINYINT NOT NULL,
```
mitruly: `TINYINT` is one byte (8 bits with MySQL). Arguably could use fewer bits (eg. `BIT` type, possible `ENUM` literals), but this should be fine.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
```diff
@@ -0,0 +1,297 @@
...
+import vinyldns.core.protobuf.ProtobufConversions
+import vinyldns.core.route.Monitored
+import vinyldns.proto.VinylDNSProto
+
+class MySqlRecordSetRepository extends RecordSetRepository with Monitored {
+  import MySqlRecordSetRepository._
+
+  private val FIND_BY_ZONEID_NAME_TYPE =
+    sql"""
+      |SELECT data
```
mitruly: Indent consistency.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
```diff
@@ -0,0 +1,297 @@
...
+        byChangeType.getOrElse(RecordSetChangeType.Delete, Nil).map(d => Seq[Any](d.recordSet.id))
+
+      // Totally don't know if this is right or not, are effectively doing the entire operation in a single transaction
+      // and from what I have read that is the correct way to do it
+      // Most zones won't have many changes on a sync once the zone is in vinyldns
+      // However, for initially loading a zone, we could have millions of inserts.  Need to fine tune
+      // both the INSERT SQL as well as how we are using scalikejdbc
+      IO {
+        DB.localTx { implicit s =>
+          inserts.grouped(1000).foreach { group =>
```
mitruly: Should we set these values as constants in the companion object?

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
```diff
@@ -0,0 +1,297 @@
...
+            i.recordSet.id,
+            i.zoneId,
+            i.recordSet.name,
+            fromRecordType(i.recordSet.typ),
+            toPB(i.recordSet).toByteArray
+          )
+      }
+
+      val updates: Seq[Seq[Any]] = byChangeType.getOrElse(RecordSetChangeType.Update, Nil).map {
+        u =>
```
mitruly: This is using the same info as `InsertRecord`. Might be better to just rename `InsertRecord` and have `updates` also be of type `Seq[InsertRecord]`?

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
```diff
@@ -0,0 +1,297 @@
...
+          Seq[Any](
+            u.zoneId,
+            u.recordSet.name,
+            fromRecordType(u.recordSet.typ),
+            toPB(u.recordSet).toByteArray,
+            u.recordSet.id)
+      }
+
+      // deletes are just the record set id
+      val deletes: Seq[Seq[Any]] =
```
mitruly: Understandable to not have a class for just one String, but shouldn't this be:
```
      val deletes: Seq[Seq[String]] =
        byChangeType.getOrElse(RecordSetChangeType.Delete, Nil).map(d => Seq[String](d.recordSet.id))
```
(rather than using `Seq[Seq[Any]]`)

modules/mysql/src/main/resources/db/migration/V3.2__RecordSet.sql
nimaeskandary: Bit would be too small no? 0 or 1 for many types

mitruly: Sorry, should have elaborated a bit. You can specify the number of bits to use:
```
11.2.4 Bit-Value Type - BIT
The BIT data type is used to store bit values. A type of BIT(M) enables storage of M-bit values. M can range from 1 to 64.

To specify bit values, b'value' notation can be used. value is a binary value written using zeros and ones. For example, b'111' and b'10000000' represent 7 and 128, respectively. See Section 9.1.5, “Bit-Value Literals”.

If you assign a value to a BIT(M) column that is less than M bits long, the value is padded on the left with zeros. For example, assigning a value of b'101' to a BIT(6) column is, in effect, the same as assigning b'000101'.
```

nimaeskandary: Additionally there are many dns types we do not support yet

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
pauljamescleary: I am unsure if we need InsertRecord at all atm.  That goes back to the open question out to scalikejdbc.

The only reason for InsertRecord is that it makes the values being inserted strongly typed (as opposed to Seq[Any])

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
pauljamescleary: It probably can be, will see if scalike is ok with it

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
pauljamescleary: This needs performance testing.  A lot may change as a result of that testing.  

As of now, we are ok with 1000 as it is in the integration test.

I think as part of performance testing, this will wind up as a configuration value.

Not a fan of hard coding it, but also don't want to do unnecesssary work before bench marking

pauljamescleary:
Closing this for the time being, as there maybe further optimizations needed as a result of perf testing.  Reviews would be premature

PR #285, opened 2018-10-11
MySql user repo

### LINKED ISSUE #178: MySQL User Repository

1. Create MySQLUserRepository that implements functionality of UserRepository
1. Create integration tests

nimaeskandary:
Fixes #178 .

Changes in this pull request:
- Implements MySql version of UserRepository
- Adds user protobuf


modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlUserRepository.scala
```diff
@@ -0,0 +1,153 @@
...
+         | SELECT data
+         |  FROM user
+         |  WHERE user_name = ?
+       """.stripMargin
+
+  private final def getUsersSqlBuilder(ids: Set[String]) =
+    sql"""
+         | SELECT data
+         |  FROM user
+         |  WHERE id IN (${ids})
```
mitruly: `{}` not needed for `ids`.

mitruly:
Looks good, just one comment.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlUserRepository.scala
```diff
@@ -0,0 +1,153 @@
...
+       """.stripMargin
+
+  private final def getUsersSqlBuilder(ids: Set[String]) =
+    sql"""
+         | SELECT data
+         |  FROM user
+         |  WHERE id IN ($ids)
+       """.stripMargin
+
+  override def getUser(userId: String): IO[Option[User]] =
```
pauljamescleary: don't use `override` for `trait` that have no implementations.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlUserRepository.scala
```diff
@@ -0,0 +1,153 @@
...
+        }
+      }
+    }
+
+  /*
+   * exclusiveStartKey and pageSize were originally made to batch the search in the dynamodb implementation
+   */
+  override def getUsers(
+      userIds: Set[String],
+      exclusiveStartKey: Option[String],
```
pauljamescleary: this is ugly.  Can we rename `exclusiveStartKey` to `startFrom` and `pageSize` to `maxItems` just like everywhere else?  I think it would be a small refactor doable in this PR

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlUserRepository.scala
```diff
@@ -0,0 +1,153 @@
...
+              'data -> toPB(user, crypto).toByteArray
+            )
+            .update()
+            .apply()
+        }
+        user
+      }
+    }
+
+  private def extractUser(colIndex: Int): WrappedResultSet => User = res => {
```
pauljamescleary: name this `toUser`

modules/core/src/main/scala/vinyldns/core/protobuf/ProtobufConversions.scala
```diff
@@ -334,4 +336,36 @@ trait ProtobufConversions {
...
+      data.getIsSuper,
+      LockStatus.withName(data.getLockStatus)
+    )
+
+  def toPB(user: User, crypto: CryptoAlgebra): VinylDNSProto.User = {
+    val builder = VinylDNSProto.User
+      .newBuilder()
+      .setUserName(user.userName)
+      .setAccessKey(user.accessKey)
+      .setSecretKey(crypto.encrypt(user.secretKey))
```
pauljamescleary: Let's not encrypt in the PB.  This should be pure serialization, and here we leaked some app details into the mix.

We should encrypt in the repository before calling `toPB`

modules/core/src/main/protobuf/VinylDNSProto.proto
```diff
@@ -179,3 +179,16 @@ message SingleChange {
...
+message User {
+    required string userName = 1;
+    required string accessKey = 2;
+    required string secretKey = 3;
+    optional string firstName = 4;
+    optional string lastName = 5;
+    optional string email = 6;
+    required int64 created = 7;
+    required string id = 8;
+    required bool isSuper = 9;
```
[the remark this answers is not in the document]
nimaeskandary: @pauljamescleary even though the field is required?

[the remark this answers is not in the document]
pauljamescleary: I was thinking for migration purposes in case those fields are not set (like dynamodb), but forgot we do not use PB for those so we should be ok just having them required.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlUserRepository.scala
```diff
@@ -0,0 +1,153 @@
...
+    sql"""
+         | SELECT data
+         |  FROM user
+         |  WHERE id IN ($ids)
+       """.stripMargin
+
+  def getUser(userId: String): IO[Option[User]] =
+    monitor("repo.User.getUser") {
+      IO {
+        DB.readOnly { implicit s =>
```
nimaeskandary: missed a log statement here

modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlUserRepositoryIntegrationSpec.scala
```diff
@@ -0,0 +1,136 @@
...
+
+    "returns None when user does not exist" in {
+      repo.getUserByName("no-existo").unsafeRunSync() shouldBe None
+    }
+  }
+
+  "MySqlUserRepository.getUsers" should {
+    "omits all non existing users" in {
+      val result = repo.getUsers(Set("no-existo", users.head.id), None, None).unsafeRunSync()
+      result.users.map(_.id) shouldBe List(testUserIds.head)
```
nimaeskandary: don't have to map ids here

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlUserRepository.scala
```diff
@@ -0,0 +1,153 @@
...
+            .map(toUser(1))
+            .first()
+            .apply()
+        }
+      }
+    }
+
+  /*
+   * startFrom and maxItems were originally made to batch the search in the dynamodb implementation,
+   * this function is not used directly as an API route
```
nimaeskandary: comment still a little vague

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlUserRepository.scala
```diff
@@ -0,0 +1,154 @@
...
+         | SELECT data
+         |  FROM user
+         |  WHERE user_name = ?
+       """.stripMargin
+
+  private final def getUsersSqlBuilder(ids: Set[String]) =
+    sql"""
+         | SELECT data
+         |  FROM user
+         |  WHERE id IN ($ids)
```
pauljamescleary: my only comment is on how this SQL is built and ultimately passed.

Does this create a prepared statement that is cacheable?

For example, "SELECT data FROM user WHERE id IN ("foo", "bar")" is **not** cacheable.

However, "SELECT data FROM user WHERE id IN (?, ?)" **is** cacheable when you bind the params to the statement.

That is why we have this pattern of "laying down the question marks" in certain queries, so some of those statements can be cached.

If this is indeed cacheable, this would be a simpler pattern than our custom sql with question marks thingie. 

Can you look into this?

nimaeskandary: Interesting, will do

PR #287, opened 2018-10-11
(WIP) MySql Record Change Repository

### LINKED ISSUE #176: MySQL Record Change Repository

1. Create MySQLRecordChangeRepository that implements functionality of DynamoDBRecordChangeRepository

pauljamescleary:
Fixes #176

### DIFF
--- a/modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlRecordChangeRepositoryIntegrationSpec.scala
+++ b/modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlRecordChangeRepositoryIntegrationSpec.scala
@@ -1,0 +1,86 @@
+package vinyldns.mysql.repository
+
+import java.util.UUID
+
+import cats.scalatest.EitherMatchers
+import org.scalatest.{BeforeAndAfterAll, BeforeAndAfterEach, Matchers, WordSpec}
+import scalikejdbc._
+import vinyldns.core.domain.record.{ChangeSet, RecordChangeRepository, RecordSetChange}
+import vinyldns.core.domain.zone.Zone
+
+class MySqlRecordChangeRepositoryIntegrationSpec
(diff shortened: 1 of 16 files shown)


modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlRecordChangeRepositoryIntegrationSpec.scala
```diff
@@ -0,0 +1,86 @@
...
+    repo = TestMySqlInstance.recordChangeRepository
+  }
+
+  override protected def beforeEach(): Unit = {
+    DB.localTx { implicit s =>
+      s.executeUpdate("DELETE FROM record_change")
+      ()
+    }
+  }
+
```
nimaeskandary: May also want to override afterAll and do a delete statement as well

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordChangeRepository.scala
```diff
@@ -0,0 +1,152 @@
...
+                .bind(zoneId, start.toLong, maxItems)
+                .map(toRecordSetChange)
+                .list()
+                .apply()
+            case None =>
+              LIST_CHANGES_NO_START.bind(zoneId, maxItems).map(toRecordSetChange).list().apply()
+          }
+
+          val nextId =
+            if (changes.size < maxItems) None
```
nimaeskandary: this is done differently in the batch change and zone change repo, in those two we make nextId None only if the table is exhausted. We increment given maxItems by 1 for the query, `take` the requested maxItems for the return list, and have some logic to determine if the table was exhausted since we queried for one more than was asked. Not sure which version is better over all, may warrant some discussion

pauljamescleary:
Closing this for the time being, as there maybe further optimizations needed as a result of perf testing.  Reviews would be premature

PR #293, opened 2018-10-16
Adding shutdown to DataStoreProvider

### LINKED ISSUE #262: Add a shutdown to the `DataStore`
**Motivation**
We do not attempt to "cleanup" when the api and portal are terminated normally.  This can leave open connections to the backing database.

**Scope of change**
* Add a method `def shutdown(): IO[Unit]` to the `DataStore` type
* Update the `MySQLDataStore` to implement the shutdown, it should call `DBs.closeAll()` to close all connection pools
* Add `def shutdown(): IO[Unit] = IO.unit` to the `DynamoDBDataStore` that does nothing
* Add a unit test for the DynamoDBDataStore, simple test to make sure it returns unit
* Update all integration tests for `MySQLRepository`, add an `afterAll` and make sure to call `TestMySQLInstance.instance.shutdown().unsafeRunSync()`

varsharun:
Fixes #262 

Changes in this pull request:
- Add `shutdown` method to `DataStoreProvider`
- Implement `shutdown` in the following
                   - `DynamoDBDataStoreProvider`
                   - `MySQLDataStoreProvider`
                   - `MockDataStoreProvider`
- Add unit test in `DynamoDBDataStoreProviderSpec`
- Modify integration test in `MySqlZoneChangeRepositoryIntegrationSpec` to shutdown in `afterAll()`


modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlDataStoreProvider.scala
```diff
@@ -126,4 +126,10 @@ class MySqlDataStoreProvider extends DataStoreProvider {
     logger.info("database init complete")
   }
 
+  override def shutdown(): IO[Unit] = IO(DBs.closeAll()).attempt.map {
```
pauljamescleary: No need to catch the `UnsupportedOperationException`, simply log any `Left` and swallow it.

Perhaps just

```scala
`IO(DBs.closeAll()).handleError(e => logger.info(...)).as(())
```

pauljamescleary: Also, we don't use the `override` keyword unless actually overriding a base class member, here we are implementing the method on the interface that is abstract

pauljamescleary: Unit testing this is going to be hard due to the nature of `DBs.closeAll()`, so without gymnastics will be hard to test the failure case.

Adding a success case in the `MySqlDataStoreProviderSpec` should be sufficient (that it succeeds as unit)

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlDataStoreProvider.scala
```diff
@@ -126,4 +127,7 @@ class MySqlDataStoreProvider extends DataStoreProvider {
     logger.info("database init complete")
   }
 
+  def shutdown(): IO[Unit] =
+    IO(DBs.closeAll())
+      .handleError(_ => logger.info("exception occurred while shutting down"))
```
pauljamescleary: Would recommend logging the actual error as well if possible.  logger.error is fine, this should only be called during shut down of the instance, not repeatedly

modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlZoneRepositoryIntegrationSpec.scala
```diff
@@ -45,6 +45,11 @@ class MySqlZoneRepositoryIntegrationSpec
       s.executeUpdate("DELETE FROM zone")
     }
 
+  override protected def afterAll(): Unit = {
+    super.afterAll()
```
pauljamescleary: don't need the super.afterAll here

modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlBatchChangeRepositoryIntegrationSpec.scala
```diff
@@ -119,6 +119,11 @@ class MySqlBatchChangeRepositoryIntegrationSpec
       s.executeUpdate("DELETE FROM single_change")
     }
 
+  override protected def afterAll(): Unit = {
+    super.afterAll()
```
pauljamescleary: don't need the super.afterAll here

modules/mysql/src/it/scala/vinyldns/mysql/queue/MySqlMessageQueueIntegrationSpec.scala
```diff
@@ -73,6 +73,11 @@ class MySqlMessageQueueIntegrationSpec extends WordSpec with Matchers
     ()
   }
 
+  override protected def afterAll(): Unit = {
+    super.afterAll()
```
pauljamescleary: don't need the super.afterAll here

modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlIntegrationSpec.scala
```diff
@@ -36,6 +36,8 @@ trait MySqlIntegrationSpec {
     instance.get[ZoneRepository](RepositoryName.zone).get
   lazy val zoneChangeRepository: ZoneChangeRepository =
     instance.get[ZoneChangeRepository](RepositoryName.zoneChange).get
+
+  def shutdown(): Unit = new MySqlDataStoreProvider().shutdown().unsafeRunSync()
```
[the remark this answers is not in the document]
varsharun: Thanks Rebecca. Yes this is the reason int tests were failing. I will fix this.

modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlZoneChangeRepositoryIntegrationSpec.scala
```diff
@@ -88,6 +88,7 @@ class MySqlZoneChangeRepositoryIntegrationSpec
       s.executeUpdate("DELETE FROM zone_change")
     }
     super.afterAll()
+    TestMySqlInstance.shutdown()
```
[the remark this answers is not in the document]
varsharun: This is used in afterAll() so isn't it after the entire suite runs?

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlDataStoreProvider.scala
```diff
@@ -126,4 +127,7 @@ class MySqlDataStoreProvider extends DataStoreProvider {
     logger.info("database init complete")
   }
 
+  def shutdown(): IO[Unit] =
+    IO(DBs.closeAll())
```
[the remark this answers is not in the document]
varsharun: Will make this change thank you for pointing it out

modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlIntegrationTestSuiteSpec.scala
```diff
@@ -0,0 +1,46 @@
...
+
+class MySqlIntegrationTestSuiteSpec
+    extends Suite
+    with MySqlIntegrationSpec
+    with BeforeAndAfterAll {
+
+  override protected def beforeAll(): Unit = {
+    // load the data store
+    TestMySqlInstance.instance
+    ()
```
mitruly: Can we move this to the up with previous line (ie. `TestMySqlInstance.instance()`)?

modules/mysql/src/it/scala/vinyldns/mysql/queue/MySqlMessageQueueIntegrationSpec.scala
```diff
@@ -46,6 +45,7 @@ final case class InvalidMessage(command: ZoneCommand) extends CommandMessage {
   def id: MessageId = MessageId(command.id)
 }
 
+@DoNotDiscover
 class MySqlMessageQueueIntegrationSpec extends WordSpec with Matchers
   with BeforeAndAfterEach with EitherMatchers with EitherValues with BeforeAndAfterAll with ProtobufConversions {
```
mitruly: No longer need to extend `BeforeAndAfterAll` here since we're removing the `beforeAll()`.

modules/mysql/src/it/scala/vinyldns/mysql/queue/MySqlMessageQueueIntegrationSpec.scala
```diff
@@ -67,12 +67,6 @@ class MySqlMessageQueueIntegrationSpec extends WordSpec with Matchers
   }
   override protected def beforeEach(): Unit = clear()
 
-  override protected def beforeAll(): Unit = {
-    // load the data store
-    TestMySqlInstance.instance
-    ()
```
varsharun: @mitruly I moved this over to the other class () was to return a Unit i think.

mitruly: Oh, I initially thought `()` was part of the function, but it passed locally with just `TestMySqlInstance.instance`. I'm okay with it if no one else has any strong feelings.

modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlIntegrationTestSuiteSpec.scala
```diff
@@ -0,0 +1,46 @@
+/*
+ * Copyright 2018 Comcast Cable Communications Management, LLC
```
rebstar6: can you move this file and MySqlIntegrationSpec up a level since theyre shared (to `modules/mysql/src/it/scala/vinyldns/mysql/`)?

modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlIntegrationTestSuiteSpec.scala
rebstar6: this is to return unit :P

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlDataStoreProvider.scala
```diff
@@ -126,4 +127,7 @@ class MySqlDataStoreProvider extends DataStoreProvider {
     logger.info("database init complete")
   }
 
+  def shutdown(): IO[Unit] =
+    IO(DBs.close())
```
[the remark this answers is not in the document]
rebstar6: you can move back to closeAll(), thatwill have the same issue though

and yes, but `def close(): Unit = throw new UnsupportedOperationException` is not actually doing anything :P.

Messing around with it now, i think having something like:
 ```
class HikariCloser(dataSource: HikariDataSource) extends DataSourceCloser {

  override def close(): Unit = dataSource.close()

}
```

then when you make the singleton, do 
```
    ConnectionPool.singleton(
      new DataSourceConnectionPool(dataSource, closer = new HikariCloser(dataSource)))
```
*should* mean its actually calling the close we want

File: modules/core/src/main/scala/vinyldns/core/repository/DataStoreLoader.scala [hunk 14]
```diff
@@ -27,19 +27,28 @@ import scala.reflect.ClassTag
 
 object DataStoreLoader {
 
+  class DataStoreInfo(
```
rebstar6: add:
```
   class DataStoreInfo(
      val dataStoreConfig: DataStoreConfig,
      val dataStore: DataStore,
      val dataStoreProvider: DataStoreProvider) {
val accessorTuple = (dataStoreConfig , dataStore)
}
```
and then you can just use that later on (rather than the map to tuple thing)

modules/core/src/test/scala/vinyldns/core/repository/DataStoreLoaderSpec.scala
```diff
@@ -105,7 +111,7 @@ class DataStoreLoaderSpec
         allDisabledReposConfig.copy(user = enabled))
 
       val loadCall = DataStoreLoader.loadAll(List(config1, config2), crypto, TestAccessorProvider)
-      loadCall.unsafeRunSync() shouldBe a[TestDataAccessor]
+      loadCall.unsafeRunSync() shouldBe a[LoaderResponse[_]]
```
rebstar6: can you do
`shouldBe a[LoaderResponse[TestDataAccessor]]`
here?

modules/mysql/src/it/scala/vinyldns/mysql/queue/MySqlMessageQueueIntegrationSpec.scala
```diff
@@ -46,8 +45,9 @@ final case class InvalidMessage(command: ZoneCommand) extends CommandMessage {
   def id: MessageId = MessageId(command.id)
 }
 
+@DoNotDiscover
```
pauljamescleary: what is this?

modules/mysql/src/it/scala/vinyldns/mysql/MySqlIntegrationTestSuiteSpec.scala
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+package vinyldns.mysql
+
+import com.typesafe.config.Config
+import org.scalatest.{BeforeAndAfterAll, Suite}
+import vinyldns.mysql.queue.MySqlMessageQueueIntegrationSpec
+import vinyldns.mysql.repository._
+
+class MySqlIntegrationTestSuiteSpec
```
pauljamescleary: unclear on why this is needed

modules/mysql/src/it/scala/vinyldns/mysql/MySqlIntegrationTestSuiteSpec.scala
```diff
@@ -0,0 +1,47 @@
...
+    with BeforeAndAfterAll {
+
+  override protected def beforeAll(): Unit = {
+    // load the data store
+    TestMySqlInstance.instance
+    ()
+  }
+
+  override def afterAll(): Unit = {
+    shutdown()
```
pauljamescleary: we really do not need to call `shutdown` for integration specs.  `fork in IntegrationTest := true` is sufficient for local testing, as it forks a child JVM and cleans it up when all integration tests are finished.  This ensures that we don't have straggling `HikariCP` instances holding onto connections to the docker mysql container.

modules/mysql/src/it/scala/vinyldns/mysql/MySqlIntegrationTestSuiteSpec.scala
rebstar6: we need a global beforeAll/afterAll, not just a beforeAll/afterAll that runs before/after a single file. If we shutdown the database connections after one test, the next will fail

modules/mysql/src/it/scala/vinyldns/mysql/MySqlIntegrationTestSuiteSpec.scala
rebstar6: its not ensuring that though, the connections keep around

rebstar6: oh, lol, we have `fork in IntegrationTest := false`...is there a reason for that?

rebstar6: oof, this was silly of me...a good alternative honestly is probably to just add `minIdle: Int` and `idleTimeout: Long` to MySqlDataStoreSettings, and then set `min-idle=0, pool-max-size = 5, idle-timeout=10000 (10000 is the min allowable)` in the integration spec config. Right now we have `pool-max-size=20`, and no min idle, so its keeping around 20 idle connections after each run.

with ^^, we still have lingering something, but no connections. You do still see a series of logs like:
```
[mysqldbPool housekeeper] DEBUG com.zaxxer.hikari.pool.HikariPool - mysqldbPool2 - After cleanup stats (total=0, active=0, idle=0, waiting=0)
```
but thats better than right now, which is:
```
[mysqldbPool housekeeper] DEBUG com.zaxxer.hikari.pool.HikariPool - mysqldbPool2 - After cleanup stats (total=20, active=0, idle=20, waiting=0)
```

pauljamescleary: we should fork in IntegrationTest := true, this is what I have in the new MySql repos I have been working on and it works swell.

We definitely can tune the idles as well.  But so far, just forking cleans everything up swimmingly.

modules/core/src/main/scala/vinyldns/core/repository/DataStoreLoader.scala
```diff
@@ -27,19 +27,30 @@ import scala.reflect.ClassTag
...
+      val dataStoreConfig: DataStoreConfig,
+      val dataStore: DataStore,
+      val dataStoreProvider: DataStoreProvider) {
+    val accessorTuple: (DataStoreConfig, DataStore) = (dataStoreConfig, dataStore)
+  }
+
+  class LoaderResponse[A](val accessor: A, shutdownHook: => List[IO[Unit]]) {
```
rebstar6: `Loader` is generic, can we just rename this to `DataStoreLoaderResponse`?

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlDataStoreProvider.scala
```diff
@@ -126,4 +130,11 @@ class MySqlDataStoreProvider extends DataStoreProvider {
     logger.info("database init complete")
   }
 
+  def shutdown(): IO[Unit] =
+    IO(DBs.closeAll())
+      .handleError(e => logger.error(s"exception=${e.getMessage} occurred while shutting down"))
```
pauljamescleary: One small change, do `logger.error(msg, e)` because we get the full stack trace.

PR #294, opened 2018-10-18
move logging dependency to core

rebstar6:
I believe this should get rid of the `Failed to load class "org.slf4j.impl.StaticLoggerBinder".` errors

For some reason the base sl4j import is still in apiDependencies, and other modules still seem able to import it. Mystery to me at the moment.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -16,13 +16,12 @@ object Dependencies {
   lazy val awsV = "1.11.423"
   lazy val jaxbV = "2.3.0"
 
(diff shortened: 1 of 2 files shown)


project/Dependencies.scala
```diff
@@ -16,13 +16,12 @@ object Dependencies {
   lazy val awsV = "1.11.423"
   lazy val jaxbV = "2.3.0"
 
-  lazy val compileDependencies = Seq(
+  lazy val apiDependencies = Seq(
```
mitruly: Subtle, good change.

PR #296, opened 2018-10-18
WIP Mysql bench

pauljamescleary:
Create Benchmark

Only need to look at the last commit.

**Benchmark Config**
Allows you to configure how many zones of each size to create as part of the benchmark.  For example `xs = 5` will create 5 "extra small zones", each zone having 10 records.

* `xs` - 10 records
* `s` - 100 records
* `m` - 1000 records
* `l` - 10,000 records
* `xl` - 100,000 records
* `xxl` - 1,000,000 records

**Bulk Loading**
Bulk loading is done lazily, as we aim to create some 2MM zones, 250MM record sets, and 250MM record set changes.  Generating all of those up front would blow the heap.

Further, "guessing" at the correct size could be similarly disastrous.  Creating more than the backend can handle will blow the heap.  Creating too little, and we are underutilizing the database.

The perfect solution here is to use FS2.  FS2 will only generate zones and records as the bulk loading process has capacity.  This limits memory while maximizing the capacity of our database.

The process of bulk loading follows:

1. Create a `Stream[ZoneGenerator]`, this will create new `ZoneGenerators` on demand. 
1. Each `ZoneGenerator` gets a `Zone`, as well as its _own_ `Stream[RecordSetChange]`.
1. Each `Stream[RecordSetChange]` will generate new record sets and record set changes _on demand_ for that zone until it exhausts the record count for that `ZoneSize` (xs, s, m, l, etc.)

Once we have our "Stream of Streams", we can then process them.  When reviewing the `benchmarkFlow`, think of it as though 1 `ZoneGenerator` is processed at a time, follow top down...

1. Save the zone to the `ZoneRepository`
1. Take the `recordStream` from the `ZoneGenerator`, and generate a "Stream of Streams".  Each Stream containing exactly 10,000 record set changes (Note: this number is arbitrary, we want it to be low enough so we do not blow the heap, but large enough to maximize our backend.  Brute force testing showed this to have good performance)
1. Take each substream of 10,000 and save the record set and record set changes to their repositories
1. We run the "Stream of Streams" in parallel using `parJoinUnbounded`.  For large zones of 1MM records, we want to run those sub-streams as fast as possible.
1. At the end of the stream, we do a _second_ `parJoinUnbounded`, why?  Because we want to run multiple _zones_ in parallel as well.  If we are loading 1MM xs smalls, we cannot have those go sequentially.

**Naming**
We have a `NameGenerator` that _predictably_ names all record sets.  This is **critical** to how the query tests work.  Those tests expect certain record names to be present in each zone.  For example, we guarantee that there will be a record set named `r1-fresh-sofa` with type `AAAA` in every zone.  Without this, our record set query tests will not run.

**Query Benchmarking**
We are only testing RecordSet queries right now.  More can easily be added later for other repositories.

We start the testing by loading one `TestZone` for each `ZoneSize` (xs, s, m, l, etc.) . For each `TestZone`, we execute each query from the `RecordSetRepository`.  Along the way, we record the latency for each query/zone size combination.

To ensure we have enough samples, we run the entire suite of tests 100 times before finishing.

**Output**
All latency is captured in dropwizard metrics.  Output goes to elastic search, so we can query, stats, and graph each run.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -301,6 +301,27 @@ lazy val mysql = (project in file("modules/mysql"))
     fork in IntegrationTest := true,
   ).dependsOn(core % "compile->compile;test->test")
 
+lazy val benchmark = (project in file("modules/benchmark"))
+  .enablePlugins(DockerComposePlugin, AutomateHeaderPlugin)
+  .configs(IntegrationTest)
+  .settings(sharedSettings)
+  .settings(headerSettings(IntegrationTest))
+  .settings(inConfig(IntegrationTest)(scalafmtConfigSettings))
+  .settings(corePublishSettings)
+  .settings(testSettings)
+  .settings(Defaults.itSettings)
+  .settings(scalaStyleCompile ++ scalaStyleTest)
+  .settings(
+    name := "benchmark",
+    fork in IntegrationTest := true,
+    libraryDependencies ++= Seq(
+      "co.fs2" %% "fs2-core" % "1.0.0",
+      "org.typelevel" %% "cats-effect" % "1.0.0",
+      "org.typelevel" %% "cats-core" % "1.4.0",
+    ),
+    fork in run := true
+  ).dependsOn(core, mysql)
+
 val preparePortal = TaskKey[Unit]("preparePortal", "Runs NPM to prepare portal for start")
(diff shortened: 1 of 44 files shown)


pauljamescleary:
Going to close this as it is rather unnecessary to review right now.  Will be posting up "real" PRs soon enough.

Issue #202, opened 2018-09-12
Create dynamic loading for message queue based on configuration


pauljamescleary:
**Background**
As part of BYOQ, users must be able to _configure_ their message queue implementation.  We need to have a dynamic loader similar to BYOR and Crypto that will bootstrap the appropriate message queue implementation.

**Requirements**

1. The message queue should follow the data store loading to allow the message queue to initialize itself.
1. If possible, the message queue should be deleted as part of message queue initialization.  **Note: SQS provides the ability to create a queue**
1. Create actual SQS Message Queue Provider implementation

**Design**
The following configuration represents what the config might look like for sqs...

```scala
message-queue {
  type = "vinyldns.sqs.MessageQueueProvider"
  settings {
    access-key = ${AWS_ACCESS_KEY}
    secret-key = ${AWS_SECRET_ACCESS_KEY}
    signing-region = ${SQS_REGION}
    service-endpoint = ${SQS_ENDPOINT}
    queue-url = ${SQS_QUEUE_URL} 
    create-if-not-exists = false
  }
}
```

PR #300, opened 2018-10-22
SQS Message Queue Provider

### LINKED ISSUE #290: SQS Message Queue Provider
Implement SQS queue provider, ensuring the following:
- [ ] If queue does not exist, create it
- [ ] Enforce type for queue URL
- [ ] If queue URL is not passed in, use queue name

mitruly:
Resolves #290.

Changes in this pull request:
- Implement `SqsMessageQueueProvider` implementation of `MessageQueueProvider`
- Slight refactor to config settings (`queue-url` → `queue-name`) and `SqsMessageQueue.apply`

### DIFF
--- a/modules/sqs/src/it/scala/vinyldns/sqs/queue/SqsMessageQueueIntegrationSpec.scala
+++ b/modules/sqs/src/it/scala/vinyldns/sqs/queue/SqsMessageQueueIntegrationSpec.scala
@@ -40,13 +40,9 @@ class SqsMessageQueueIntegrationSpec extends WordSpec
   private val sqsMessageQueueSettings: MessageQueueConfig =
     pureconfig.loadConfigOrThrow[MessageQueueConfig](ConfigFactory.load().getConfig("sqs"))
 
-  private val sqsConfig = pureconfig.loadConfigOrThrow[SqsMessageQueueSettings](
-    sqsMessageQueueSettings.settings)
-
   private val provider = new SqsMessageQueueProvider()
-  private val client = provider.setupClient(sqsConfig).unsafeRunSync()
-  private val queueUrl = provider.setupQueue(client, sqsConfig.queueName).unsafeRunSync()
-  private val queue: SqsMessageQueue = new SqsMessageQueue(queueUrl, client)
+  private val queue: SqsMessageQueue =
+    provider.load(sqsMessageQueueSettings).map(_.asInstanceOf[SqsMessageQueue]).unsafeRunSync()
 
   override protected def afterEach(): Unit = {
     // Remove items from queue after each test
--- a/modules/sqs/src/it/scala/vinyldns/sqs/queue/SqsMessageQueueIntegrationSpec.scala
+++ b/modules/sqs/src/it/scala/vinyldns/sqs/queue/SqsMessageQueueIntegrationSpec.scala
@@ -43,9 +43,10 @@ class SqsMessageQueueIntegrationSpec extends WordSpec
   private val sqsConfig = pureconfig.loadConfigOrThrow[SqsMessageQueueSettings](
     sqsMessageQueueSettings.settings)
 
-  private val queueName = SqsMessageQueueProvider.DEFAULT_QUEUE_NAME
-
-  private val queue: SqsMessageQueue = SqsMessageQueue(sqsConfig, queueName)
+  private val provider = new SqsMessageQueueProvider()
+  private val client = provider.setupClient(sqsConfig).unsafeRunSync()
+  private val queueUrl = provider.setupQueue(client, sqsConfig.queueName).unsafeRunSync()
+  private val queue: SqsMessageQueue = SqsMessageQueue(queueUrl, client)
 
   // Re-create queue before tests
   override protected def beforeAll(): Unit = {
@@ -194,7 +195,6 @@ class SqsMessageQueueIntegrationSpec extends WordSpec
     "throw an error if there are issues parsing the message" in {
       val message = new Message()
 
-      val queue = SqsMessageQueue(sqsConfig, queueName)
       assertThrows[AmazonSQSException] {
         queue.parse(message).unsafeRunSync()
       }
--- a/modules/sqs/src/it/scala/vinyldns/sqs/queue/SqsMessageQueueIntegrationSpec.scala
+++ b/modules/sqs/src/it/scala/vinyldns/sqs/queue/SqsMessageQueueIntegrationSpec.scala
@@ -21,13 +21,14 @@ import java.util.concurrent.TimeUnit.SECONDS
 import cats.data.NonEmptyList
 import cats.scalatest.EitherMatchers
 import com.amazonaws.services.sqs.model._
+import com.typesafe.config.ConfigFactory
 import org.scalatest._
 import org.scalatest.mockito.MockitoSugar
 import vinyldns.core.TestRecordSetData._
 import vinyldns.core.TestZoneData._
 import vinyldns.core.domain.record.RecordSetChange
 import vinyldns.core.protobuf.ProtobufConversions
-import vinyldns.core.queue.{MessageCount, MessageId}
+import vinyldns.core.queue.{MessageCount, MessageId, MessageQueueConfig}
 import vinyldns.sqs.queue.SqsMessageQueue.InvalidMessageTimeout
 
 import scala.concurrent.duration.FiniteDuration
@@ -36,7 +37,15 @@ class SqsMessageQueueIntegrationSpec extends WordSpec
   with MockitoSugar with BeforeAndAfterAll with BeforeAndAfterEach with Matchers with EitherMatchers with EitherValues
   with ProtobufConversions {
 
-  private val queue: SqsMessageQueue = SqsMessageQueue()
+  private val sqsMessageQueueSettings: MessageQueueConfig =
+    pureconfig.loadConfigOrThrow[MessageQueueConfig](ConfigFactory.load().getConfig("sqs"))
+
+  private val sqsConfig = pureconfig.loadConfigOrThrow[SqsMessageQueueSettings](
+    sqsMessageQueueSettings.settings)
+
+  private val queueName = SqsMessageQueueProvider.DEFAULT_QUEUE_NAME
+
+  private val queue: SqsMessageQueue = SqsMessageQueue(sqsConfig, queueName)
 
   // Re-create queue before tests
   override protected def beforeAll(): Unit = {
@@ -185,7 +194,7 @@ class SqsMessageQueueIntegrationSpec extends WordSpec
     "throw an error if there are issues parsing the message" in {
       val message = new Message()
 
-      val queue = SqsMessageQueue()
+      val queue = SqsMessageQueue(sqsConfig, queueName)
       assertThrows[AmazonSQSException] {
         queue.parse(message).unsafeRunSync()
       }
--- a/modules/sqs/src/it/scala/vinyldns/sqs/queue/SqsMessageQueueIntegrationSpec.scala
+++ b/modules/sqs/src/it/scala/vinyldns/sqs/queue/SqsMessageQueueIntegrationSpec.scala
@@ -46,12 +46,7 @@ class SqsMessageQueueIntegrationSpec extends WordSpec
   private val provider = new SqsMessageQueueProvider()
   private val client = provider.setupClient(sqsConfig).unsafeRunSync()
   private val queueUrl = provider.setupQueue(client, sqsConfig.queueName).unsafeRunSync()
-  private val queue: SqsMessageQueue = SqsMessageQueue(queueUrl, client)
-
-  // Re-create queue before tests
-  override protected def beforeAll(): Unit = {
-    queue.client.createQueue("sqs")
-  }
+  private val queue: SqsMessageQueue = new SqsMessageQueue(queueUrl, client)
 
   override protected def afterEach(): Unit = {
     // Remove items from queue after each test
--- a/modules/sqs/src/it/scala/vinyldns/sqs/queue/SqsMessageQueueProviderIntegrationSpec.scala
+++ b/modules/sqs/src/it/scala/vinyldns/sqs/queue/SqsMessageQueueProviderIntegrationSpec.scala
@@ -17,12 +17,12 @@
 package vinyldns.sqs.queue
 import com.typesafe.config.{Config, ConfigFactory}
 import org.scalatest.{Matchers, WordSpec}
-import vinyldns.core.queue.MessageQueueConfig
+import vinyldns.core.queue.{MessageQueueConfig, MessageQueueLoader}
 
 class SqsMessageQueueProviderIntegrationSpec extends WordSpec with Matchers {
   val sqsConfig: Config = ConfigFactory.load().getConfig("sqs")
 
-  import SqsMessageQueueProvider._
+  val undertest = new SqsMessageQueueProvider()
 
   "load" should {
     "fail if a required setting is not provided" in {
@@ -40,7 +40,7 @@ class SqsMessageQueueProviderIntegrationSpec extends WordSpec with Matchers {
 
       val badSettings = pureconfig.loadConfigOrThrow[MessageQueueConfig](badConfig)
 
-      a[pureconfig.error.ConfigReaderException[MessageQueueConfig]] should be thrownBy load(badSettings)
+      a[pureconfig.error.ConfigReaderException[MessageQueueConfig]] should be thrownBy undertest.load(badSettings)
         .unsafeRunSync()
     }
 
@@ -59,12 +59,34 @@ class SqsMessageQueueProviderIntegrationSpec extends WordSpec with Matchers {
           |    """.stripMargin)
 
       val messageConfig = pureconfig.loadConfigOrThrow[MessageQueueConfig](nonExistentQueueConfig)
-      val messageQueue = load(messageConfig).unsafeRunSync()
+      val messageQueue = undertest.load(messageConfig).unsafeRunSync()
 
       noException should be thrownBy messageQueue
         .asInstanceOf[SqsMessageQueue]
         .client
         .getQueueUrl("new-queue")
+    }
+  }
+
+  "MessageQueueLoader" should {
+    "invoke SQS provider properly" in {
+      val nonExistentQueueConfig =
+        ConfigFactory.parseString("""
+          |    class-name = "vinyldns.sqs.queue.SqsMessageQueueProvider"
+          |
+          |    settings {
+          |      access-key = "x"
+          |      secret-key = "x"
+          |      signing-region = "x"
+          |      service-endpoint = "http://localhost:19005/"
+          |      queue-name = "new-queue"
+          |    }
+          |    """.stripMargin)
+
+      val messageConfig = pureconfig.loadConfigOrThrow[MessageQueueConfig](nonExistentQueueConfig)
+      val queue = MessageQueueLoader.load(messageConfig).unsafeRunSync()
+
+      queue shouldBe a[SqsMessageQueue]
     }
   }
 }
--- a/modules/sqs/src/it/scala/vinyldns/sqs/queue/SqsMessageQueueProviderIntegrationSpec.scala
+++ b/modules/sqs/src/it/scala/vinyldns/sqs/queue/SqsMessageQueueProviderIntegrationSpec.scala
@@ -1,0 +1,70 @@
+/*
+ * Copyright 2018 Comcast Cable Communications Management, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package vinyldns.sqs.queue
+import com.typesafe.config.{Config, ConfigFactory}
+import org.scalatest.{Matchers, WordSpec}
+import vinyldns.core.queue.MessageQueueConfig
+
+class SqsMessageQueueProviderIntegrationSpec extends WordSpec with Matchers {
+  val sqsConfig: Config = ConfigFactory.load().getConfig("sqs")
+
+  import SqsMessageQueueProvider._
+
+  "load" should {
+    "fail if a required setting is not provided" in {
+      val badConfig =
+        ConfigFactory.parseString("""
+          |    class-name = "vinyldns.sqs.queue.SqsMessageQueueProvider"
+          |
+          |    settings {
+          |      access-key = "x"
+          |      signing-region = "x"
+          |      service-endpoint = "http://localhost:19005/"
+          |      queue-name = "queue-name"
+          |    }
+          |    """.stripMargin)
+
+      val badSettings = pureconfig.loadConfigOrThrow[MessageQueueConfig](badConfig)
+
+      a[pureconfig.error.ConfigReaderException[MessageQueueConfig]] should be thrownBy load(badSettings)
+        .unsafeRunSync()
+    }
+
+    "create the queue if the queue is non-existent" in {
+      val nonExistentQueueConfig =
+        ConfigFactory.parseString("""
+          |    class-name = "vinyldns.sqs.queue.SqsMessageQueueProvider"
+          |
+          |    settings {
+          |      access-key = "x"
+          |      secret-key = "x"
+          |      signing-region = "x"
+          |      service-endpoint = "http://localhost:19005/"
+          |      queue-name = "new-queue"
+          |    }
+          |    """.stripMargin)
+
+      val messageConfig = pureconfig.loadConfigOrThrow[MessageQueueConfig](nonExistentQueueConfig)
+      val messageQueue = load(messageConfig).unsafeRunSync()
+
+      noException should be thrownBy messageQueue
+        .asInstanceOf[SqsMessageQueue]
+        .client
+        .getQueueUrl("new-queue")
+    }
+  }
+}
--- a/modules/sqs/src/it/scala/vinyldns/sqs/queue/SqsMessageQueueProviderIntegrationSpec.scala
+++ b/modules/sqs/src/it/scala/vinyldns/sqs/queue/SqsMessageQueueProviderIntegrationSpec.scala
@@ -83,6 +83,24 @@ class SqsMessageQueueProviderIntegrationSpec extends WordSpec with Matchers {
       val messageConfig = pureconfig.loadConfigOrThrow[MessageQueueConfig](invalidQueueNameConfig)
       assertThrows[InvalidQueueName](undertest.load(messageConfig).unsafeRunSync())
     }
+
+    "fail with InvalidQueueName if a FIFO queue is specified" in {
+      val fifoQueueName =
+        ConfigFactory.parseString("""
+          |    class-name = "vinyldns.sqs.queue.SqsMessageQueueProvider"
+          |
+          |    settings {
+          |      access-key = "x"
+          |      secret-key = "x"
+          |      signing-region = "x"
+          |      service-endpoint = "http://localhost:19005/"
+          |      queue-name = "queue.fifo"
+          |    }
+          |    """.stripMargin)
+
+      val messageConfig = pureconfig.loadConfigOrThrow[MessageQueueConfig](fifoQueueName)
+      assertThrows[InvalidQueueName](undertest.load(messageConfig).unsafeRunSync())
+    }
   }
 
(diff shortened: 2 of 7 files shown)


modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueue.scala
```diff
@@ -192,11 +187,23 @@ object SqsMessageQueue extends ProtobufConversions {
...
+          sqsMessageQueueSettings.signingRegion))
+        .withCredentials(
+          new AWSStaticCredentialsProvider(
+            new BasicAWSCredentials(
+              sqsMessageQueueSettings.accessKey,
+              sqsMessageQueueSettings.secretKey)))
         .build()
 
+    // Create queue if it doesn't exist
+    val queueUrl = try {
```
mitruly: Would like to know of a more functional way to do this, if possible.

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueueProvider.scala
```diff
@@ -0,0 +1,74 @@
...
+
+package vinyldns.sqs.queue
+import cats.effect.IO
+import org.slf4j.LoggerFactory
+import pureconfig.module.catseffect.loadConfigF
+import vinyldns.core.queue.{MessageQueue, MessageQueueConfig, MessageQueueProvider}
+
+import scala.util.matching.Regex
+
+object SqsMessageQueueProvider extends MessageQueueProvider {
```
mitruly: Figured that `object` made more sense since there aren't any instance-specific attributes that would benefit from this being a `class`.

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueueProvider.scala
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  def validateQueueName(queueName: String): Either[SqsMessageQueueProviderError, String] = {
+
+    /** Queue name restrictions:
+        - Up to 80 characters (alphanumeric, hyphens and underscores)
+        - Name of a FIFO queue must end with ".fifo" suffix
+
+      see: https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-limits.html#limits-queues */
+    val validQueueNameRegex: Regex = """^([\w\-]{1,80})$|^([\w\-]{1,75}\.fifo)$""".r
```
mitruly: Due to the info about queue limits (see: https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-limits.html#limits-queues) and the fact that the queue URL can be derived by knowing the name, it made more sense to me to use `queue-name` over `queue-url`.

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueueProvider.scala
rebstar6: look at message queue loader:
```
 def load(config: MessageQueueConfig): IO[MessageQueue] =
    for {
      _ <- IO(logger.error(s"Attempting to load queue ${config.className}"))
      provider <- IO(
        Class
          .forName(config.className)
          .getDeclaredConstructor()
          .newInstance()
          .asInstanceOf[MessageQueueProvider])
      queue <- provider.load(config)
    } yield queue
```

not sure if Class.forName will find an object. Need to check that

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueueSettings.scala
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+package vinyldns.sqs.queue
+
+case class SqsMessageQueueSettings(
+    accessKey: String,
+    secretKey: String,
+    serviceEndpoint: String,
+    signingRegion: String,
+    queueName: Option[String])
```
rebstar6: i think if you just give queueName a default, if it doesnt exist it will store as the default. then you can remove that defaulting logic from the provider.

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueueProvider.scala
```diff
@@ -0,0 +1,74 @@
...
+import org.slf4j.LoggerFactory
+import pureconfig.module.catseffect.loadConfigF
+import vinyldns.core.queue.{MessageQueue, MessageQueueConfig, MessageQueueProvider}
+
+import scala.util.matching.Regex
+
+object SqsMessageQueueProvider extends MessageQueueProvider {
+  final val DEFAULT_QUEUE_NAME = "sqs"
+
+  sealed abstract class SqsMessageQueueProviderError(message: String) extends Throwable(message)
```
rebstar6: in this and the other error, if there is only one error extending this, making a sealed abstract class feels like overkill.

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueue.scala
```diff
@@ -192,11 +187,23 @@ object SqsMessageQueue extends ProtobufConversions {
...
+      client.getQueueUrl(queueName).getQueueUrl
+    } catch {
+      case _: QueueDoesNotExistException => client.createQueue(queueName).getQueueUrl
```
rebstar6: createQueue throws, which means this should live inside something (IO likely)

I would probably move this apply logic into the provider within an IO. or have apply return an IO directly

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueueProvider.scala
```diff
@@ -0,0 +1,74 @@
...
+    for {
+      settingsConfig <- loadConfigF[IO, SqsMessageQueueSettings](config.settings)
+      queueName <- IO.fromEither(getQueueName(settingsConfig))
+      queue <- setupQueue(settingsConfig, queueName)
+    } yield queue
+
+  def getQueueName(sqsMessageQueueSettings: SqsMessageQueueSettings)
+    : Either[SqsMessageQueueProviderError, String] = {
+    logger.info(s"Retrieving queue name...")
+    sqsMessageQueueSettings.queueName match {
```
pauljamescleary: Why do we assume a default queue name?

A default value should be provided via `reference.conf`, not hard-coded.

pauljamescleary: Right, I would make the queue name required, then you can get rid of the `Option` stuff, makes the `SqsMessageQueue.apply` simpler as well, as you only need to pass in the settings.

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueue.scala
pauljamescleary: This should whole function should return `IO`, as the `client.getQueueUrl` could fail, as could `client.createQueue` for any number of reasons.

```scala
IO(client.getQueueUrl(n).getQueueUrl)
  .recoverWith {
    case _: QueueDoesNotExistException => client.createQueue(queueName).getQueueUrl 
  }
```

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueueSettings.scala
pauljamescleary: I don't even think we need a default.  You gotta configure it.  If it isn;t there, blow up.

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueueProvider.scala
mitruly: Good call @rebstar6 - just tested this and you are right.

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueue.scala
```diff
@@ -173,32 +168,8 @@ object SqsMessageQueue extends ProtobufConversions {
...
-              new ExponentialBackoffStrategy(2, 64), // Base delay and max back-off delay of 64
-              100, // Max error retry count (set to dead-letter count); default is 3
-              true
-            )))
-        .withEndpointConfiguration(new EndpointConfiguration(serviceEndpoint, signingRegion))
-        .withCredentials(new AWSStaticCredentialsProvider(
-          new BasicAWSCredentials(accessKey, secretKey)))
-        .build()
-
+  def apply(queueUrl: String, client: AmazonSQSAsync): SqsMessageQueue =
```
rebstar6: do we need this apply? doesnt hurt, was just wondering

modules/sqs/src/it/scala/vinyldns/sqs/queue/SqsMessageQueueIntegrationSpec.scala
```diff
@@ -36,7 +37,16 @@ class SqsMessageQueueIntegrationSpec extends WordSpec
...
+  private val sqsConfig = pureconfig.loadConfigOrThrow[SqsMessageQueueSettings](
+    sqsMessageQueueSettings.settings)
+
+  private val provider = new SqsMessageQueueProvider()
+  private val client = provider.setupClient(sqsConfig).unsafeRunSync()
+  private val queueUrl = provider.setupQueue(client, sqsConfig.queueName).unsafeRunSync()
+  private val queue: SqsMessageQueue = SqsMessageQueue(queueUrl, client)
 
   // Re-create queue before tests
   override protected def beforeAll(): Unit = {
```
rebstar6: you no longer need this beforeAll, it should be created by the provider

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueueProvider.scala
```diff
@@ -0,0 +1,98 @@
...
+        .build()
+    }
+
+  def setupQueue(client: AmazonSQSAsync, queueName: String): IO[String] = {
+    logger.info(s"Setting up queue...")
+    // Create queue if it doesn't exist
+    IO {
+      client.getQueueUrl(queueName).getQueueUrl
+    }.attempt.unsafeRunSync() match {
+      case Left(_: QueueDoesNotExistException) => IO.pure(client.createQueue(queueName).getQueueUrl)
```
rebstar6: createQueue isnt pure

rebstar6: also we shouldnt be running anything here

mitruly: Is there a way to use `recoverWith`? `.recoverWith` wasn't resolving when I tried to do:
```
IO(client.getQueueUrl(queueName).getQueueUrl)
  .recoverWith {
    case _: QueueDoesNotExistException => client.createQueue(queueName).getQueueUrl 
  }
```
Also tried to use `IO.fromEither(...)` but that didn't seem to resolve `recover` either.

rebstar6: i *think* the attempt part is ok though:

```
    IO {
      client.getQueueUrl(queueName).getQueueUrl
    }.attempt.flatMap {
      case Right(queueUrl) => IO.pure(queueUrl)
      case Left(_: QueueDoesNotExistException) => IO(client.createQueue(queueName).getQueueUrl)
      case Left(e) => IO.raiseError(e)
    }
```

rebstar6: import cats.implicits._

mitruly: Thanks, @rebstar6. That sneaky `cats.implicits._`...

modules/sqs/src/it/scala/vinyldns/sqs/queue/SqsMessageQueueIntegrationSpec.scala
```diff
@@ -36,12 +37,16 @@ class SqsMessageQueueIntegrationSpec extends WordSpec
...
 
-  private val queue: SqsMessageQueue = SqsMessageQueue()
+  private val sqsMessageQueueSettings: MessageQueueConfig =
+    pureconfig.loadConfigOrThrow[MessageQueueConfig](ConfigFactory.load().getConfig("sqs"))
 
-  // Re-create queue before tests
-  override protected def beforeAll(): Unit = {
-    queue.client.createQueue("sqs")
-  }
+  private val sqsConfig = pureconfig.loadConfigOrThrow[SqsMessageQueueSettings](
```
rebstar6: load with the loader here - 
```
  private val provider = new SqsMessageQueueProvider()
  private val queue: SqsMessageQueue =
    provider.load(sqsMessageQueueSettings).map(_.asInstanceOf[SqsMessageQueue]).unsafeRunSync()
```

then you can get rid of some of the other vals

modules/sqs/src/it/scala/vinyldns/sqs/queue/SqsMessageQueueProviderIntegrationSpec.scala
```diff
@@ -0,0 +1,92 @@
...
+ * limitations under the License.
+ */
+
+package vinyldns.sqs.queue
+import com.typesafe.config.{Config, ConfigFactory}
+import org.scalatest.{Matchers, WordSpec}
+import vinyldns.core.queue.{MessageQueueConfig, MessageQueueLoader}
+
+class SqsMessageQueueProviderIntegrationSpec extends WordSpec with Matchers {
+  val sqsConfig: Config = ConfigFactory.load().getConfig("sqs")
```
rebstar6: unused config

modules/sqs/src/it/scala/vinyldns/sqs/queue/SqsMessageQueueProviderIntegrationSpec.scala
```diff
@@ -0,0 +1,92 @@
...
+
+  val undertest = new SqsMessageQueueProvider()
+
+  "load" should {
+    "fail if a required setting is not provided" in {
```
rebstar6: add a "fail if an invalid queue name is given" test here

modules/sqs/src/it/scala/vinyldns/sqs/queue/SqsMessageQueueProviderIntegrationSpec.scala
mitruly: Good catch. IntelliJ doesn't catch certain unused vars 😞

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueueProvider.scala
```diff
@@ -0,0 +1,97 @@
...
+      .getOrElse(Left(InvalidQueueName(queueName)))
+  }
+
+  def setupClient(sqsMessageQueueSettings: SqsMessageQueueSettings): IO[AmazonSQSAsync] =
+    IO {
+      AmazonSQSAsyncClientBuilder
+        .standard()
+        .withClientConfiguration(
+          new ClientConfiguration()
+            .withRetryPolicy(new RetryPolicy(
```
rebstar6: i know this is copy paste - wanna validate thought that this retry policy is working as we think. If we are setting message timeout explicitly, it seems a little odd to have this logic in here, as I would think it would go unused, no?

rebstar6: mm looked at docs, i think these are separate concerns

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueueProvider.scala
```diff
@@ -0,0 +1,97 @@
...
+        .build()
+    }
+
+  def setupQueue(client: AmazonSQSAsync, queueName: String): IO[String] = {
+    logger.info(s"Setting up queue...")
+    // Create queue if it doesn't exist
+    IO {
+      client.getQueueUrl(queueName).getQueueUrl
+    }.recoverWith {
+      case _: QueueDoesNotExistException => IO(client.createQueue(queueName).getQueueUrl)
```
rebstar6: note on the regex - .fifo means a fifo queue was created, but this is specifically creating a standard queue.

probably fine, i guess fifo people will have to use this to connect, but we may wanna either exclude .fifo as an option, or if it gets to this point, make it 
`case _: QueueDoesNotExistException if ! queueName.endsWith(".fifo")`
since this create wont work for them

modules/sqs/src/test/resources/application.conf
```diff
@@ -0,0 +1,9 @@
+sqs {
+  class-name = "vinyldns.sqs.queue.SqsMessageQueueProvider"
+  settings = {
+    access-key = "x"
+    secret-key = "x"
+    signing-region = "x"
+    service-endpoint = "endpoint"
```
rebstar6: no queue-name? is this config even used?

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueueProvider.scala
```diff
@@ -0,0 +1,97 @@
...
+    }
+  }
+}
+
+object SqsMessageQueueProvider {
+  final case class InvalidQueueName(queueName: String)
+      extends Throwable(
+        s"Invalid queue name: $queueName. Must be 1-80 alphanumeric, hyphen or underscore characters.")
+
+  private val logger = LoggerFactory.getLogger("SqsMessageQueueProvider")
```
rebstar6: `LoggerFactory.getLogger[classOf[SqsMessageQueueProvider]]` here i think

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueueProvider.scala
```diff
@@ -0,0 +1,98 @@
...
+
+  def setupClient(sqsMessageQueueSettings: SqsMessageQueueSettings): IO[AmazonSQSAsync] =
+    IO {
+      AmazonSQSAsyncClientBuilder
+        .standard()
+        .withClientConfiguration(
+          new ClientConfiguration()
+            .withRetryPolicy(new RetryPolicy(
+              RetryPolicy.RetryCondition.NO_RETRY_CONDITION,
+              new ExponentialBackoffStrategy(2, 64), // Base delay and max back-off delay of 64
```
pauljamescleary: Can you add a unit test to verify these settings?

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueueProvider.scala
```diff
@@ -0,0 +1,98 @@
...
+              100, // Max error retry count (set to dead-letter count); default is 3
```
pauljamescleary: does this _have_ to align with the settings on the SQS message queue?

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueueProvider.scala
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+  def setupQueue(client: AmazonSQSAsync, queueName: String): IO[String] = {
+    logger.info(s"Setting up queue...")
+    // Create queue if it doesn't exist
+    IO {
+      client.getQueueUrl(queueName).getQueueUrl
+    }.recoverWith {
+      case _: QueueDoesNotExistException if !queueName.endsWith(".fifo") =>
+        IO(client.createQueue(queueName).getQueueUrl)
```
pauljamescleary: this seems to be a branch untested.  `endsWith.fifo`.  So if it ends with `.fifo` we don't create it?  Is there a reason why?

mitruly: Queue names ending with `.fifo` are for FIFO queues, which I think are standard queues but do exactly-once processing. I think it might be easier to just exclude FIFO queues from valid names instead of throwing an `InvalidParameterValue`.

pauljamescleary: Yea, that is better for sure.   Exclude it in the `validateQueueName`, and then remove the `if` guard in the pattern match here.

rebstar6: that works too. @pauljamescleary this was based on a comment i made earlier - the create for a fifo queue doesnt look like that. in theory, if you have pre-created a fifo, and are just connecting to it here, this is fine, but we shouldnt be creating new fifo (or that should be a separate case here).

removing .fifo entirely works too.

rebstar6:
good with this PR minus the fifo change 👍

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueueProvider.scala
```diff
@@ -45,35 +45,41 @@ class SqsMessageQueueProvider extends MessageQueueProvider {
...
-    val validQueueNameRegex: Regex = """^([\w\-]{1,80})$|^([\w\-]{1,75}\.fifo)$""".r
+    val validQueueNameRegex: Regex = """^([\w\-]{1,80})$""".r
 
     validQueueNameRegex
       .findFirstIn(queueName)
       .map(Right(_))
       .getOrElse(Left(InvalidQueueName(queueName)))
   }
 
+  def clientConfiguration: ClientConfiguration =
```
rebstar6: why is this pulled out?

mitruly: I think Paul wanted me to test that the client configuration that gets passed in is correct: https://github.com/vinyldns/vinyldns/pull/300/files/f33336f98850228cdf87a4dd5283ca9334d51de5#diff-653ba200d2fab0167265636f9fc7a25c

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueueProvider.scala
pauljamescleary: Can we support FIFO queues at all?  i.e., is there any reason we cannot?

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueueProvider.scala
rebstar6: that feels silly, were testing that AWS is reading the settings we gave it. not our code.

rebstar6: @pauljamescleary what were you wanting for this testing?

pauljamescleary: @rebstar6 Not silly if these settings impacted functionally how the system operates.  For example, someone could change these settings in a future PR and potentially break things in unpredictable ways.

At first, I thought the Retry policy had to do with SQS messages.  In which case, we do indeed want to test it out as it could impact how SQS behaves from a functional standpoint (how many retries a message goes through before dead-lettering).

I went through the AWS code, and it looks like the retry policy only impacts how the client recovers from errors talking to AWS.  In that case, I don't think we need our own `RetryPolicy`.  Use the `DEFAULT_BACKOFF_STRATEGY` in `PredefinedBackoffStrategies` instead here @mitruly

modules/sqs/src/test/scala/vinyldns/sqs/queue/SqsMessageQueueProviderSpec.scala
```diff
@@ -0,0 +1,91 @@
...
+import com.amazonaws.AmazonClientException
+import com.amazonaws.retry.RetryPolicy
+import com.amazonaws.services.sqs.model.ReceiveMessageRequest
+import org.scalatest.{Matchers, WordSpec}
+
+class SqsMessageQueueProviderSpec extends WordSpec with Matchers {
+  val undertest = new SqsMessageQueueProvider()
+  import SqsMessageQueueProvider._
+
+  "clientConfiguration" should {
```
pauljamescleary: @mitruly my apologies, the retry policy has nothing to do with SQS message retries / dead-letter policies.  This does not need to be tested.

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueueProvider.scala
mitruly: If that's the case, I can just use the default client configuration here (ie. remove the passed-in one).

mitruly:
@pauljamescleary I don't know if we can use FIFOs since we are using AmazonSQSAsync:
```
Compatibility
Clients
The Amazon SQS Buffered Asynchronous Client doesn't currently support FIFO queues.
```
(source: https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/FIFO-queues.html#FIFO-compatibility)

pauljamescleary:
I'm ok if we don't support FIFO tbh.

If we make that decision, gotta make sure we update the docs.  Make sure we have a unit test that if we create a queue named ".fifo" it fails too.

mitruly:
Attempting to create the queue throws the following exception:
```
[info] - should invoke SQS provider properly *** FAILED *** (1 second, 368 milliseconds)
[info]   com.amazonaws.services.sqs.model.AmazonSQSException: InvalidParameterValue; see the SQS docs. (Service: AmazonSQS; Status Code: 400; Error Code: InvalidParameterValue; Request ID: 00000000-0000-0000-0000-000000000000)
[info]   at com.amazonaws.http.AmazonHttpClient$RequestExecutor.handleErrorResponse(AmazonHttpClient.java:1660)
[info]   at com.amazonaws.http.AmazonHttpClient$RequestExecutor.executeOneRequest(AmazonHttpClient.java:1324)
[info]   at com.amazonaws.http.AmazonHttpClient$RequestExecutor.executeHelper(AmazonHttpClient.java:1074)
[info]   at com.amazonaws.http.AmazonHttpClient$RequestExecutor.doExecute(AmazonHttpClient.java:745)
[info]   at com.amazonaws.http.AmazonHttpClient$RequestExecutor.executeWithTimer(AmazonHttpClient.java:719)
[info]   at com.amazonaws.http.AmazonHttpClient$RequestExecutor.execute(AmazonHttpClient.java:701)
[info]   at com.amazonaws.http.AmazonHttpClient$RequestExecutor.access$500(AmazonHttpClient.java:669)
[info]   at com.amazonaws.http.AmazonHttpClient$RequestExecutionBuilderImpl.execute(AmazonHttpClient.java:651)
[info]   at com.amazonaws.http.AmazonHttpClient.execute(AmazonHttpClient.java:515)
[info]   at com.amazonaws.services.sqs.AmazonSQSClient.doInvoke(AmazonSQSClient.java:2229)
```

This is also when using attributes specified by https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-create-queue.html#create-queue-java-FIFO

mitruly:
@pauljamescleary Sounds good - I think doc changes are included with #202.

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueueProvider.scala
```diff
@@ -0,0 +1,89 @@
...
+    }.recoverWith {
+      case _: QueueDoesNotExistException => IO(client.createQueue(queueName).getQueueUrl)
+    }
+  }
+}
+
+object SqsMessageQueueProvider {
+  final case class InvalidQueueName(queueName: String)
+      extends Throwable(
+        s"Invalid queue name: $queueName. Must be 1-80 alphanumeric, hyphen or underscore characters.")
```
pauljamescleary: update the message, mention that it could be a fifo queue

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueueProvider.scala
pauljamescleary: I am good with not supporting them

rebstar6: in theory we can, we just need to change our create statement. leave out for now, can deal with it if we need it, wouldnt be hard to add

mitruly: @rebstar6 I tried changing to use a `CreateQueueRequest` and passed in the attribute `Map` as directed in https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-create-queue.html#create-queue-java-FIFO, but got the same error message (since I think `AmazonSQSClientAsync` doesn't support FIFO.

PR #301, opened 2018-10-23
Update to latest cats and fs2

pauljamescleary:
**Motivation**
We are on an old version of both cats, cats-effect, and fs2.  The primary issue is that `parSequence` and other parallel commands don't actually run in parallel.

**Changes**
Running in parallel now requires in some instances a `Timer` to be available in implicit scope, and others `ContextShift`.

Added the necessary implicits.


mitruly:
Would it be better to add:
```
protected implicit val cs: ContextShift[IO] = IO.contextShift(scala.concurrent.ExecutionContext.global)
``` to `DynamoDBIntegrationSpec` instead of the individual DynamoDB integration specs?

pauljamescleary:
@mitruly Not a fan of implicits sitting in base classes.  Implicits tend to be a pain as it is, hiding them can be painful.

PR #304, opened 2018-10-24
remove config defaults

rebstar6:
related to https://github.com/vinyldns/vinyldns/pull/283

one of the mysql fields (connection-timeout-millis, idle-timeout, etc) was excluded from our deploy application.conf, but still being applied because in reference.conf.

took it out, should fall back to hikari defaults, not what we set

also updated docs

### DIFF
--- a/docker/api/docker.conf
+++ b/docker/api/docker.conf
@@ -66,6 +66,7 @@ vinyldns {
       max-lifetime = 600000
       maximum-pool-size = 20
       minimum-idle = 20
+      register-mbeans = true
     }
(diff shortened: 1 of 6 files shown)


docker/api/docker.conf
```diff
@@ -66,6 +66,7 @@ vinyldns {
       max-lifetime = 600000
       maximum-pool-size = 20
       minimum-idle = 20
```
rebstar6: may wanna remove minimum-idle/idle-timeout from anything real (should keep fixed size connection pool). do want them in testing though.

Issue #302, opened 2018-10-23
Remove machine specific creds file from build.sbt


nimaeskandary:
**Motivation**
Currently sonatype settings in build.sbt use this line to get credentials:

```credentials += Credentials(Path.userHome / ".sbt" / "1.0" / "vinyldns-gpg-credentials")```

This could be removed and we could expect creds to be in global settings @ `~/.sbt/1.0/credentials.sbt`

This would reduce sbt warnings on machines that don't have `vinyldns-gpg-credentials`

**Scope of change**
* Remove `credentials += Credentials(Path.userHome / ".sbt" / "1.0" / "vinyldns-gpg-credentials")` from build.sbt
* Update MAINTAINERS.md, in the section that talks about where to put sonatype creds, these should be put in `~/.sbt/1.0/credentials.sbt` now

remerle:
No longer an issue - the code mentioned no longer exists.

PR #306, opened 2018-10-24
Remove credentials setting and update docs

ildac:
related to #302 

Changes in this pull request:
- remove credentials line from build.sbt
- update "Sonatype Credentials" section in MAINTANERS.md documentation


build.sbt
```diff
@@ -253,8 +253,7 @@ lazy val corePublishSettings = Seq(
     Developer(id="mitruly", name="Michael Ly", email="michaeltrulyng@gmail.com", url=url("https://github.com/mitruly")),
     Developer(id="britneywright", name="Britney Wright", email="blw06g@gmail.com", url=url("https://github.com/britneywright")),
   ),
-  sonatypeProfileName := "io.vinyldns",
-  credentials += Credentials(Path.userHome / ".sbt" / "1.0" / "vinyldns-gpg-credentials")
+  sonatypeProfileName := "io.vinyldns"
```
pauljamescleary: should we also remove the `.dockerCredentials`?

pauljamescleary:
👋  @ildac , thanks for helping us out.  Left a question, also maybe @nimaeskandary can answer as well.

build.sbt
ildac: Yes I think that could use the same global setup as the Sonatype credential, I can give it a trt and if the case update the pull request.

pauljamescleary: Thanks for looking into this.  I _think_ ideally we don't want _any_ credentials in the build, and just assume they live in global

nimaeskandary: This global route makes sense to me

Issue #291, opened 2018-10-13
Performance Tuning for MySql


pauljamescleary:
**Background**
Just a tracker issue for doing some performance tuning for mysql.  Looking to capture notes on this card along the way.

As we move all repositories to mysql, it is imperative to understand the operational characteristics of MySQL especially as it pertains to vinyldns.

The critical aspects from a size and performance standpoint are as follows.

**Bulk inserting for record set and record changes**
When a zone is loaded into VinylDNS for the first time, we run a AXFR full zone transfer and store all record sets into the vinyldns system.  We also store a corresponding record change for each record set at the same time.

For most zones, performance is a non-issue.  However, when zones have 100,000s or 1MMs of record sets in them, this operation can be understandably expensive.

**Archiving of record changes**
With DynamoDB, it is relatively cheap to store long-term all data in the record change table.  However, with MySQL, disk space is limited wrt DynamoDB.  We need to have a plan for archiving record changes, potentially configuration driven, with background jobs / processes / stored procedures and triggers that will move data past a certain age into some kind of long-term backup.

This may not be necessary for all vinyldns users, as it depends on data size.

**Query performance for record set and record changes**
VinylDNS installations may have 1MMs of record sets and record changes.  For example, we anticipate Comcast to grow to 100MMs of record sets and 100GB of record changes.  For this size, it is imperative to have an archiving strategy for old record changes.

Indexing and query performance will be critical to the responsiveness of the application.  We have to fine tune indexes and understand query execution plans for record and record change data.

## Notes
**Bulk Insert MySql.**

Bulk insert in MySQL is achieved using sql `INSERT INTO table(a, b, c) VALUES (?, ?, ?), (?, ?, ?), (?, ?, ?)`

The SQL shows that you can keep appending additional rows under the `VALUES` clause seemingly indefinitely.

The number of rows you can insert in a single query is depending on a number of factors, crucially the `max_allowed_packet`

There are a number of ways to do bulk insert using mysql, including file loading.  [High-speed inserts with MySQL](https://medium.com/@benmorel/high-speed-inserts-with-mysql-9d3dcd76f723) covers the different ways to bulk load data.  **Key takeaway: extended inserts achieve 201,000 inserts per second with 26 bytes per row average**

**Proposed recordset table**

```sql
CREATE TABLE recordset (
  id CHAR(36) NOT NULL,
  zone_id CHAR(36) NOT NULL,
  name VARCHAR(256) NOT NULL,
[...]

PR #309, opened 2018-10-25
Add MySqlRecordSetRepository

### LINKED ISSUE #177: MySQL Record Set Repository

pauljamescleary:
Fixes #177

* Updated docker for mysql to use general_log for fun sql debug times
* Made sure to use rewriteBatchStatements to acheive new hights for bulk inserts
* `MySqlDataStoreProvider` support for the record set repo


pauljamescleary:
**INDEX RESULTS**
These are the results of running `EXPLAIN` on sample queries.

- [x] `getByZoneIdNameType` - confirmed hits the index

```
MySQL [vinyldns]> explain select data from recordset where zone_id = "77f4a0b0-286c-454a-b173-fea5d3458a1f" and name = "r532-nippy-baby" and type = 3;
+----+-------------+-----------+------------+------+--------------------+--------------------+---------+-------------------+------+----------+-------+
| id | select_type | table     | partitions | type | possible_keys      | key                | key_len | ref               | rows | filtered | Extra |
+----+-------------+-----------+------------+------+--------------------+--------------------+---------+-------------------+------+----------+-------+
|  1 | SIMPLE      | recordset | NULL       | ref  | zone_id_name_index | zone_id_name_index | 295     | const,const,const |    1 |   100.00 | NULL  |
+----+-------------+-----------+------------+------+--------------------+--------------------+---------+-------------------+------+----------+-------+
1 row in set, 1 warning (0.05 sec)
```

- [x] `getByZoneIdAndName` - hits the index

```
MySQL [vinyldns]> explain select data from recordset where zone_id = "77f4a0b0-286c-454a-b173-fea5d3458a1f" and name = "r532-nippy-baby";
+----+-------------+-----------+------------+------+--------------------+--------------------+---------+-------------+------+----------+-------+
| id | select_type | table     | partitions | type | possible_keys      | key                | key_len | ref         | rows | filtered | Extra |
+----+-------------+-----------+------------+------+--------------------+--------------------+---------+-------------+------+----------+-------+
|  1 | SIMPLE      | recordset | NULL       | ref  | zone_id_name_index | zone_id_name_index | 294     | const,const |    1 |   100.00 | NULL  |
+----+-------------+-----------+------------+------+--------------------+--------------------+---------+-------------+------+----------+-------+
1 row in set, 1 warning (0.00 sec)
```

- [x] `getById` - confirmed hits PRIMARYKEY index

```
MySQL [vinyldns]> explain select data from recordset where id = "0000002a-435d-46f0-a18e-6aa85e71c176";
+----+-------------+-----------+------------+-------+---------------+---------+---------+-------+------+----------+-------+
| id | select_type | table     | partitions | type  | possible_keys | key     | key_len | ref   | rows | filtered | Extra |
+----+-------------+-----------+------------+-------+---------------+---------+---------+-------+------+----------+-------+
|  1 | SIMPLE      | recordset | NULL       | const | PRIMARY       | PRIMARY | 36      | const |    1 |   100.00 | NULL  |
+----+-------------+-----------+------------+-------+---------------+---------+---------+-------+------+----------+-------+
1 row in set, 1 warning (0.00 sec)
```

- [x] `getRecordSetCount` - confirmed hits index

```
MySQL [vinyldns]> explain select count(*) from recordset where zone_id = "77f4a0b0-286c-454a-b173-fea5d3458a1f";
+----+-------------+-----------+------------+------+--------------------+--------------------+---------+-------+------+----------+-------------+
| id | select_type | table     | partitions | type | possible_keys      | key                | key_len | ref   | rows | filtered | Extra       |
+----+-------------+-----------+------------+------+--------------------+--------------------+---------+-------+------+----------+-------------+
|  1 | SIMPLE      | recordset | NULL       | ref  | zone_id_name_index | zone_id_name_index | 36      | const |  100 |   100.00 | Using index |
+----+-------------+-----------+------------+------+--------------------+--------------------+---------+-------+------+----------+-------------+
1 row in set, 1 warning (0.00 sec)
```

- [x] `getByFQDN` - confirmed hits index

```
MySQL [vinyldns]> explain select data from recordset where fqdn IN ("r532-nippy-baby.z5.s.33605.uproot2.she.", "r51-wrong-boat.z5.l.1.floaty2.com.");
+----+-------------+-----------+------------+-------+---------------+------------+---------+------+------+----------+-----------------------+
| id | select_type | table     | partitions | type  | possible_keys | key        | key_len | ref  | rows | filtered | Extra                 |
+----+-------------+-----------+------------+-------+---------------+------------+---------+------+------+----------+-----------------------+
|  1 | SIMPLE      | recordset | NULL       | range | fqdn_index    | fqdn_index | 258     | NULL |    2 |   100.00 | Using index condition |
+----+-------------+-----------+------------+-------+---------------+------------+---------+------+------+----------+-----------------------+
1 row in set, 1 warning (0.00 sec)
```

modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlRecordSetRepositoryIntegrationSpec.scala
```diff
@@ -0,0 +1,247 @@
...
+    with BeforeAndAfterEach
+    with Matchers
+    with Inspectors
+    with OptionValues with EitherMatchers {
+
+  import vinyldns.core.TestRecordSetData._
+  import vinyldns.core.TestZoneData._
+  private val repo = TestMySqlInstance.recordSetRepository.asInstanceOf[MySqlRecordSetRepository]
+
+  override protected def beforeEach(): Unit =
```
nimaeskandary: Think an afterAll that does the same thing would be good as well

pauljamescleary: so as to not leaving straggling data

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
```diff
@@ -0,0 +1,262 @@
...
+          val inClause = " IN (" + fqdns.as("?").mkString(",") + ")"
+          val query = FIND_BY_FQDN + inClause
+          SQL(query).bind(fqdns.map(_.value): _*).map(toRecordSet).list().apply()
+        }
+      }
+    }
+}
+
+object MySqlRecordSetRepository extends ProtobufConversions {
+  final case class InvalidRecordType(value: Int)
```
nimaeskandary: is this used?

pauljamescleary: you tell me ;) I don't know, may not be, I will check

build.sbt
```diff
@@ -299,6 +298,7 @@ lazy val mysql = (project in file("modules/mysql"))
   .settings(scalaStyleCompile ++ scalaStyleTest)
   .settings(
     organization := "io.vinyldns",
+    fork in IntegrationTest := true,
```
rebstar6: this is in `testSettings` already

modules/core/src/main/scala/vinyldns/core/domain/record/FQDN.scala
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package vinyldns.core.domain.record
+
+final case class FQDN private (value: String) extends AnyVal
+object FQDN {
+  // zoneName cannot be apex!
+  def apply(recordName: String, zoneName: String): FQDN = {
```
rebstar6: mmm apex records wont work im thinking?

i think this will store FQDN(test.com.test.com.) for apex in test.com, no?

modules/mysql/src/it/resources/application.conf
```diff
@@ -7,8 +7,8 @@ mysql {
     # assumes a docker or mysql instance running locally
     name = "vinyldns"
     driver = "org.mariadb.jdbc.Driver"
-    migration-url = "jdbc:mariadb://localhost:19004/?user=root&password=pass"
-    url = "jdbc:mariadb://localhost:19004/vinyldns?user=root&password=pass"
+    migration-url = "jdbc:mariadb://localhost:19004/?user=root&password=pass&rewriteBatchedStatements=true"
+    url = "jdbc:mariadb://localhost:19004/vinyldns?user=root&password=pass&rewriteBatchedStatements=true"
```
rebstar6: why do we need this in the url if its in the settings?

modules/core/src/main/scala/vinyldns/core/domain/record/FQDN.scala
pauljamescleary: Yea, that is what the comment was about :) If we have a more full proof way will add.  Maybe "if recordName matches zoneName without trailing dots then just recordName"?

rebstar6: look at relativize in DnsConversions

pauljamescleary: Right, cannot use `DNS` library since this is core.  I have this built though, I Think it is easy enough to do.

rebstar6: yes. id probably put in `object DomainHelpers` (in core)

modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlRecordSetRepositoryIntegrationSpec.scala
```diff
@@ -0,0 +1,252 @@
...
+import vinyldns.core.domain.record.{ChangeSet, FQDN, RecordSetChange}
+import vinyldns.core.domain.zone.Zone
+import vinyldns.mysql.TestMySqlInstance
+
+class MySqlRecordSetRepositoryIntegrationSpec
+  extends WordSpec
+    with BeforeAndAfterEach
+    with BeforeAndAfterAll
+    with Matchers
+    with Inspectors
```
rebstar6: i believe Inspectors and OptionValues are both unused

modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlRecordSetRepositoryIntegrationSpec.scala
```diff
@@ -0,0 +1,252 @@
...
+      results should contain theSameElementsAs expected
+    }
+    "return none when there is no match" in {
+      insert(okZone, 1).map(_.recordSet)
+      val results = repo.getRecordSetsByName(okZone.id, "not-there").unsafeRunSync()
+      results shouldBe empty
+    }
+  }
+  "get by fqdn" should {
+    "return all record sets matching an fqdn" in {
```
rebstar6: add an apex in this list once you have that FQDN thing fixed

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
```diff
@@ -0,0 +1,259 @@
...
+  private val INSERT_RECORDSET =
+    sql"INSERT IGNORE INTO recordset(id, zone_id, name, fqdn, type, data) VALUES (?, ?, ?, ?, ?, ?)"
+
+  private val UPDATE_RECORDSET =
+    sql"UPDATE recordset SET zone_id = ?, name = ?, fqdn = ?, type = ?, data = ? WHERE id = ?"
+
+  private val DELETE_RECORDSET =
+    sql"DELETE FROM recordset WHERE id = ?"
+
+  private val FIND_BY_FQDN =
```
rebstar6: this will be interesting actually - 

right now we are not deleting recordsets in our func tests. those old records are never returned, however, because we only ever get by name with zone ID (and the zone is deleted, so never getting that old ID).

Moving to fqdn is right, but also means a bunch of old data can be returned. Need to consider this if/when we switch to using this

rebstar6: per ^^, may wanna expand abandonZone to clear out all associated recordsets? For that, may wanna add in a strictly zoneId index

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
```diff
@@ -0,0 +1,259 @@
...
+  def apply(changeSet: ChangeSet): IO[ChangeSet] =
+    monitor("repo.RecordSet.apply") {
+      val byChangeType = changeSet.changes.groupBy(_.changeType)
+      val inserts: Seq[Seq[Any]] = byChangeType.getOrElse(RecordSetChangeType.Create, Nil).map {
+        i =>
+          Seq[Any](
+            i.recordSet.id,
+            i.recordSet.zoneId,
+            i.recordSet.name,
+            FQDN(i.recordSet.name, i.zone.name).value,
```
rebstar6: i kinda think we should have FQDN on the recordset itself. i imagine eventually we will want to display that to the user anyway.

that said, backing into it from where we are with dynamo isnt pretty, so this is probably fine

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
```diff
@@ -0,0 +1,259 @@
...
+    RecordType.SSHFP -> 9,
+    RecordType.TXT -> 10,
+    RecordType.UNKNOWN -> unknownRecordType
+  )
+
+  def toRecordSet(rs: WrappedResultSet): RecordSet =
+    fromPB(VinylDNSProto.RecordSet.parseFrom(rs.bytes(1)))
+
+  def fromRecordType(typ: RecordType): Int =
+    recordTypeLookup.getOrElse(typ, unknownRecordType)
```
rebstar6: hmm, theres nothing in here really letting us know if we forget to update the map above when adding a new record type. Not to say we will, but I am a little nervous about it (more places we have to remember the change is no good).

Maybe this should Either instead of just quietly converting?

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
```diff
@@ -0,0 +1,259 @@
...
+          FIND_BY_ZONEID_NAME
+            .bindByName('zoneId -> zoneId, 'name -> name)
+            .map(toRecordSet)
+            .list()
+            .apply()
+        }
+      }
+    }
+
+  def getRecordSetsByFQDN(fqdns: List[FQDN]): IO[List[RecordSet]] =
```
rebstar6: nitpick: getRecordSetsByFQDNs plural

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
```diff
@@ -0,0 +1,259 @@
...
+
+  /**
+    * TODO: There is a potential issue with the way we do this today.  We load all record sets eagerly, potentially
+    * causing memory pressure for the app depending on the number of records in the zone.
+    *
+    * This needs to change in the future; however, for right now we just need to tune
+    * the JVM as we have the same issue in the DynamoDB repository.  Until
+    * we create a better sync and load process that is better for memory, this should
+    * be the same as the other repo.
+    */
```
rebstar6: this call has 2 uses:
- user list rs: we only let them list at max 100 records at once, so should be fine
- zone sync: we wanna load/compare all the things (i assume this is what youre referring to with the memory comment).

If needed, this can be split off into separate database if there is a better way - sync certainly doesnt need the filters and things.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
```diff
@@ -0,0 +1,259 @@
...
+            i.recordSet.id,
+            i.recordSet.zoneId,
+            i.recordSet.name,
+            FQDN(i.recordSet.name, i.zone.name).value,
+            fromRecordType(i.recordSet.typ),
+            toPB(i.recordSet).toByteArray
+          )
+      }
+
+      val updates: Seq[Seq[Any]] = byChangeType.getOrElse(RecordSetChangeType.Update, Nil).map {
```
rebstar6: yicky, Seq[Seq[Any]].

I see why, could in theory make a class to hold each of the parameter types, but I can deal either way

project/Dependencies.scala
```diff
@@ -70,10 +70,11 @@ object Dependencies {
 
   lazy val mysqlDependencies = Seq(
     "org.flywaydb"              %  "flyway-core"                    % "5.1.4",
-    "org.mariadb.jdbc"          %  "mariadb-java-client"            % "2.2.6",
+    "org.mariadb.jdbc"          %  "mariadb-java-client"            % "2.3.0",
     "org.scalikejdbc"           %% "scalikejdbc"                    % scalikejdbcV,
     "org.scalikejdbc"           %% "scalikejdbc-config"             % scalikejdbcV,
-    "com.zaxxer"                %  "HikariCP"                       % "3.2.0"
+    "com.zaxxer"                %  "HikariCP"                       % "3.2.0",
+    "ch.qos.logback"            %  "logback-classic"                % "1.0.7"
```
rebstar6: this is in core, shouldnt need here (flows down)

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
pauljamescleary: can also leave it out for the time being and implement later.  It is not yet in the `RecordSetRepository` trait, and will remain unused.  Realistically, we cannot use it until we either add support in DynamoDB, or we remove DynamoDB repositories all together.

It was useful for performance testing purposes.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
pauljamescleary: I think we should leave it out for the time being.  If we want FQDN in the repository, we have to support it in DynamoDB as well.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
pauljamescleary: Yea, this is scalikejdbc and one of the many reasons I am not a fan of it.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
pauljamescleary: There were recommendations that came out of #291 .  Zone loading is likely easier to "stream".  Zone syncing requires some thought.

I confirmed that our largest zones will be < 200,000 records (which tbh is a ton for any DNS installation) so I think we are ok for the time being.  Will prioritize those performance tuning takeaways as needed

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
pauljamescleary: Yes, you are correct, I will amend

project/Dependencies.scala
pauljamescleary: I think that went in after I started this, will update

modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlRecordSetRepositoryIntegrationSpec.scala
pauljamescleary: I took out FQDN support.  Comments are below, we _must_ support it in DynamoDB.  It makes sense to implement them as part of the same effort / epic.  Adding it back in the future is simple.  The exercise was valuable from a performance perspective.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
pauljamescleary: taking out FQDN for now

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
pauljamescleary: making this a pattern match, which is "total" for an enumeration.  We don't need to worry about `Option` or `Either`.

PR #316, opened 2018-10-30
Add MySql Record Change Repository

### LINKED ISSUE #176: MySQL Record Change Repository

pauljamescleary:
Fixes #176


modules/mysql/src/main/resources/db/migration/V3.4__RecordChange.sql
```diff
@@ -0,0 +1,14 @@
+CREATE SCHEMA IF NOT EXISTS ${dbName};
+
+USE ${dbName};
+
+CREATE TABLE record_change (
+  id CHAR(36) NOT NULL,
+  zone_id CHAR(36) NOT NULL,
+  created BIGINT(13) NOT NULL,
```
mitruly: Is it possible to use `DATETIME` here for consistency with other SQL scripts?

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordChangeRepository.scala
```diff
@@ -0,0 +1,142 @@
...
+    extends RecordChangeRepository
+    with Monitored
+    with ProtobufConversions {
+  import MySqlRecordChangeRepository._
+
+  private val LIST_CHANGES_WITH_START =
+    sql"""
+      |SELECT data
+      |  FROM record_change
+      | WHERE zone_id = ?
```
mitruly: Is it possible to map `?` to variable names for these queries with multiple parameters and use `bindByName`? I think it's easier to read and less having to check back for reference to make sure that ordering matches.

modules/mysql/src/main/resources/db/migration/V3.4__RecordChange.sql
nimaeskandary: believe this is since the paging is just doing lexicographic comparison anyway, how the zone change was done as well

mitruly: Makes sense. Just noticed that we are using `DATETIME` in `MessageQueue` and assumed that `DATETIME` was numeric.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordChangeRepository.scala
pauljamescleary: We avoid those in the `INSERT` since those happen in bulk and `?` is faster / better for batch SQL.  But using `names` for everything else is reasonable.

mitruly: Makes sense - appreciate the explanation!

nimaeskandary:
migration number should be 3.5 now since groups went in

pauljamescleary:
> migration number should be 3.5 now since groups went in

Renamed

Issue #270, opened 2018-10-07
docker-up-vinyldns.sh fails to reliably start VinylDNS services


mdb:
**Describe the bug**

`bin/docker-up-vinyldns.sh` frequently fails to start VinylDNS services.

**VinylDNS Version**

`master`

**To Reproduce**
1. Run `bin/docker-up-vinyldns.sh` from within the root of the `vinyldns` repository
2. Observe that the script frequently exits with messages such as `Exceeded retries waiting for VinylDNS API to be ready, failing`.

**Expected behavior**
I would expect `bin/docker-up-vinyldns.sh` to reliably start VinylDNS services.

**Stack trace or error log output**

```
bin/docker-up-vinyldns.sh
Starting portal server and all dependencies in the background...
Creating network "docker_default" with the default driver
Creating vinyldns-bind9     ... done
Creating vinyldns-dynamodb  ... done
Creating vinyldns-mysql     ... done
Creating vinyldns-elasticmq ... done
Creating vinyldns-api       ... done
Creating vinyldns-portal    ... done
Waiting for API to be ready at http://localhost:9000 ...
Retrying Again
Retrying Again
Retrying Again
Retrying Again
Retrying Again
Retrying Again
Retrying Again
Retrying Again
Retrying Again
Retrying Again
Retrying Again
Retrying Again
Retrying Again
Retrying Again
Retrying Again
Retrying Again
Retrying Again
Retrying Again
Retrying Again
Retrying Again
Retrying Again
Retrying Again
Retrying Again
Retrying Again
Retrying Again
Retrying Again
Retrying Again
Retrying Again
Retrying Again
Retrying Again
Retrying Again
Retrying Again
Retrying Again
Retrying Again
Retrying Again
Retrying Again
Retrying Again
Retrying Again
Retrying Again
Retrying Again
Exceeded retries waiting for VinylDNS API to be ready, failing
```

**Additional context**
I have observed this problem on my local Macbook Pro, as well as in TravisCI.

nimaeskandary:
Hm, we set that retry with all the vinyl devs pretty much using the same machines, we could increase the retry another minute or two to accommodate for other machines. 

Could also add an arg variable to override timeout?

pauljamescleary:
Right, I think increasing the timeout should be fine.  We have to wait for mysql to start up, then run our migrations (which run on an empty environment), and finally wait for the app to start up which itself takes time (akka-http most of it).

pauljamescleary:
[...]

PR #318, opened 2018-10-31
create a simple way to start Dockerized API

mdb:
This seeks to fix issue #270 and, specifically, address these comments:
https://github.com/vinyldns/vinyldns/issues/270#issuecomment-434758538

While it's not a particularly DRY solution, it seeks to provide users
a simple way to start a Dockerized VinylDNS API _without_ the need to either...

* start the VinylDNS portal as well
* build a VinylDNS jar file

There is likely room to refactor and simplify the `bin/*.sh` files.

Fixes #.

Changes in this pull request:
-
-

### DIFF
--- a/bin/docker-up-api-server-local.sh
+++ b/bin/docker-up-api-server-local.sh
@@ -1,0 +1,58 @@
+#!/bin/bash
+######################################################################
+# Copies the contents of `docker` into target/scala-2.12
+# to start up dependent services via docker compose.  Once
+# dependent services are started up, the fat jar built by sbt assembly
+# is loaded into a docker container.  The api will be available
+# by default on port 9000
+######################################################################
+
+
+DIR=$( cd $(dirname $0) ; pwd -P )
+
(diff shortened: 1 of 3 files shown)


nimaeskandary:
If https://github.com/vinyldns/vinyldns/pull/319/files goes in then the image tags would have to change to `0.8.0-SNAPSHOT`, this is because I just did a release but we only update latest on a non snapshot. We'll be releasing our first official release for reals and from then on we can change the tags in the script to `latest`

nimaeskandary:
Looking at it again, since this is only for the api, the `0.8.0` tag will still work. Will keep a note of it and once we doing the real 0.8.0 we can probably update all these docker scripts to use the `latest` tag

nimaeskandary:
@mdb curious, would https://github.com/vinyldns/vinyldns/pull/319 satisfy the issue such that this pr could be closed?

mdb:
@nimaeskandary Yep; I've closed this PR. Thanks!

PR #319, opened 2018-10-31
fix snapshot image

### LINKED ISSUE #270: docker-up-vinyldns.sh fails to reliably start VinylDNS services

nimaeskandary:
Fixes #270 .

Changes in this pull request:
- moves portal conf settings to env
- added a flag to only start up api


nimaeskandary:
pull and run `bin/docker-compose-up.sh` to test

docker/docker-compose-build.yml
```diff
@@ -51,14 +51,14 @@ services:
...
   portal:
-    image: "vinyldns/portal:0.8.0"
+    image: "vinyldns/portal:0.8.0-SNAPSHOT"
     env_file:
       .env
     ports:
       - "${PORTAL_PORT}:${PORTAL_PORT}"
     container_name: "vinyldns-portal"
     volumes:
-      - ./portal/application.conf:/opt/docker/conf/application.conf
```
nimaeskandary: did this because we already include the local.conf in the default configuration, and not everything has to be overriden

modules/portal/conf/application.conf
```diff
@@ -8,7 +8,7 @@
 # This must be changed for production, but we recommend not changing it in this file.
 #
 # See http://www.playframework.com/documentation/latest/ApplicationSecret for more details.
-play.http.secret.key = "changeme"
+play.http.secret.key = "this-secret-must-be-changed-for-prod"
```
nimaeskandary: the default changeme auto-fails

docker/docker-compose-build.yml
```diff
@@ -38,7 +38,7 @@ services:
       - ./elasticmq/custom.conf:/etc/elasticmq/elasticmq.conf
 
   api:
-    image: "vinyldns/api:0.8.0"
+    image: "vinyldns/api:0.8.0-SNAPSHOT"
```
nimaeskandary: when we do our first release for realsies, we will start updating the latest tag on docker, at that point this can be changed to `vinyldns/api:latest`

bin/docker-up-vinyldns.sh
```diff
@@ -1,26 +1,60 @@
...
 function usage {
     printf "usage: docker-up-vinyldns.sh [OPTIONS]\n\n"
     printf "starts up a local VinylDNS installation using docker compose\n\n"
     printf "options:\n"
-    printf "\t-t, --timeout seconds: overwrite the timeout used when waiting for components to startup, default of 60\n"
+    printf "\t-t, --timeout seconds: overwrite ping timeout of 60\n"
+    printf "\t-a, --api-only: do not start up vinyldns-portal\n"
 }
 
 while [ "$1" != "" ]; do
```
mdb: Naive question: if this only evaluates `$1`, will it function as desired if _both_ `--timeout` and `--api-only` are passed?

nimaeskandary: Should do, tested out a few cases that seem good, since there's an extra `shift` in the case that its `-t`, then it will make sure to shift both the flag and the arg when it comes by it, in the case that `-a` is first, it only shifts the flag

nimaeskandary:
@mdb currently `--api-only` doesn't start up the api service at all, in addition to not pinging it, when that flag is present it calls docker-compose up with the `api` argument, and since portal is not an api dependency in the config it doesn't start the portal up

bin/docker-up-vinyldns.sh
nimaeskandary: shift by the way is moving args to the left, so what was $2 will now be $1, $3-$2, etc

bin/docker-up-vinyldns.sh
```diff
@@ -1,26 +1,60 @@
 #!/usr/bin/env bash
-######################################################################
+#####################################################################################################
 # Starts up the api, portal, and dependent services via
 # docker-compose. The api will be available on localhost:9000 and the
 # portal will be on localhost:9001
 #
+# Relevant variable overrides can be found at ./.env and ../docker/.env
+#
```
mitruly: Might possibly want to just provide the two vars from `.env` we are using, as they are not specified in this file.

nimaeskandary: Good point, but seeing us using the .env files much more in the future, assuming they are light right now because they were recently implemented and not flushed out

mitruly: Oh, sorry. Meant in comments like in `docker.conf`.

nimaeskandary: Could see that changing a lot too!

docker/portal/local.conf
```diff
@@ -0,0 +1,20 @@
+# overwrites urls to use docker-compose service names, for example, to reach the api the url is
```
rebstar6: my only concern with this is that we generally git ignore local.confs so people can configure for their app specifically. 

Can you call it something else and include that too in the default?

nimaeskandary: good call, sure

docker/portal/docker.conf
```diff
@@ -0,0 +1,26 @@
+# Overrides settings in modules/portal/conf/application.conf
+
+# ldap searchbase
+
+#LDAP.searchbase = [
```
nimaeskandary: Env overrides and hocon conf arrays don't play nicely, we could change our parser class in the portal to parse a string for these tokens as opposed to using an array, but don't think that should be in this pr

nimaeskandary: Env overrides and hocon conf arrays don't play nicely, we could change our parser class in the portal to parse a string for these tokens as opposed to using an array, but don't think that should be in this pr

nimaeskandary:
Need to publish new snapshots to docker hub from my other laptop. To test before then:

```
./bin/remove-vinyl-containers.sh
docker rmi vinyldns/portal:0.8.0-SNAPSHOT
docker rmi vinyldns/api:0.8.0-SNAPSHOT
sbt ";project api;docker:publishLocal;project portal;docker:publishLocal"

./bin/docker-up-vinyldns.sh
```

docker/portal/docker.conf
pauljamescleary: this whole file appears to be commented out? don't think that should stay that way right?

nimaeskandary: Right now the only things you'd need the docker.conf for is the LDAP search base and out custom link configs, since they are both arrays right now, figured it may make things easier for adopters but could remove

modules/portal/conf/reference.conf
```diff
@@ -9,6 +9,7 @@
 #
```
nimaeskandary: tested this with an application.conf mounted and no application.conf mounted, works as expected, reference.conf is the default and the applicaiton.conf overrides

modules/portal/conf/reference.conf
```diff
@@ -119,3 +159,6 @@ play.modules.enabled += "modules.VinylDNSModule"
 
 // Local.conf has files specific to your environment, for example your own LDAP settings
 include "local.conf"
+
+// Additional overrides
+include "application.conf"
```
pauljamescleary: shouldn't have to do this.  The way the config works is that it will load an application.conf if found, and then fall back to reference .conf

modules/portal/conf/reference.conf
```diff
@@ -44,46 +45,78 @@ play.i18n.langs = [ "en" ]
...
-  searchBase = [{organization = "someDomain", domainName = "DC=test,DC=test,DC=com"}, {organization = "anotherDomain", domainName = "DC=test,DC=com"}]
+  user = "test"
+  user = ${?LDAP_USER}
+  password = "test"
+  password = ${?LDAP_PASSWORD}
+  domain = "test"
+  domain = ${?LDAP_DOMAIN}
+
+  searchBase = [
+    {
```
pauljamescleary: just curious if we should have a default of a single search domain, with overrides for environment variables.  The thinking is that most installations will have a single LDAP domain tbh.

nimaeskandary: added this and tested it

modules/portal/conf/reference.conf
```diff
@@ -92,18 +123,24 @@ play.filters.enabled += "play.filters.csrf.CSRFFilter"
...
+play.http.session.secure = ${?SESSION_SECURE}
+
 play.http.session.httpOnly = true
+play.http.session.httpOnly = ${?SESSION_HTTP_ONLY}
 
 // MUST be the same as the API!!!
 crypto {
-  type = "vinyldns.core.crypto.JavaCrypto"
-  secret = "8B06A7F3BC8A2497736F1916A123AA40E88217BE9264D8872597EF7A6E5DCE61"
+  type = "vinyldns.core.crypto.NoOpCrypto"
```
nimaeskandary: this is the default in the api's universal conf

nimaeskandary:
By the way to test this cleanly:

```
./bin/remove-vinyl-containers.sh
docker rmi vinyldns/portal:0.8.0-SNAPSHOT
docker rmi vinyldns/api:0.8.0-SNAPSHOT
./bin/docker-up-vinyldns.sh
```

PR #325, opened 2018-11-01
Add healthchecking dynamically for databases

### LINKED ISSUE #313: Add health checks for DynamoDB, SQS
**Background**
Presently, we only health check that we can talk to MySQL.  This is done via the `/health` HealthRoute.  We will not be system aware of an outage on other dependencies.

**Solution**
Use drop wizard metrics health checking to check all dependencies.  This should be in the module itself (MySQL, DynamoDB).

Will need to add LDAP authentication checks to the portal.

**Notes**
https://metrics.dropwizard.io/3.1.0/manual/healthchecks/

If we run health checks on demand (by the Load Balancer), it maybe too frequent.

I would recommend having a daemon service that polls those health checks once every few seconds, and have the `HealthRoute` just ping that daemon.  Can use FS2 and things like `Ref` for concurrently setting and fetching state.

rebstar6:
Fixes #313 

add dynamo/mysql database healthchecking

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/route/HealthService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/route/HealthService.scala
@@ -19,21 +19,23 @@ package vinyldns.api.route
 import cats.effect.{ContextShift, IO}
 import cats.implicits._
 import org.slf4j.LoggerFactory
+import vinyldns.core.route.HealthCheck.{HealthCheckError, HealthCheckResponse}
 
-class HealthService(dataStoreHealthChecks: List[IO[Unit]]) {
+class HealthService(healthChecks: List[HealthCheckResponse]) {
 
   private val logger = LoggerFactory.getLogger(classOf[HealthService])
 
   private implicit val cs: ContextShift[IO] =
     IO.contextShift(scala.concurrent.ExecutionContext.global)
 
-  def checkHealth(): IO[Either[Throwable, Unit]] =
-    dataStoreHealthChecks.parSequence
-      .as((): Unit)
-      .handleErrorWith { e =>
-        IO {
-          logger.error(s"DataStore health check error raised: $e")
-        }.flatMap(_ => IO.raiseError[Unit](e))
+  def checkHealth(): IO[List[HealthCheckError]] =
+    healthChecks.parSequence
+      .map {
+        _.collect {
+          case Left(err) => {
+            logger.error(s"Health Check Failure: ${err.message}")
+            err
+          }
+        }
       }
-      .attempt
 }
--- a/modules/api/src/main/scala/vinyldns/api/route/HealthService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/route/HealthService.scala
@@ -16,15 +16,24 @@
 
 package vinyldns.api.route
 
+import cats.effect.{ContextShift, IO}
 import cats.implicits._
-import vinyldns.api.Interfaces._
-import vinyldns.core.domain.zone.ZoneRepository
+import org.slf4j.LoggerFactory
 
-class HealthService(zoneRepository: ZoneRepository) {
+class HealthService(dataStoreHealthChecks: List[IO[Unit]]) {
 
-  def checkHealth(): Result[Unit] =
-    zoneRepository
-      .getZone("notFound")
-      .map(_ => ().asRight)
-      .toResult
+  private val logger = LoggerFactory.getLogger(classOf[HealthService])
+
+  private implicit val cs: ContextShift[IO] =
+    IO.contextShift(scala.concurrent.ExecutionContext.global)
+
+  def checkHealth(): IO[Either[Throwable, Unit]] =
+    dataStoreHealthChecks.parSequence
+      .as((): Unit)
+      .handleErrorWith { e =>
+        IO {
+          logger.error(s"DataStore health check error raised: $e")
+        }.flatMap(_ => IO.raiseError[Unit](e))
+      }
+      .attempt
 }
--- a/modules/api/src/main/scala/vinyldns/api/route/HealthService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/route/HealthService.scala
@@ -19,9 +19,9 @@ package vinyldns.api.route
 import cats.effect.{ContextShift, IO}
 import cats.implicits._
 import org.slf4j.LoggerFactory
-import vinyldns.core.route.HealthCheck.{HealthCheckError, HealthCheckResponse}
+import vinyldns.core.health.HealthCheck.{HealthCheck, HealthCheckError}
 
-class HealthService(healthChecks: List[HealthCheckResponse]) {
+class HealthService(healthChecks: List[HealthCheck]) {
 
   private val logger = LoggerFactory.getLogger(classOf[HealthService])
 
--- a/modules/api/src/main/scala/vinyldns/api/route/HealthService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/route/HealthService.scala
@@ -32,10 +32,9 @@ class HealthService(healthChecks: List[HealthCheck]) {
     healthChecks.parSequence
       .map {
         _.collect {
-          case Left(err) => {
+          case Left(err) =>
             logger.error(s"Health Check Failure: ${err.message}")
             err
-          }
         }
       }
 }
--- a/modules/api/src/test/scala/vinyldns/api/route/BatchChangeRoutingSpec.scala
+++ b/modules/api/src/test/scala/vinyldns/api/route/BatchChangeRoutingSpec.scala
@@ -18,7 +18,7 @@ package vinyldns.api.route
 
 import akka.http.scaladsl.model.StatusCodes._
 import akka.http.scaladsl.model.{ContentTypes, HttpEntity}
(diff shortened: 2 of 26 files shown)


modules/core/src/main/scala/vinyldns/core/repository/DataStore.scala
```diff
@@ -24,6 +25,11 @@ import vinyldns.core.repository.RepositoryName.RepositoryName
 
 import scala.reflect.ClassTag
 
+class LoadedDataStore(
+    val dataStore: DataStore,
+    val shutdownHook: IO[Unit],
+    val healthCheck: IO[Unit])
```
rebstar6: ^^ is the return value from each load now. needed to move healthcheck in as returned (rather than a function) because it relies on config in the dynamo world.

moved shutdown too for consistency

modules/dynamodb/src/main/scala/vinyldns/dynamodb/repository/DynamoDBDataStoreProvider.scala
```diff
@@ -116,5 +116,9 @@ class DynamoDBDataStoreProvider extends DataStoreProvider {
     ).parMapN { DataStore.apply }
   }
 
-  def shutdown(): IO[Unit] = IO.unit
+  private def checkHealth(dynamoConfig: DynamoDBDataStoreSettings): IO[Unit] =
+    IO {
+      val client = DynamoDBClient(dynamoConfig)
+      client.listTables(1)
+    }.map(_ => ())
```
rebstar6: per ^^, what do we think is a good dynamo test? listTables just ensures theres some level of connectivity, but does not get at all the specific tables (which could have individual settings)

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlDataStoreProvider.scala
```diff
@@ -99,7 +104,22 @@ class MySqlDataStoreProvider extends DataStoreProvider {
...
+         |  FROM zone
+         |
+      """.stripMargin
+
+  private def checkHealth(): IO[Unit] =
+    IO {
+      DB.readOnly { implicit s =>
+        HEALTH_CHECK.map(_ => ()).first().apply()
+      }
+    }.map(_ => ())
```
rebstar6: same here, do we know what we want this HC to actually do?

modules/api/src/test/scala/vinyldns/api/route/BatchChangeRoutingSpec.scala
```diff
@@ -38,7 +38,6 @@ import vinyldns.core.domain.batch._
 class BatchChangeRoutingSpec
     extends WordSpec
     with ScalatestRouteTest
-    with Directives
```
rebstar6: just cleanup, for some reason all these routing specs extend Directives but dont need to

modules/api/src/main/scala/vinyldns/api/route/HealthService.scala
```diff
@@ -16,15 +16,26 @@
...
-      .getZone("notFound")
-      .map(_ => ().asRight)
-      .toResult
+  private val logger = LoggerFactory.getLogger(classOf[HealthService])
+
+  private implicit val cs: ContextShift[IO] =
+    IO.contextShift(scala.concurrent.ExecutionContext.global)
+
+  def checkHealth(): IO[List[HealthCheckError]] =
+    healthChecks.parSequence
```
nimaeskandary: May be worth adding a comment here explaining how the healthcheck IO will be re-executed every time this route is called, took a bit for that to sink in for me

rebstar6: yeah this is probably the 1st time that ive noticed the IO being lazily evaluated making a difference in practice.

because of that, it should evaluate each time it's called. i *think* this would be a problem with future.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlDataStoreProvider.scala
```diff
@@ -99,7 +102,21 @@ class MySqlDataStoreProvider extends DataStoreProvider {
     }
   }
 
-  def shutdown(): IO[Unit] =
+  private def shutdown(): IO[Unit] =
     IO(DBs.close())
       .handleError(e => logger.error(s"exception occurred while shutting down", e))
+
+  private final val HEALTH_CHECK =
```
nimaeskandary: Since the zone table doesn't have to be MySQL this could be something table agnostic like `SELECT version()`

rebstar6: doesnt matter, because all migrations are run, the zone table will always exist if mysql is on, even if no data is put into it

nimaeskandary: Gotcha

modules/core/src/main/scala/vinyldns/core/route/HealthCheck.scala
```diff
@@ -0,0 +1,36 @@
+/*
```
pauljamescleary: this should not be in `route`.  The `route` layer deals with translation to/from the domain to HTTP things.

`core.health` or `core.domain.health` would be better

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlDataStoreProvider.scala
pauljamescleary: `SELECT 1 FROM DUAL` is the usual database ping.

modules/core/src/main/scala/vinyldns/core/route/HealthCheck.scala
```diff
@@ -0,0 +1,36 @@
...
+ * limitations under the License.
+ */
+
+package vinyldns.core.route
+
+import cats.effect.IO
+
+object HealthCheck {
+
+  type HealthCheckResponse = IO[Either[HealthCheckError, Unit]]
```
pauljamescleary: This was really confusing to follow.  In many applications, you see a `Request/Response` type of interaction (for example _all_ of the AWS clients).

When I see `HealthCheckResponse`, I think "the response of executing some method.

Here, it is actually an `IO` that we repeatedly evaluate.

This should be renamed so it does not collide with the common request/response pattern.

Can it just be `HealthCheck`?  `type HealthCheck = IO[Either[HealthCheckError, Unit]]`

modules/core/src/main/scala/vinyldns/core/health/HealthCheck.scala
```diff
@@ -0,0 +1,36 @@
...
+  type HealthCheck = IO[Either[HealthCheckError, Unit]]
+
+  case class HealthCheckError(message: String) extends Throwable(message)
+
+  implicit class HealthCheckImprovements(io: IO[Either[Throwable, _]]) {
+    def asHealthCheck: HealthCheck =
+      io.map {
+        case Left(err) =>
+          Left(HealthCheckError(Option(err.getMessage).getOrElse("no message from error")))
+        case _ => Right((): Unit)
```
mitruly: Typing isn't needed for this; `case _ => Right(())` or `case _ => Right(Unit)` both worked for me.

modules/dynamodb/src/main/scala/vinyldns/dynamodb/repository/DynamoDBDataStoreProvider.scala
```diff
@@ -113,8 +114,14 @@ class DynamoDBDataStoreProvider extends DataStoreProvider {
...
+    ).parMapN {
+      DataStore.apply
+    }
   }
 
-  def shutdown(): IO[Unit] = IO.unit
+  private def checkHealth(dynamoConfig: DynamoDBDataStoreSettings): HealthCheck =
+    IO {
+      val client = DynamoDBClient(dynamoConfig)
+      client.listTables(1)
```
mitruly: More of a question - should we try to keep health checks consistent (eg. accessing a non-existent table/listing available tables)?

modules/api/src/main/scala/vinyldns/api/route/HealthService.scala
```diff
@@ -16,15 +16,26 @@
...
+      .map {
+        _.collect {
+          case Left(err) => {
```
mitruly: IntelliJ says these `{}` not needed, but doesn't matter to me much.

pauljamescleary: for sure, remove the unnecessary curly braces.

modules/dynamodb/src/main/scala/vinyldns/dynamodb/repository/DynamoDBDataStoreProvider.scala
pauljamescleary: This is good imo.  We are testing 1) connectivity and 2) credentials here.

rebstar6: yeah i think its just different with dynamo. leaving as is.

PR #327, opened 2018-11-02
prevent duplicate zone creation

### LINKED ISSUE #310: Zones can be created multiple times (api)
If multiple requests are made fast enough you could create duplicate zones in the API. If the second request starts processing before the first request actually saves the zone in the database the second request will pass the "zoneDoesNotExist" check.

britneywright:
Fixes #310.

Changes in this pull request:
- Add a check to see if a zone exists during the ZoneChange processing
- If a zone exists, change the ZoneChangeStatus to Failed and raise a ZoneAlreadyExistsError
- The error is swallowed as message flow continues, but no duplicate zone is created.

Expect the tests to fail at the moment, need to test this at the point two zone requests are on the message queue, under the assumption they passed the validations in the ZoneService. I did create one test that conditionally passes right now in the func tests. "test_create_zone_same_name_fails_during_processing" should pass if you comment out `_ <- zoneDoesNotExist(zone)` in the ZoneService.

To kind of test in the portal:
- Comment out `_ <- zoneDoesNotExist(zone)` in the ZoneService
- Create a zone in the portal and click the refresh button. You should see the new zone.
- Create a zone with the same name and click the refresh button. You should still only see one zone.

### DIFF
--- a/modules/api/functional_test/live_tests/zones/create_zone_test.py
+++ b/modules/api/functional_test/live_tests/zones/create_zone_test.py
@@ -471,3 +471,66 @@ def test_user_cannot_create_zone_with_failed_validations(shared_zone_test_contex
     assert_that(result['errors'], contains_inanyorder(
         contains_string("not-approved.thing.com. is not an approved name server")
     ))
+
+@pytest.mark.skip_production
+def test_create_zone_same_name_fails_during_processing(shared_zone_test_context):
+    """
+    Test create a zone errors if the name is found during processing
+    """
+    client = shared_zone_test_context.ok_vinyldns_client
+    result_zone = None
+    duplicate_result_zone = None
+    try:
+        zone_name = 'one-time'
+
+        zone = {
+            'name': zone_name,
+            'email': 'test@test.com',
+            'adminGroupId': shared_zone_test_context.ok_group['id'],
+            'connection': {
+                'name': 'vinyldns.',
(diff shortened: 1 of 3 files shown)


modules/api/functional_test/live_tests/zones/create_zone_test.py
```diff
@@ -471,3 +471,66 @@ def test_user_cannot_create_zone_with_failed_validations(shared_zone_test_contex
...
+                'primaryServer': VinylDNSTestContext.dns_ip
+            },
+            'transferConnection': {
+                'name': 'vinyldns.',
+                'keyName': VinylDNSTestContext.dns_key_name,
+                'key': VinylDNSTestContext.dns_key,
+                'primaryServer': VinylDNSTestContext.dns_ip
+            }
+        }
+        result = client.create_zone(zone, status=202)
```
nimaeskandary: I wonder if this can be reliably func tested because a the local setup may be too quick, also the create_zone of the client will blocks until the api reports back a 202, so it may be more realistic to make a custom create_zone that does not wait for the response

pauljamescleary: I agree that a func test is likely not reliable as timing is always tricky.  I would recommend that we do not use a func test here.

britneywright:
closing. taking different approach in #331

PR #331, opened 2018-11-06
Prevent duplicate zone creations

### LINKED ISSUE #310: Zones can be created multiple times (api)

britneywright:
Fixes #310.

Changes in this pull request:
- Handle duplicate zones in `MySqlZoneRepository` and `ZoneChangeHandler` only


modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneRepository.scala
```diff
@@ -357,19 +363,41 @@ class MySqlZoneRepository extends ZoneRepository with ProtobufConversions with M
...
+            case (Some(_), None) => saveZoneProcess(zone).asRight
+            case (None, Some(_)) =>
+              DuplicateZoneError(s"Zone with name ${zone.name} already exists.").asLeft
+          }
         }
       }
     }
 
-  def retryWithBackoff[A](f: A => IO[A], a: A, delay: FiniteDuration, maxRetries: Int): IO[A] =
+  def saveZoneProcess(zone: Zone): Zone =
```
mitruly: I believe that zone access is part of the zone; the `zone_access` table seems to be accessed by joining with `zone_id`. ie. zone access does not have to be handled specifically.

I think that using `putZone(zone)` instead of `saveZoneProcess(zone)` should work, but there should be a test for making sure that updating a zone name retains the same ACL rules.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneRepository.scala
```diff
@@ -120,35 +121,40 @@ class MySqlZoneRepository extends ZoneRepository with ProtobufConversions with M
...
     */
-  def save(zone: Zone): IO[Zone] =
+  def save(zone: Zone): IO[Either[DuplicateZoneError, Zone]] =
     zone.status match {
       case ZoneStatus.Deleted =>
-        retryWithBackoff(deleteTx, zone, INITIAL_RETRY_DELAY, MAX_RETRIES)
-      case _ => retryWithBackoff(saveTx, zone, INITIAL_RETRY_DELAY, MAX_RETRIES)
+        val doDelete: Zone => IO[Either[DuplicateZoneError, Zone]] = z => deleteTx(z).map(Right(_))
+        retryWithBackoff(doDelete, zone, INITIAL_RETRY_DELAY, MAX_RETRIES)
+      case _ => {
```
mitruly: Curly braces aren't necessary here.

modules/api/src/main/scala/vinyldns/api/engine/ZoneChangeHandler.scala
```diff
@@ -25,13 +25,15 @@ import vinyldns.core.domain.zone.{
...
-      zoneChangeRepository: ZoneChangeRepository): ZoneChange => IO[ZoneChange] = zoneChange => {
-    for {
-      _ <- zoneRepository.save(zoneChange.zone)
-      savedChange <- zoneChangeRepository.save(zoneChange.copy(status = ZoneChangeStatus.Synced))
-    } yield savedChange
-  }
+      zoneChangeRepository: ZoneChangeRepository): ZoneChange => IO[ZoneChange] =
+    zoneChange =>
+      zoneRepository.save(zoneChange.zone).flatMap {
+        case Left(ZoneRepository.DuplicateZoneError(msg)) => zoneChangeRepository.save(
```
mitruly: Importing `vinyldns.core.domain.zone.ZoneRepository.DuplicateZoneError` at the top and using `DuplicateZoneError` directly looks cleaner.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneRepository.scala
```diff
@@ -357,19 +361,41 @@ class MySqlZoneRepository extends ZoneRepository with ProtobufConversions with M
...
     monitor("repo.ZoneJDBC.save") {
       IO {
         DB.localTx { implicit s =>
-          deleteZoneAccess(zone)
-          putZone(zone)
-          putZoneAccess(zone)
-          zone
+          val zoneById = getZoneByIdInSession(zone.id)
+          val zoneByName = getZoneByNameInSession(zone.name)
+          (zoneById, zoneByName) match {
```
nimaeskandary: seems this could also be achieved with a primary composite key (id, name) as opposed to just id. Wonder if the performance hit from a composite key is better or worse than these pre-queries

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneRepository.scala
```diff
@@ -357,19 +363,41 @@ class MySqlZoneRepository extends ZoneRepository with ProtobufConversions with M
...
+            case (Some(foundZoneById), Some(foundZoneByName)) => {
+              if (foundZoneById.id == foundZoneByName.id) {
+                saveZoneProcess(zone).asRight
+              } else
+                DuplicateZoneError(s"Incorrect ID for Zone with name ${zone.name}").asLeft
```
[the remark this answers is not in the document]
britneywright: We want distinct messages for each of the failure cases. Those messages are passed into the ZoneChange and saved.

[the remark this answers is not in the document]
pauljamescleary: @mitruly @britneywright encoded the rules here.  It is rather tough, because in theory zone name should have been unique all along.

I do agree that we want to encode different error messages.

From a repo perspective, we cannot make assumptions of the state of the data in the table.  We can only code to handle the possible failure conditions.

```
1. Zone ID not exists, Name not exists - INSERT (happy path)
2. Zone ID exists, Name not exists - UPDATE all fields including name
3. Zone ID not exists, Name exists - Fail as we cannot overwrite the ID
4. Zone ID exists, Name exists SAME record - UPDATE all fields (name is ok too since it is the same)
5. Zone ID exists, Name exists DIFFERENT record - Fail as we cannot update the name, and cannot overwrite the id of the other record
```

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneRepository.scala
pauljamescleary: There are some odd foreign key constraints that come into play with just putZone @mitruly, I would keep the process as is.

In theory, we should drop the foreign key constraint, and just `deleteZoneAccess` on zone delete as well.  That foreign key causes headaches.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneRepository.scala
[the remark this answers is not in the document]
mitruly: Thanks for the explanation, @pauljamescleary.

If we want different error messages, I think it would be better to create different cases classes:
```
sealed abstract class DuplicateZoneError(message: String)
  final case class ZoneAlreadyExists(zoneName: String) extends DuplicateZoneError(
    s"Zone with name '$zoneName' already exists.")
  final case class IncorrectId(zoneName: String) extends DuplicateZoneError(
    s"Incorrect ID for zone with name '$zoneName'.")
```

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneRepository.scala
```diff
@@ -357,19 +358,35 @@ class MySqlZoneRepository extends ZoneRepository with ProtobufConversions with M
...
-  def saveTx(zone: Zone): IO[Zone] =
+  def saveTx(zone: Zone): IO[Either[DuplicateZoneError, Zone]] =
     monitor("repo.ZoneJDBC.save") {
       IO {
         DB.localTx { implicit s =>
-          deleteZoneAccess(zone)
-          putZone(zone)
-          putZoneAccess(zone)
-          zone
+          getZoneByNameInSession(zone.name) match {
```
mitruly: This can be simplified:

```
          getZoneByNameInSession(zone.name) match {
            case Some(foundZone) if zone.id != foundZone.id => DuplicateZoneError(zone.name).asLeft
            case _ => saveZoneProcess(zone).asRight
          }
```

modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlZoneRepositoryIntegrationSpec.scala
```diff
@@ -109,7 +110,19 @@ class MySqlZoneRepositoryIntegrationSpec
...
   "MySqlZoneRepository" should {
     "return the zone when it is saved" in {
-      repo.save(okZone).unsafeRunSync() shouldBe okZone
+      repo.save(okZone).unsafeRunSync() shouldBe Right(okZone)
+    }
+
+    "return the updated zone when it is saved with a new name" in {
+      repo.save(okZone).unsafeRunSync() shouldBe Right(okZone)
+      val newZone = okZone.copy(name = "newname.")
+      repo.save(newZone).unsafeRunSync() shouldBe Right(newZone)
```
pauljamescleary: Do a `getZoneByName` and `getZone(id)` after this save of `newZone`

Verify that the results of both calls `shouldBe newZone`

 We want to verify that we did not change the id.  That would be really bad.  It is probably a good idea to just check

`val retrievedZone.id shouldBe okZone.id` just to be 100%

britneywright: In this case we're copying the okZone so the ID never changes. I also changed this test to change the adminGroup rather than the name. We discussed that we shouldn't support changing a zone name in an update further upstream. I need to double check whether we currently allow it. If so we'll create an issue to stop that.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneRepository.scala
pauljamescleary: Agreed @britneywright the guard is better than the if/else

britneywright: of course, bad habit i can't break.

modules/mysql/src/main/resources/db/migration/V3.7__ZoneNameConstraint.sql
```diff
@@ -0,0 +1,5 @@
+CREATE SCHEMA IF NOT EXISTS ${dbName};
+
+USE ${dbName};
+
+ALTER TABLE zone ADD CONSTRAINT zone_name_unique UNIQUE (name);
```
rebstar6: hypothetically, if we have dups in there currently, what will this do? not positive we can ensure we dont

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneRepository.scala
```diff
@@ -357,19 +358,31 @@ class MySqlZoneRepository extends ZoneRepository with ProtobufConversions with M
...
+            case Some(foundZone) if zone.id != foundZone.id => DuplicateZoneError(zone.name).asLeft
+            case _ => saveZoneProcess(zone).asRight
+          }
         }
       }
     }
 
-  def retryWithBackoff[A](f: A => IO[A], a: A, delay: FiniteDuration, maxRetries: Int): IO[A] =
+  def saveZoneProcess(zone: Zone): Zone =
+    DB.localTx { implicit s =>
```
rebstar6: should this be in tx? called from saveTx, meaning we have nexted DB.localtx's happening

modules/mysql/src/main/resources/db/migration/V3.7__ZoneNameConstraint.sql
pauljamescleary: The constraint cannot be added if there are duplicate records, this will fail

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneRepository.scala
britneywright: @pauljamescleary thoughts here?

pauljamescleary: I think this is rather harmless, but we should probably not nest Tx, a card to cleanup is good

PR #333, opened 2018-11-07
Allow skipping of production tags from tests

pauljamescleary:
Some of our func tests are not meant to be run against "real" environments.  Since we now run the functest container against non docker environments, we have to skip these tests.

### DIFF
--- a/docker/functest/run.sh
+++ b/docker/functest/run.sh
@@ -15,6 +15,13 @@ else
   TEST_PATTERN="-k ${TEST_PATTERN}"
 fi
 
+# If we are to skip tests not suitable for prod environments, set to true or omit
+if [ -z "${FOR_PROD}" ]; then
+  SKIP_TAGS=
+else
+  SKIP_TAGS="-m \"not skip_production\""
+fi
+
 echo "Waiting for API to be ready at ${VINYLDNS_URL} ..."
 DATA=""
 RETRY=60
@@ -42,4 +49,4 @@ done
 echo "Running live tests against ${VINYLDNS_URL} and DNS server ${DNS_IP}"
 
 cd /app
-./run-tests.py live_tests -v ${TEST_PATTERN} --url=${VINYLDNS_URL} --dns-ip=${DNS_IP}
+./run-tests.py live_tests -v ${TEST_PATTERN} ${SKIP_TAGS} --url=${VINYLDNS_URL} --dns-ip=${DNS_IP}
--- a/docker/functest/run.sh
+++ b/docker/functest/run.sh
@@ -44,11 +44,11 @@ echo "Running live tests against ${VINYLDNS_URL} and DNS server ${DNS_IP}"
 cd /app
 
 # If PROD_ENV is not true, we are in a local docker environment so do not skip anything
-if [ "${PROD_ENV}" != "true"]; then
-    echo "./run-tests.py live_tests -v --url=${VINYLDNS_URL} --dns-ip=${DNS_IP} ${TEST_PATTERN}"
-    ./run-tests.py live_tests -v --url=${VINYLDNS_URL} --dns-ip=${DNS_IP} ${TEST_PATTERN}
-else
+if [ "${PROD_ENV}" = "true"]; then
     # -m plays havoc with -k, using variables is a headache, so doing this by hand
     echo "./run-tests.py live_tests -m \"not skip_production\" -v --url=${VINYLDNS_URL} --dns-ip=${DNS_IP} ${TEST_PATTERN}"
     ./run-tests.py live_tests -v -m "not skip_production" --url=${VINYLDNS_URL} --dns-ip=${DNS_IP} ${TEST_PATTERN}
+else
+    echo "./run-tests.py live_tests -v --url=${VINYLDNS_URL} --dns-ip=${DNS_IP} ${TEST_PATTERN}"
+    ./run-tests.py live_tests -v --url=${VINYLDNS_URL} --dns-ip=${DNS_IP} ${TEST_PATTERN}
 fi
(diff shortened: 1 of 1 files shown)


docker/functest/run.sh
```diff
@@ -15,6 +15,13 @@ else
   TEST_PATTERN="-k ${TEST_PATTERN}"
 fi
 
+# If we are to skip tests not suitable for prod environments, set to true or omit
+if [ -z "${FOR_PROD}" ]; then
+  SKIP_TAGS=
+else
+  SKIP_TAGS="-m \"not skip_production\""
```
mitruly: Since `SKIP_TAGS` would be a set value (rather than passing something in), it feels weird to use `-z` with an env var. Would prefer to see a parameter passed into `run.sh` (eg. `-s`/`--skip-production`) which would set the command.

nimaeskandary:
Hm getting a "vinyldns-functest | collecting ... ERROR: file not found: skip_production" when trying this out locally, investigating

docker/functest/run.sh
```diff
@@ -15,6 +15,13 @@ else
   TEST_PATTERN="-k ${TEST_PATTERN}"
 fi
 
+# If we are to skip tests not suitable for prod environments, set to true or omit
```
britneywright:
```suggestion
# Assume we are not in a production environment unless this is set. Suggest setting to true if in a production environment. 
```

docker/functest/run.sh
```diff
@@ -15,6 +15,13 @@ else
...
+if [ -z "${FOR_PROD}" ]; then
```
britneywright: can we rename this to PROD or PROD_ENV

docker/functest/run.sh
```diff
@@ -42,4 +49,4 @@ done
 echo "Running live tests against ${VINYLDNS_URL} and DNS server ${DNS_IP}"
 
 cd /app
-./run-tests.py live_tests -v ${TEST_PATTERN} --url=${VINYLDNS_URL} --dns-ip=${DNS_IP}
+./run-tests.py live_tests -v ${TEST_PATTERN} ${SKIP_TAGS} --url=${VINYLDNS_URL} --dns-ip=${DNS_IP}
```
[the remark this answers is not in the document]
pauljamescleary: cool, thanks

docker/functest/run.sh
```diff
@@ -42,4 +42,13 @@ done
 echo "Running live tests against ${VINYLDNS_URL} and DNS server ${DNS_IP}"
 
 cd /app
-./run-tests.py live_tests -v ${TEST_PATTERN} --url=${VINYLDNS_URL} --dns-ip=${DNS_IP}
+
+# If PROD_ENV is unset, we are in a local docker environment so do not skip anything
+if [ -z "${PROD_ENV}" ]; then
+    echo "./run-tests.py live_tests -v --url=${VINYLDNS_URL} --dns-ip=${DNS_IP} ${TEST_PATTERN}"
+    ./run-tests.py live_tests -v --url=${VINYLDNS_URL} --dns-ip=${DNS_IP} ${TEST_PATTERN}
```
britneywright: the "not skip_production" should be here right, since this is the prod environment?

docker/functest/run.sh
```diff
@@ -42,4 +42,13 @@ done
...
+else
+    # -m plays havoc with -k, using variables is a headache, so doing this by hand
+    echo "./run-tests.py live_tests -m \"not skip_production\" -v --url=${VINYLDNS_URL} --dns-ip=${DNS_IP} ${TEST_PATTERN}"
```
britneywright: and "not skip_production" shouldn't be here because this is the local environment.

docker/functest/run.sh
pauljamescleary: `-z` means "if not"

britneywright: thanks :)

docker/functest/run.sh
pauljamescleary: We have to get the value in from the `Dockerfile`.  ENV vars are the easiest for that

docker/functest/run.sh
pauljamescleary: yea, if we are not in the prod environment, then run everything.

Otherwise, we are not in a prod environment, so skip certain tests.

docker/functest/run.sh
```diff
@@ -42,4 +42,13 @@ done
 echo "Running live tests against ${VINYLDNS_URL} and DNS server ${DNS_IP}"
 
 cd /app
-./run-tests.py live_tests -v ${TEST_PATTERN} --url=${VINYLDNS_URL} --dns-ip=${DNS_IP}
+
+# If PROD_ENV is unset, we are in a local docker environment so do not skip anything
+if [ -z "${PROD_ENV}" ]; then
```
nimaeskandary:
```suggestion
if [ "${PROD_ENV}" != "true"]; then
```

nimaeskandary: Could be confusing because if someone sets PROD_ENV to false it would still treat it the same

docker/functest/run.sh
```diff
@@ -42,4 +42,13 @@ done
 echo "Running live tests against ${VINYLDNS_URL} and DNS server ${DNS_IP}"
 
 cd /app
-./run-tests.py live_tests -v ${TEST_PATTERN} --url=${VINYLDNS_URL} --dns-ip=${DNS_IP}
+
+# If PROD_ENV is unset, we are in a local docker environment so do not skip anything
```
britneywright: comment needs to be updated. Should be something like "if PROD_ENV is unset or not set to true".

pauljamescleary:
@rebstar6 I updated, basically if the PROD_ENV was unset, that first != was still firing.

I switched up the order.  You _HAVE_ to set PROD_ENV=true for that to be skipped now.

PR #338, opened 2018-11-13
Queue healthchecks

### LINKED ISSUE #313: Add health checks for DynamoDB, SQS

rebstar6:
Fixes #313

Changes in this pull request:
- healthcheck mysql queue, SQS queue

### DIFF
--- a/modules/mysql/src/main/scala/vinyldns/mysql/queue/MySqlMessageQueue.scala
+++ b/modules/mysql/src/main/scala/vinyldns/mysql/queue/MySqlMessageQueue.scala
@@ -24,6 +24,7 @@ import org.slf4j.LoggerFactory
 import scalikejdbc._
 import vinyldns.core.domain.record.RecordSetChange
 import vinyldns.core.domain.zone.{ZoneChange, ZoneCommand}
+import vinyldns.core.health.HealthCheck._
 import vinyldns.core.protobuf.ProtobufConversions
 import vinyldns.core.queue._
 import vinyldns.core.route.Monitored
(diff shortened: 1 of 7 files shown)


modules/mysql/src/main/scala/vinyldns/mysql/queue/MySqlMessageQueue.scala
```diff
@@ -88,6 +89,12 @@ class MySqlMessageQueue extends MessageQueue with Monitored with ProtobufConvers
       | WHERE id in (?)
     """.stripMargin
 
+  private final val HEALTH_CHECK =
+    sql"""
+         |SELECT 1
```
mitruly: Think the `|` should align with the `l` in `sql`.

modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueue.scala
```diff
@@ -157,6 +159,12 @@ class SqsMessageQueue(val queueUrl: String, val client: AmazonSQSAsync)
         )
       }
     }.as(())
+
+  def healthCheck(): HealthCheck =
+    sqsAsync[ListQueuesRequest, ListQueuesResult](
+      new ListQueuesRequest(),
```
mitruly: More of a PSA: Seems like this is a good way to check health status.

```
You can make occasional API calls (ListQueues, GetQueueUrl) to SQS in other regions to validate that you can access it.
```

(Source: https://forums.aws.amazon.com/thread.jspa?threadID=135828)

mitruly:
Restarted openjdk11 build since that errored out.

PR #359, opened 2018-11-19
Wait for zone and record set promises in portal before displaying record sets

### LINKED ISSUE #141: Records with same name as zone sometimes show up as a dotted host (invalid)
**Describe the bug**
Sometimes, when loading the main zone page with record sets, records with the same name as the zone (which is valid) get flagged as a dotted host.

**VinylDNS Version**
Latest

**Additional context**
In the records controller, we call both getZone and listRecordSets on page load. After listRecordSets returns, we determine if each recordset is dotted based on comparison with the zone name. If getZone returns last, this comparison fails for records that are the same as the zone name, resulting in the dotted host warning when it should not.

We probably still want these calls to happen at the same time, but need to wait to do the comparison until both have completed.

mitruly:
Need to wait for promises from multiple asynchronous calls to return before displaying records so that dotted host warning message does not erroneously appear.

Fixes #141.

Changes in this pull request:
- Update `controller.records.js` so that calls to `toDisplayRecord` wait for resolved zone and records promises.
- Update `batch-change-new-controller.js` and batch change view to remove unnecessary `event.preventDefault()` calls.

----
Note to reviewers: 
- Tested this by increasing the timeout of `loadZonesPromise = $timeout($scope.refreshZone, 0);` from `0` to `2000`.
- Information about the `$q` service/promise construction ca be found at: https://docs.angularjs.org/api/ng/service/$q

### DIFF
--- a/modules/portal/public/lib/controllers/controller.records.js
+++ b/modules/portal/public/lib/controllers/controller.records.js
@@ -15,7 +15,7 @@
  */
 
 angular.module('controller.records', [])
-    .controller('RecordsController', function ($scope, $timeout, $log, recordsService, groupsService, pagingService, utilityService) {
+    .controller('RecordsController', function ($scope, $timeout, $log, recordsService, groupsService, pagingService, utilityService, $q) {
 
     /**
       * Scope data initial setup
@@ -34,6 +34,9 @@ angular.module('controller.records', [])
     $scope.recordsetChanges = {};
     $scope.currentRecord = {};
     $scope.zoneInfo = {};
+
+    var loadZonesPromise;
+    var loadRecordsPromise;
 
     $scope.recordModalState = {
         CREATE: 0,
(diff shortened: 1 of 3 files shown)


modules/portal/public/lib/controllers/controller.records.js
```diff
@@ -391,17 +394,20 @@ angular.module('controller.records', [])
...
-            newRecords.push(recordsService.toDisplayRecord(record, $scope.zoneInfo.name));
-        });
-        $scope.records = newRecords;
-        if($scope.records.length > 0) {
-          $("td.dataTables_empty").hide();
-        } else {
-          $("td.dataTables_empty").show();
-        }
-    }
+        $q.all([loadZonesPromise, loadRecordsPromise])
```
rebstar6: did you look into testing this in controller.records.spec.js?

rebstar6:
added an explicit wait on the api side, and confirmed this works, so 👍 on the code. would be great if possible to get this unit tested

modules/portal/public/lib/controllers/controller.records.js
mitruly: I'll look into adding one now.

mitruly:
Having issues writing a test that works as expected, as the promises in `updateRecordDisplay` are not resolving as expected. Arguably this would be testing `$q.all()`, so not sure if a new test is super necessary.

PR #363, opened 2018-11-20
Implement MySQL user change repository

### LINKED ISSUE #360: MySQL UserChangeRepo
refer to #16

- Create MySQL UserChangeRepository that implements functionality of DynamoDBUserChangeRepository
- Create unit and integration tests
- Create flyway sql

mitruly:
Fixes #360.

Changes in this pull request:
- Update protocol buffer (protobuf).
- Implement MySQL user change repository.

### DIFF
--- a/modules/core/src/main/protobuf/VinylDNSProto.proto
+++ b/modules/core/src/main/protobuf/VinylDNSProto.proto
@@ -192,3 +192,12 @@ message User {
     optional string lastName = 9;
     optional string email = 10;
 }
+
+message UserChange {
+    required User newUser = 1;
+    required string madeByUserId = 2;
+    required int64 created = 3;
+    optional User oldUser = 4;
+    required string id = 5;
+    required string changeType = 6;
+}
--- a/modules/core/src/main/protobuf/VinylDNSProto.proto
+++ b/modules/core/src/main/protobuf/VinylDNSProto.proto
@@ -157,8 +157,8 @@ message GroupChange {
 }
 
 message SingleAddChange {
-    required int64 ttl = 1;
-    required RecordData recordData = 2;
+  required int64 ttl = 1;
+  required RecordData recordData = 2;
 }
 
 message SingleChangeData {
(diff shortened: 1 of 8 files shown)


modules/core/src/main/protobuf/VinylDNSProto.proto
```diff
@@ -192,3 +192,12 @@ message User {
     optional string lastName = 9;
     optional string email = 10;
 }
+
+message UserChange {
+    required User newUser = 1;
+    required string madeByUserId = 2;
+    required int64 created = 3;
+    optional User oldUser = 4;
```
pauljamescleary: please add optional arguments at the end (after all required)

modules/mysql/src/main/resources/db/migration/V3.9__UserChanges.sql
```diff
@@ -0,0 +1,17 @@
...
+*/
+
+CREATE TABLE user_change (
+  change_id CHAR(36) NOT NULL,
+  user_id CHAR(36) NOT NULL,
+  data BLOB NOT NULL,
+  created_timestamp BIGINT(13) NOT NULL,
+  PRIMARY KEY (change_id),
+  INDEX user_id_index (user_id),
+  INDEX created_timestamp_index (created_timestamp)
```
pauljamescleary: we are only ever getting by change id right now.  No need for the index.  If the index is not used, you are just wasting memory.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlUserChangeRepository.scala
```diff
@@ -0,0 +1,82 @@
...
+class MySqlUserChangeRepository
+    extends UserChangeRepository
+    with Monitored
+    with ProtobufConversions {
+  private final val logger = LoggerFactory.getLogger(classOf[MySqlUserChangeRepository])
+
+  private final val PUT_USER_CHANGE =
+    sql"""
+         |  REPLACE INTO user_change (change_id, user_id, data, created_timestamp)
+         |   VALUES ({changeId}, {userId}, {data}, {createdTimestamp})
```
pauljamescleary: I thought we landed on the side of `INSERT INTO x ON DUPLICATE KEY UPDATE` instead of using `REPLACE INTO`.

`REPLACE INTO` will do a delete and then an insert, whereas the other will do an update.

mitruly: Good call. Should I make a card to update the other MySQL repos to use `INSERT INTO` instead of `REPLACE INTO` as well?

modules/mysql/src/main/resources/db/migration/V3.9__UserChanges.sql
mitruly: Good point. Removing both of these indices.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlUserChangeRepository.scala
```diff
@@ -0,0 +1,85 @@
...
+    extends UserChangeRepository
+    with Monitored
+    with ProtobufConversions {
+  private final val logger = LoggerFactory.getLogger(classOf[MySqlUserChangeRepository])
+
+  private final val PUT_USER_CHANGE =
+    sql"""
+         |  INSERT INTO user_change (change_id, user_id, data, created_timestamp)
+         |       VALUES ({changeId}, {userId}, {data}, {createdTimestamp}) ON DUPLICATE KEY
+         |       UPDATE user_id=VALUES(user_id),
```
rebstar6: do we actually need to do any updates on duplicate key? i feel like no

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlUserChangeRepository.scala
```diff
@@ -0,0 +1,85 @@
...
+            .bind(changeId)
+            .map(toUserChange(1))
+            .first()
+            .apply()
+        }
+      }
+    }
+
+  def save(change: UserChange): IO[UserChange] =
+    monitor("repo.UserChanges.save") {
```
rebstar6: this is super nitpicky, but id monitor all the repos as singular, thats how weve been naming them (repo.UserChange.save)

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlUserChangeRepository.scala
mitruly: I think this is the "last-write-wins" scenario - @pauljamescleary would have to confirm.

rebstar6: hrmm, zonechanges still uses `REPLACE INTO` (wrong), groupchanges is like you have it here, and recordsetchanges is `INSERT IGNORE INTO`

double hrmm. weird inconsistency that happened

PR #372, opened 2018-11-26
Adding version to login

pauljamescleary:
- Added a config entry for the portal `version`
- Added a `Meta` type to hold the version
- Added an implicit val to the `FrontendController` following the pattern of `CustomLinks`
- Show the version on the login screen


pauljamescleary:
image

modules/portal/app/views/login.scala.html
```diff
@@ -42,7 +42,7 @@ <h3 class="vinyldns-login-title">Welcome to VinylDNS</h3>
                 </div>
                 <div class="vinyldns-login-footer">
                     <div class="pull-left">
-                        VinylDNS, for all your DNS needs.
+                        VinylDNS (version @{meta.version})
```
britneywright: can we wrap this in an if statement to display if the version is given, rather than the version unknown?

mitruly: Also, is VinylDNS no longer for all our DNS needs?

pauljamescleary: > can we wrap this in an if statement to display if the version is given, rather than the version unknown?

It should always be set.  I am working on making sure that it is always there.  For internal releases, it will be there for sure.  It is a little trickier for the public images.

pauljamescleary: > Also, is VinylDNS no longer for all our DNS needs?

Heh, Not a fan of that tagline, idk

pauljamescleary:
This has been updated now so there will always be a version in there for the portal and the api, and they will be the same.  We have to have an uncomfortable default of "unset", as the actual setting of the thing is done in a separate process through sbt native packager, but it will always be there now.

rebstar6:
should sbt run set this? its not for me

pauljamescleary:
> should sbt run set this? its not for me

@rebstar6 you will not see this in `sbt run`.  The version is laid down during the build of the docker images (the build.sbt changes are in apiDockerSettings and portalDockerSettings).

pauljamescleary:
image

pauljamescleary:
> should sbt run set this? its not for me

@rebstar6 The intent is for _users_ of the system to see their version clearly on the login screen, not for developers to see when working locally.  However, I was able to add the jvm params in the portal _and_ the api to set the version correctly.  Updated PR and added screen shots.

build.sbt
```diff
@@ -332,6 +341,9 @@ lazy val portal = (project in file("modules/portal")).enablePlugins(PlayScala, A
     routesGenerator := InjectedRoutesGenerator,
     coverageExcludedPackages := "<empty>;views.html.*;router.*",
     javaOptions in Test += "-Dconfig.file=conf/application-test.conf",
+    
+    // ads the version when working locally with sbt run
```
mitruly: `ads` → `adds`

PR #373, opened 2018-11-26
Remove zone history check

### LINKED ISSUE #362: Remove calls to /zones/id/history from the python tests
**Describe the bug**
vinyldns_python has a function `get_zone_history` calling /zones/id/history. That endpoint doesnt exist in the api.

We are calling the above in `wait_until_zone_change_status`, so any calls to that function are waiting the full time, not until the status has changed.

In short, we need to remove get_zone_history and update or remove anything calling it.

britneywright:
Fixes #362

Changes in this pull request:
- removed `get_zone_history method` and `get_zone_change` methods
- update `wait_until_zone_change_status` method to check the status of the most recent zone change

### DIFF
--- a/modules/api/functional_test/live_tests/zones/create_zone_test.py
+++ b/modules/api/functional_test/live_tests/zones/create_zone_test.py
@@ -63,7 +63,7 @@ def test_create_zone_success(shared_zone_test_context):
         }
         result = client.create_zone(zone, status=202)
         result_zone = result['zone']
-        client.wait_until_zone_change_status(result, 'Synced')
+        client.wait_until_zone_change_status_synced(result)
 
         get_result = client.get_zone(result_zone['id'])
 
--- a/modules/api/functional_test/live_tests/zones/create_zone_test.py
+++ b/modules/api/functional_test/live_tests/zones/create_zone_test.py
@@ -63,8 +63,7 @@ def test_create_zone_success(shared_zone_test_context):
         }
         result = client.create_zone(zone, status=202)
         result_zone = result['zone']
-        client.wait_until_zone_change_status_synced(result)
-        client.wait_until_zone_synced(result)
+        client.wait_until_zone_active(result)
 
         get_result = client.get_zone(result_zone['id'])
 
--- a/modules/api/functional_test/live_tests/zones/create_zone_test.py
+++ b/modules/api/functional_test/live_tests/zones/create_zone_test.py
@@ -65,6 +65,8 @@ def test_create_zone_success(shared_zone_test_context):
         result_zone = result['zone']
         client.wait_until_zone_change_status_synced(result)
 
+        time.sleep(1)
+
         get_result = client.get_zone(result_zone['id'])
 
         get_zone = get_result['zone']
--- a/modules/api/functional_test/live_tests/zones/create_zone_test.py
+++ b/modules/api/functional_test/live_tests/zones/create_zone_test.py
@@ -35,57 +35,57 @@ records_in_dns = [
      'type': u'A',
      'records': [{u'address': u'6.6.6.6'}]}]
 
-# def test_create_zone_success(shared_zone_test_context):
-#     """
-#     Test successfully creating a zone
-#     """
-#     client = shared_zone_test_context.ok_vinyldns_client
-#     result_zone = None
-#     try:
-#         zone_name = 'one-time'
-#
-#         zone = {
-#             'name': zone_name,
-#             'email': 'test@test.com',
-#             'adminGroupId': shared_zone_test_context.ok_group['id'],
-#             'connection': {
-#                 'name': 'vinyldns.',
-#                 'keyName': VinylDNSTestContext.dns_key_name,
-#                 'key': VinylDNSTestContext.dns_key,
-#                 'primaryServer': VinylDNSTestContext.dns_ip
-#             },
-#             'transferConnection': {
-#                 'name': 'vinyldns.',
-#                 'keyName': VinylDNSTestContext.dns_key_name,
-#                 'key': VinylDNSTestContext.dns_key,
-#                 'primaryServer': VinylDNSTestContext.dns_ip
-#             }
-#         }
-#         result = client.create_zone(zone, status=202)
-#         result_zone = result['zone']
-#         client.wait_until_zone_change_status_synced(result)
-#
-#         get_result = client.get_zone(result_zone['id'])
-#
-#         get_zone = get_result['zone']
-#         assert_that(get_zone['name'], is_(zone['name']+'.'))
-#         assert_that(get_zone['email'], is_(zone['email']))
-#         assert_that(get_zone['adminGroupId'], is_(zone['adminGroupId']))
-#         assert_that(get_zone['latestSync'], is_not(none()))
-#         assert_that(get_zone['status'], is_('Active'))
-#
-#         # confirm that the recordsets in DNS have been saved in vinyldns
-#         recordsets = client.list_recordsets(result_zone['id'])['recordSets']
-#
-#         assert_that(len(recordsets), is_(7))
-#         for rs in recordsets:
-#             small_rs = dict((k, rs[k]) for k in ['name', 'type', 'records'])
-#             small_rs['records'] = sorted(small_rs['records'])
-#             assert_that(records_in_dns, has_item(small_rs))
-#
-#     finally:
-#         if result_zone:
-#             client.abandon_zones([result_zone['id']], status=202)
+def test_create_zone_success(shared_zone_test_context):
+    """
+    Test successfully creating a zone
+    """
+    client = shared_zone_test_context.ok_vinyldns_client
+    result_zone = None
+    try:
+        zone_name = 'one-time'
+
+        zone = {
+            'name': zone_name,
+            'email': 'test@test.com',
+            'adminGroupId': shared_zone_test_context.ok_group['id'],
+            'connection': {
+                'name': 'vinyldns.',
+                'keyName': VinylDNSTestContext.dns_key_name,
+                'key': VinylDNSTestContext.dns_key,
+                'primaryServer': VinylDNSTestContext.dns_ip
+            },
+            'transferConnection': {
+                'name': 'vinyldns.',
+                'keyName': VinylDNSTestContext.dns_key_name,
+                'key': VinylDNSTestContext.dns_key,
+                'primaryServer': VinylDNSTestContext.dns_ip
+            }
+        }
+        result = client.create_zone(zone, status=202)
+        result_zone = result['zone']
+        client.wait_until_zone_change_status_synced(result)
+
+        get_result = client.get_zone(result_zone['id'])
+
+        get_zone = get_result['zone']
+        assert_that(get_zone['name'], is_(zone['name']+'.'))
+        assert_that(get_zone['email'], is_(zone['email']))
+        assert_that(get_zone['adminGroupId'], is_(zone['adminGroupId']))
+        assert_that(get_zone['latestSync'], is_not(none()))
+        assert_that(get_zone['status'], is_('Active'))
+
+        # confirm that the recordsets in DNS have been saved in vinyldns
+        recordsets = client.list_recordsets(result_zone['id'])['recordSets']
+
+        assert_that(len(recordsets), is_(7))
+        for rs in recordsets:
+            small_rs = dict((k, rs[k]) for k in ['name', 'type', 'records'])
+            small_rs['records'] = sorted(small_rs['records'])
+            assert_that(records_in_dns, has_item(small_rs))
+
+    finally:
+        if result_zone:
+            client.abandon_zones([result_zone['id']], status=202)
 
 
 @pytest.mark.skip_production
--- a/modules/api/functional_test/live_tests/zones/create_zone_test.py
+++ b/modules/api/functional_test/live_tests/zones/create_zone_test.py
@@ -65,8 +65,6 @@ def test_create_zone_success(shared_zone_test_context):
         result_zone = result['zone']
         client.wait_until_zone_change_status_synced(result)
 
-        time.sleep(1)
-
         get_result = client.get_zone(result_zone['id'])
 
         get_zone = get_result['zone']
--- a/modules/api/functional_test/live_tests/zones/create_zone_test.py
+++ b/modules/api/functional_test/live_tests/zones/create_zone_test.py
@@ -35,57 +35,57 @@ records_in_dns = [
      'type': u'A',
      'records': [{u'address': u'6.6.6.6'}]}]
 
-def test_create_zone_success(shared_zone_test_context):
-    """
-    Test successfully creating a zone
-    """
-    client = shared_zone_test_context.ok_vinyldns_client
-    result_zone = None
-    try:
-        zone_name = 'one-time'
-
-        zone = {
-            'name': zone_name,
-            'email': 'test@test.com',
-            'adminGroupId': shared_zone_test_context.ok_group['id'],
-            'connection': {
-                'name': 'vinyldns.',
-                'keyName': VinylDNSTestContext.dns_key_name,
-                'key': VinylDNSTestContext.dns_key,
-                'primaryServer': VinylDNSTestContext.dns_ip
-            },
-            'transferConnection': {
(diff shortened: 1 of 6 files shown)


modules/api/functional_test/vinyldns_python.py
```diff
@@ -704,22 +671,20 @@ def wait_until_zone_change_status(self, zone_change, expected_status):
         """
         Waits until the zone change status matches the expected status
```
rebstar6: looking at usage, we are only ever calling this with status='Synced'. Because of that, i would probably change this to `wait_until_zone_change_synced`, meaning you can cut out early if the status is `Failed`

modules/api/functional_test/vinyldns_python.py
```diff
@@ -704,22 +671,20 @@ def wait_until_zone_change_status(self, zone_change, expected_status):
...
+        def change_id_match(change):
+            return change[u'id'] == zone_change[u'id']
 
+        while change[u'status'] != expected_status and retries > 0:
+            latest_changes = self.list_zone_changes(change['zone']['id'])
+            if type(latest_changes).__name__  == 'dict' and latest_changes[u'zoneChanges']:
+                matching_changes = filter(change_id_match, latest_changes[u'zoneChanges'])
                 if len(matching_changes) > 0:
                     change = matching_changes[0]
+            else:
```
rebstar6: shouldnt need this else, its not doing anything

modules/api/functional_test/vinyldns_python.py
```diff
@@ -704,22 +671,20 @@ def wait_until_zone_change_status(self, zone_change, expected_status):
...
 
-            if u'zoneChanges' in history:
-                zone_changes = history[u'zoneChanges']
-                matching_changes = filter(change_id_match, zone_changes)
+        def change_id_match(change):
+            return change[u'id'] == zone_change[u'id']
 
+        while change[u'status'] != expected_status and retries > 0:
+            latest_changes = self.list_zone_changes(change['zone']['id'])
+            if type(latest_changes).__name__  == 'dict' and latest_changes[u'zoneChanges']:
```
rebstar6: do we need this type check? id think ` if u'zoneChanges' in latest_changes:` would be ok

modules/api/functional_test/vinyldns_python.py
```diff
@@ -704,22 +671,20 @@ def wait_until_zone_change_status(self, zone_change, expected_status):
...
-
         change = zone_change
         retries = MAX_RETRIES
-        while change[u'status'] != expected_status and retries > 0:
-            history = self.get_zone_history(zone_change[u'zone'][u'id'])
 
-            if u'zoneChanges' in history:
-                zone_changes = history[u'zoneChanges']
-                matching_changes = filter(change_id_match, zone_changes)
+        def change_id_match(change):
```
mitruly: This isn't explicitly needed and `change` here shadows the variable in this function. Piggybacking off of what @rebstar6 said, this is a bit cleaner:

```
    def wait_until_zone_change_status(self, zone_change, expected_status):
        """
        Waits until the zone change status matches the expected status
        """
        change = zone_change
        retries = MAX_RETRIES

        while change[u'status'] != expected_status and retries > 0:
            latest_changes = self.list_zone_changes(change['zone']['id'])
            if u'zoneChanges' in latest_changes:
                matching_changes = filter(lambda c: c[u'id'] == zone_change[u'id'], latest_changes[u'zoneChanges'])
                if len(matching_changes) > 0:
                    change = matching_changes[0]
            time.sleep(RETRY_WAIT)
            retries -= 1

        return change[u'status'] == expected_status
```

modules/api/functional_test/vinyldns_python.py
mitruly: +1 here. The scripting aspect of Python allows us to simply check if a value is defined before using it rather than checking a type.

modules/api/functional_test/vinyldns_python.py
```diff
@@ -700,30 +667,23 @@ def wait_until_recordset_deleted(self, zone_id, record_set_id, **kwargs):
...
+        latest_change = zone_change
         retries = MAX_RETRIES
-        while change[u'status'] != expected_status and retries > 0:
-            history = self.get_zone_history(zone_change[u'zone'][u'id'])
-
-            if u'zoneChanges' in history:
-                zone_changes = history[u'zoneChanges']
-                matching_changes = filter(change_id_match, zone_changes)
 
+        while latest_change[u'status'] != 'Synced' and retries > 0:
```
rebstar6: ^^ OR `latest_change[u'status'] != 'Failed'`. if its failed we can cut out

modules/api/functional_test/vinyldns_python.py
```diff
@@ -700,30 +667,23 @@ def wait_until_recordset_deleted(self, zone_id, record_set_id, **kwargs):
...
+            if u'zoneChanges' in changes:
+                matching_changes = filter(lambda change: change[u'id'] == zone_change[u'id'], changes[u'zoneChanges'])
                 if len(matching_changes) > 0:
-                    change = matching_changes[0]
+                    latest_change = matching_changes[0]
             time.sleep(RETRY_WAIT)
             retries -= 1
 
-        return change[u'status'] == expected_status
+        return latest_change[u'status'] == 'Synced'
```
rebstar6: i kinda think we should `assert(latest_change[u'status'] == 'Synced')` here (or whatever that syntax is). weirdly, we call this wait_for but it does not fail if the condition is not met, just returns false

britneywright:
Looks like the func tests are now taking 15 minutes to complete in Travis https://travis-ci.org/vinyldns/vinyldns/builds/463580219. It was down to 11 minutes before the most recent commits. For added reference Master takes about 13 minutes to run the func tests now.

modules/api/functional_test/vinyldns_python.py
britneywright: for some reason adding the assertion makes `test_sync_zone_success` fail in Travis (https://travis-ci.org/vinyldns/vinyldns/jobs/463879028#L9042). It never fails locally though. But it didn't fail when I also had the `latest_change[u'status'] != 'Failed'` in the while statement of the function though, maybe it's a timing issue?

modules/api/functional_test/live_tests/zones/create_zone_test.py
```diff
@@ -63,7 +63,9 @@ def test_create_zone_success(shared_zone_test_context):
         }
         result = client.create_zone(zone, status=202)
         result_zone = result['zone']
-        client.wait_until_zone_change_status(result, 'Synced')
+        client.wait_until_zone_change_status_synced(result)
+
+        time.sleep(1)
```
rebstar6: why are we sleeping?

britneywright: this was one of the tests that kept randomly failing in Travis. I thought it might be a timing issue.

modules/api/functional_test/vinyldns_python.py
```diff
@@ -764,6 +724,23 @@ def wait_until_zone_exists(self, zone_change, **kwargs):
 
         return response == 200
 
+    def wait_until_zone_synced(self, zone_change):
```
mitruly: Can we consolidate this with `wait_until_zone_change_status_synced`?

modules/api/functional_test/live_tests/zones/create_zone_test.py
```diff
@@ -63,7 +63,8 @@ def test_create_zone_success(shared_zone_test_context):
         }
         result = client.create_zone(zone, status=202)
         result_zone = result['zone']
-        client.wait_until_zone_change_status(result, 'Synced')
+        client.wait_until_zone_change_status_synced(result)
```
rebstar6: this is still super odd. were you able to get to the bottom of why the change is appearing completed before the zone has finished syncing?

rebstar6: ahh i see it. look here https://github.com/vinyldns/vinyldns/blob/5a02347cfb5a8bd92b503586d7ca992a7dabf1f1/modules/api/src/main/scala/vinyldns/api/backend/CommandHandler.scala#L139

That 1st call to zoneChangeProcessor(sync) not only saves the zone in syncing status, but also saves the change as synced even though it's not complete. hrmm. should probably dig more, most likely that initial save should save the zone status, but store the change as pending still.

rebstar6: ^^ said, i think what were doing to address that in this PR is reasonable. i do wonder if it's causing any other transient failures though. can you make a card to investigate further?

modules/api/functional_test/vinyldns_python.py
```diff
@@ -764,6 +724,23 @@ def wait_until_zone_exists(self, zone_change, **kwargs):
...
+        """
+        Waits a period of time for the zone sync to complete.
+
+        :param zone_change: the create zone change for the zone that has been created.
+        """
+        zone_id = zone_change[u'zone'][u'id']
+        retries = MAX_RETRIES
+        something = self.get_zone(zone_id)
```
rebstar6: probably wanna call this something other than `something` :P

modules/api/functional_test/live_tests/zones/create_zone_test.py
britneywright: Correct me if I'm misunderstanding, but in the `processChangeRequests` https://github.com/vinyldns/vinyldns/blob/master/modules/api/src/main/scala/vinyldns/api/backend/CommandHandler.scala#L139 we change the zone_change status to Synced, then actually sync the zone in the next step.

rebstar6: thats what im seeing too - we save the change as synced 2x. on the zone status side, i believe the 1st save is saving with `zone status=Syncing` and the final one is `zone status=Active`. so a wait for zone status=Active (which should occur at the same time the latestSync happens) should work.

all this said, i think remove the wait for change here, and just update `wait_until_zone_synced` to be called `wait_until_zone_active` which waits for the full create (create + sync) to occur.

britneywright:
`test_zone_sync_success` is failing here https://github.com/vinyldns/vinyldns/blob/master/modules/api/functional_test/live_tests/zones/sync_zone_test.py#L157 instead of returning a 403 on the zone sync it returns a 202. My thought is the previous sync is happening fast enough that the next sync is allowed to go through as well.

modules/api/functional_test/vinyldns_python.py
nimaeskandary: Potential for race conditions where the latest change in the db is already synced and the last request is slow to be processed?

nimaeskandary: nevermind didn't notice its going after a specific zone change

nimaeskandary: Plusone-ing rebeccas comment on exiting loop early if status is failed, `while (latest_change[u'status'] != 'Synced' and retries > 0) and latest_change[u'status'] != 'Failed':`

nimaeskandary:
thumbs except rebeccas comment regarding early exit on sync status

modules/api/functional_test/vinyldns_python.py
britneywright: I forgot to comment or maybe I put it in an outdated place, for some reason when I added the latest_change[u'status'] != 'Failed' the func tests took 5 minutes longer. That happened multiple times so removed it.

britneywright: ah yeah that was what this comment was in reference too https://github.com/vinyldns/vinyldns/pull/373#issuecomment-444347708

nimaeskandary: @britneywright with an `OR` I could see it adding time depending on how its combined with retries, have you tried with an `AND`?

britneywright: We have no tests that are set up to have a Failed zone change status so we don't find ourselves getting a Failed and don't need to exit early. If we're still looping it's because the status is still Pending and not Synced yet.

britneywright: I added it, same run time, all the tests still pass.

modules/api/functional_test/live_tests/zones/sync_zone_test.py
```diff
@@ -151,10 +150,7 @@ def test_sync_zone_success(shared_zone_test_context):
 
         # sync again
         change = client.sync_zone(zone['id'], status=202)
-        client.wait_until_zone_change_status(change, 'Synced')
-
-        # confirm cannot again sync without waiting
-        client.sync_zone(zone['id'], status=403)
+        client.wait_until_zone_change_status_synced(change)
```
rebstar6: even this technically isnt waiting enough logically. could be moving on too early

modules/api/functional_test/vinyldns_python.py
```diff
@@ -764,6 +726,23 @@ def wait_until_zone_exists(self, zone_change, **kwargs):
...
+        """
+        Waits a period of time for the zone sync to complete.
+
+        :param zone_change: the create zone change for the zone that has been created.
+        """
+        zone_id = zone_change[u'zone'][u'id']
+        retries = MAX_RETRIES
+        zone = self.get_zone(zone_id)
+
+        while u'latestSync' not in zone and retries > 0:
```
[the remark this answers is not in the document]
britneywright: change is here https://github.com/vinyldns/vinyldns/pull/373/files#diff-db0cbd6424228193b3a89af595a286a8R744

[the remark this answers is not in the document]
britneywright: ah I see what you're saying, get the zone in the test rather than in this function.

modules/api/functional_test/live_tests/zones/sync_zone_test.py
rebstar6: remove the wait_until_zone_change_status_synced, wait_for_zone_sync_to_complete will cover it

modules/api/functional_test/vinyldns_python.py
```diff
@@ -2,6 +2,7 @@
 import time
 import logging
 import collections
+import datetime
```
rebstar6: dont think this is used

modules/api/functional_test/vinyldns_python.py
```diff
@@ -700,30 +667,23 @@ def wait_until_recordset_deleted(self, zone_id, record_set_id, **kwargs):
...
+        latest_change = zone_change
         retries = MAX_RETRIES
-        while change[u'status'] != expected_status and retries > 0:
-            history = self.get_zone_history(zone_change[u'zone'][u'id'])
-
-            if u'zoneChanges' in history:
-                zone_changes = history[u'zoneChanges']
-                matching_changes = filter(change_id_match, zone_changes)
 
+        while (latest_change[u'status'] != 'Synced' and retries > 0) and latest_change[u'status'] != 'Failed':
```
rebstar6: parens shouldnt matter for `and`

modules/api/functional_test/vinyldns_python.py
```diff
@@ -736,7 +696,9 @@ def wait_until_zone_deleted(self, zone_id, **kwargs):
         retries = MAX_RETRIES
         url = urljoin(self.index_url, u'/zones/{0}'.format(zone_id))
         response, data = self.make_request(url, u'GET', self.headers, not_found_ok=True, status=(200, 404), **kwargs)
+        print("RESPONSE: " + str(response))
         while response != 404 and retries > 0:
+            print("RESPONSE: " + str(response))
```
rebstar6: do we need these prints? ^^

modules/api/functional_test/vinyldns_python.py
```diff
@@ -764,6 +726,22 @@ def wait_until_zone_exists(self, zone_change, **kwargs):
...
+    def wait_until_zone_active(self, zone_id):
+        """
+        Waits a period of time for the zone sync to complete.
+
+        :param zone_id: the ID for the zone.
+        """
+        retries = MAX_RETRIES
+        zone_request = self.get_zone(zone_id)
+
+        while ('status' not in zone_request or zone_request[u'zone'][u'status'] != 'Active') and retries > 0:
```
rebstar6: should this be
`'zone' not in zone_request`
or
`'status' not in zone_request[u'zone']`
?

PR #375, opened 2018-11-26
portal health endpoint

### LINKED ISSUE #339: Add portal healthcheck endpoint with DB check
split off of #313 
Similar to the API, we need a /health endpoint in the portal. At this endpoint, we should have:

- The portal should have the same database healthchecking as the api for the configured database. This should just have to be hooked in properly, see https://github.com/vinyldns/vinyldns/pull/325
- a connectivity check for LDAP (not handled in this card)

For this card, just setup the endpoint with the database check

rebstar6:
Fixes #339

### DIFF
--- a/modules/portal/app/controllers/HealthController.scala
+++ b/modules/portal/app/controllers/HealthController.scala
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2018 Comcast Cable Communications Management, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package controllers
+
+import javax.inject.{Inject, Singleton}
+import play.api.mvc.{AbstractController, Action, AnyContent, ControllerComponents}
+import vinyldns.core.health.HealthService
+
+@Singleton
+class HealthController @Inject()(
+    // configuration: Configuration,
+    //  authenticator: Authenticator,
+    //  userAccountAccessor: UserAccountAccessor,
+    //   wsClient: WSClient,
+    components: ControllerComponents,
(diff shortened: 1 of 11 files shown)


modules/portal/app/controllers/HealthService.scala
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+package controllers
+
+import cats.effect.{ContextShift, IO}
+import cats.implicits._
+import org.slf4j.LoggerFactory
+import vinyldns.core.health.HealthCheck.{HealthCheck, HealthCheckError}
+
+class HealthService(healthChecks: List[HealthCheck]) {
```
rebstar6: this is straight copy/paste from the api. i guess i will move into core? a little weird since we dont have services in there, but i think that ultimately will be the move

modules/portal/app/controllers/VinylDNS.scala
```diff
@@ -125,6 +126,17 @@ class VinylDNS @Inject()(
     processLogin(username, password)
   }
 
+  def health(): Action[AnyContent] = Action { implicit request =>
+    healthService
+      .checkHealth()
+      .map {
+        case head :: _ =>
+          InternalServerError("There was an internal server error.").withHeaders(cacheHeaders: _*)
+        case _ => Ok("OK").withHeaders(cacheHeaders: _*)
```
rebstar6: not certain that we need the withHeaders

mitruly: I think it's better to do this (and not cache the response) since users could potentially view stale health checks.

Source: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control

modules/portal/app/controllers/HealthController.scala
```diff
@@ -0,0 +1,38 @@
...
+@Singleton
+class HealthController @Inject()(components: ControllerComponents, healthService: HealthService)
+    extends AbstractController(components)
+    with CacheHeader {
+
+  def health(): Action[AnyContent] = Action { implicit request =>
+    healthService
+      .checkHealth()
+      .map {
+        case head :: _ =>
```
mitruly: `head` isn't used here, so we could do:

```
        case _ :: _ =>
          InternalServerError("There was an internal server error.").withHeaders(cacheHeaders: _*)
        case _ => Ok("OK").withHeaders(cacheHeaders: _*)
```

I slightly prefer:
```
        case Nil => Ok("OK").withHeaders(cacheHeaders: _*)
        case _ =>
          InternalServerError("There was an internal server error.").withHeaders(cacheHeaders: _*)
```

Both worked locally for me.

Issue #235, opened 2018-09-24
Use ip4s for CIDR and IP validating


rebstar6:
https://github.com/Comcast/ip4s

This is something we could use both for our CIDR rules logic, as well as for any ip validations we do, which is awesome. Some thoughts on where:

- A record data should be IPv4 type
- AAAA record data should be IPv6 type
- Batch PTR input should be IPAddress
- Look at ReverseZoneHelpers file for misc cidr things (classless reverse zone delegations, and ACL rules use these)

Scope of this card includes:
- Remove dependency on orchard CIDR library, use ip4s, make all tests pass.
- Replace validations with ip4s where needed
- Check the portal, there may be no changes there

pauljamescleary:
Hundo-P - we should normalize on IP4S, we are already using it in most places and should do it in other places as well.  Rather low hanging fruit, could touch a fair bit of code, shouldn't be too complex

PR #376, opened 2018-11-26
High Value Domains

### LINKED ISSUE #254: High Value Domains
We need to support the idea of "blacklists", at a global level and a zone level.  However, there are instances where you don't want anyone to ever modify certain domains.

Allow admins to configure a list of high value domains.  These domains should _never_ be managed through VinylDNS.  All attempts, even super users, should be prevented to make any changes to these DNS records.

It maybe sufficient to have this just be configuration driven initially until we have an "Admin" section of VinylDNS.

**Design**
1. Add a config section called "high-value-domains".  Have it be a list of FQDN based regular expressions.
1. Add a validation for any record update.  If the record fqdn (record label + zone name) is a high value domain, immediately reject.  This should just be in `RecordSetValidations` and `BatchChangeValidations`
1. Be sure to call the new `isHighValueDomain` for all adds, creates, deletes in the `RecordSetService`and `BatchChangeService`

nimaeskandary:
Fixes #254 .

Changes in this pull request:
- Adds a section to config for high value domain regexes
- Checks if recordset name conflict with high value domains in batch changes, as well CRUD recordset services


modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetValidations.scala
```diff
@@ -192,4 +193,12 @@ object RecordSetValidations {
   private def isOriginRecord(recordSetName: String, zoneName: String): Boolean =
     recordSetName == "@" || omitTrailingDot(recordSetName) == omitTrailingDot(zoneName)
 
+  def isNotHighValueDomain(recordName: String, zoneName: String): Either[Throwable, Unit] = {
```
nimaeskandary: copy how the approvedNameServer is converted from a validNel to an Either

modules/api/functional_test/live_tests/zones/create_zone_test.py
```diff
@@ -466,8 +466,6 @@ def test_user_cannot_create_zone_with_failed_validations(shared_zone_test_contex
     }
 
     result = shared_zone_test_context.ok_vinyldns_client.create_zone(zone, status=400)
-    import json
```
nimaeskandary: stray print?

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneRecordValidations.scala
```diff
@@ -16,18 +16,24 @@
...
+import cats.data._
+import vinyldns.api.domain.{DomainValidationError, HighValueDomainError}
 import vinyldns.core.domain.record.{NSData, RecordSet, RecordType}
 
 import scala.util.matching.Regex
 
 object ZoneRecordValidations {
 
+  /* Checks to see if an individual string is part of the regex list */
+  def isStringInRegexList(regexList: List[Regex], string: String): Boolean =
```
nimaeskandary: made this more generic to accommodate for the approved name serves and high value domain functions

modules/api/functional_test/live_tests/batch/create_batch_change_test.py
```diff
@@ -625,6 +625,52 @@ def test_create_batch_change_with_missing_input_name_fails(shared_zone_test_cont
     assert_error(errors, error_messages=["Missing BatchChangeInput.changes.inputName"])
 
 
+def test_create_batch_change_with_high_value_domain_fails(shared_zone_test_context):
```
britneywright: These two tests could be combined into one right? with both an add and delete in the changes.

modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetValidations.scala
```diff
@@ -192,4 +193,12 @@ object RecordSetValidations {
...
+    val name = DnsConversions.recordDnsName(recordName, zoneName).toString
```
mitruly: I think the name here has to be the record FQDN (see second item from https://github.com/vinyldns/vinyldns/issues/254):

```
2. Add a validation for any record update. If the record fqdn (record label + zone name) is a high value domain, immediately reject. This should just be in RecordSetValidations and BatchChangeValidations
```

We should make sure that we have a test for record name + zone name combo for High Value Domain (eg. if we ban `bad.ok.*`, record `bad.ok.` should fail but `bad.old-vinyldns2.` should succeed.

modules/api/functional_test/live_tests/batch/create_batch_change_test.py
mitruly: Arguably they should remain separate, since they would otherwise be interpreted as an update; this does bring up a valid point that we should test update.

modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetValidations.scala
nimaeskandary: I can add some more tests, but that `DnsConversions.recordDnsName` is constructing the full FQDN, and can be seen by tests in their output, for example, for a recordset create, it can be seen in the error message the full record name is `dont-touch-me.ok.`

mitruly: I took another look and agree with you here about the implementation. Could arguably just update existing tests to be a combination of the record and zone name rather than creating new tests.

modules/api/src/it/resources/application.conf
```diff
@@ -10,6 +10,10 @@ vinyldns {
     "172.17.42.1."
   ]
 
+  high-value-domains = [
```
rebstar6: shouldnt need this here if its in ref.conf

modules/api/functional_test/live_tests/batch/create_batch_change_test.py
britneywright: The input names just have to match the regex, they don't have to be exactly the same, except for the update example. So could test add, delete and update in one batch change test.

modules/api/functional_test/live_tests/batch/create_batch_change_test.py
```diff
@@ -625,6 +625,52 @@ def test_create_batch_change_with_missing_input_name_fails(shared_zone_test_cont
...
+    """
+    Test creating a batch change with a high value domain as an inputName fails
+    """
+    client = shared_zone_test_context.ok_vinyldns_client
+    batch_change_input = {
+        "comments": "this is optional",
+        "changes": [
+            {
+                "changeType": "Add",
```
rebstar6: `get_change_A_AAAA_json("dont-touch-me.example.bar.", address="4.5.6.7")`

modules/api/functional_test/live_tests/batch/create_batch_change_test.py
```diff
@@ -625,6 +625,52 @@ def test_create_batch_change_with_missing_input_name_fails(shared_zone_test_cont
...
+def test_delete_batch_change_with_high_value_domain_fails(shared_zone_test_context):
+    """
+    Test creating a batch change with a high value domain as an inputName fails
+    """
+    client = shared_zone_test_context.ok_vinyldns_client
+    batch_change_input = {
+        "comments": "this is optional",
+        "changes": [
+            {
+                "changeType": "DeleteRecordSet",
```
rebstar6: same here, use the helpers

modules/api/functional_test/live_tests/batch/create_batch_change_test.py
rebstar6: +1 @britneywright - can have one test with 
```
dont-touch-me-add.example.bar
dont-touch-me-update.example.bar
dont-touch-me-delete.example.bar
something-valid.ok
```

modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
```diff
@@ -55,6 +55,8 @@ object VinylDNSConfig {
   lazy val system: ActorSystem = ActorSystem("VinylDNS", VinylDNSConfig.config)
   lazy val approvedNameServers: List[Regex] =
     vinyldnsConfig.getStringList("approved-name-servers").asScala.toList.map(n => n.r)
+  lazy val highValueDomains: List[Regex] =
```
rebstar6: does 
`loadConfigF[IO, List[String]](vinyldnsConfig, "high-value-domains").map(lst => lst.map(_.r))`
work here? getStringList isnt ideal, should IO that

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
```diff
@@ -89,9 +91,12 @@ class BatchChangeValidations(changeLimit: Int, accessValidation: AccessValidatio
   def validateInputName(change: ChangeInput): SingleValidation[Unit] =
     change.typ match {
```
rebstar6: since its everywhere, add at the end

```
  def validateInputName(change: ChangeInput): SingleValidation[Unit] = {
    val typedChecks = change.typ match {
      case A | AAAA | MX =>
        validateHostName(change.inputName).asUnit |+| notInReverseZone(change)
      case CNAME | TXT =>
        validateHostName(change.inputName).asUnit
      case PTR =>
        validatePtrIp(change.inputName)
      case other => InvalidBatchRecordType(other.toString).invalidNel[Unit]
    }

    typedChecks |+| isNotHighValueDomain(change.inputName)
  }
```

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneRecordValidations.scala
```diff
@@ -64,4 +70,13 @@ object ZoneRecordValidations {
     }
     validations.sequence
   }
+
+  def isNotHighValueDomain(
+      highValueDomainList: List[Regex],
+      name: String): ValidatedNel[DomainValidationError, Unit] =
```
rebstar6: call name `fqdn`

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneRecordValidations.scala
```diff
@@ -64,4 +70,13 @@ object ZoneRecordValidations {
...
+    if (!ZoneRecordValidations.isStringInRegexList(highValueDomainList, name)) {
```
rebstar6: youre in ZoneRecordValidations, can just call isStringInRegexList

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneRecordValidations.scala
nimaeskandary: @rebstar6 this could be an IP address too

nimaeskandary: Oof that made me realize I need to do some IP address tests

rebstar6: mm not sure about that. it should be on fqdn, so ptrs would get caught as `1.2.3.4.in-addr.arpa`

that said, i guess by-IP could be a nice-to-have here

modules/api/src/it/resources/application.conf
```diff
@@ -6,10 +6,6 @@ akka {
 }
 
 vinyldns {
-  approved-name-servers = [
```
nimaeskandary: figured this should use reference.conf since there was already something there for approved name servers

modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
nimaeskandary: Iffy about the expected behavior for this if the config is not correct, right now i suppose the application crashes?

modules/api/src/test/scala/vinyldns/api/domain/zone/ZoneConnectionValidatorSpec.scala
```diff
@@ -108,7 +108,7 @@ class ZoneConnectionValidatorSpec
     RecordSetStatus.Active,
     DateTime.now,
     None,
-    List(NSData("some.test.ns.")))
+    List(NSData("172.17.42.1.")))
```
nimaeskandary: using the one from reference.conf instead

modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
rebstar6: it wont start up. you pull it out of the IO in boot

also of note, make sure it works if you dont have any high value domains (consider that empty list then)

modules/api/src/it/resources/application.conf
rebstar6: honestly there isnt any sane default here. these should be empty lists in ref.conf if they are needed, and filled out in application.conf

rebstar6: i think... :P

modules/api/src/main/resources/reference.conf
```diff
@@ -32,8 +32,14 @@ vinyldns {
...
   approved-name-servers = [
-    "172.17.42.1.",
-    "ns1.parent.com."
+    "172\\.17\\.42\\.1\\.",
+    "ns1\\.parent\\.com\\.*"
+  ]
+
+  high-value-domains = [
+    "dont-touch-me.*",
+    "192\\.0\\.2\\.199\\.*",
```
rebstar6: why * on this one?

i think "192\\.0\\.2\\.199" is probably what we would want for an IP, the trailing dot shouldnt impact

modules/api/src/main/resources/reference.conf
```diff
@@ -32,8 +32,14 @@ vinyldns {
...
+    "0:0:0:0:0:0:0:1"
```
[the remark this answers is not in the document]
rebstar6: oh oops, nvm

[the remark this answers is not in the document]
rebstar6: should add some v6 tests though so you may want the example to be in out v6 test zone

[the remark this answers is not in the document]
nimaeskandary: have a unit test with ipv6 and writing out the nibbles was really annoying so did all zeros except a one at the end

modules/api/src/main/scala/vinyldns/api/domain/dns/DnsConversions.scala
```diff
@@ -70,6 +70,16 @@ trait DnsConversions {
...
     Try(DNS.ReverseMap.fromAddress(ip, DNS.Address.IPv6)).toOption.map(_.toString)
 
+  /*
+    normalizes ipv6 and ipv4 to use all fields with minimal zeros, e.g. 1:0::b8 -> 1:0:0:0:0:0:0:b8. according
+    to inetaddress spec InetAddress.getByName on a literal ip does not issue a DNS lookup, just validates that it is a
+    valid ip. If the inputted ip is not valid it will return None
+
+    https://docs.oracle.com/javase/8/docs/api/java/net/InetAddress.html#getByName-java.lang.String-
+   */
+  def getValidNormalizedIpAddress(ip: String): Option[String] =
```
rebstar6: may be a good time to pull in https://github.com/Comcast/ip4s/blob/master/docs/guide.md

weve been meaning to incorporate it anyway. i believe if you wrap both the IP and whats in config as IpAddress they should be comparable

modules/api/src/main/resources/reference.conf
rebstar6: hmm actually this is interesting, we shouldnt be treating the IPs as regex's anyway. i wonder if they should be in their own config. maybe 
```
high-value-domains {
   # regex matching on all record types
    regex = []    
    # specific IP matching for PTR records only 
    ips = []
}
```

...would also allow us to expand to CIDR rules rather than specific match in the IP list

modules/api/src/main/scala/vinyldns/api/domain/dns/DnsConversions.scala
nimaeskandary: interesting, that is cool will replace then

modules/api/src/test/scala/vinyldns/api/domain/zone/ZoneConnectionValidatorSpec.scala
rebstar6: bring it back! NS to IP isnt actually how its supposed to be i believe

nimaeskandary: oh whoopsie

modules/api/src/main/scala/vinyldns/api/domain/dns/DnsConversions.scala
rebstar6: cc https://github.com/vinyldns/vinyldns/issues/235

nimaeskandary: hm that seems like a lot to pull into this pr, is there a happy medium where that library is used for the high value domain comparison for now?

modules/api/functional_test/live_tests/batch/create_batch_change_test.py
```diff
@@ -650,6 +650,35 @@ def test_create_batch_change_with_unsupported_record_type_fails(shared_zone_test
...
+    }
+
+    response = client.create_batch_change(batch_change_input, status=400)
+
+    assert_error(response[0], error_messages=['Record name "dont-touch-me-add.ok." is configured as a High Value Domain, cannot be modified'])
+    assert_error(response[1], error_messages=['Record name "dont-touch-me-update.ok." is configured as a High Value Domain, cannot be modified'])
+    assert_error(response[2], error_messages=['Record name "dont-touch-me-update.ok." is configured as a High Value Domain, cannot be modified'])
+    assert_error(response[3], error_messages=['Record name "dont-touch-me-delete.ok." is configured as a High Value Domain, cannot be modified'])
+    assert_error(response[4], error_messages=['Record name "192.0.2.199" is configured as a High Value Domain, cannot be modified'])
+     assert_that(response[5], is_not(has_key("errors")))
```
mitruly: Unexpected indent.

modules/api/src/it/resources/application.conf
nimaeskandary: still need to put this back

modules/api/src/test/scala/vinyldns/api/domain/record/RecordSetValidationsSpec.scala
```diff
@@ -237,7 +237,7 @@ class RecordSetValidationsSpec
       "return ok if the record is an NS record but not origin" in {
         val valid = invalidNsRecordToOrigin.copy(
           name = "this-is-not-origin-mate",
-          records = List(NSData("some.test.ns.")))
+          records = List(NSData("172.17.42.1")))
```
nimaeskandary: have to change back

modules/api/src/main/scala/vinyldns/api/domain/DomainValidationErrors.scala
```diff
@@ -121,4 +121,9 @@ final case class RecordInReverseZoneError(name: String, typ: String) extends Dom
     "Invalid Record Type In Reverse Zone: record with name " +
       s""""$name" and type "$typ" is not allowed in a reverse zone."""
 }
+
+final case class HighValueDomainError(name: String) extends DomainValidationError {
+  def message: String =
+    s"""Record name "$name" is configured as a High Value Domain, cannot be modified"""
```
britneywright: can we change this to `s"""Record name "$name" is configured as a High Value Domain, so it cannot be modified."""`

modules/api/functional_test/live_tests/zones/create_zone_test.py
nimaeskandary: was in this file originally for a test, figured id keep this edit

docker/bind9/zones/1.9.e.f.c.c.7.2.9.6.d.f.ip6.arpa
```diff
@@ -8,3 +8,4 @@ $ttl 38400
 1.9.e.f.c.c.7.2.9.6.d.f.ip6.arpa.	IN	NS	172.17.42.1.
 4.2.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0	IN	PTR	www.vinyldns.
 5.2.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0	IN	PTR	mail.vinyldns.
+0.0.0.0.f.f.f.f.0.0.0.0.0.0.0.0.0.0.0.0 IN  PTR test.foo.
```
rebstar6: point this to "some.high.value" or something so we can track it in logs if something goes wrong

modules/api/functional_test/live_tests/batch/create_batch_change_test.py
```diff
@@ -1813,7 +1858,7 @@ def test_ipv6_ptr_recordtype_add_checks(shared_zone_test_context):
     """
     client = shared_zone_test_context.ok_vinyldns_client
 
-    existing_ptr = get_recordset_json(shared_zone_test_context.ip6_reverse_zone, "f.f.f.f.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0", "PTR", [{"ptrdname": "test.com."}], 100)
+    existing_ptr = get_recordset_json(shared_zone_test_context.ip6_reverse_zone, "a.a.a.a.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0", "PTR", [{"ptrdname": "test.com."}], 100)
```
rebstar6: i will let this slide this time, but in the future please make sure the new records you are creating dont conflict with existing tests. this is a big PR, so sorting through what tests got changed because the HVD conflicts with an existing record, and whats a real change isnt ideal

modules/api/src/main/resources/reference.conf
```diff
@@ -36,6 +36,20 @@ vinyldns {
     "ns1.parent.com."
   ]
 
+  # FQDNs / IPs that cannot be modified via VinylDNS
+  high-value-domains = {
```
rebstar6: remove this from reference.conf (or just make empty). I dont think we want these defaults making it out to prod. I would just store each item as an empty list

modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
```diff
@@ -54,7 +57,20 @@ object VinylDNSConfig {
   lazy val cryptoConfig: Config = vinyldnsConfig.getConfig("crypto")
   lazy val system: ActorSystem = ActorSystem("VinylDNS", VinylDNSConfig.config)
   lazy val approvedNameServers: List[Regex] =
```
rebstar6: this isnt optional, there has to be at least one approved backend

modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
```diff
@@ -54,7 +57,20 @@ object VinylDNSConfig {
...
-    vinyldnsConfig.getStringList("approved-name-servers").asScala.toList.map(n => n.r)
+    Try(vinyldnsConfig.getStringList("approved-name-servers").asScala.toList.map(n => n.r))
+      .getOrElse(List[Regex]())
+
+  lazy val highValueRegexList: List[Regex] =
+    Try(vinyldnsConfig.getStringList("high-value-domains.regex-list").asScala.toList.map(n => n.r))
+      .getOrElse(List[Regex]())
+
+  lazy val highValueIpList: List[Option[IpAddress]] =
```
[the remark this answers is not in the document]
rebstar6: dont love it. docs indicate the approach is to check for the field existence. id probably go with a helper you can share:
```  def getOptionalStringList(key: String): List[String] = {
    if (vinyldnsConfig.hasPath(key)) {
      vinyldnsConfig.getStringList(key).asScala.toList
    } else List()
  }

  lazy val highValueRegexList: List[Regex] =
    getOptionalStringList("high-value-domains.regex-list").map(_.r)```

File: modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala [hunk 26]
```diff
@@ -54,7 +57,20 @@ object VinylDNSConfig {
...
+    Try(
+      vinyldnsConfig
```
rebstar6: flatten this out, working with a `List[Option]` doesnt make sense. Just need a List with the existing values (List[IpAddress])

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneRecordValidations.scala
```diff
@@ -16,18 +16,39 @@
...
 
+  /* Checks to see if an individual string is part of the regex list */
+  def isStringInRegexList(regexList: List[Regex], string: String): Boolean =
+    regexList.exists(rx => rx.findAllIn(string).contains(string))
+
+  /* Checks to see if an ip address is part of the ip address list */
+  def isIpInIpList(ipList: List[Option[IpAddress]], ip: String): Boolean = {
+    val ipToTest = IpAddress(ip)
+    if (ipToTest.isDefined) {
+      ipList.exists { ipOption =>
```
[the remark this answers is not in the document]
rebstar6: this will get cleaned up once ipList is instead a `List[IpAddress]` - working with the List[Option] makes it ugly

modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
mitruly: On another note - should we be trying to use `loadConfigF` here instead of `Try`?

modules/api/src/test/scala/vinyldns/api/domain/record/RecordSetServiceSpec.scala
```diff
@@ -115,6 +115,23 @@ class RecordSetServiceSpec
       val result = leftResultOf(underTest.addRecordSet(record, okAuth).value)
       result shouldBe a[InvalidRequest]
     }
+    "fail if the record is a high value domain" in {
+      val record = aaaa.copy(
+        name = "high-value-domain",
+        zoneId = zoneAuthorized.id,
+        status = RecordSetStatus.Active)
+
+      doReturn(IO.pure(List()))
```
rebstar6: you shouldnt need these mocks, i think this fails before it hits those lines

modules/api/src/test/scala/vinyldns/api/domain/record/RecordSetServiceSpec.scala
```diff
@@ -275,6 +292,25 @@ class RecordSetServiceSpec
...
+        name = "high-value-domain",
+        zoneId = zoneAuthorized.id,
+        status = RecordSetStatus.Active)
+
+      val newRecord = oldRecord.copy(ttl = oldRecord.ttl + 1000)
+
+      doReturn(IO.pure(Some(oldRecord)))
+        .when(mockRecordRepo)
+        .getRecordSet(zoneAuthorized.id, newRecord.id)
+      doReturn(IO.pure(List()))
```
rebstar6: same here, 1st mock needed, i think 2nd is not

modules/api/src/it/resources/application.conf
```diff
@@ -37,6 +37,20 @@ vinyldns {
     }
   }
 
+  # FQDNs / IPs that cannot be modified via VinylDNS
+  high-value-domains = {
+    regex-list = [
+      "high-value-domain.*"
+    ]
+    ip-list = [
+      // these are using reverse zones in the vinyldns/bind9 docker image
```
rebstar6: update the commenting here:

- the regex list will be used for all record types except PTR
- IP list will be used exclusively for PTR records

can you also, in a separate PR, add this info into the docs.

docker/api/docker.conf
```diff
@@ -172,6 +172,22 @@ vinyldns {
   }
 
   batch-change-limit = 20
+
+  # FQDNs / IPs that cannot be modified via VinylDNS
+  # regex-list used for all record types except PTR
+  # ip-list used exclusivley for PTR records
```
mitruly: `exclusively`

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneRecordValidations.scala
```diff
@@ -32,16 +32,12 @@ object ZoneRecordValidations {
...
-      ipList.exists { ipOption =>
-        ipOption match {
-          case None => false
-          case ipInList => ipInList === ipToTest
-        }
-      }
-    } else false
+  def isIpInIpList(ipList: List[IpAddress], ip: String): Boolean = {
+    val ipToTestOption = IpAddress(ip)
+    ipToTestOption match {
```
mitruly: Replacing this `match` clause with `ipToTestOption.exists(ipToTest => ipList.exists(_ === ipToTest))` passes tests for me.

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneRecordValidations.scala
```diff
@@ -16,18 +16,30 @@
...
 object ZoneRecordValidations {
 
+  /* Checks to see if an individual string is part of the regex list */
+  def isStringInRegexList(regexList: List[Regex], string: String): Boolean =
+    regexList.exists(rx => rx.findAllIn(string).contains(string))
+
+  /* Checks to see if an ip address is part of the ip address list */
+  def isIpInIpList(ipList: List[IpAddress], ipToTest: String): Boolean =
+    IpAddress(ipToTest).exists(ip => ipList.exists(_ === ip))
+
```
rebstar6: this is dumb. `validate` is failing because there is whitespace at the end of this line.

modules/api/src/main/resources/reference.conf
```diff
@@ -36,6 +36,14 @@ vinyldns {
     "ns1.parent.com."
   ]
 
+  # FQDNs / IPs that cannot be modified via VinylDNS
+  # regex-list used for all record types except PTR
+  # ip-list used exclusively for PTR records
+  high-value-domains = {
+    regex-list = []
```
rebstar6: i think you can actually take this out completely (not have empty lists) and it should be fine with how we are doing the config load

modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetValidations.scala
```diff
@@ -192,4 +193,25 @@ object RecordSetValidations {
   private def isOriginRecord(recordSetName: String, zoneName: String): Boolean =
     recordSetName == "@" || omitTrailingDot(recordSetName) == omitTrailingDot(zoneName)
 
+  def isNotHighValueDomain(recordSet: RecordSet, zone: Zone): Either[Throwable, Unit] = {
+    val result = recordSet.typ match {
+      case RecordType.PTR =>
+        val ip = reverseNameToIp(recordSet.name, zone)
+        ZoneRecordValidations.isNotHighValueIp(VinylDNSConfig.highValueIpList, ip)
+      case _ =>
+        val fqdn = DnsConversions.recordDnsName(recordSet.name, zone.name).toString()
```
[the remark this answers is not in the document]
nimaeskandary: ooph good call, so a if its apex it should just use zone name

modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetValidations.scala
```diff
@@ -192,4 +193,25 @@ object RecordSetValidations {
...
+        ZoneRecordValidations.isNotHighValueFqdn(VinylDNSConfig.highValueRegexList, fqdn)
+    }
+
+    result.toEither
+      .map(_ => ())
+      .leftMap(errors => InvalidRequest(errors.toList.map(_.message).mkString(", ")))
+  }
+
+  def reverseNameToIp(recordName: String, zone: Zone): String =
```
rebstar6: put this directly in ReverseZoneHelpers

PR #380, opened 2018-12-03
Support admin

### LINKED ISSUE #368: Most basic read only user
- Add boolean on the user type for `supportUser` (or whatever we wanna call it). This will need to be stored in the protobuf, so note that. Default to false.
- Treat all GETs for this user as if they are super users
- Treat all PUTs for this user as if they are a normal user (no extra access)

britneywright:
Fixes #368.

Changes in this pull request:
- add isSupport attribute to Users. Boolean like isSuper.
- Group services are set up a bit differently than Zones and records. For groups add the support check to view a group and also to list_my_groups so it returns all groups, like the behavior of a super user.
- For Zones and Records added support user to `canSeeZones`. Also added support user in `getPrioritizedAccessLevel` so their existing access rules will apply first but if they have no access rules then they'd get read access.

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/AccessValidations.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/AccessValidations.scala
@@ -102,11 +102,12 @@ object AccessValidations extends AccessValidationAlgebra {
       }
 
       recordSets.map { rs =>
+        val aclAccessLevel = getAccessFromUserAcls(rs.name, rs.typ)
         val accessLevel = {
-          if ((getAccessFromUserAcls(rs.name, rs.typ) == AccessLevel.NoAccess) && auth.signedInUser.isSupport)
+          if ((aclAccessLevel == AccessLevel.NoAccess) && auth.canReadAll)
             AccessLevel.Read
           else
-            getAccessFromUserAcls(rs.name, rs.typ)
+            aclAccessLevel
         }
         RecordSetInfo(rs, accessLevel)
       }
--- a/modules/api/src/main/scala/vinyldns/api/domain/AccessValidations.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/AccessValidations.scala
@@ -17,12 +17,7 @@
 package vinyldns.api.domain
 
 import vinyldns.api.Interfaces.ensuring
-import vinyldns.api.domain.zone.{
-  ACLRuleOrdering,
-  NotAuthorizedError,
-  PTRACLRuleOrdering,
-  RecordSetInfo
-}
+import vinyldns.api.domain.zone.{ACLRuleOrdering, NotAuthorizedError, PTRACLRuleOrdering, RecordSetInfo}
 import vinyldns.core.domain.auth.AuthPrincipal
 import vinyldns.core.domain.record.{RecordSet, RecordType}
 import vinyldns.core.domain.record.RecordType.RecordType
@@ -34,19 +29,19 @@ object AccessValidations extends AccessValidationAlgebra {
   def canSeeZone(auth: AuthPrincipal, zone: Zone): Either[Throwable, Unit] =
     ensuring(
       NotAuthorizedError(s"User ${auth.signedInUser.userName} cannot access zone '${zone.name}'"))(
-      hasReadZoneAccess(auth, zone) || userHasAclRules(auth, zone))
+      auth.canReadAll(zone.adminGroupId) || userHasAclRules(auth, zone))
 
   def canChangeZone(auth: AuthPrincipal, zone: Zone): Either[Throwable, Unit] =
     ensuring(
       NotAuthorizedError(s"User ${auth.signedInUser.userName} cannot modify zone '${zone.name}'"))(
-      hasZoneAdminAccess(auth, zone))
+      auth.canEditAll(zone.adminGroupId))
 
   def canAddRecordSet(
       auth: AuthPrincipal,
       recordName: String,
       recordType: RecordType,
       zone: Zone): Either[Throwable, Unit] = {
-    val accessLevel = getAccessLevel(auth, recordName, recordType, zone)
+    val accessLevel = getFullAccessLevel(auth, recordName, recordType, zone)
     ensuring(
       NotAuthorizedError(s"User ${auth.signedInUser.userName} does not have access to create " +
         s"$recordName.${zone.name}"))(
@@ -58,7 +53,7 @@ object AccessValidations extends AccessValidationAlgebra {
       recordName: String,
       recordType: RecordType,
       zone: Zone): Either[Throwable, Unit] = {
-    val accessLevel = getAccessLevel(auth, recordName, recordType, zone)
+    val accessLevel = getFullAccessLevel(auth, recordName, recordType, zone)
     ensuring(
       NotAuthorizedError(s"User ${auth.signedInUser.userName} does not have access to update " +
         s"$recordName.${zone.name}"))(
@@ -73,7 +68,7 @@ object AccessValidations extends AccessValidationAlgebra {
     ensuring(
       NotAuthorizedError(s"User ${auth.signedInUser.userName} does not have access to delete " +
         s"$recordName.${zone.name}"))(
-      getAccessLevel(auth, recordName, recordType, zone) == AccessLevel.Delete)
+      getFullAccessLevel(auth, recordName, recordType, zone) == AccessLevel.Delete)
 
   def canViewRecordSet(
       auth: AuthPrincipal,
@@ -83,13 +78,13 @@ object AccessValidations extends AccessValidationAlgebra {
     ensuring(
       NotAuthorizedError(s"User ${auth.signedInUser.userName} does not have access to view " +
         s"$recordName.${zone.name}"))(
-      getAccessLevel(auth, recordName, recordType, zone) != AccessLevel.NoAccess)
+      getReadAccessLevel(auth, recordName, recordType, zone) != AccessLevel.NoAccess)
 
   def getListAccessLevels(
       auth: AuthPrincipal,
       recordSets: List[RecordSet],
       zone: Zone): List[RecordSetInfo] =
-    if (hasZoneAdminAccess(auth, zone)) recordSets.map(RecordSetInfo(_, AccessLevel.Delete))
+    if (auth.canEditAll(zone.adminGroupId)) recordSets.map(RecordSetInfo(_, AccessLevel.Delete))
     else {
       val rulesForUser = zone.acl.rules.filter(ruleAppliesToUser(auth, _))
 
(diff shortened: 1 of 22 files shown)


modules/core/src/main/scala/vinyldns/core/domain/auth/AuthPrincipal.scala
```diff
@@ -23,6 +23,9 @@ case class AuthPrincipal(signedInUser: User, memberGroupIds: Seq[String]) {
   def isAuthorized(adminGroupId: String): Boolean =
```
rebstar6: we now have auth stuff in a bunch of places, which is confusing. I think get rid of both isAuthorized and readOnlyAuthorization - that logic can all live in the AccessValidations directly

modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipValidations.scala
```diff
@@ -38,8 +38,13 @@ object MembershipValidations {
...
 
+  def isSupportAdmin(authPrincipal: AuthPrincipal): Either[Throwable, Unit] =
+    ensuring(NotAuthorizedError("Not authorized")) {
+      authPrincipal.signedInUser.isSupport
+    }
+
   def canSeeGroup(groupId: String, authPrincipal: AuthPrincipal): Either[Throwable, Unit] =
     ensuring(NotAuthorizedError("Not authorized")) {
-      authPrincipal.isAuthorized(groupId)
+      authPrincipal.isAuthorized(groupId) || authPrincipal.signedInUser.isSupport
```
rebstar6: kill the isSupportAdmin/authprinciple stuff.

just do `authPrincipal.canReadAll || authPrincipal.isInGroup(groupId)`

modules/core/src/main/scala/vinyldns/core/domain/auth/AuthPrincipal.scala
rebstar6: that said, may be helpful to have a `canReadAll: Boolean = super + support` and `canEditAll: Boolean = super` so you dont have to always do the ORs everywhere.

and a `isInGroup: Bool = memberGroupIds.contains(groupId)`

modules/api/src/main/scala/vinyldns/api/domain/AccessValidations.scala
```diff
@@ -159,9 +162,14 @@ object AccessValidations extends AccessValidationAlgebra {
   def ruleAppliesToRecordType(recordType: RecordType, rule: ACLRule): Boolean =
     rule.recordTypes.isEmpty || rule.recordTypes.contains(recordType)
 
-  def getPrioritizedAccessLevel(recordType: RecordType, rules: Set[ACLRule]): AccessLevel =
+  def getPrioritizedAccessLevel(
```
rebstar6: I dont think you need to edit the ACLs. they should be able to see the records because they have access to the zone already

modules/api/src/main/scala/vinyldns/api/domain/AccessValidations.scala
```diff
@@ -34,19 +34,19 @@ object AccessValidations extends AccessValidationAlgebra {
   def canSeeZone(auth: AuthPrincipal, zone: Zone): Either[Throwable, Unit] =
     ensuring(
       NotAuthorizedError(s"User ${auth.signedInUser.userName} cannot access zone '${zone.name}'"))(
-      (hasZoneAdminAccess(auth, zone) || zone.shared) || userHasAclRules(auth, zone))
```
rebstar6: woah theres shared zone logic in here already? uh oh

modules/api/src/main/scala/vinyldns/api/domain/AccessValidations.scala
```diff
@@ -34,19 +34,19 @@ object AccessValidations extends AccessValidationAlgebra {
...
+      (auth.canReadAll(zone.adminGroupId) || zone.shared) || userHasAclRules(auth, zone))
```
rebstar6: weird parentheses here, remove the extra around the 1st 2 things

modules/api/src/main/scala/vinyldns/api/domain/AccessValidations.scala
```diff
@@ -169,11 +171,19 @@ object AccessValidations extends AccessValidationAlgebra {
       topRule.accessLevel
     }
 
-  def getAccessLevel(
+  def getFullAccessLevel(
+      auth: AuthPrincipal,
+      recordName: String,
+      recordType: RecordType,
+      zone: Zone): AccessLevel =
+    if (auth.canEditAll(zone.adminGroupId)) AccessLevel.Delete
```
rebstar6: when you add in the other cases here, if it gets crazy with the if/else, i believe this should work:

```
  def getAccessLevel(
      auth: AuthPrincipal,
      recordName: String,
      recordType: RecordType,
      zone: Zone): AccessLevel = auth match {
    case superUser if superUser.canEditAll => AccessLevel.Delete
    case zoneAdmin if zoneAdmin.inAdminGroup(zone.adminGroupId) => AccessLevel.Delete
    case supportUser if supportUser.canReadAll =>
      if(aclAccess == AccessLevel.NoAccess) AccessLevel.Read else aclAccess
  }
```

modules/api/src/main/scala/vinyldns/api/domain/AccessValidations.scala
britneywright: the canReadAll covers the inAdminGroup

rebstar6: referring to it with https://github.com/vinyldns/vinyldns/pull/380#discussion_r240313465

modules/api/src/test/scala/vinyldns/api/domain/membership/MembershipServiceSpec.scala
```diff
@@ -629,6 +639,31 @@ class MembershipServiceSpec
...
+          rightResultOf(underTest.listMembers(testGroup.id, None, 100, supportAuth).value)
+
+        result.members should contain theSameElementsAs expectedMembers
+        result.nextId shouldBe testListUsersResult.lastEvaluatedId
+        result.maxItems shouldBe 100
+        result.startFrom shouldBe None
+      }
+
+      "return a list of members if the user is a support admin" in {
         // make sure that the ok user is the admin user, would be an admin member
```
mitruly: Not sure if this exact comment applies.

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneService.scala
```diff
@@ -77,7 +77,7 @@ class ZoneService(
       _ <- validateZoneConnectionIfChanged(newZone, existingZone)
       _ <- canChangeZone(auth, existingZone).toResult
       _ <- adminGroupExists(newZone.adminGroupId)
-      _ <- userIsMemberOfGroup(newZone.adminGroupId, auth).toResult
+      _ <- canChangeZone(auth, newZone).toResult
```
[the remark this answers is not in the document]
britneywright: this is to ensure if the group changes it's to a group the user is a member of.

modules/core/src/main/protobuf/VinylDNSProto.proto
```diff
@@ -192,6 +192,7 @@ message User {
   optional string firstName = 8;
   optional string lastName = 9;
   optional string email = 10;
+  required bool isSupport = 11;
```
rebstar6: i wonder if this should instead be `optional bool isSupport = 11 [default = false];`

rebstar6: i think so, im worried making it required will break the existing users

modules/dynamodb/src/main/scala/vinyldns/dynamodb/repository/DynamoDBUserRepository.scala
```diff
@@ -134,7 +136,8 @@ object DynamoDBUserRepository {
       isSuper = if (item.get(IS_SUPER) == null) false else item.get(IS_SUPER).getBOOL,
       lockStatus =
         if (item.get(LOCK_STATUS) == null) LockStatus.Unlocked
-        else userStatus(item.get(LOCK_STATUS).getS)
+        else userStatus(item.get(LOCK_STATUS).getS),
```
rebstar6: need to add to DDBUserRepoIntegrationSpec on these changes

modules/api/src/main/scala/vinyldns/api/domain/AccessValidations.scala
```diff
@@ -106,15 +102,17 @@ object AccessValidations extends AccessValidationAlgebra {
       }
 
       recordSets.map { rs =>
-        val accessLevel = getAccessFromUserAcls(rs.name, rs.typ)
+        val accessLevel = {
+          if ((getAccessFromUserAcls(rs.name, rs.typ) == AccessLevel.NoAccess) && auth.signedInUser.isSupport)
```
rebstar6: store `getAccessFromUserAcls(rs.name, rs.typ)` in a val so were not calling it twice

modules/api/src/main/scala/vinyldns/api/domain/AccessValidations.scala
```diff
@@ -106,15 +106,17 @@ object AccessValidations extends AccessValidationAlgebra {
       }
 
       recordSets.map { rs =>
-        val accessLevel = getAccessFromUserAcls(rs.name, rs.typ)
+        val accessLevel = {
```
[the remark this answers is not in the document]
rebstar6: makes sense

modules/api/src/main/scala/vinyldns/api/domain/AccessValidations.scala
rebstar6: also use `auth.canReadAll` here for clarity

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
```diff
@@ -325,7 +325,7 @@ class BatchChangeValidations(changeLimit: Int, accessValidation: AccessValidatio
   def canGetBatchChange(
       batchChange: BatchChange,
       auth: AuthPrincipal): Either[BatchChangeErrorResponse, Unit] =
-    if (auth.signedInUser.isSuper || auth.userId == batchChange.userId) {
+    if (auth.signedInUser.isSuper || auth.userId == batchChange.userId || auth.signedInUser.isSupport) {
```
rebstar6: auth.canReadAll || auth.userId == batchChange.userId

modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
```diff
@@ -131,11 +131,12 @@ class MembershipService(
       startFrom: Option[String],
       maxItems: Int,
       authPrincipal: AuthPrincipal): Result[ListMyGroupsResponse] = {
-    val groupsCall = if (authPrincipal.signedInUser.isSuper) {
-      groupRepo.getAllGroups()
-    } else {
-      groupRepo.getGroups(authPrincipal.memberGroupIds.toSet)
-    }
+    val groupsCall =
+      if (authPrincipal.signedInUser.isSuper || authPrincipal.signedInUser.isSupport) {
```
rebstar6: authPrincipal.canReadAll

modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipValidations.scala
```diff
@@ -30,16 +30,16 @@ object MembershipValidations {
 
   def isGroupAdmin(group: Group, authPrincipal: AuthPrincipal): Either[Throwable, Unit] =
```
rebstar6: can we rename this to `canEditGroup`?

modules/api/src/main/scala/vinyldns/api/domain/AccessValidations.scala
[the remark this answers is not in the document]
mitruly: This refactor comment still stands.

modules/api/src/main/scala/vinyldns/api/domain/AccessValidations.scala
```diff
@@ -173,7 +172,13 @@ object AccessValidations extends AccessValidationAlgebra {
       auth: AuthPrincipal,
       recordName: String,
       recordType: RecordType,
-      zone: Zone): AccessLevel =
-    if (hasZoneAdminAccess(auth, zone)) AccessLevel.Delete
-    else getAccessFromAcl(auth, recordName, recordType, zone)
+      zone: Zone): AccessLevel = auth match {
+    case superUser if superUser.canEditAll => AccessLevel.Delete
```
rebstar6: i suppose these can be merged --

`case admin if admin.canEditAll || admin.isGroupMember(zone.adminGroupId) => AccessLevel.Delete`

PR #382, opened 2018-12-03
Remove SQS access and secret key from logging

### LINKED ISSUE #381: Remove SQS secret and access keys from logging
**Motivation**
Logs shouldn't contain sensitive information

**Scope of change**
Secret and access keys should not be logged

mitruly:
Break out SQS settings logging so access and secret key aren't printed.

Fixes #381.

Changes in this pull request:
- Update logging message in `SqsMessageQueueProvider.scala`.

### DIFF
--- a/modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueueProvider.scala
+++ b/modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueueProvider.scala
@@ -56,7 +56,10 @@ class SqsMessageQueueProvider extends MessageQueueProvider {
 
   def setupClient(sqsMessageQueueSettings: SqsMessageQueueSettings): IO[AmazonSQSAsync] =
     IO {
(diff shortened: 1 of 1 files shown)


modules/sqs/src/main/scala/vinyldns/sqs/queue/SqsMessageQueueProvider.scala
```diff
@@ -56,7 +56,10 @@ class SqsMessageQueueProvider extends MessageQueueProvider {
 
   def setupClient(sqsMessageQueueSettings: SqsMessageQueueSettings): IO[AmazonSQSAsync] =
     IO {
-      logger.error(s"Setting up queue client SqsMessageQueueSettings: $sqsMessageQueueSettings")
```
rebstar6: probably better just overriding toString in the SqsMessageQueueSettings type

rebstar6: that said, im good with it either way 👍

PR #383, opened 2018-12-03
Ldap healthcheck

### LINKED ISSUE #367: Add LDAP healthcheck in the portal
see https://github.com/vinyldns/vinyldns/issues/357

validate connectivity to LDAP from the portal. do test user get, not found or found is ok, just make sure it's not timing out.

rebstar6:
Fixes #367

### DIFF
--- a/modules/portal/app/controllers/LdapAuthenticator.scala
+++ b/modules/portal/app/controllers/LdapAuthenticator.scala
@@ -18,9 +18,12 @@ package controllers
 
 import java.util
 
+import cats.effect.IO
+import cats.implicits._
 import controllers.LdapAuthenticator.LdapByDomainAuthenticator
 import javax.naming.Context
 import javax.naming.directory._
+import vinyldns.core.health.HealthCheck._
 
 import scala.collection.JavaConverters._
 import scala.util.{Failure, Try}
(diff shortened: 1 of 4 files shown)


modules/portal/app/controllers/LdapAuthenticator.scala
```diff
@@ -198,11 +201,28 @@ class LdapAuthenticator(
   def lookup(username: String): Try[UserDetails] =
     findUserDetails(searchBase, username, authenticator.lookup(_, username, serviceAccount))
 
+  def checkHealth(): HealthCheck =
+    IO {
```
rebstar6: this is gross. i need to figure out some better scala for this (suggestions welcome).

basically, `authenticator.lookup(domain, "healthlookup", serviceAccount)` is a Try. In the case that the Try is successful, or it's a UserDoesNotExistException, we are fine. Any other error means failure.

mitruly: This makes sense; not sure if there is a more concise way to convey this; playing locally, the following worked (but is functionally the same):

```
  def healthCheck(): HealthCheck =
    IO {
      searchBase.headOption
        .map { domain =>
          authenticator.lookup(domain, "healthlookup", serviceAccount).as(()).recover {
            case _: UserDoesNotExistException => ()
          }
        } match {
        case Some(Failure(exception)) => exception.asLeft
        case _ => ().asRight
      }
    }.asHealthCheck
```

rebstar6: re-pushed more similar to what you have. thoughts @mitruly?

mitruly: Me likey

modules/portal/app/controllers/LdapAuthenticator.scala
```diff
@@ -185,9 +188,9 @@ class LdapAuthenticator(
     extends Authenticator {
 
   private def findUserDetails(
-      domains: List[LdapSearchDomain],
-      userName: String,
-      f: LdapSearchDomain => Try[UserDetails]): Try[UserDetails] = domains match {
+                               domains: List[LdapSearchDomain],
```
mitruly: Looks like some weird IntelliJ formatting.

rebstar6: i hate when it does this.

britneywright:
this is an edge case, but if you run the portal locally without using LDAP the health endpoint errors. Do we want to handle that or at least make note of it somewhere?

rebstar6:
hmm a good question @britneywright. im inclined to take the dummy ldap domain out of reference.conf now that I look at it. will dig.

PR #384, opened 2018-12-04
maven-cache

nimaeskandary:
Noticed we're not caching maven dependencies, also, according to https://docs.travis-ci.com/user/caching/, npm should be cached this way instead of `node_modules`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,10 +18,11 @@ services:
 cache:
   pip: true
   bundler: true
+  npm: true
   directories:
     - $HOME/.ivy2/cache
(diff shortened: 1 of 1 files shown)


britneywright:
will we need to clear the existing cache after this is merged in?

nimaeskandary:
@britneywright don't believe so, we can validate after with the next pr since we can check their logs and if they needed maven dependencies when using masters cache

nimaeskandary:
nuking the cache is a pain though so we should wait to see, remembering that builds took upwards of 40 minutes when we've done it in the past

PR #385, opened 2018-12-04
groups scan timing

rebstar6:
adding some logs

### DIFF
--- a/modules/dynamodb/src/main/scala/vinyldns/dynamodb/repository/DynamoDBGroupRepository.scala
+++ b/modules/dynamodb/src/main/scala/vinyldns/dynamodb/repository/DynamoDBGroupRepository.scala
@@ -183,9 +183,9 @@ class DynamoDBGroupRepository private[repository] (
         .withExpressionAttributeValues(Map(":del" -> new AttributeValue("Deleted")).asJava)
 
       val scan = for {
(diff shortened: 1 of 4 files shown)


modules/dynamodb/src/main/scala/vinyldns/dynamodb/repository/DynamoDBGroupRepository.scala
```diff
@@ -174,14 +174,22 @@ class DynamoDBGroupRepository private[repository] (
     monitor("repo.Group.getAllGroups") {
       log.info(s"getting all group IDs")
       val scanRequest = new ScanRequest().withTableName(groupTableName)
-      dynamoDBHelper.scanAll(scanRequest).map { results =>
+
+      val groups = for {
```
britneywright: Is it okay to have this groups variable and the groups variable assigned inside the `groups.map`?

rebstar6: mm bummer. technically yes, but its confusing to read. will change

mitruly: IntelliJ doesn't throw any warnings about suspicious shadowing and this seems fine due to scope.

PR #386, opened 2018-12-05
Add health check for DNS backend

### LINKED ISSUE #357: Add healthcheck on default DNS backend
split off of #313

/health now tests connectivity to all of the configured databases and queues. Need to add in some kind of test of connectivity to the default DNS backend.

Need to be careful about implementation here, may need to build in some layer of protection here so we are not over-hitting the backend

mitruly:
Add health check for DNS backend and tie into `Boot`.

Fixes #357.

Changes in this pull request:
- Split off IO component of `resolve` into `queryDnsBackend` (to avoid converting `IO` → `Result` → `IO` → `HealthCheck`)
- Create health check on `ZoneConnectionValidator` to test using the default zone connection info

### DIFF
--- a/modules/api/src/main/resources/reference.conf
+++ b/modules/api/src/main/resources/reference.conf
@@ -8,6 +8,17 @@ vinyldns {
   # if we should start up polling for change requests, set this to false for the inactive cluster
   processing-disabled = false
 
+  health-check {
+    dns-backend {
+      # connection information for DNS backend
+      address = "localhost"
+      port = 19001
+
+      # timeout in millis to test connection
+      timeout = 20000
+    }
+  }
+
   queue {
     class-name = "vinyldns.sqs.queue.SqsMessageQueueProvider"
 
--- a/modules/api/src/main/resources/reference.conf
+++ b/modules/api/src/main/resources/reference.conf
@@ -8,16 +8,7 @@ vinyldns {
   # if we should start up polling for change requests, set this to false for the inactive cluster
   processing-disabled = false
 
-  health-check {
-    dns-backend {
-      # connection information for DNS backend
-      address = "localhost"
-      port = 19001
-
-      # timeout in millis to test connection
-      timeout = 20000
-    }
-  }
+  health-check.dns-backend.timeout = 10000
 
   queue {
     class-name = "vinyldns.sqs.queue.SqsMessageQueueProvider"
--- a/modules/api/src/main/resources/reference.conf
+++ b/modules/api/src/main/resources/reference.conf
@@ -8,8 +8,6 @@ vinyldns {
   # if we should start up polling for change requests, set this to false for the inactive cluster
   processing-disabled = false
 
-  health-check.dns-backend.timeout = 10000
-
   queue {
     class-name = "vinyldns.sqs.queue.SqsMessageQueueProvider"
 
--- a/modules/api/src/main/scala/vinyldns/api/Boot.scala
+++ b/modules/api/src/main/scala/vinyldns/api/Boot.scala
@@ -20,7 +20,6 @@ import akka.actor.ActorSystem
 import akka.http.scaladsl.Http
 import akka.stream.{ActorMaterializer, Materializer}
 import cats.effect.{ContextShift, IO, Timer}
-import pureconfig.module.catseffect.loadConfigF
 import fs2.concurrent.SignallingRef
 import io.prometheus.client.CollectorRegistry
 import io.prometheus.client.dropwizard.DropwizardExports
@@ -80,9 +79,7 @@ object Boot extends App {
       batchChangeLimit <- IO(VinylDNSConfig.vinyldnsConfig.getInt("batch-change-limit"))
       syncDelay <- IO(VinylDNSConfig.vinyldnsConfig.getInt("sync-delay"))
       msgsPerPoll <- IO.fromEither(MessageCount(queueConfig.messagesPerPoll))
-      healthCheckTimeout <- loadConfigF[IO, Option[Int]](
-        VinylDNSConfig.vinyldnsConfig,
-        "health-check-timeout")
+      healthCheckTimeout <- VinylDNSConfig.healthCheckTimeout
       _ <- CommandHandler
         .run(
           messageQueue,
@@ -111,8 +108,7 @@ object Boot extends App {
         zoneValidations,
         AccessValidations)
       val healthService = new HealthService(
-        messageQueue.healthCheck :: connectionValidator.healthCheck(
-          healthCheckTimeout.getOrElse(10000)) ::
+        messageQueue.healthCheck :: connectionValidator.healthCheck(healthCheckTimeout) ::
           loaderResponse.healthChecks)
       val batchChangeConverter =
         new BatchChangeConverter(repositories.batchChangeRepository, messageQueue)
--- a/modules/api/src/main/scala/vinyldns/api/Boot.scala
+++ b/modules/api/src/main/scala/vinyldns/api/Boot.scala
@@ -20,6 +20,7 @@ import akka.actor.ActorSystem
 import akka.http.scaladsl.Http
 import akka.stream.{ActorMaterializer, Materializer}
 import cats.effect.{ContextShift, IO, Timer}
+import pureconfig.module.catseffect.loadConfigF
 import fs2.concurrent.SignallingRef
 import io.prometheus.client.CollectorRegistry
 import io.prometheus.client.dropwizard.DropwizardExports
@@ -79,6 +80,9 @@ object Boot extends App {
       batchChangeLimit <- IO(VinylDNSConfig.vinyldnsConfig.getInt("batch-change-limit"))
       syncDelay <- IO(VinylDNSConfig.vinyldnsConfig.getInt("sync-delay"))
       msgsPerPoll <- IO.fromEither(MessageCount(queueConfig.messagesPerPoll))
+      healthCheckTimeout <- loadConfigF[IO, Option[Int]](
+        VinylDNSConfig.vinyldnsConfig,
+        "health-check-timeout")
       _ <- CommandHandler
         .run(
           messageQueue,
@@ -107,7 +111,8 @@ object Boot extends App {
         zoneValidations,
         AccessValidations)
       val healthService = new HealthService(
-        messageQueue.healthCheck :: connectionValidator.healthCheck ::
+        messageQueue.healthCheck :: connectionValidator.healthCheck(
+          healthCheckTimeout.getOrElse(10000)) ::
           loaderResponse.healthChecks)
       val batchChangeConverter =
         new BatchChangeConverter(repositories.batchChangeRepository, messageQueue)
--- a/modules/api/src/main/scala/vinyldns/api/Boot.scala
+++ b/modules/api/src/main/scala/vinyldns/api/Boot.scala
@@ -106,7 +106,9 @@ object Boot extends App {
         messageQueue,
         zoneValidations,
         AccessValidations)
-      val healthService = new HealthService(messageQueue.healthCheck :: loaderResponse.healthChecks)
+      val healthService = new HealthService(
+        messageQueue.healthCheck :: connectionValidator.healthCheck ::
+          loaderResponse.healthChecks)
       val batchChangeConverter =
         new BatchChangeConverter(repositories.batchChangeRepository, messageQueue)
       val batchChangeService =
--- a/modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
+++ b/modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
@@ -73,7 +73,4 @@ object VinylDNSConfig {
     val primaryServer = connectionConfig.getString("primaryServer")
     ZoneConnection(name, keyName, key, primaryServer).encrypted(Crypto.instance)
   }
-
-  lazy val dnsBackendHealthCheckConfig: Config =
-    vinyldnsConfig.getConfig("health-check.dns-backend")
 }
--- a/modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
+++ b/modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
@@ -73,4 +73,8 @@ object VinylDNSConfig {
     val primaryServer = connectionConfig.getString("primaryServer")
     ZoneConnection(name, keyName, key, primaryServer).encrypted(Crypto.instance)
   }
+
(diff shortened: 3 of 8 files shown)


modules/api/src/main/scala/vinyldns/api/Boot.scala
```diff
@@ -106,7 +106,9 @@ object Boot extends App {
         messageQueue,
         zoneValidations,
         AccessValidations)
-      val healthService = new HealthService(messageQueue.healthCheck :: loaderResponse.healthChecks)
+      val healthService = new HealthService(
+        messageQueue.healthCheck :: connectionValidator.healthCheck ::
```
mitruly: All elements `E` have to proceed the `List[E]` with `::`.

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneConnectionValidator.scala
```diff
@@ -85,5 +87,14 @@ class ZoneConnectionValidator(defaultConnection: ZoneConnection)
     }
   }
 
+  def healthCheck(): HealthCheck =
+    dnsConnection(defaultConnection)
+      .queryDnsBackend(healthCheckRecordName, defaultConnection.name, RecordType.A)
```
mitruly: `queryDnsBackend` calls `runQuery` which takes care of successful cases for us (`SUCCESSFUL`, `TYPE_NOT_FOUND` and `HOST_NOT_FOUND`.

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneConnectionValidator.scala
```diff
@@ -85,5 +87,14 @@ class ZoneConnectionValidator(defaultConnection: ZoneConnection)
     }
   }
 
+  def healthCheck(): HealthCheck =
```
mitruly: Not sure if there is a better place for this, but `ZoneConnectionValidator` made sense to me.

modules/api/src/main/scala/vinyldns/api/domain/dns/DnsConnection.scala
```diff
@@ -194,6 +189,17 @@ class DnsConnection(val resolver: DNS.SimpleResolver) extends DnsConversions {
       case DNS.Lookup.SUCCESSFUL => Right(toFlattenedRecordSets(answers, query.zoneName))
     }
   }
+
+  private[domain] def queryDnsBackend(
```
pauljamescleary: I don't see the reason for doing this.  Just use `resolve` the way it is, and figure out usage at the call site.

In general, avoid `private` helper functions

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneConnectionValidator.scala
pauljamescleary: There is no guarantee that at the `defaultConnection.name` is any kind of zone name.  It actually maybe an _invalid_ zone name, which would result in a failure for other reasons.

You want to build the query to ensure that it is well-formed so it doesn't fail

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneConnectionValidator.scala
```diff
@@ -85,5 +87,14 @@ class ZoneConnectionValidator(defaultConnection: ZoneConnection)
...
+    dnsConnection(defaultConnection)
+      .queryDnsBackend(healthCheckRecordName, defaultConnection.name, RecordType.A)
+      .asHealthCheck
+
   private[domain] def dnsConnection(conn: ZoneConnection): DnsConnection = DnsConnection(conn)
 }
+
+object ZoneConnectionValidator {
+  private[zone] val healthCheckRecordName = "maybe-exists"
```
pauljamescleary: I don't see a reason for having this.  Hard-coding the query to the call-site is better

modules/api/src/main/scala/vinyldns/api/domain/dns/DnsConnection.scala
mitruly: Using `resolve` would yield a `EitherT[IO, Throwable, A]` and `asHealthCheck` takes an `IO[Either[Throwable, _]]`. Seemed to make the most sense to not unwrap `IO` from `EitherT` for `resolve`, though types are cheap. Are we sure we want to use `resolve` directly?

mitruly: Thanks for the feedback, @pauljamescleary - I'll update accordingly.

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneConnectionValidator.scala
```diff
@@ -85,5 +86,11 @@ class ZoneConnectionValidator(defaultConnection: ZoneConnection)
     }
   }
 
+  def healthCheck(): HealthCheck =
+    dnsConnection(defaultConnection)
+      .resolve("maybe-exists-record", "vinyldns.", RecordType.A)
```
rebstar6: dont use vinyldns. for the zone, the real backends wont have that.

when i change this to look for `some-zone.` this fails

rebstar6: thinking about this more, it seems unlikely this would work without a known zone - in theory, if you tried the default keys on a zone setup with different keys, i imagine that would also be refused.

I imagine we may want to allow people to configure a validation zone. then in the operators guide we'd have to be clear that they need to create that zone in their backend for this to run.

In theory, could make it optional, so if not configured, we just dont check it (or force people to, i'm indifferent t the options here)

pauljamescleary: Was hoping that the `resolve` would work.  If it cannot succeed deterministically, than something like the following sample maybe sufficient.

```scala
    val socket = new java.net.Socket()
    socket.connect(new java.net.InetSocketAddress(address, port),socketTimeout)
    socket.close()
```

rebstar6: fine with ^. I do think a configured test zone (default key) would work, and may be needed, to do any kind of key checks. That said, were probably then also getting in an iffy realm where we also want to have some kind of protection in place so we're not ddos-ing ourselves.

lets just do the basic connect for this PR.

modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
```diff
@@ -73,4 +73,7 @@ object VinylDNSConfig {
     val primaryServer = connectionConfig.getString("primaryServer")
     ZoneConnection(name, keyName, key, primaryServer).encrypted(Crypto.instance)
   }
+
+  lazy val dnsBackendHealthCheckConfig: Config =
```
mitruly: Could potentially make this optional.

modules/api/src/main/resources/reference.conf
```diff
@@ -8,6 +8,17 @@ vinyldns {
...
   processing-disabled = false
 
+  health-check {
+    dns-backend {
+      # connection information for DNS backend
+      address = "localhost"
+      port = 19001
+
+      # timeout in millis to test connection
+      timeout = 20000
```
mitruly: Could probably lower this timeout, but not sure of a good value.

modules/api/src/main/scala/vinyldns/api/domain/dns/DnsConnection.scala
```diff
@@ -104,6 +109,21 @@ class DnsConnection(val resolver: DNS.SimpleResolver) extends DnsConversions {
...
     }.toResult
 
+  def healthCheck(): Try[Unit] = {
+    val socket = new java.net.Socket()
+    val address = VinylDNSConfig.dnsBackendHealthCheckConfig.getString("address")
+    val port = VinylDNSConfig.dnsBackendHealthCheckConfig.getInt("port")
+    val socketTimeout = VinylDNSConfig.dnsBackendHealthCheckConfig.getInt("timeout")
+
+    val result = for {
+      socketConnectionAttempt <- Try(
```
pauljamescleary: you do not need a for comprehension here.  Use a for comprehension when there are multiple steps (at least two) in a process that requires sequential processing.

modules/api/src/main/scala/vinyldns/api/domain/dns/DnsConnection.scala
```diff
@@ -104,6 +109,21 @@ class DnsConnection(val resolver: DNS.SimpleResolver) extends DnsConversions {
...
+        socket.connect(new InetSocketAddress(address, port), socketTimeout))
+    } yield socketConnectionAttempt
+    socket.close()
```
pauljamescleary: should always ensure that we close.  I think `Try` has a `recoverWith` or something similar.

pauljamescleary: If you were to use `IO` instead of `Try`, you could use `Resource` as away to making sure the socket is always closed.  Maybe overkill, but food for thought ... https://typelevel.org/cats-effect/datatypes/resource.html

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneConnectionValidator.scala
```diff
@@ -87,11 +93,16 @@ class ZoneConnectionValidator(defaultConnection: ZoneConnection)
   }
 
   def healthCheck(): HealthCheck =
-    IO {
-      dnsConnection(defaultConnection)
-        .healthCheck()
-        .toEither
-    }.asHealthCheck
+    Resource
+      .fromAutoCloseable(IO(new Socket()))
```
mitruly: `.fromAutoCloseable` is used for `Closeable` Java classes, including `Socket` (since it takes care of the closing for us; meaning that `Resource.make` and specifying a follow-up release function is not needed).

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneConnectionValidator.scala
```diff
@@ -38,6 +40,10 @@ class ZoneConnectionValidator(defaultConnection: ZoneConnection)
 
   val opTimeout: FiniteDuration = 6.seconds
 
+  val (healthCheckAddress, healthCheckPort) =
```
mitruly: Setting values here since they are static (in the sense that they shouldn't be changing once the application is running). Will save unneeded processing for when health checks are being called continuously.

modules/api/src/test/scala/vinyldns/api/domain/zone/ZoneConnectionValidatorSpec.scala
```diff
@@ -66,7 +64,8 @@ class ZoneConnectionValidatorSpec
       IO.pure(mockZoneView)
   }
 
-  private def testDefaultConnection = mock[ZoneConnection]
+  private def testDefaultConnection: ZoneConnection =
```
mitruly: Need to specify an actual `ZoneConnection` here based on new (`healthCheckAddress` and `healthCheckPort`) variables.

modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
```diff
@@ -73,4 +73,7 @@ object VinylDNSConfig {
...
+    vinyldnsConfig.getConfig("health-check.dns-backend")
```
rebstar6: dont need to get a whole config here, just a single value

modules/api/src/main/resources/reference.conf
```diff
@@ -8,6 +8,8 @@ vinyldns {
   # if we should start up polling for change requests, set this to false for the inactive cluster
   processing-disabled = false
 
+  health-check.dns-backend.timeout = 10000
```
rebstar6: seems unnecessarily very nested

modules/api/src/test/scala/vinyldns/api/domain/zone/ZoneConnectionValidatorSpec.scala
rebstar6: why? is anything calling those vals?

mitruly: Unless I make them lazy, a null exception will be thrown from `parseHostAndPort`.

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneConnectionValidator.scala
```diff
@@ -37,6 +40,10 @@ class ZoneConnectionValidator(defaultConnection: ZoneConnection)
...
+    DnsConnection.parseHostAndPort(defaultConnection.primaryServer)
+  val healthCheckSocketTimeout: Int = VinylDNSConfig.dnsBackendHealthCheckConfig.getInt("timeout")
```
rebstar6: dont love accessing config like this (even though we do it elsewhere)

i almost wonder if we should `def healthCheck(timeout: Int):` later and then we avoid it

modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
rebstar6: definitely optional

modules/api/src/main/scala/vinyldns/api/Boot.scala
```diff
@@ -79,6 +80,9 @@ object Boot extends App {
       batchChangeLimit <- IO(VinylDNSConfig.vinyldnsConfig.getInt("batch-change-limit"))
       syncDelay <- IO(VinylDNSConfig.vinyldnsConfig.getInt("sync-delay"))
       msgsPerPoll <- IO.fromEither(MessageCount(queueConfig.messagesPerPoll))
+      healthCheckTimeout <- loadConfigF[IO, Option[Int]](
```
rebstar6: still store the val in VinylDNSConfig:

```
lazy val healthCheckTimeout: IO[Int] = loadConfigF[IO, Option[Int]](vinyldnsConfig,"health-check-timeout").map(_.getOrElse(10000))
```


than in boot can just do `healthCheckTimeout <- VinylDNSConfig.healthCheckTimeout`

rebstar6: also snuck that map/getOrElse in there -- just cleanup

PR #393, opened 2018-12-12
Batch ownerGroupId

### LINKED ISSUE #341: Add ownerGroupID field to batchChange, databases
The ownerGroupID represents the group that will be assigned as the owner for all recordsets created in a shared zone in the batch. We need a new field to be stored on the BatchChange type:

- this field should be optional, default to None
- field needs to be added to the batchchange model, and stored in the blob (protobuf change)

Note: this card should not update the batch endpoints. those will be changed in #352

--------
edit: we dont protobuf batch changes. this actually involves a DB migration to add the field

rebstar6:
Fixes #341 

Changes in this pull request:
- add ownerGroupID field to batch
- store new field in database

### DIFF
--- a/modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlBatchChangeRepositoryIntegrationSpec.scala
+++ b/modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlBatchChangeRepositoryIntegrationSpec.scala
@@ -85,7 +85,8 @@ class MySqlBatchChangeRepositoryIntegrationSpec
         sc2.copy(id = UUID.randomUUID().toString),
         sc3.copy(id = UUID.randomUUID().toString),
         deleteChange.copy(id = UUID.randomUUID().toString)
-      )
+      ),
+      Some("some-owner-group")
     )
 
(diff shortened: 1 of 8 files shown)


modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlBatchChangeRepositoryIntegrationSpec.scala
```diff
@@ -85,7 +85,8 @@ class MySqlBatchChangeRepositoryIntegrationSpec
         sc2.copy(id = UUID.randomUUID().toString),
         sc3.copy(id = UUID.randomUUID().toString),
         deleteChange.copy(id = UUID.randomUUID().toString)
-      )
+      ),
+      Some("some-owner-group")
```
mitruly: Would feel more comfortable if we had a mix of at least one more owner group ID for the `getBatchChangeSummariesByUserId` in the tests. Otherwise I am good with these changes 👍

rebstar6: made random ID instead

PR #394, opened 2018-12-12
correct zoneChange status in ZoneChangeHandler

### LINKED ISSUE #391: zoneChange status changes to 'Synced' before zone sync completes
**Describe the bug**
In the CommandHandler whenever a zone is created or synced we update the zoneChange status and the zone status, however we are updating the zoneChange status before the zone actually syncs or updates. https://github.com/vinyldns/vinyldns/blob/947de02f4a2202bb106bee019a6e46f543e671ca/modules/api/src/main/scala/vinyldns/api/backend/CommandHandler.scala#L135 This is causing intermittent failures in our func tests.

We also should check if a zone status is Syncing we shouldn't allow another sync to go through

britneywright:
Fixes #391.

If a zoneChange is of type Sync or Create we shouldn't change the status to Synced until after the zone syncs.

Changes in this pull request:
-  check the zoneChange status and the zone status, if the zoneChange type is Sync or Create and the zone status is Syncing we won't change the zoneChange status to Synced.

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/engine/ZoneChangeHandler.scala
+++ b/modules/api/src/main/scala/vinyldns/api/engine/ZoneChangeHandler.scala
@@ -17,12 +17,7 @@
 package vinyldns.api.engine
 
 import cats.effect.IO
-import vinyldns.core.domain.zone.{
-  ZoneChange,
-  ZoneChangeRepository,
-  ZoneChangeStatus,
-  ZoneRepository
-}
+import vinyldns.core.domain.zone._
 
 object ZoneChangeHandler {
   def apply(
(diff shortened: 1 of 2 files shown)


modules/api/src/main/scala/vinyldns/api/engine/ZoneChangeHandler.scala
```diff
@@ -36,6 +31,10 @@ object ZoneChangeHandler {
               status = ZoneChangeStatus.Failed,
               systemMessage = Some(duplicateZoneError.message))
           )
+        case Right(_)
```
rebstar6: i dont like the way this logic is spread all over. i think the cleanest thing to do is to just keep all the sync logic in one place. so:

- revert this change
- update the ZoneSyncHandler to do the full save/sync/save for the syncs (and adds)
- in the CommandHandler, rather than calling zoneChangeProcessor/zoneSyncProcessor/zoneChangeProcessor for this, just call zoneSyncProcessor

PR #397, opened 2018-12-13
Allow super users to create shared zones

### LINKED ISSUE #345: Allow super users to create a shared zone
For now, this functionality is super user only.

In order to not overwrite shared accidentally, should create a ZoneInput type with only the fields a user can enter. shared should be Option[Boolean] so if not explicitly set, it will remain as stored in an update.

```
case class ZoneInput(
    name: String,
    email: String,
    id: Option[String] = None,
    connection: Option[ZoneConnection] = None,
    transferConnection: Option[ZoneConnection] = None,
    shared: Option[Boolean] = None,
    acl: ZoneACL = ZoneACL(),
    adminGroupId: String) 
```

Note: ID should be None for a create, exist for an update.

On the create zone endpoint, allow users to add a shared flag (true or false). Default is false if missing. This information will be stored in the existing `shared` field on the zone.

If a non-super user includes shared=true, respond with an error.

--
Testing note: cant func test super user stuff, so have good unit tests. Can func test normal user explicitly sets shared == true (fail), shared == false (succeed), and shared not included (succeed, should already be in our suite now though)

mitruly:
Allow super users to create shared zones.
Fixes #345.

Changes in this pull request:
- Add `CreateZoneInput` class to restrict what is parsed into a zone creation request. Implement corresponding serializer and tie-in into routing.
- Add zone validation to check whether zone is shared, and if so, whether user is a super user.
- Updated existing create zone tests. Added new tests for shared flag.

### DIFF
--- a/modules/api/functional_test/live_tests/shared_zone_test_context.py
+++ b/modules/api/functional_test/live_tests/shared_zone_test_context.py
@@ -87,7 +87,7 @@ class SharedZoneTestContext(object):
                 {
                     'name': '1.9.e.f.c.c.7.2.9.6.d.f.ip6.arpa.',
                     'email': 'test@test.com',
-                    'shared': True,
+                    'shared': False,
                     'adminGroupId': self.ok_group['id'],
                     'connection': {
                         'name': 'ip6.',
@@ -109,7 +109,7 @@ class SharedZoneTestContext(object):
                 {
                     'name': '30.172.in-addr.arpa.',
                     'email': 'test@test.com',
-                    'shared': True,
+                    'shared': False,
                     'adminGroupId': self.ok_group['id'],
                     'connection': {
                         'name': 'ip4.',
@@ -175,7 +175,7 @@ class SharedZoneTestContext(object):
                 {
                     'name': 'system-test.',
                     'email': 'test@test.com',
-                    'shared': True,
+                    'shared': False,
                     'adminGroupId': self.ok_group['id'],
                     'connection': {
                         'name': 'system-test.',
--- a/modules/api/functional_test/live_tests/zones/create_zone_test.py
+++ b/modules/api/functional_test/live_tests/zones/create_zone_test.py
@@ -165,11 +165,11 @@ def test_create_invalid_zone_data(shared_zone_test_context):
     zone = {
         'name': zone_name,
         'email': 'test@test.com',
-        'status': 'invalid_status'
+        'shared': 'invalid_value'
     }
 
     errors = client.create_zone(zone, status=400)['errors']
-    assert_that(errors, contains_inanyorder('Invalid ZoneStatus'))
+    assert_that(errors, contains_inanyorder('Do not know how to convert JString(invalid_value) into boolean'))
 
 
 def test_create_zone_with_connection_failure(shared_zone_test_context):
--- a/modules/api/functional_test/live_tests/zones/create_zone_test.py
+++ b/modules/api/functional_test/live_tests/zones/create_zone_test.py
@@ -1,3 +1,4 @@
+import copy
 import pytest
 import uuid
 
@@ -151,7 +152,7 @@ def test_create_missing_zone_data(shared_zone_test_context):
     }
 
     errors = client.create_zone(zone, status=400)['errors']
-    assert_that(errors, contains_inanyorder('Missing Zone.name', 'Missing Zone.email'))
+    assert_that(errors, contains_inanyorder('Missing Zone.name', 'Missing Zone.email', 'Missing Zone.adminGroupId'))
 
 
 def test_create_invalid_zone_data(shared_zone_test_context):
@@ -165,7 +166,8 @@ def test_create_invalid_zone_data(shared_zone_test_context):
     zone = {
         'name': zone_name,
         'email': 'test@test.com',
-        'shared': 'invalid_value'
+        'shared': 'invalid_value',
+        'adminGroupId': 'admin-group-id'
     }
 
     errors = client.create_zone(zone, status=400)['errors']
@@ -471,3 +473,12 @@ def test_user_cannot_create_zone_with_failed_validations(shared_zone_test_contex
     assert_that(result['errors'], contains_inanyorder(
         contains_string("not-approved.thing.com. is not an approved name server")
     ))
+
+def test_normal_user_cannot_create_shared_zone(shared_zone_test_context):
+    """
+    Test that a normal user cannot create a shared zone
+    """
+    super_zone = copy.deepcopy(shared_zone_test_context.ok_zone)
+    super_zone['shared'] = True
+
+    shared_zone_test_context.ok_vinyldns_client.create_zone(super_zone, status=403)
--- a/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneProtocol.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneProtocol.scala
@@ -200,3 +200,5 @@ case class ConnectionFailed(zone: Zone, message: String) extends Throwable(messa
 
 case class ZoneValidationFailed(zone: Zone, errors: List[String], message: String)
     extends Throwable(message)
+
+case class UnauthorizedSharedZoneAction(msg: String) extends Throwable(msg)
--- a/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneProtocol.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneProtocol.scala
@@ -200,5 +200,3 @@ case class ConnectionFailed(zone: Zone, message: String) extends Throwable(messa
 
 case class ZoneValidationFailed(zone: Zone, errors: List[String], message: String)
     extends Throwable(message)
-
-case class UnauthorizedSharedZoneAction(msg: String) extends Throwable(msg)
--- a/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneService.scala
@@ -66,6 +66,7 @@ class ZoneService(
       _ <- zoneDoesNotExist(zone)
       _ <- adminGroupExists(zone.adminGroupId)
       _ <- canChangeZone(auth, zone).toResult
+      _ <- checkSharedZone(zone, auth.signedInUser).toResult
       createZoneChange <- ZoneChangeGenerator.forAdd(zone, auth).toResult
       _ <- messageQueue.send(createZoneChange).toResult[Unit]
     } yield createZoneChange
--- a/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneService.scala
@@ -62,11 +62,11 @@ class ZoneService(
   def connectToZone(zone: Zone, auth: AuthPrincipal): Result[ZoneCommandResult] =
     for {
       _ <- isValidZoneAcl(zone.acl).toResult
+      _ <- validateSharedZoneAuthorized(zone, auth.signedInUser).toResult
       _ <- connectionValidator.validateZoneConnections(zone)
       _ <- zoneDoesNotExist(zone)
       _ <- adminGroupExists(zone.adminGroupId)
       _ <- canChangeZone(auth, zone).toResult
-      _ <- checkSharedZone(zone, auth.signedInUser).toResult
       createZoneChange <- ZoneChangeGenerator.forAdd(zone, auth).toResult
       _ <- messageQueue.send(createZoneChange).toResult[Unit]
     } yield createZoneChange
--- a/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneValidations.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneValidations.scala
@@ -73,6 +73,6 @@ class ZoneValidations(syncDelayMillis: Int) {
 
   // Validates that the zone is either not shared or shared and the user is a super user
   def validateSharedZoneAuthorized(zone: Zone, user: User): Either[Throwable, Unit] =
-    ensuring(UnauthorizedSharedZoneAction("Only super users are allowed to create shared zones."))(
+    ensuring(NotAuthorizedError("Not authorized to create shared zones."))(
       !zone.shared || user.isSuper)
 }
--- a/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneValidations.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneValidations.scala
@@ -20,6 +20,7 @@ import cats.syntax.either._
 import com.aaronbedra.orchard.CIDR
 import org.joda.time.DateTime
 import vinyldns.api.Interfaces.ensuring
+import vinyldns.core.domain.membership.User
 import vinyldns.core.domain.record.RecordType
 import vinyldns.core.domain.zone.{ACLRule, Zone, ZoneACL}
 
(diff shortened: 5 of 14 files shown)


modules/api/functional_test/live_tests/shared_zone_test_context.py
```diff
@@ -87,7 +87,7 @@ def __init__(self):
                 {
                     'name': '1.9.e.f.c.c.7.2.9.6.d.f.ip6.arpa.',
                     'email': 'test@test.com',
-                    'shared': True,
```
rebstar6: lol

mitruly: Yeah, took me a bit longer than I am comfortable admitting to find this, heh.

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneService.scala
```diff
@@ -66,6 +66,7 @@ class ZoneService(
       _ <- zoneDoesNotExist(zone)
       _ <- adminGroupExists(zone.adminGroupId)
       _ <- userIsMemberOfGroup(zone.adminGroupId, auth).toResult
+      _ <- checkSharedZone(zone, auth.signedInUser).toResult
```
rebstar6: move this earlier in the logic; its a quick check so no need to get through all the other stuff if it fails

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneValidations.scala
```diff
@@ -74,4 +75,8 @@ class ZoneValidations(syncDelayMillis: Int) {
         }
       case None => ().asRight
     }
+
+  def checkSharedZone(zone: Zone, user: User): Either[Throwable, Unit] =
```
rebstar6: i wanna word this differently...

maybe "validateSharedZoneAuthorized" or something? i dunno, ill brainstorm

modules/api/src/main/scala/vinyldns/api/route/DnsJsonProtocol.scala
```diff
@@ -88,6 +89,21 @@ trait DnsJsonProtocol extends JsonValidation {
...
+  case object CreateZoneInputSerializer extends ValidationSerializer[CreateZoneInput] {
+    override def fromJson(js: JValue): ValidatedNel[String, CreateZoneInput] =
+      (
+        (js \ "name")
+          .required[String]("Missing Zone.name")
+          .map(name => if (name.endsWith(".")) name else s"$name."),
+        (js \ "email").required[String]("Missing Zone.email"),
+        (js \ "connection").optional[ZoneConnection],
+        (js \ "transferConnection").optional[ZoneConnection],
+        (js \ "shared").optional[Boolean],
```
rebstar6: `(js \ "shared").default[Boolean](false)`

modules/api/src/main/scala/vinyldns/api/route/DnsJsonProtocol.scala
```diff
@@ -88,6 +89,21 @@ trait DnsJsonProtocol extends JsonValidation {
...
+        (js \ "acl").default[ZoneACL](ZoneACL()),
+        (js \ "adminGroupId").default[String]("system")
```
rebstar6: this really is now required, we just stop it in later validations if missing. i think we should be able to require it here

modules/api/src/main/scala/vinyldns/api/route/ZoneRouting.scala
```diff
@@ -43,8 +43,8 @@ trait ZoneRoute extends Directives {
 
   val zoneRoute = { authPrincipal: AuthPrincipal =>
     (post & path("zones") & monitor("Endpoint.createZone")) {
-      entity(as[Zone]) { zone =>
-        execute(zoneService.connectToZone(encrypt(zone), authPrincipal)) { chg =>
+      entity(as[CreateZoneInput]) { createZoneInput =>
+        execute(zoneService.connectToZone(encrypt(Zone(createZoneInput)), authPrincipal)) { chg =>
```
rebstar6: id rather just have `zoneService.connectToZone` take in `encrypt(createZoneInput)` and construct the zone at the end only when it's needed.

if this turns into a huge change though, this is fine

modules/api/src/test/scala/vinyldns/api/domain/zone/ZoneServiceSpec.scala
```diff
@@ -95,23 +95,24 @@ class ZoneServiceSpec
...
 
-    "returns a InvalidZoneAdminError error if the zone admin group does not exist" in {
+    "return a InvalidZoneAdminError error if the zone admin group does not exist" in {
       doReturn(IO.pure(None)).when(mockZoneRepo).getZoneByName(anyString)
       doReturn(IO.pure(None)).when(mockGroupRepo).getGroup(anyString)
 
       val error = leftResultOf(underTest.connectToZone(zoneAuthorized, okAuth).value)
 
-      error shouldBe a[InvalidZoneAdminError]
+      error shouldBe an[InvalidZoneAdminError]
```
rebstar6: real grammar focus in this pr :P

modules/api/src/test/scala/vinyldns/api/domain/zone/ZoneServiceSpec.scala
```diff
@@ -127,7 +128,31 @@ class ZoneServiceSpec
...
+
+    "succeed if zone is shared and user is a super user" in {
+      val newZone = zoneAuthorized.copy(shared = true)
+      val superAuth = okAuth.copy(signedInUser = okUser.copy(isSuper = true))
+      doReturn(IO.pure(None)).when(mockZoneRepo).getZoneByName(anyString)
+
+      val resultZone = rightResultOf(
+        underTest.connectToZone(newZone, superAuth).map(_.asInstanceOf[ZoneChange]).value).zone
+
+      Option(resultZone.id) shouldBe defined
```
rebstar6: huh? it is by definition right?

if the worry here is null do `resultZone.id should not be null` (or however thats worded)

modules/api/functional_test/live_tests/shared_zone_test_context.py
```diff
@@ -87,7 +87,7 @@ def __init__(self):
...
+                    'shared': False,
                     'adminGroupId': self.ok_group['id'],
                     'connection': {
```
rebstar6: add a test in create_zone_test that a normal user cant create a shared zone

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneValidations.scala
mitruly: Using suggested `validateSharedZoneAuthorized` for now.

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneService.scala
mitruly: Makes sense - moved it before all of the logic that require extra data aggregation.

modules/api/functional_test/live_tests/shared_zone_test_context.py
mitruly: Added test.

modules/api/src/main/scala/vinyldns/api/route/ZoneRouting.scala
mitruly: I think this would work, but would delegate refactoring to the update shared zone PR since there are a lot of shared functions.

modules/api/src/test/scala/vinyldns/api/domain/zone/ZoneServiceSpec.scala
```diff
@@ -95,23 +95,24 @@ class ZoneServiceSpec
       resultZone.name shouldBe zoneAuthorized.name
       resultZone.status shouldBe ZoneStatus.Syncing
       resultZone.connection shouldBe zoneAuthorized.connection
+      resultZone.shared shouldBe false
     }
 
-    "returns a ZoneAlreadyExists error if the zone exists" in {
+    "return a ZoneAlreadyExists error if the zone exists" in {
```
britneywright: why "return" instead of "returns"?

mitruly: That's the grammar aspect. `Creating Zones should return a ZoneAlreadyExists error if the zone exists`

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneProtocol.scala
```diff
@@ -200,3 +200,5 @@ case class ConnectionFailed(zone: Zone, message: String) extends Throwable(messa
 
 case class ZoneValidationFailed(zone: Zone, errors: List[String], message: String)
     extends Throwable(message)
+
+case class UnauthorizedSharedZoneAction(msg: String) extends Throwable(msg)
```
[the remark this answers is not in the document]
britneywright: do we need this if there's already `NotAuthorizedError`?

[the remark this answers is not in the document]
mitruly: I'm indifferent to reusing or making a new one, so would like more input. My reasoning was that a user may be able to create/update a zone other than the shared flag, so having a distinction may be less confusing when working with the API side.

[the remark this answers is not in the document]
mitruly: The behavior from the client side would look the same since the user would still get a 403 Forbidden.

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneService.scala
```diff
@@ -62,6 +62,7 @@ class ZoneService(
   def connectToZone(zone: Zone, auth: AuthPrincipal): Result[ZoneCommandResult] =
     for {
       _ <- isValidZoneAcl(zone.acl).toResult
+      _ <- validateSharedZoneAuthorized(zone, auth.signedInUser).toResult
```
britneywright: can this be combined with what's in `canChangeZone` in a new `canCreateZone` function.

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneValidations.scala
```diff
@@ -69,4 +70,9 @@ class ZoneValidations(syncDelayMillis: Int) {
         }
       case None => ().asRight
     }
+
+  // Validates that the zone is either not shared or shared and the user is a super user
+  def validateSharedZoneAuthorized(zone: Zone, user: User): Either[Throwable, Unit] =
+    ensuring(UnauthorizedSharedZoneAction("Only super users are allowed to create shared zones."))(
```
nimaeskandary: Maybe change this "not authorized to create shared zones" since we would have to remember to change this error message

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneService.scala
mitruly: Would like some feedback on this, as I think it is a good discussion point. With creates, we can perform a validation check before trying to retrieve the zone and membership data. Updates would definitely be combined in some fashion since we need the current zone context.

rebstar6: tricky. feels like a preference question to me, and I dont have strong feelings about where the logic lives. it is definitely an auth/access concern, which generally lives in `AccessValidations`, but then again, i wouldnt necessarily want the "if zone is shared" part of the logic there.

i net out being fine with either

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneValidations.scala
mitruly: I'm definitely interested in name recommendations. Wanted to define an error that can also be used with shared zone updates (with slightly different message text).

modules/api/functional_test/live_tests/zones/create_zone_test.py
```diff
@@ -165,11 +166,12 @@ def test_create_invalid_zone_data(shared_zone_test_context):
...
         'name': zone_name,
         'email': 'test@test.com',
-        'status': 'invalid_status'
+        'shared': 'invalid_value',
+        'adminGroupId': 'admin-group-id'
     }
 
     errors = client.create_zone(zone, status=400)['errors']
-    assert_that(errors, contains_inanyorder('Invalid ZoneStatus'))
+    assert_that(errors, contains_inanyorder('Do not know how to convert JString(invalid_value) into boolean'))
```
rebstar6: why did we change this test?

mitruly: The create zone endpoint used to take in `Zone`, so we were testing passing in an invalid enumeration for `ZoneStatus`. Since `CreateZoneInput` doesn't accept a status parameter, I had to update it to something similar.

rebstar6: right, but why add shared into this test? if you leave it without that value, it still status, right?

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneProtocol.scala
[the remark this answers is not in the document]
rebstar6: use NotAuthorizedError, it takes a string anyway, so you can just feed the specific description for this case

modules/api/functional_test/live_tests/zones/create_zone_test.py
mitruly: If we don't add it explicitly with an incorrect value type, `shared` would default to `false`. JSON deserialization treats missing and mismatched values differently and this test was designated for invalid zone data.

modules/api/src/main/scala/vinyldns/api/route/DnsJsonProtocol.scala
```diff
@@ -88,6 +89,21 @@ trait DnsJsonProtocol extends JsonValidation {
...
+        (js \ "connection").optional[ZoneConnection],
+        (js \ "transferConnection").optional[ZoneConnection],
+        (js \ "shared").default[Boolean](false),
+        (js \ "acl").default[ZoneACL](ZoneACL()),
+        (js \ "adminGroupId").required[String]("Missing Zone.adminGroupId")
+      ).mapN(CreateZoneInput.apply)
+  }
+
   case object ZoneSerializer extends ValidationSerializer[Zone] {
     override def fromJson(js: JValue): ValidatedNel[String, Zone] =
```
rebstar6: make a note on the update PR, i believe once thats in we can remove the ZoneSerializer.fromJson

modules/api/src/main/scala/vinyldns/api/route/ZoneRouting.scala
rebstar6: ok

modules/api/src/test/scala/vinyldns/api/route/VinylDNSJsonProtocolSpec.scala
```diff
@@ -22,70 +22,130 @@ import org.json4s._
...
+          ("transferConnection" -> transferConnection) ~~
+          ("adminGroupId" -> "admin-group-id")
+
       val actual = zone.extract[Zone]
-      anonymize(actual) shouldBe anonymize(expected)
+      anonymize(actual) shouldBe anonymize(completeZone)
+    }
+  }
+
+  "CreateZoneInputSerializer" should {
```
rebstar6: shouldnt we test failure cases here?

modules/api/functional_test/live_tests/zones/create_zone_test.py
rebstar6: im saying its fine to default it to false, the test should still work as it used to because it will fail on the invalid status youve given it. like, no updates to this test (other than adding adminGroupId) means the test still succeeds, right? shared isnt a required field

mitruly: The test fails because no errors are thrown now that `status` will no longer attempt deserialization since it's not part of `CreateZoneInput`. The type mismatch check will only trigger for fields that get picked up.

rebstar6: oooooh duh. silly me, ok got it

PR #398, opened 2018-12-14
move zone change processing into zone sync handler

### LINKED ISSUE #391: zoneChange status changes to 'Synced' before zone sync completes

britneywright:
Fixes #391.

Changes in this pull request:
- move the zone change handling for zone creates and syncs into the ZoneSyncHandler

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/backend/CommandHandler.scala
+++ b/modules/api/src/main/scala/vinyldns/api/backend/CommandHandler.scala
@@ -134,14 +134,7 @@ object CommandHandler {
       message.command match {
         case sync: ZoneChange
             if sync.changeType == ZoneChangeType.Sync || sync.changeType == ZoneChangeType.Create =>
-          val doSync =
-            for {
-              _ <- zoneChangeProcessor(sync) // make sure zone is updated to a syncing status
-              syncChange <- zoneSyncProcessor(sync)
-              _ <- zoneChangeProcessor(syncChange) // update zone to Active
-            } yield syncChange
-
-          outcomeOf(message)(doSync)
+          outcomeOf(message)(zoneSyncProcessor(sync))
 
         case zoneChange: ZoneChange =>
           outcomeOf(message)(zoneChangeProcessor(zoneChange))
@@ -196,7 +189,7 @@ object CommandHandler {
     val recordChangeHandler =
       RecordSetChangeHandler(recordSetRepo, recordChangeRepo, batchChangeRepo)
     val zoneSyncHandler =
-      ZoneSyncHandler(recordSetRepo, recordChangeRepo)
+      ZoneSyncHandler(recordSetRepo, recordChangeRepo, zoneChangeRepo, zoneRepo)
 
     CommandHandler
       .mainFlow(
--- a/modules/api/src/main/scala/vinyldns/api/engine/ZoneSyncHandler.scala
+++ b/modules/api/src/main/scala/vinyldns/api/engine/ZoneSyncHandler.scala
@@ -69,8 +69,6 @@ object ZoneSyncHandler extends DnsConversions with Monitored {
             status = ZoneChangeStatus.Failed,
             systemMessage = Some(duplicateZoneError.message))
         )
-      case Right(_) if zoneChange.zone.status == ZoneStatus.Active =>
-        zoneChangeRepository.save(zoneChange.copy(status = ZoneChangeStatus.Synced))
       case Right(_) =>
         zoneChangeRepository.save(zoneChange)
     }
@@ -131,7 +129,8 @@ object ZoneSyncHandler extends DnsConversions with Monitored {
               _ <- saveRecordSets
             } yield
               zoneChange.copy(
-                zone.copy(status = ZoneStatus.Active, latestSync = Some(DateTime.now)))
+                zone.copy(status = ZoneStatus.Active, latestSync = Some(DateTime.now)),
+                status = ZoneChangeStatus.Synced)
           }
         }
       }
--- a/modules/api/src/main/scala/vinyldns/api/engine/ZoneSyncHandler.scala
+++ b/modules/api/src/main/scala/vinyldns/api/engine/ZoneSyncHandler.scala
@@ -48,14 +48,16 @@ object ZoneSyncHandler extends DnsConversions with Monitored {
         VinylDNSZoneViewLoader.apply): ZoneChange => IO[ZoneChange] =
     zoneChange =>
       for {
-        _ <- saveZoneAndChange(zoneRepository, zoneChangeRepository, zoneChange) // zone status: Syncing
+        _ <- saveZoneAndChange(zoneRepository, zoneChangeRepository, zoneChange) // initial save to store zone status
+        // as Syncing
         syncChange <- runSync(
           recordSetRepository,
           recordChangeRepository,
           zoneChange,
           dnsLoader,
           vinyldnsLoader)
-        _ <- saveZoneAndChange(zoneRepository, zoneChangeRepository, syncChange) // zone status: Active
+        _ <- saveZoneAndChange(zoneRepository, zoneChangeRepository, syncChange) // final save to store zone status
+        // as Active
       } yield syncChange
 
   def saveZoneAndChange(
--- a/modules/api/src/main/scala/vinyldns/api/engine/ZoneSyncHandler.scala
+++ b/modules/api/src/main/scala/vinyldns/api/engine/ZoneSyncHandler.scala
@@ -23,8 +23,14 @@ import org.slf4j.LoggerFactory
 import vinyldns.api.domain.dns.DnsConversions
 import vinyldns.api.domain.zone.{DnsZoneViewLoader, VinylDNSZoneViewLoader}
 import vinyldns.core.domain.record._
-import vinyldns.core.domain.zone.{Zone, ZoneChange, ZoneChangeStatus, ZoneStatus}
+import vinyldns.core.domain.zone.{Zone, ZoneStatus}
 import vinyldns.core.route.Monitored
+import vinyldns.core.domain.zone.{
+  ZoneChange,
+  ZoneChangeRepository,
+  ZoneChangeStatus,
+  ZoneRepository
+}
 
 object ZoneSyncHandler extends DnsConversions with Monitored {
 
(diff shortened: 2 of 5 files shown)


modules/api/src/main/scala/vinyldns/api/engine/ZoneSyncHandler.scala
```diff
@@ -35,71 +41,108 @@ object ZoneSyncHandler extends DnsConversions with Monitored {
...
+      zoneChangeRepository: ZoneChangeRepository,
+      zoneChange: ZoneChange): IO[ZoneChange] =
+    zoneRepository.save(zoneChange.zone).flatMap {
+      case Left(duplicateZoneError) =>
+        zoneChangeRepository.save(
+          zoneChange.copy(
+            status = ZoneChangeStatus.Failed,
+            systemMessage = Some(duplicateZoneError.message))
+        )
+      case Right(_) if zoneChange.zone.status == ZoneStatus.Active =>
```
rebstar6: just do `case Right(_) => save...` here with no status changes. 

let the proper status be passed in (so have runSync change the status to `ZoneChangeStatus.Synced` in that logic)

modules/api/src/main/scala/vinyldns/api/engine/ZoneSyncHandler.scala
```diff
@@ -35,71 +41,108 @@ object ZoneSyncHandler extends DnsConversions with Monitored {
...
-              } yield
-                zoneChange.copy(
-                  zone.copy(status = ZoneStatus.Active, latestSync = Some(DateTime.now)))
-            }
+            // join together the results of saving both the record changes as well as the record sets
+            for {
+              _ <- saveRecordChanges
+              _ <- saveRecordSets
+            } yield
+              zoneChange.copy(
```
rebstar6: ah yeah right here! make it active here

modules/api/src/main/scala/vinyldns/api/engine/ZoneSyncHandler.scala
```diff
@@ -35,71 +41,108 @@ object ZoneSyncHandler extends DnsConversions with Monitored {
...
-                zoneChange.copy(
-                  zone.copy(status = ZoneStatus.Active, latestSync = Some(DateTime.now))))
-            } else {
-              logger.info(
-                s"Zone sync for change $zoneChange found ${changes.size} changes to be saved")
-              val changeSet = ChangeSet(changes).copy(status = ChangeSetStatus.Applied)
+          if (changes.isEmpty) {
+            logger.info(s"Zone sync for change $zoneChange had no records to sync")
+            IO.pure(
+              zoneChange.copy(
```
rebstar6: and here, synced

mitruly:
I think we should try to keep coverage consistent or at least pass the 0.1% threshold.

britneywright:
new tests are still a work in progress, but I think the zoneSyncHandler logic is all there.

modules/api/src/main/scala/vinyldns/api/engine/ZoneSyncHandler.scala
```diff
@@ -35,71 +41,107 @@ object ZoneSyncHandler extends DnsConversions with Monitored {
...
+        _ <- saveZoneAndChange(zoneRepository, zoneChangeRepository, syncChange) // zone status: Active
+      } yield syncChange
 
-          val dnsView = time(s"zone.sync.loadDnsView(${zoneChange.id})")(dnsLoader(zone).load())
-          val vinyldnsView = time(s"zone.sync.loadVinylDNSView(${zoneChange.id})")(
-            vinyldnsLoader(zone, recordSetRepository).load())
-          val recordSetChanges = (dnsView, vinyldnsView).parTupled.map {
-            case (dnsZoneView, vinylDnsZoneView) => vinylDnsZoneView.diff(dnsZoneView)
-          }
+  def saveZoneAndChange(
```
mitruly: Discussed this with @britneywright a little bit and would like to open up the discussion to others: Think that bringing in `saveZoneAndChange` here is not needed since we could just update `ZoneChangeHandler.apply` to have a success case of `case Right(_) => zoneChangeRepository.save(zoneChange)`.

This would require that `ZoneChangeGenerator` to set the expected `ZoneChangeStatus` (potentially having different statuses for the front and back engines).

modules/api/src/main/scala/vinyldns/api/backend/CommandHandler.scala
```diff
@@ -134,14 +134,7 @@ object CommandHandler {
       message.command match {
         case sync: ZoneChange
             if sync.changeType == ZoneChangeType.Sync || sync.changeType == ZoneChangeType.Create =>
-          val doSync =
-            for {
-              _ <- zoneChangeProcessor(sync) // make sure zone is updated to a syncing status
```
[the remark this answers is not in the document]
rebstar6: while it's a bit copy/paste-y, i think the benefits of having the concerns separated outweigh the added lines. right now, the sharing is what made it easy for this bug to be introduced.

the alternative would be to have a `ZoneChangeSaver` that does a clean save and pass that into both the ZoneSyncHandler and ZoneChangeHandler, but may not be worth the overhead

modules/api/src/main/scala/vinyldns/api/engine/ZoneSyncHandler.scala
```diff
@@ -35,71 +41,107 @@ object ZoneSyncHandler extends DnsConversions with Monitored {
...
+      zoneRepository: ZoneRepository,
       dnsLoader: Zone => DnsZoneViewLoader = DnsZoneViewLoader.apply,
       vinyldnsLoader: (Zone, RecordSetRepository) => VinylDNSZoneViewLoader =
         VinylDNSZoneViewLoader.apply): ZoneChange => IO[ZoneChange] =
     zoneChange =>
-      monitor("zone.sync") {
-        time(s"zone.sync(${zoneChange.zoneId})") {
-          val zone = zoneChange.zone
+      for {
+        _ <- saveZoneAndChange(zoneRepository, zoneChangeRepository, zoneChange) // zone status: Syncing
```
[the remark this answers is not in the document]
rebstar6: agreed - just update the comment. can say something like `initial save to store zone status as syncing` or something.

modules/api/src/test/scala/vinyldns/api/engine/ZoneSyncHandlerSpec.scala
```diff
@@ -164,31 +166,154 @@ class ZoneSyncHandlerSpec
...
+  "ZoneSyncHandler" should {
+    "process successful zone sync" in {
+      doReturn(IO.pure(Right(testZoneChange)))
+        .when(zoneRepo)
+        .save(any[Zone])
+
+      val dnsLoader = mock[Zone => DnsZoneViewLoader]
+      doReturn(mockDNSLoader).when(dnsLoader).apply(any[Zone])
+
+      val syncer = ZoneSyncHandler(
```
mitruly: Feel that most of these tests can consolidate these sync handlers as well, since there are like three variations that can be defined for all tests to use:
```
  private val mockDnsZoneViewLoader = mock[Zone => DnsZoneViewLoader]

  private val zoneSync = ZoneSyncHandler(
    recordSetRepo,
    recordChangeRepo,
    zoneChangeRepo,
    zoneRepo,
    mockDnsZoneViewLoader,
    (_, _) => mockVinylDNSLoader)
  private val zoneSyncDnsLoader = ZoneSyncHandler(
    recordSetRepo,
    recordChangeRepo,
    zoneChangeRepo,
    zoneRepo,
    _ => mockDNSLoader,
    (_, _) => mockVinylDNSLoader)
  private val runZoneSync = ZoneSyncHandler.runSync(
    recordSetRepo,
    recordChangeRepo,
    testZoneChange,
    _ => mockDNSLoader,
    (_, _) => mockVinylDNSLoader)
```

modules/api/src/test/scala/vinyldns/api/engine/ZoneSyncHandlerSpec.scala
```diff
@@ -164,31 +181,134 @@ class ZoneSyncHandlerSpec
...
+  "runSync" should {
+    "send the correct zone to the DNSZoneViewLoader" in {
       val captor = ArgumentCaptor.forClass(classOf[Zone])
 
       val dnsLoader = mock[Zone => DnsZoneViewLoader]
       doReturn(mockDNSLoader).when(dnsLoader).apply(any[Zone])
 
       val syncer =
-        ZoneSyncHandler(recordSetRepo, recordChangeRepo, dnsLoader, (_, _) => mockVinylDNSLoader)
+        ZoneSyncHandler.runSync(
```
mitruly: We generally don't need to create intermediate variables (eg. `syncer` in most places) if they aren't being used. This can just be:
```
     ZoneSyncHandler.runSync(
          recordSetRepo,
          recordChangeRepo,
          testZoneChange,
          dnsLoader,
          (_, _) => mockVinylDNSLoader).unsafeRunSync()
```

modules/api/src/test/scala/vinyldns/api/engine/ZoneSyncHandlerSpec.scala
```diff
@@ -164,31 +181,132 @@ class ZoneSyncHandlerSpec
...
+      result.zone.status shouldBe ZoneStatus.Active
+      result.zone.latestSync should not be defined
+    }
+  }
+
+  "saveZoneAndChange" should {
+    "save zone and zoneChange with given statuses" in {
+      doReturn(IO.pure(Right(testZoneChange))).when(zoneRepo).save(testZoneChange.zone)
+
+      val saver = ZoneSyncHandler.saveZoneAndChange(zoneRepo, zoneChangeRepo, testZoneChange)
```
mitruly: Can we also remove `saver` here?

modules/api/src/test/scala/vinyldns/api/engine/ZoneSyncHandlerSpec.scala
```diff
@@ -164,31 +181,132 @@ class ZoneSyncHandlerSpec
...
+
+      savedChange.status shouldBe ZoneChangeStatus.Pending
+      savedChange.zone.status shouldBe ZoneStatus.Syncing
+      savedChange.zone.latestSync shouldBe testZoneChange.zone.latestSync
+    }
+
+    "handle duplicateZoneError" in {
+      doReturn(IO.pure(Left(DuplicateZoneError("error")))).when(zoneRepo).save(testZoneChange.zone)
+
+      val saver = ZoneSyncHandler.saveZoneAndChange(zoneRepo, zoneChangeRepo, testZoneChange)
```
mitruly: `saver` here, too.

PR #403, opened 2018-12-17
minor logging changes

rebstar6:


mitruly:
Looks like the `build a string to be logged` test in `VinylDNSServiceSpec.scala` needs to change the text from `status=200` to `status_code=200`.

Issue #364, opened 2018-11-20
Add test flag to block test users from shared zones


rebstar6:
With shared zone support, our test users (along with all normal users), would have access to all the shared zones in the system. We really only want them to have access to the test shared zones.

To accomplish this, should create some kind of test flag on users and zones. If a recordset change is made, and the user is a test user, the change should only go through if the zone is flagged as a test zone. This applies in batch and the regular recordset endpoints

---

EDIT: this is the full scope of this card:
- add test flag on users
- add a test flag on zones
- disallow recordset edits (batch or in zone) by test users in non-test zones

there will be a followup to address a load endpoint

rebstar6:
Here's what im thinking:
- add test flag on users
- add a test flag on zones
- disallow recordset edits (batch or in zone) by test users in non-test zones
- add an endpoint to load the test data. This will load users and optionally some groups and zones. the shared zones will have to be loaded in this way, normal zones can be loaded with either. at least one group will have to exist in this endpoint in order to load the zone (for the admin group)

for now:
- dont think we need any changes to groups. the test flag on the user should stop changes from going through anyway
- dont think we need changes to zone edits: shared zones cant be edited by anyone outside of the admin group. so long as our test users are not in our shared zones groups, this is fine

rebstar6:
we can potentially limit this further and still preload the zone without an endpoint (add a group and zone in TestDataLoader)

mitruly:
I'm on-board with most of these points. Regarding the endpoint to load the test data - is there specific data/request types that we are supporting? I'm kind of leaning toward the approach of using a TestDataLoader and not using an endpoint.

rebstar6:
updating the details for this card to include the above in the loader. can followup with an endpoint later

PR #407, opened 2018-12-19
add isTest field to zone

rebstar6:
part of https://github.com/vinyldns/vinyldns/issues/364

adds the isTest field to the Zone type

### DIFF
--- a/modules/core/src/main/scala/vinyldns/core/domain/zone/Zone.scala
+++ b/modules/core/src/main/scala/vinyldns/core/domain/zone/Zone.scala
@@ -23,12 +23,12 @@ import vinyldns.core.crypto.CryptoAlgebra
 
 object ZoneStatus extends Enumeration {
   type ZoneStatus = Value
-  val Active, Deleted, PendingUpdate, PendingDelete, Syncing = Value
+  val Active, Deleted, Syncing = Value
 }
 
 import vinyldns.core.domain.zone.ZoneStatus._
 
-case class Zone(
+final case class Zone(
     name: String,
     email: String,
     status: ZoneStatus = ZoneStatus.Active,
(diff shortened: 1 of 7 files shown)


modules/core/src/main/scala/vinyldns/core/domain/zone/Zone.scala
```diff
@@ -60,10 +61,14 @@ case class Zone(
     sb.append("account=\"").append(account).append("\"; ")
     sb.append("adminGroupId=\"").append(adminGroupId).append("\"; ")
     sb.append("status=\"").append(status.toString).append("\"; ")
-    sb.append("shared=\"").append(shared.toString).append("\"; ")
+    sb.append("shared=\"").append(shared).append("\"; ")
     sb.append("connection=\"").append(connection.toString).append("\"; ")
```
mitruly: Noticed that `key`s for `connection` and `transferConnection` aren't encrypted when they print here, but we should take care of that in a separate card.

modules/core/src/main/scala/vinyldns/core/protobuf/ProtobufConversions.scala
```diff
@@ -81,7 +81,8 @@ trait ProtobufConversions {
       name = zn.getName,
       email = zn.getEmail,
       status =
-        if (zn.getStatus() == "Pending") ZoneStatus.Active else ZoneStatus.withName(zn.getStatus()),
+        if (zn.getStatus().startsWith("Pending")) ZoneStatus.Active
```
mitruly: Could get rid of the parens on `.getStatus` on these two calls.

modules/core/src/main/scala/vinyldns/core/domain/zone/Zone.scala
rebstar6: yeah, generally at this point they should be encrypted anyway, but it would be worth just updating ZoneConnection.toString to just print name/keyName/primaryServer...encrypted key is meaningless anyway

mitruly: Yeah, that's also a good point where we could remove it from the `toString`.

Issue #346, opened 2018-11-14
Shared Zones support


rebstar6:
Epic tracking shared zones related cards

RFC: https://github.com/vinyldns/rfcs/blob/master/text/0002-shared-zones.md

nimaeskandary:
Note, a group can be deleted though it is set to be the owner of recordsets. For now we are allowing this, and future updates to those recordsets must change the owner group if it no longer exists

britneywright:
can we close this epic?

Issue #354, opened 2018-11-14
Incorporate owner group into create/update recordset endpoints


rebstar6:
In create and update recordset, allow users to input an ownerGroupId on the recordset (optional).

Create/update of owner group id will allow super users to do some management before making a zone shared. Because of that, this will be available regardless of if the zone is shared state.

Only super users should be able to update this field to any existing group in the system. For now, if any other users includes an ownerGroupId in their request that is not the same as the existing one, respond with an error.

nimaeskandary:
this is being opened up to zone admins and members of the existing owner group

nimaeskandary:
Believe this is temporary for some pre-management, and logic will be removed and replaced with https://github.com/vinyldns/vinyldns/issues/422 at a later date

PR #412, opened 2018-12-21
add data to test shared zones in func tests

britneywright:
Part of #346.

This is initial func test data setup for the shared zone work we're doing.

Changes in this pull request:
- create a shared zone in the test data loader, along with an admin group for it and a user for the admin group
- test that the zone can be retrieved 
- test that the user can create a recordset in the shared zone with an owner group ID


modules/api/src/main/scala/vinyldns/api/repository/TestDataLoader.scala
```diff
@@ -139,10 +151,39 @@ object TestDataLoader {
...
+          userRepo.save(user)
+      }.parSequence
+      _ <- groupRepo.save(sharedZoneGroup)
+      _ <- membershipRepo.addMembers(
+        groupId = "shared-zone-group",
+        memberUserIds = Set(sharedZoneUser.id))
+      _ <- zoneRepo.save(sharedZone)
+    } yield ()
 }
+
```
mitruly: Can remove this new line.

modules/api/functional_test/utils.py
```diff
@@ -393,6 +393,7 @@ def clear_groups(client, exclude=[]):
 
     for group_id in group_ids:
         if not group_id in exclude:
+            print "group to delete: " + group_id
```
mitruly: More of a question: do we need this debug statement?

modules/api/functional_test/live_tests/zones/get_zone_test.py
```diff
@@ -19,6 +19,17 @@ def test_get_zone_by_id(shared_zone_test_context):
     assert_that(retrieved['id'], is_(shared_zone_test_context.system_test_zone['id']))
     assert_that(retrieved['adminGroupName'], is_(shared_zone_test_context.ok_group['name']))
 
+def test_get_shared_zone_by_id(shared_zone_test_context):
+    """
+    Test get an existings shared zone by id
```
mitruly: Should be `existing` here.

modules/api/functional_test/live_tests/recordsets/create_recordset_test.py
```diff
@@ -1741,3 +1741,78 @@ def test_create_high_value_domain_fails_for_ip6_ptr(shared_zone_test_context):
 
     error_ptr = client.create_recordset(ptr, status=422)
     assert_that(error_ptr, is_('Record name "fd69:27cc:fe91:0000:0000:0000:0000:ffff" is configured as a High Value Domain, so it cannot be modified.'))
+
+
+def test_create_recordset_with_record_ownergroup(shared_zone_test_context):
+    """
+    Test creating a new record set in an existing zone with an assigned record ownergroup
```
mitruly: Might make more sense to change "ownergroup" to "owner group ID" here.

modules/api/functional_test/live_tests/recordsets/create_recordset_test.py
```diff
@@ -1741,3 +1741,78 @@ def test_create_high_value_domain_fails_for_ip6_ptr(shared_zone_test_context):
...
+        assert_that('10.2.2.2', is_in(rdata_strings))
+    finally:
+        if result_rs:
+            try:
+                delete_result = client.delete_recordset(result_rs['zoneId'], result_rs['id'], status=202)
+                client.wait_until_recordset_change_status(delete_result, 'Complete')
+            except:
+                pass
+
+        result = client.delete_group(created_group['id'], status=200)
```
mitruly: I think this block should be within the `try` in `finally` since it's clean up that might fail (since `create_group` might fail).

modules/api/functional_test/live_tests/recordsets/create_recordset_test.py
```diff
@@ -1741,3 +1741,82 @@ def test_create_high_value_domain_fails_for_ip6_ptr(shared_zone_test_context):
...
+            'ttl': 100,
+            'records': [
+                {
+                    'address': '10.1.1.1'
+                },
+                {
+                    'address': '10.2.2.2'
+                }
+            ],
+            'ownerGroupId': created_group['id']
```
nimaeskandary: it seems this conflicts with #354 actually, wonder how we should handle func testing if only super user can edit the field

britneywright: We talked about this before the holidays. I believe we settled on super users and zone admins being able to modify the owner group ID to a record set. That's how the current logic works at least, so this test passes under current constraints. But if you make it only accessible to super users this test can be changed. This was just a one off test I added to make sure the data worked as expected. @rebstar6 can chime in if I'm recalling our discussion correctly.

rebstar6: i believe were all good here with the update at https://github.com/vinyldns/vinyldns/issues/354#issuecomment-451266100

let me know if not

modules/api/functional_test/live_tests/shared_zone_test_context.py
```diff
@@ -224,6 +225,26 @@ def __init__(self):
                 }, status=202)
             self.parent_zone = parent_zone_change['zone']
 
+            get_shared_zone = self.shared_zone_vinyldns_client.get_zone('shared-zone')
```
rebstar6: add comments here about why we are doing this

modules/api/functional_test/live_tests/shared_zone_test_context.py
```diff
@@ -234,12 +255,16 @@ def __init__(self):
             self.ok_vinyldns_client.wait_until_zone_exists(classless_zone_delegation_change)
             self.ok_vinyldns_client.wait_until_zone_exists(system_test_zone_change)
             self.ok_vinyldns_client.wait_until_zone_exists(parent_zone_change)
+            self.shared_zone_vinyldns_client.sync_zone(shared_zone['id'])
```
rebstar6: need to reorder the sync and wait. the sync cant happen until after the keys have been successfully updated

modules/api/functional_test/live_tests/zones/get_zone_test.py
```diff
@@ -19,6 +19,17 @@ def test_get_zone_by_id(shared_zone_test_context):
...
+    """
+    Test get an existing shared zone by id
+    """
+    client = shared_zone_test_context.shared_zone_vinyldns_client
+
+    result = client.get_zone(shared_zone_test_context.shared_zone['id'], status=200)
+    retrieved = result['zone']
+
+    assert_that(retrieved['id'], is_(shared_zone_test_context.shared_zone['id']))
+    assert_that(retrieved['adminGroupName'], is_('testSharedZoneGroup'))
```
rebstar6: assert_that retrieved['shared'] is true

modules/api/functional_test/live_tests/recordsets/create_recordset_test.py
```diff
@@ -1741,3 +1741,82 @@ def test_create_high_value_domain_fails_for_ip6_ptr(shared_zone_test_context):
 
     error_ptr = client.create_recordset(ptr, status=422)
     assert_that(error_ptr, is_('Record name "fd69:27cc:fe91:0000:0000:0000:0000:ffff" is configured as a High Value Domain, so it cannot be modified.'))
+
+
+def test_create_recordset_with_owner_group_id(shared_zone_test_context):
```
rebstar6: this test adds a lot that we dont need. i think take it out of this PR, it should be up to the later logic PRs to validate their logic works.

PR #413, opened 2018-12-27
clean images and set version in bin/docker-up-vinyldns.sh

nimaeskandary:
Changes in this pull request:
- allows a user to purge vinyldns docker images used by `bin/docker-up-vinyldns.sh`
- allows a user to overwrite the version tag used by `bin/docker-up-vinyldns.sh`

### DIFF
--- a/bin/docker-up-vinyldns.sh
+++ b/bin/docker-up-vinyldns.sh
@@ -9,7 +9,7 @@
 # Options:
 #	-t, --timeout seconds: overwrite ping timeout, default of 60
 #	-a, --api-only: only starts up vinyldns-api and its dependencies, excludes vinyldns-portal
-#	-c, --clean: delete local vinyldns docker images so they are re-pulled from docker hub
+#	-c, --clean: re-pull vinyldns/api and vinyldns/portal images from docker hub
 #	-v, --version tag: overwrite vinyldns/api and vinyldns/portal docker tags
 #####################################################################################################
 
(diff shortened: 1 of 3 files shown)


bin/docker-up-vinyldns.sh
```diff
@@ -49,36 +45,67 @@ function usage {
...
-        -t | --timeout ) TIMEOUT="$2";  shift;;
-        -a | --api-only ) SERVICE="api";;
-        * ) usage; exit;;
-    esac
-    shift
-done
+function clean_images {
+	if (( $CLEAN == 1 )); then
+		echo "cleaning docker images..."
+		docker rmi vinyldns/api:"$VINYLDNS_VERSION"
```
mitruly: Would it make more sense to remove all `vinyldns/api` and `vinyldns/portal` images? Could pass in `$(docker images -a | grep vinyldns/api | awk '{print $3}')` (image ID list) to `docker rmi` if that's the case.

docker/.env
```diff
@@ -14,3 +14,6 @@ DYNAMODB_ENDPOINT=http://vinyldns-dynamodb:8000
 USER_TABLE_NAME=users
 USER_CHANGE_TABLE_NAME=userChange
 TEST_LOGIN=true
+
+# api and portal docker tag
+VINYLDNS_VERSION=0.8.0
```
rebstar6: should this instead work on the `latest` tag?

bin/docker-up-vinyldns.sh
nimaeskandary: hm I would prefer deleting only the images used since others installed wouldn't be involved in the run

mitruly: Yeah, I was thinking about that as well. Realistically, people would want to be using a single image (either snapshot or release) and wouldn't have multiple.

With that said, I'm fine with having `clean` remove only the version that matches the script execution if we update the help text to something like:
```
-c, --clean: re-pull image versions from docker hub
```

PR #418, opened 2019-01-02
test data cleanup

rebstar6:
Been slowly working on consolidating our test data in a reasonable way.

These files were heavily impacted by my next PR, but didnt want this changing around to result in impossible reviews, so this PR is just test cleanup

### DIFF
--- a/modules/api/src/test/scala/vinyldns/api/VinylDNSTestData.scala
+++ b/modules/api/src/test/scala/vinyldns/api/VinylDNSTestData.scala
@@ -22,6 +22,7 @@ import org.joda.time.DateTime
 import vinyldns.api.domain.record.{ListRecordSetChangesResponse, RecordSetChangeGenerator}
 import vinyldns.api.domain.zone._
 import vinyldns.api.repository.TestDataLoader
+import vinyldns.core.TestRecordSetData
 import vinyldns.core.domain.auth.AuthPrincipal
 import vinyldns.core.domain.membership.{Group, User}
 import vinyldns.core.domain.record._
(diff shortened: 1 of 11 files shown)


modules/core/src/main/scala/vinyldns/core/domain/record/RecordSet.scala
```diff
@@ -56,6 +56,7 @@ case class RecordSet(
     val sb = new StringBuilder
     sb.append("RecordSet: [")
     sb.append("id=\"").append(id).append("\"; ")
+    sb.append("zoneId=\"").append(zoneId).append("\"; ")
```
mitruly: Noticed that there is a `RecordSetSpec` that tests `toString`.

modules/api/src/test/scala/vinyldns/api/VinylDNSTestData.scala
```diff
@@ -58,45 +57,13 @@ trait VinylDNSTestData {
...
-  val rsOk: RecordSet = RecordSet(
-    okZone.id,
-    "ok",
-    RecordType.A,
-    200,
-    RecordSetStatus.Active,
-    DateTime.now,
-    None,
-    List(AData("10.1.1.1")))
+  val rsOk: RecordSet = TestRecordSetData.rsOk
```
mitruly: Hmm, it might make more sense to just `import vinyldns.core.TestRecordSetData._` so that we don't have to redeclare the same variable names.

rebstar6: i left it intentionally - i find `val rsOk = rsOk` to be confusing

rebstar6: updated

mitruly: Oh, I meant that if we import `TestRecordSetData._`, we don't have to do `val rsOk = ...` at all since `rsOk` will resolve to the declaration in `TestRecordSetData`.

rebstar6: oh lol, yeah for some reason it doesnt flow down like that. shrugs.

PR #425, opened 2019-01-07
include recordset ownergroup name in getRecordSet response

### LINKED ISSUE #350: Incorporate owner group into recordset endpoints
- Return the owner group ID AND name in the getRecordset/listRecordsets endpoints
- in create and update recordset, allow users to input an ownerGroupId on the recordset (optional). 

Note that create/update of owner group id will not be surfaced in the portal, but will allow users to do some management before making a zone shared. Because of that, this will be available regardless of if the zone is shared state.

Logically, the zone admin and super users should be able to create/update this field to any existing group in the system. For now, lock down any other user from making this kind of update.

britneywright:
Fixes part of #350.

Changes in this pull request:
- only changes getRecordSet right now

### DIFF
--- a/modules/api/functional_test/live_tests/recordsets/get_recordset_test.py
+++ b/modules/api/functional_test/live_tests/recordsets/get_recordset_test.py
@@ -136,18 +136,6 @@ def test_get_recordset_from_shared_zone(shared_zone_test_context):
     client = shared_zone_test_context.shared_zone_vinyldns_client
     retrieved_rs = None
     try:
-        new_group = {
-            'name': 'record-ownergroup',
-            'email': 'test@test.com',
-            'description': 'this is a description',
-            'members': [ { 'id': 'sharedZoneUser'}, { 'id': 'ok'} ],
-            'admins': [ { 'id': 'sharedZoneUser'}, { 'id': 'ok'} ]
-        }
-        created_group = client.create_group(new_group, status=200)
-
-        assert_that(created_group['members'], has_length(2))
-        assert_that(created_group['admins'], has_length(2))
-
         new_rs = {
             'zoneId': shared_zone_test_context.shared_zone['id'],
             'name': 'test_get_recordset',
@@ -161,7 +149,7 @@ def test_get_recordset_from_shared_zone(shared_zone_test_context):
                     'address': '10.2.2.2'
                 }
             ],
-            'ownerGroupId': created_group['id']
+            'ownerGroupId': shared_zone_test_context.shared_record_group['id']
         }
         result = client.create_recordset(new_rs, status=202)
         result_rs = client.wait_until_recordset_change_status(result, 'Complete')['recordSet']
@@ -181,5 +169,3 @@ def test_get_recordset_from_shared_zone(shared_zone_test_context):
         if retrieved_rs:
             delete_result = client.delete_recordset(retrieved_rs['zoneId'], retrieved_rs['id'], status=202)
             client.wait_until_recordset_change_status(delete_result, 'Complete')
-        result = client.delete_group(created_group['id'], status=200)
-        assert_that(result['status'], is_('Deleted'))
--- a/modules/api/functional_test/live_tests/recordsets/get_recordset_test.py
+++ b/modules/api/functional_test/live_tests/recordsets/get_recordset_test.py
@@ -210,21 +210,10 @@ def test_get_owned_recordset_from_not_shared_zone(shared_zone_test_context):
     client = shared_zone_test_context.ok_vinyldns_client
     result_rs = None
     try:
-        new_rs = {
-            'zoneId': shared_zone_test_context.ok_zone['id'],
-            'name': 'test_cant_get_owned_recordset',
-            'type': 'A',
-            'ttl': 100,
-            'records': [
-                {
-                    'address': '10.1.1.1'
-                },
-                {
-                    'address': '10.2.2.2'
-                }
-            ],
-            'ownerGroupId': shared_zone_test_context.shared_record_group['id']
-        }
+        new_rs = get_recordset_json(shared_zone_test_context.ok_zone,
+                                    "test_cant_get_owned_recordset", "TXT", [{'text':'should-work'}],
+                                    100,
+                                    shared_zone_test_context.shared_record_group['id'])
         result = client.create_recordset(new_rs, status=202)
         result_rs = client.wait_until_recordset_change_status(result, 'Complete')['recordSet']
 
--- a/modules/api/functional_test/live_tests/recordsets/get_recordset_test.py
+++ b/modules/api/functional_test/live_tests/recordsets/get_recordset_test.py
@@ -136,21 +136,11 @@ def test_get_recordset_from_shared_zone(shared_zone_test_context):
     client = shared_zone_test_context.shared_zone_vinyldns_client
     retrieved_rs = None
     try:
-        new_rs = {
-            'zoneId': shared_zone_test_context.shared_zone['id'],
-            'name': 'test_get_recordset',
-            'type': 'A',
-            'ttl': 100,
-            'records': [
-                {
-                    'address': '10.1.1.1'
-                },
-                {
-                    'address': '10.2.2.2'
-                }
-            ],
-            'ownerGroupId': shared_zone_test_context.shared_record_group['id']
-        }
+        new_rs = get_recordset_json(shared_zone_test_context.shared_zone,
+                                    "test_get_recordset", "TXT", [{'text':'should-work'}],
+                                    100,
+                                    shared_zone_test_context.shared_record_group['id'])
+
         result = client.create_recordset(new_rs, status=202)
         result_rs = client.wait_until_recordset_change_status(result, 'Complete')['recordSet']
 
@@ -159,11 +149,6 @@ def test_get_recordset_from_shared_zone(shared_zone_test_context):
         retrieved = ok_client.get_recordset(result_rs['zoneId'], result_rs['id'])
         retrieved_rs = retrieved['recordSet']
         verify_recordset(retrieved_rs, new_rs)
-
-        records = [x['address'] for x in retrieved_rs['records']]
-        assert_that(records, has_length(2))
-        assert_that('10.1.1.1', is_in(records))
-        assert_that('10.2.2.2', is_in(records))
 
     finally:
         if retrieved_rs:
@@ -177,20 +162,9 @@ def test_get_unowned_recordset_from_shared_zone(shared_zone_test_context):
     client = shared_zone_test_context.shared_zone_vinyldns_client
     result_rs = None
     try:
-        new_rs = {
-            'zoneId': shared_zone_test_context.shared_zone['id'],
-            'name': 'test_get_unowned_recordset',
-            'type': 'A',
-            'ttl': 100,
-            'records': [
-                {
-                    'address': '10.1.1.1'
-                },
-                {
-                    'address': '10.2.2.2'
-                }
-            ]
-        }
+        new_rs = get_recordset_json(shared_zone_test_context.shared_zone,
+                                    "test_get_unowned_recordset", "TXT", [{'text':'should-not-work'}])
+
         result = client.create_recordset(new_rs, status=202)
         result_rs = client.wait_until_recordset_change_status(result, 'Complete')['recordSet']
 
--- a/modules/api/functional_test/live_tests/recordsets/get_recordset_test.py
+++ b/modules/api/functional_test/live_tests/recordsets/get_recordset_test.py
@@ -150,6 +150,9 @@ def test_get_recordset_from_shared_zone(shared_zone_test_context):
         retrieved_rs = retrieved['recordSet']
         verify_recordset(retrieved_rs, new_rs)
 
+        assert_that(retrieved_rs['ownerGroupId'], is_(shared_zone_test_context.shared_record_group['id']))
+        assert_that(retrieved_rs['ownerGroupName'], is_('record-ownergroup'))
+
     finally:
         if retrieved_rs:
             delete_result = client.delete_recordset(retrieved_rs['zoneId'], retrieved_rs['id'], status=202)
--- a/modules/api/functional_test/live_tests/recordsets/get_recordset_test.py
+++ b/modules/api/functional_test/live_tests/recordsets/get_recordset_test.py
@@ -169,3 +169,37 @@ def test_get_recordset_from_shared_zone(shared_zone_test_context):
         if retrieved_rs:
             delete_result = client.delete_recordset(retrieved_rs['zoneId'], retrieved_rs['id'], status=202)
             client.wait_until_recordset_change_status(delete_result, 'Complete')
+
+def test_get_owned_recordset_from_not_shared_zone(shared_zone_test_context):
+    """
+    Test getting a recordset as the record group owner not in a shared zone fails
+    """
+    client = shared_zone_test_context.ok_vinyldns_client
+    result_rs = None
+    try:
+        new_rs = {
+            'zoneId': shared_zone_test_context.ok_zone['id'],
+            'name': 'test_cant_get_owned_recordset',
+            'type': 'A',
+            'ttl': 100,
+            'records': [
+                {
+                    'address': '10.1.1.1'
+                },
+                {
+                    'address': '10.2.2.2'
+                }
+            ],
+            'ownerGroupId': shared_zone_test_context.shared_record_group['id']
+        }
+        result = client.create_recordset(new_rs, status=202)
+        result_rs = client.wait_until_recordset_change_status(result, 'Complete')['recordSet']
+
+        # Get the recordset we just made and verify
+        shared_client = shared_zone_test_context.shared_zone_vinyldns_client
+        shared_client.get_recordset(result_rs['zoneId'], result_rs['id'], status=403)
+
+    finally:
+        if result_rs:
+            delete_result = client.delete_recordset(result_rs['zoneId'], result_rs['id'], status=202)
+            client.wait_until_recordset_change_status(delete_result, 'Complete')
--- a/modules/api/functional_test/live_tests/recordsets/get_recordset_test.py
+++ b/modules/api/functional_test/live_tests/recordsets/get_recordset_test.py
@@ -128,3 +128,58 @@ def test_at_get_recordset(shared_zone_test_context):
         if result_rs:
             delete_result = client.delete_recordset(result_rs['zoneId'], result_rs['id'], status=202)
             client.wait_until_recordset_change_status(delete_result, 'Complete')
+
+def test_get_recordset_from_shared_zone(shared_zone_test_context):
+    """
+    Test getting a recordset as the record group owner
+    """
+    client = shared_zone_test_context.shared_zone_vinyldns_client
+    retrieved_rs = None
+    try:
+        new_group = {
+            'name': 'record-ownergroup',
+            'email': 'test@test.com',
+            'description': 'this is a description',
+            'members': [ { 'id': 'sharedZoneUser'}, { 'id': 'ok'} ],
+            'admins': [ { 'id': 'sharedZoneUser'}, { 'id': 'ok'} ]
+        }
+        created_group = client.create_group(new_group, status=200)
+
+        assert_that(created_group['members'], has_length(2))
+        assert_that(created_group['admins'], has_length(2))
+
+        new_rs = {
+            'zoneId': shared_zone_test_context.shared_zone['id'],
+            'name': 'test_get_recordset',
+            'type': 'A',
+            'ttl': 100,
+            'records': [
+                {
+                    'address': '10.1.1.1'
+                },
+                {
+                    'address': '10.2.2.2'
+                }
+            ],
+            'ownerGroupId': created_group['id']
+        }
+        result = client.create_recordset(new_rs, status=202)
+        result_rs = client.wait_until_recordset_change_status(result, 'Complete')['recordSet']
+
+        # Get the recordset we just made and verify
+        ok_client = shared_zone_test_context.ok_vinyldns_client
+        retrieved = ok_client.get_recordset(result_rs['zoneId'], result_rs['id'])
+        retrieved_rs = retrieved['recordSet']
+        verify_recordset(retrieved_rs, new_rs)
+
+        records = [x['address'] for x in retrieved_rs['records']]
+        assert_that(records, has_length(2))
+        assert_that('10.1.1.1', is_in(records))
+        assert_that('10.2.2.2', is_in(records))
+
+    finally:
+        if retrieved_rs:
+            delete_result = client.delete_recordset(retrieved_rs['zoneId'], retrieved_rs['id'], status=202)
+            client.wait_until_recordset_change_status(delete_result, 'Complete')
+        result = client.delete_group(created_group['id'], status=200)
+        assert_that(result['status'], is_('Deleted'))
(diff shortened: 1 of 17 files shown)


modules/api/src/main/scala/vinyldns/api/domain/AccessValidations.scala
```diff
@@ -185,4 +190,21 @@ object AccessValidations extends AccessValidationAlgebra {
     }
     case _ => getAccessFromAcl(auth, recordName, recordType, zone)
   }
+
+  def getAccessLevel(
```
nimaeskandary: I don't think you need two separate `getAccessLevel` functions for this, you could have one, and do 

```
case admin if admin.canEditAll || admin.isGroupMember(zone.adminGroupId) ||
      ownerGroupId.exists(admin.isGroupMember) => AccessLevel.Delete
```

modules/api/src/test/scala/vinyldns/api/domain/AccessValidationsSpec.scala
```diff
@@ -274,7 +274,12 @@ class AccessValidationsSpec
```
nimaeskandary: think there should be a test in this file for a success in the case of being in the ownergroup and zone being shared

modules/api/functional_test/live_tests/recordsets/get_recordset_test.py
```diff
@@ -128,3 +128,58 @@ def test_at_get_recordset(shared_zone_test_context):
...
+            'ttl': 100,
+            'records': [
+                {
+                    'address': '10.1.1.1'
+                },
+                {
+                    'address': '10.2.2.2'
+                }
+            ],
+            'ownerGroupId': created_group['id']
```
rebstar6: should we just use ok_group here so we dont have to add a new group?

modules/api/src/main/scala/vinyldns/api/domain/AccessValidations.scala
```diff
@@ -185,4 +190,21 @@ object AccessValidations extends AccessValidationAlgebra {
...
+      auth: AuthPrincipal,
+      recordName: String,
+      recordType: RecordType,
+      zone: Zone,
+      recordOwnerGroupId: Option[String] = None): AccessLevel = auth match {
+    case admin
```
rebstar6: you should just need one getAccessLevel. i think logically it looks like:

```
  def getAccessLevel(
      auth: AuthPrincipal,
      recordName: String,
      recordType: RecordType,
      zone: Zone,
      recordOwnerGroupId: Option[String] = None): AccessLevel = auth match {
    case admin if admin.canEditAll || admin.isGroupMember(zone.adminGroupId) => AccessLevel.Delete
    case sharedAccess if zone.shared && recordOwnerGroupId.exists(sharedAccess.isGroupMember) => AccessLevel.Delete
    case supportUser if supportUser.canReadAll => {
    }
  }
```

modules/api/src/main/scala/vinyldns/api/domain/AccessValidations.scala
```diff
@@ -79,11 +79,16 @@ object AccessValidations extends AccessValidationAlgebra {
       auth: AuthPrincipal,
       recordName: String,
       recordType: RecordType,
-      zone: Zone): Either[Throwable, Unit] =
+      zone: Zone,
+      recordOwnerGroupId: Option[String]): Either[Throwable, Unit] =
     ensuring(
       NotAuthorizedError(s"User ${auth.signedInUser.userName} does not have access to view " +
         s"$recordName.${zone.name}"))(
-      getAccessLevel(auth, recordName, recordType, zone) != AccessLevel.NoAccess)
```
rebstar6: see later change, dont need this split

modules/api/functional_test/live_tests/recordsets/get_recordset_test.py
rebstar6: ugh nvm, need shared user in it. lame.

rebstar6: may consider adding the group in shared_zone_test_context instead though assuming we plan to reuse that logic

modules/api/src/test/scala/vinyldns/api/domain/AccessValidationsSpec.scala
britneywright: got em added now

modules/api/functional_test/live_tests/recordsets/get_recordset_test.py
britneywright: i moved it to shared_zone_test_context

modules/api/src/main/scala/vinyldns/api/domain/AccessValidations.scala
britneywright: removed the old function and updated the check for record owner in a shared zone

modules/api/functional_test/live_tests/recordsets/get_recordset_test.py
```diff
@@ -128,3 +128,44 @@ def test_at_get_recordset(shared_zone_test_context):
         if result_rs:
             delete_result = client.delete_recordset(result_rs['zoneId'], result_rs['id'], status=202)
             client.wait_until_recordset_change_status(delete_result, 'Complete')
+
+def test_get_recordset_from_shared_zone(shared_zone_test_context):
```
nimaeskandary: can we also have a failure case that validates a `403` for a zone that is not set to shared but the record has a recordOwnerId set?

modules/api/functional_test/live_tests/shared_zone_test_context.py
```diff
@@ -42,6 +43,15 @@ def __init__(self):
             # in theory this shouldn't be needed, but getting 'user is not in group' errors on zone creation
             self.confirm_member_in_group(self.dummy_vinyldns_client, self.dummy_group)
 
+            shared_record_group = {
```
nimaeskandary: hm is this necessary? seems using the ok_group id would achieve the same goal

modules/api/src/main/scala/vinyldns/api/route/DnsJsonProtocol.scala
```diff
@@ -245,6 +246,26 @@ trait DnsJsonProtocol extends JsonValidation {
         ("ownerGroupId" -> rs.ownerGroupId)
   }
 
+  case object RecordSetSummaryInfoSerializer extends ValidationSerializer[RecordSetSummaryInfo] {
+    override def fromJson(js: JValue): ValidatedNel[String, RecordSetSummaryInfo] =
```
nimaeskandary: is the fromJson used?

modules/api/functional_test/live_tests/shared_zone_test_context.py
britneywright: I would need to add the shared user to the ok group and see. didn't want to mess with existing test data, but i can try it.

nimaeskandary: hm if you need to add the shared user then I would support keeping the new group, but was thinking you may be able to achieve the get test with only ok user, since they are the only one that has to be in the owner group given that the shared client is a zone admin

modules/api/src/main/scala/vinyldns/api/route/DnsJsonProtocol.scala
britneywright: yes, the recordset routing relies on it.

modules/api/functional_test/live_tests/recordsets/get_recordset_test.py
britneywright: added

modules/api/functional_test/live_tests/shared_zone_test_context.py
britneywright: so currently a non-super user has to be a member of the group they are assigning as a recordset owner. If that changes in the future the group can be removed, but for now it's needed.

modules/api/src/main/scala/vinyldns/api/domain/AccessValidations.scala
```diff
@@ -177,12 +179,15 @@ object AccessValidations extends AccessValidationAlgebra {
...
       recordName: String,
       recordType: RecordType,
-      zone: Zone): AccessLevel = auth match {
-    case admin if admin.canEditAll || admin.isGroupMember(zone.adminGroupId) => AccessLevel.Delete
-    case supportUser if supportUser.canReadAll => {
+      zone: Zone,
+      recordOwnerGroupId: Option[String] = None): AccessLevel = auth match {
+    case admin if admin.canEditAll || admin.isGroupMember(zone.adminGroupId) =>
+      AccessLevel.Delete
+    case recordOwner if zone.shared && recordOwnerGroupId.exists(recordOwner.isGroupMember) =>
```
mitruly: Could technically include this within the `admin` clause.

Also want to confirm that a non-admin is not able to get a record set that has no `ownerGroupId` from a shared zone.

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneProtocol.scala
```diff
@@ -130,6 +131,38 @@ object RecordSetInfo {
   )
 }
 
+case class RecordSetSummaryInfo(
```
mitruly: Once `listRecordsets` with `ownerGroupId` is implemented, may be worth having a discussion as to whether this should be separate from `RecordSetInfo` or merged together.

modules/api/src/test/scala/vinyldns/api/domain/record/RecordSetServiceSpec.scala
```diff
@@ -83,13 +85,31 @@ class RecordSetServiceSpec
...
+    }
+
     "fail when the account is not authorized" in {
       doReturn(IO.pure(Some(aaaa)))
         .when(mockRecordRepo)
         .getRecordSet(zoneNotAuthorized.id, aaaa.id)
       val result = leftResultOf(underTest.getRecordSet(aaaa.id, zoneNotAuthorized.id, okAuth).value)
       result shouldBe a[NotAuthorizedError]
     }
+    "fail if the record already exists" in {
```
mitruly: Is this test for coverage? I was kinda confused as this is testing `addRecordSet`.

modules/api/src/test/scala/vinyldns/api/domain/record/RecordSetServiceSpec.scala
```diff
@@ -339,21 +359,101 @@ class RecordSetServiceSpec
   }
 
   "getRecordSet" should {
-    "return the recordSet" in {
+    doReturn(IO.pure(Some(sharedZone))).when(mockZoneRepo).getZone(sharedZone.id)
+
+    "return the recordSet if user is a zone admin" in {
```
mitruly: This is super nitpicky, but can we change the text from "recordSet" to "record" or "record set" for consistency? I don't think it makes sense to camel-case in the test description.

modules/api/src/test/scala/vinyldns/api/domain/record/RecordSetServiceSpec.scala
```diff
@@ -339,21 +359,101 @@ class RecordSetServiceSpec
...
+          .value)
+      result shouldBe a[NotAuthorizedError]
+    }
+  }
+
+  "getGroupName" should {
+    "return the group name if a record owner group ID is present" in {
+      doReturn(IO.pure(Some(okGroup))).when(mockGroupRepo).getGroup(any[String])
+
+      val result = underTest.getGroupName(Some(okGroup.id))
```
mitruly: The following is less complicated to follow:
```
      val result = rightResultOf(underTest.getGroupName(Some(okGroup.id)).value)
      result shouldBe Some("ok")
```

modules/api/src/test/scala/vinyldns/api/domain/record/RecordSetServiceSpec.scala
```diff
@@ -339,21 +359,101 @@ class RecordSetServiceSpec
...
+      val resultToOption = result.value.unsafeRunSync().right.toOption.get
+      resultToOption shouldBe Some("ok")
+    }
+
+    "return None if a record owner group ID is not present" in {
+      val result = underTest.getGroupName(None)
```
mitruly:
```
      val result = rightResultOf(underTest.getGroupName(None).value)
      result shouldBe None
```

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneProtocol.scala
britneywright: RecordSetInfo includes access rules making it more complicated. That's why I'm splitting it into separate PRs

modules/api/src/test/scala/vinyldns/api/domain/record/RecordSetServiceSpec.scala
britneywright: yeah for test coverage

modules/api/src/it/scala/vinyldns/api/domain/record/RecordSetServiceIntegrationSpec.scala
```diff
@@ -159,7 +160,7 @@ class RecordSetServiceIntegrationSpec
     recordSetRepo =
       DynamoDBRecordSetRepository(recordSetStoreConfig, dynamoIntegrationConfig).unsafeRunSync()
     zoneRepo = zoneRepository
-
+    groupRepo = groupRepository
```
rebstar6: our IT application.conf (https://github.com/vinyldns/vinyldns/blob/master/modules/api/src/it/resources/application.conf) uses the dynamo group repo, not mysql. Because of how mysql starts up, this still technically works, but we should keep the settings here consistent with that config

modules/api/src/it/scala/vinyldns/api/domain/record/RecordSetServiceIntegrationSpec.scala
```diff
@@ -177,6 +178,7 @@ class RecordSetServiceIntegrationSpec
 
     testRecordSetService = new RecordSetService(
       zoneRepo,
+      groupRepo,
```
rebstar6: are there any test changes here that even need the real repo? based on the updates it seems you may be able to just use a mock[GroupRepository] also

modules/api/src/main/scala/vinyldns/api/domain/AccessValidations.scala
rebstar6: i like it split out, this person isnt an admin, its a very different access model even though they still have Delete access

modules/api/functional_test/live_tests/recordsets/get_recordset_test.py
```diff
@@ -128,3 +128,111 @@ def test_at_get_recordset(shared_zone_test_context):
...
+            ],
+            'ownerGroupId': shared_zone_test_context.shared_record_group['id']
+        }
+        result = client.create_recordset(new_rs, status=202)
+        result_rs = client.wait_until_recordset_change_status(result, 'Complete')['recordSet']
+
+        # Get the recordset we just made and verify
+        ok_client = shared_zone_test_context.ok_vinyldns_client
+        retrieved = ok_client.get_recordset(result_rs['zoneId'], result_rs['id'])
+        retrieved_rs = retrieved['recordSet']
```
rebstar6: cut out all the address verification, thats not relevant. What we want to test here is that:

retrieved_rs['ownerGroupId'] is  shared_zone_test_context.shared_record_group['id']
retrieved_rs['ownerGroupName'] is ^^'s name

modules/api/functional_test/live_tests/recordsets/get_recordset_test.py
```diff
@@ -128,3 +128,111 @@ def test_at_get_recordset(shared_zone_test_context):
...
+            client.wait_until_recordset_change_status(delete_result, 'Complete')
+
+def test_get_unowned_recordset_from_shared_zone(shared_zone_test_context):
+    """
+    Test getting an unowned recordset with no admin rights fails
+    """
+    client = shared_zone_test_context.shared_zone_vinyldns_client
+    result_rs = None
+    try:
+        new_rs = {
```
rebstar6: use the `get_recordset_json` helper in utils

new_rs = get_recordset_json(zone, name, type, [{'address': '10.1.1.1', ...}])

modules/api/src/it/scala/vinyldns/api/domain/record/RecordSetServiceIntegrationSpec.scala
```diff
@@ -192,7 +192,7 @@ class RecordSetServiceIntegrationSpec
         .getRecordSet(apexTestRecordA.id, apexTestRecordA.zoneId, auth)
         .value
         .unsafeToFuture()
-        .mapTo[Either[Throwable, RecordSet]]
+        .mapTo[Either[Throwable, RecordSetSummaryInfo]]
```
rebstar6: should probably have added to the RecordSetServiceIntegrationSpec, so back to unmocked.

2 cases: 1 where the record has an ownerGroupId, and that group exists in the group repo, and 2 where has an ownerGroupId and the group does not exist.

in case 2, we should not fail, just return "unknown" or something.

modules/api/functional_test/utils.py
```diff
@@ -493,13 +493,14 @@ def get_change_MX_json(input_name, ttl=200, preference=1, exchange="foo.bar.", c
...
-def get_recordset_json(zone, rname, type, rdata_list, ttl=200):
+def get_recordset_json(zone, rname, type, rdata_list, ttl=200, ownergroup_id=""):
     json = {
         "zoneId": zone['id'],
         "name": rname,
         "type": type,
         "ttl": ttl,
-        "records": rdata_list
+        "records": rdata_list,
+        "ownerGroupId": ownergroup_id
```
rebstar6: do ownergroup_id = Undefined and dont explicitly include it in the request unless specified

modules/api/src/main/scala/vinyldns/api/domain/AccessValidationsAlgebra.scala
```diff
@@ -53,7 +53,8 @@ trait AccessValidationAlgebra {
       auth: AuthPrincipal,
       recordName: String,
       recordType: RecordType,
-      zone: Zone): Either[Throwable, Unit]
+      zone: Zone,
+      recordOwnerGroupId: Option[String]): Either[Throwable, Unit]
```
rebstar6: given that we are now pulling 3 items from the recordset, it may be easier to just have this (

```
  def canViewRecordSet(
      auth: AuthPrincipal,
      recordSet: RecordSet,
      zone: Zone): Either[Throwable, Unit]
```

not a priority if it makes other things conflict

modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetService.scala
```diff
@@ -206,4 +216,12 @@ class RecordSetService(
         RecordSetChangeInfo(change, userMap.get(change.userId)))
     } yield recordSetChangesInfo
   }
+
+  def getGroupName(groupId: Option[String]): Result[Option[String]] = {
+    val groupName = for {
+      input <- OptionT.fromOption[IO](groupId)
+      dbGet <- OptionT(groupRepository.getGroup(input))
```
rebstar6: to the earlier testing point, we cant fail here in the case that we cant get the group name. ultimately it should be logically teh same as zoneService where we do

```
  def getGroupName(groupId: String): Result[String] = {
    groupRepository.getGroup(groupId).map {
      case Some(group) => group.name
      case None => "Unknown group name"
    }
  }.toResult
```

rebstar6: just need to deal with the groupId as option instead of existing

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneProtocol.scala
rebstar6: yeah if separate we should probably word them more logically. this one should be RecordSetInfo, and i think the other maybe RecordSetListInfo or something

modules/api/src/main/scala/vinyldns/api/domain/AccessValidationsAlgebra.scala
britneywright: i tried switching to recordSet early on, but batch uses the access validations too with the record name and type coming from ChangeInput or something like that.

modules/api/src/it/scala/vinyldns/api/domain/record/RecordSetServiceIntegrationSpec.scala
```diff
@@ -53,19 +53,25 @@ class RecordSetServiceIntegrationSpec
 
   private var recordSetRepo: DynamoDBRecordSetRepository = _
   private var zoneRepo: ZoneRepository = _
+  private var groupRepo: GroupRepository = _
 
   private var testRecordSetService: RecordSetServiceAlgebra = _
 
   private val user = User("live-test-user", "key", "secret")
+  private val user2 = User("live-test-user", "key", "secret")
```
rebstar6: give user2 different name/key/secret

modules/api/src/it/scala/vinyldns/api/domain/record/RecordSetServiceIntegrationSpec.scala
```diff
@@ -351,6 +394,36 @@ class RecordSetServiceIntegrationSpec
...
+
+    "get a shared record when user is in assigned ownerGroup" in {
+      val result =
+        testRecordSetService
+          .getRecordSet(sharedTestRecord.id, sharedTestRecord.zoneId, auth2)
+          .value
+          .unsafeToFuture()
+          .mapTo[Either[Throwable, RecordSetInfo]]
+      whenReady(result, timeout) { out =>
+        print("OUT")
```
rebstar6: prints :P

modules/api/src/test/scala/vinyldns/api/domain/record/RecordSetServiceSpec.scala
```diff
@@ -339,21 +359,99 @@ class RecordSetServiceSpec
...
+      val result = leftResultOf(
+        underTest
+          .getRecordSet(notSharedZoneRecordWithOwnerGroup.id, zoneNotAuthorized.id, okAuth)
+          .value)
+      result shouldBe a[NotAuthorizedError]
+    }
+  }
+
+  "getGroupName" should {
+    "return the group name if a record owner group ID is present" in {
```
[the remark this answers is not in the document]
rebstar6: i know you have it in func tests, but still add here for good measure ^^

modules/api/src/main/scala/vinyldns/api/domain/AccessValidationsAlgebra.scala
rebstar6: ah ok

modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetService.scala
rebstar6: nvm, per tests this seems addressed :P

modules/api/src/test/scala/vinyldns/api/domain/record/RecordSetServiceSpec.scala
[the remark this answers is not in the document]
britneywright: here it is: https://github.com/vinyldns/vinyldns/blob/190e2043ed00015d9e44222d3664be01b7391511/modules/api/src/test/scala/vinyldns/api/domain/record/RecordSetServiceSpec.scala#L402

PR #426, opened 2019-01-07
ownergroup id logic for create and update recordset endpoints

### LINKED ISSUE #354: Incorporate owner group into create/update recordset endpoints

nimaeskandary:
Fixes #354.

Changes in this pull request:
- being in the ownerGroup only allows updates in a shared zone
- those with normal write, update, privileges can alter owner group id freely
- for non super users, the owner group must exist and the user must be in it

### DIFF
--- a/modules/api/functional_test/live_tests/recordsets/create_recordset_test.py
+++ b/modules/api/functional_test/live_tests/recordsets/create_recordset_test.py
@@ -1769,6 +1769,7 @@ def test_create_with_owner_group_in_private_zone_by_admin_passes(shared_zone_tes
         record_json['ownerGroupId'] = group['id']
         create_response = client.create_recordset(record_json, status=202)
         create_rs = client.wait_until_recordset_change_status(create_response, 'Complete')['recordSet']
+        assert_that(create_rs['ownerGroupId'], is_(group['id']))
 
     finally:
         if create_rs:
@@ -1791,6 +1792,7 @@ def test_create_with_owner_group_in_shared_zone_by_admin_passes(shared_zone_test
         record_json['ownerGroupId'] = group['id']
         create_response = client.create_recordset(record_json, status=202)
         create_rs = client.wait_until_recordset_change_status(create_response, 'Complete')['recordSet']
+        assert_that(create_rs['ownerGroupId'], is_(group['id']))
 
     finally:
         if create_rs:
@@ -1816,6 +1818,7 @@ def test_create_with_owner_group_in_private_zone_by_acl_passes(shared_zone_test_
         record_json['ownerGroupId'] = group['id']
         create_response = client.create_recordset(record_json, status=202)
         create_rs = client.wait_until_recordset_change_status(create_response, 'Complete')['recordSet']
+        assert_that(create_rs['ownerGroupId'], is_(group['id']))
 
     finally:
         clear_ok_acl_rules(shared_zone_test_context)
@@ -1842,6 +1845,7 @@ def test_create_with_owner_group_in_shared_zone_by_acl_passes(shared_zone_test_c
         record_json['ownerGroupId'] = group['id']
         create_response = client.create_recordset(record_json, status=202)
         create_rs = client.wait_until_recordset_change_status(create_response, 'Complete')['recordSet']
+        assert_that(create_rs['ownerGroupId'], is_(group['id']))
 
     finally:
         clear_shared_zone_acl_rules(shared_zone_test_context)
--- a/modules/api/functional_test/live_tests/recordsets/create_recordset_test.py
+++ b/modules/api/functional_test/live_tests/recordsets/create_recordset_test.py
@@ -1754,9 +1754,9 @@ def test_no_add_access_non_test_zone(shared_zone_test_context):
     client.create_recordset(record, status=403)
 
 
-def test_create_with_owner_group_passes(shared_zone_test_context):
+def test_create_with_owner_group_in_private_zone_by_admin_passes(shared_zone_test_context):
     """
-    Test that creating a record with an owner group passes
+    Test that creating a record with an owner group in a non shared zone by a zone admin passes
     """
 
     client = shared_zone_test_context.ok_vinyldns_client
@@ -1765,7 +1765,7 @@ def test_create_with_owner_group_passes(shared_zone_test_context):
     create_rs = None
 
     try:
-        record_json = get_recordset_json(zone, 'test_shared_success', 'A', [{'address': '1.1.1.1'}])
+        record_json = get_recordset_json(zone, 'test_shared_owner_group_success', 'A', [{'address': '1.1.1.1'}])
         record_json['ownerGroupId'] = group['id']
         create_response = client.create_recordset(record_json, status=202)
         create_rs = client.wait_until_recordset_change_status(create_response, 'Complete')['recordSet']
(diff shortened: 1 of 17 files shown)


modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetService.scala
```diff
@@ -81,7 +85,7 @@ class RecordSetService(
       // because changes happen to the RS in forUpdate itself, converting 1st and validating on that
       rsForValidations = change.recordSet
       _ <- isNotHighValueDomain(recordSet, zone).toResult
-      _ <- canUpdateRecordSet(auth, existing.name, existing.typ, zone).toResult
+      _ <- canUpdateRecordSet(auth, existing.name, existing.typ, zone, recordSet.ownerGroupId).toResult
```
mitruly: This should probably be using `existing.ownerGroupId` here with a different validation for whether the passed-in `ownerGroupId`:
1. Is valid
1. Has the user as a member or the user is a super user

I worked on something similar for batch: https://github.com/vinyldns/vinyldns/pull/427/files#diff-3d42091f87b70df0dd77babb5a2a5a8f.

modules/api/src/test/scala/vinyldns/api/domain/record/RecordSetServiceSpec.scala
```diff
@@ -296,6 +316,46 @@ class RecordSetServiceSpec
       result shouldBe InvalidRequest(
         HighValueDomainError(s"high-value-domain.${okZone.name}").message)
     }
+    "fail if user is in ownerGroup but zone is not shared" in {
```
mitruly: Super nitpicky, but would prefer not to camel-case in test description unless necessary.

modules/api/src/test/scala/vinyldns/api/domain/AccessValidationsSpec.scala
```diff
@@ -230,7 +240,36 @@ class AccessValidationsSpec
...
+      accessValidationTest.canUpdateRecordSet(
+        userAuth,
+        record.name,
+        record.typ,
+        zone,
+        Some(oneUserDummyGroup.id)
+      ) should be(right)
+    }
+
+    "return false if the user is in the ownerGroup and the zone is not shared" in {
```
mitruly: "owner group" and "record set" rather than their camel-case counterparts would be preferable.

modules/api/functional_test/live_tests/recordsets/create_recordset_test.py
```diff
@@ -1741,3 +1741,52 @@ def test_create_high_value_domain_fails_for_ip6_ptr(shared_zone_test_context):
...
+
+    client = shared_zone_test_context.ok_vinyldns_client
+    zone = shared_zone_test_context.ok_zone
+
+    record_json = get_recordset_json(zone, 'test_shared_success', 'A', [{'address': '1.1.1.1'}])
+    record_json['ownerGroupId'] = 'no-existo'
+    client.create_recordset(record_json, status=422)
+
+
+def test_create_with_owner_group_when_not_member(shared_zone_test_context):
```
mitruly: Should add `_fails` to the end of this test name for consistency.

modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetValidations.scala
```diff
@@ -207,4 +208,14 @@ object RecordSetValidations {
       .map(_ => ())
       .leftMap(errors => InvalidRequest(errors.toList.map(_.message).mkString(", ")))
   }
+
+  def canUseOwnerGroup(
+      groupId: Option[String],
+      authPrincipal: AuthPrincipal): Either[Throwable, Unit] =
+    groupId match {
+      case Some(id) if authPrincipal.signedInUser.isSuper || authPrincipal.isGroupMember(id) =>
```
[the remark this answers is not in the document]
nimaeskandary: I believe having edit access to the zone allows this, while only being in the record owner group while the zone is not shared does not allow it

modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetService.scala
```diff
@@ -224,4 +229,17 @@ class RecordSetService(
     } yield dbGet.name
     groupName.value.toResult[Option[String]]
   }
+
+  def ownerGroupExists(groupId: Option[String]): Result[Option[String]] =
```
mitruly: I think we need to account for `zone.shared` with this check (at least for create record set) as well. I was able to add a record to a shared zone without `ownerGroupId`, which I believe should not happen.

nimaeskandary: may be wrong, but I believe in our discussion during scrum we decided that the record set endpoints that is fine because the people using the interface are zone admins and super users @mitruly

modules/api/functional_test/live_tests/recordsets/create_recordset_test.py
```diff
@@ -1752,3 +1752,52 @@ def test_no_add_access_non_test_zone(shared_zone_test_context):
     zone = shared_zone_test_context.non_test_shared_zone
     record = get_recordset_json(zone, 'non-test-zone-A', 'A', [{'address': '1.2.3.4'}])
     client.create_recordset(record, status=403)
+
+
+def test_create_with_owner_group_passes(shared_zone_test_context):
+    """
+    Test that creating a record with an owner group passes
```
rebstar6: need to test the same in a shared zone

modules/api/functional_test/live_tests/recordsets/update_recordset_test.py
```diff
@@ -1992,3 +1992,113 @@ def test_no_update_access_non_test_zone(shared_zone_test_context):
...
+        update = create_rs
+        update['ownerGroupId'] = None
+        update_response = ok_client.update_recordset(update, status=202)
+        update_rs = shared_client.wait_until_recordset_change_status(update_response, 'Complete')['recordSet']
+
+    finally:
+        if update_rs:
+            delete_result = shared_client.delete_recordset(zone['id'], update_rs['id'], status=202)
+            shared_client.wait_until_recordset_change_status(delete_result, 'Complete')
+
```
rebstar6: similar to create, validate the cross of the 3 user levels (zone admin, acl, ownerGroupId) for both assigned and unassigned records within shared zones.

validating once that ownerGroupId people cant make changes in non-shared zones is good enough for that side

modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetService.scala
```diff
@@ -224,4 +229,17 @@ class RecordSetService(
...
+    groupId match {
+      case Some(id) =>
```
rebstar6: split this up. make a `def getGroupIfProvided(groupId: Option[String]): Result[Option[Group]]` (basically the same as getGroupName). call that and then have a ownerGroupExists in the service that takes in the group returned from the call and the id in the recordset

rebstar6: ^^ that validation should just return `Either[Throwable, Unit]`, not using the string here

modules/api/functional_test/live_tests/recordsets/create_recordset_test.py
```diff
@@ -1752,3 +1752,143 @@ def test_no_add_access_non_test_zone(shared_zone_test_context):
...
+    client = shared_zone_test_context.ok_vinyldns_client
+    zone = shared_zone_test_context.ok_zone
+    group = shared_zone_test_context.shared_record_group
+    create_rs = None
+
+    try:
+        record_json = get_recordset_json(zone, 'test_shared_owner_group_success', 'A', [{'address': '1.1.1.1'}])
+        record_json['ownerGroupId'] = group['id']
+        create_response = client.create_recordset(record_json, status=202)
+        create_rs = client.wait_until_recordset_change_status(create_response, 'Complete')['recordSet']
```
rebstar6: need to assert on all of these that create_rs['ownerGroupId] is group['id']

modules/api/src/it/scala/vinyldns/api/domain/record/RecordSetServiceIntegrationSpec.scala
```diff
@@ -408,18 +422,82 @@ class RecordSetServiceIntegrationSpec
...
           .unsafeToFuture()
           .mapTo[Either[Throwable, RecordSetInfo]]
       whenReady(result, timeout) { out =>
-        rightValue(out).name shouldBe "shared-record"
+        rightValue(out).name shouldBe "shared-record-bad-owner-group"
         rightValue(out).ownerGroupName shouldBe None
       }
     }
+
+    "fail if user is in owner group but zone is not shared" in {
```
britneywright: can these descriptions include the word update to make them more clear when skimming? Like "fail to update if user is in owner group but zone is not shared"

modules/api/src/it/scala/vinyldns/api/domain/record/RecordSetServiceIntegrationSpec.scala
```diff
@@ -408,18 +422,82 @@ class RecordSetServiceIntegrationSpec
...
+
+      val result = testRecordSetService
+        .updateRecordSet(newRecord, auth2)
+        .value
+        .unsafeRunSync()
+
+      leftValue(result) shouldBe a[InvalidRequest]
+    }
+
+    "succeed if user is in owner group and zone is shared" in {
```
britneywright: "update" or "update successfully"

modules/api/src/it/scala/vinyldns/api/domain/record/RecordSetServiceIntegrationSpec.scala
nimaeskandary: good suggestion will do, am used to writing these with the tests broken up by function but can see how it is not nice when the tests are generalized for the whole class

PR #427, opened 2019-01-07
[WIP] Shared zone batch change logic

### LINKED ISSUE #342: Incorporate shared zone logic into batch

**OwnerGroupID**

**Access Rules**

mitruly:
Fixes #342.

Changes in this pull request:
-
-


modules/api/src/main/scala/vinyldns/api/domain/AccessValidationsAlgebra.scala
```diff
@@ -52,6 +52,34 @@ trait AccessValidationAlgebra {
       recordType: RecordType,
       zone: Zone): Either[Throwable, Unit]
 
+  def canAddRecordSet(
```
mitruly: Added a set of access validations with `ownerGroupId` so that the record set endpoints can still use the ones without `ownerGroupId`.

modules/api/src/main/scala/vinyldns/api/domain/AccessValidations.scala
```diff
@@ -178,11 +226,35 @@ object AccessValidations extends AccessValidationAlgebra {
...
+    case admin if admin.canEditAll || admin.isGroupMember(zone.adminGroupId) =>
+      AccessLevel.Delete
+    case supportUser if supportUser.canReadAll =>
       val aclAccess = getAccessFromAcl(auth, recordName, recordType, zone)
       if (aclAccess == AccessLevel.NoAccess) AccessLevel.Read else aclAccess
-    }
     case _ => getAccessFromAcl(auth, recordName, recordType, zone)
   }
+
+  def getAccessLevel(
```
mitruly: Copied `getAccessLevel` so that existing record set validations can use a version without `ownerGroupId`.

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -60,16 +65,22 @@ class BatchChangeService(
       auth: AuthPrincipal): BatchResult[BatchChange] =
     for {
       _ <- validateBatchChangeInputSize(batchChangeInput).toBatchResult
+      _ <- validateOwnerGroupId(batchChangeInput.ownerGroupId, auth).toBatchResult
```
mitruly: Initial validation of `ownerGroupId`, if provided. We wait until `validateChangesWithContext` to determine whether record set is in a shared zone, and if so, whether or not an `ownerGroupId` has been provided.

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -153,6 +164,33 @@ class BatchChangeService(
     allSeq.map(lst => ExistingRecordSets(lst.flatten))
   }
 
+  def validateOwnerGroupId(
```
mitruly: Ideally only the group retrieval should be in the service and the other validations should be in `BatchChangeValidations.scala`, but it did not appear nicely.

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
```diff
@@ -264,17 +270,6 @@ class BatchChangeValidations(changeLimit: Int, accessValidation: AccessValidatio
       case None => ().validNel
     }
 
-  def recordExists(
```
mitruly: `recordExists` was removed here since we need to fetch the existing record set's `ownerGroupId` and determine if user is authorized for updates/deletes.

PR #430, opened 2019-01-08
stuck syncing bug

rebstar6:
if a zone sync results in no changes in the zone, it will be stuck in syncing state. sad face.

### DIFF
--- a/modules/api/src/test/scala/vinyldns/api/engine/ZoneSyncHandlerSpec.scala
+++ b/modules/api/src/test/scala/vinyldns/api/engine/ZoneSyncHandlerSpec.scala
@@ -210,6 +210,36 @@ class ZoneSyncHandlerSpec
       result.zone.latestSync shouldBe defined
     }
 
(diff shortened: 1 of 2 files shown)


modules/api/src/test/scala/vinyldns/api/engine/ZoneSyncHandlerSpec.scala
```diff
@@ -210,6 +210,36 @@ class ZoneSyncHandlerSpec
       result.zone.latestSync shouldBe defined
     }
 
+    "process successful zone sync with no changes" in {
+      doReturn(IO.pure(Right(testZoneChange)))
+        .when(zoneRepo)
+        .save(any[Zone])
+      doReturn(() => IO(testDnsView)).when(mockDNSLoader).load
+      doReturn(() => IO(testDnsView)).when(mockVinylDNSLoader).load
```
rebstar6: returning same thing, no changes ^

PR #431, opened 2019-01-09
Test user rules

### LINKED ISSUE #364: Add test flag to block test users from shared zones

rebstar6:
Fixes #364 

adds in the logic blocking test users from record updates in shared zones. note:

- test users can still update zones (no change to logic there, need to be in admin group)
- test users cannot create/update/delete records in non-test zones, regardless of if they have other access (zone admins even)
- once created, a zone flagged with 'isTest' will remain a test zone forever (no updates to that field)
- we do not return the isTest flag to the user on Create/GetZone - this is mostly hidden behind the scenes. you can still see it if you inspect a zone change, since that is returning the full zone as stored for now


modules/api/functional_test/live_tests/recordsets/create_recordset_test.py
```diff
@@ -1741,3 +1741,24 @@ def test_create_high_value_domain_fails_for_ip6_ptr(shared_zone_test_context):
...
     assert_that(error_ptr, is_('Record name "fd69:27cc:fe91:0000:0000:0000:0000:ffff" is configured as a High Value Domain, so it cannot be modified.'))
+
+
+def test_no_add_access_non_test_zone(shared_zone_test_context):
+    """
+    Test that a test user cannot create a record in a non-test zone (even if admin)
+    """
+
+    client = shared_zone_test_context.shared_zone_vinyldns_client
+    record = {
```
mitruly: `get_recordset_json` (from `utils.py`) can be used to simplify record set JSON creation here.

PR #433, opened 2019-01-10
preserve recordset ownerGroupId during zone sync

### LINKED ISSUE #409: Update sync logic to not override ownerGroupId
If a recordset change is synced in, in the case of an update, the ownergroupid should not be overwritten

update the sync logic to ensure this

britneywright:
Fixes #409

Changes in this pull request:
- add `ownerGroupId` in forSyncUpdate function in RecordSetChangeGenerator with value set as the current recordSet.ownerGroupId value


modules/api/src/test/scala/vinyldns/api/domain/zone/ZoneViewSpec.scala
```diff
@@ -267,6 +269,7 @@ class ZoneViewSpec extends WordSpec with Matchers with VinylDNSTestData {
           val record = rsc.recordSet
           record.id shouldBe expected.id
           record.account shouldBe expected.account
+          record.ownerGroupId shouldBe expected.ownerGroupId
```
mitruly: May make more sense to move this check to the `compute the correct updates` test above (or update the test name).

modules/api/functional_test/live_tests/zones/sync_zone_test.py
```diff
@@ -130,13 +131,23 @@ def test_sync_zone_success(shared_zone_test_context):
 
         assert_that(len(recordsets), is_(10))
         for rs in recordsets:
+            if rs['name'] == 'foo':
```
mitruly: The content that is being tested is good, though it might be easier to contain in its own test. I don't particularly feel strongly about this in either way, as there's only one sync test.

rebstar6: i think its fine, for most stuff id agree, but the sync tests are annoying since you need to respect the wait time.

modules/api/functional_test/live_tests/zones/sync_zone_test.py
```diff
@@ -178,7 +189,13 @@ def test_sync_zone_success(shared_zone_test_context):
 
         changes = client.list_recordset_changes(zone['id'])
         for c in changes['recordSetChanges']:
-            assert_that(c['systemMessage'], is_('Change applied via zone sync'))
+            if c['id'] == foo_rs_change['id']:
+                assert_that(c, is_not(has_key('systemMessage')))
```
rebstar6: dont need this assertion, just do
```
if c['id'] != foo_rs_change['id']:
    assert_that(c['systemMessage'], is_('Change applied via zone sync'))
```

modules/api/src/test/scala/vinyldns/api/domain/zone/ZoneViewSpec.scala
rebstar6: +1 on ^, validate that:

- Diff views the record [Test/A/200/1.1.1.1] from db as the same as [Test/A/200/1.1.1.1/some-owner-group] and does not generate a change there
-  Diff views the record [Test/A/200/2.2.2.2] from db as different than [Test/A/200/1.1.1.1/some-owner-group] and generates an update that includes `some-owner-group` (db) and `2.2.2.2` (backend)

PR #434, opened 2019-01-10
Add owner group ID to batch change (without context validations)

### LINKED ISSUE #342: Incorporate shared zone logic into batch

mitruly:
PR 1 of 2. Fixes #342.

Changes in this pull request:
- Update JSON serializers/deserializers to accept/return owner group for batch change.
- Add group validation for:
  1. valid owner group ID
  2. owner group membership in batch service.
- Set owner group ID when applicable in batch change converter:
  - For creates: if zone is shared and owner group ID is provided
  - For updates: if zone is shared and existing record set doesn't have owner group ID set
- Add unit and functional tests

### DIFF
--- a/modules/api/functional_test/live_tests/batch/create_batch_change_test.py
+++ b/modules/api/functional_test/live_tests/batch/create_batch_change_test.py
@@ -2528,6 +2528,7 @@ def test_create_batch_change_for_shared_zone_updates_recordset_owner_group_id_su
         completed_batch = shared_client.wait_until_batch_change_completed(result)
         to_delete = [(change['zoneId'], change['recordSetId']) for change in completed_batch['changes']]
 
+        assert_that(result['ownerGroupId'], is_('shared-zone-group'))
         assert_change_success_response_values(result['changes'], zone=shared_zone, index=0, record_name="no-owner-group-id",
                                               input_name="no-owner-group-id.shared.", record_data="4.3.2.1")
         assert_change_success_response_values(result['changes'], zone=shared_zone, index=1, record_name="update",
@@ -2573,4 +2574,4 @@ def test_create_batch_change_for_shared_zone_with_unauthorized_owner_group_id_fa
     }
 
     errors = shared_client.create_batch_change(batch_change_input, status=403)
-    assert_that(errors, is_('User "sharedZoneUser" is not a member of group "' + ok_group['id'] + '". Owner group ID is only required for record set changes in shared zones.'))
+    assert_that(errors, is_('User "sharedZoneUser" must be a member of group ' + ok_group['id'] + ' to apply this group to batch changes'))
--- a/modules/api/functional_test/live_tests/batch/create_batch_change_test.py
+++ b/modules/api/functional_test/live_tests/batch/create_batch_change_test.py
@@ -451,6 +451,56 @@ def test_create_batch_change_without_comments_succeeds(shared_zone_test_context)
     finally:
         clear_zoneid_rsid_tuple_list(to_delete, client)
 
+def test_create_batch_change_with_owner_group_id_succeeds(shared_zone_test_context):
+    """
+    Test successfully creating a batch change with owner group ID specified
+    """
+    client = shared_zone_test_context.ok_vinyldns_client
+    ok_zone = shared_zone_test_context.ok_zone
+    batch_change_input = {
+        "changes": [
+            get_change_A_AAAA_json("owner-group-id.ok.", address="4.3.2.1")
+        ],
+        "ownerGroupId": shared_zone_test_context.ok_group['id']
+    }
+    to_delete = []
+
+    try:
+        result = client.create_batch_change(batch_change_input, status=202)
+        completed_batch = client.wait_until_batch_change_completed(result)
+        to_delete = [(change['zoneId'], change['recordSetId']) for change in completed_batch['changes']]
+
+        assert_change_success_response_values(result['changes'], zone=ok_zone, index=0, record_name="owner-group-id",
+                                              input_name="owner-group-id.ok.", record_data="4.3.2.1")
+        assert_that(completed_batch['ownerGroupId'], is_(shared_zone_test_context.ok_group['id']))
+
+    finally:
+        clear_zoneid_rsid_tuple_list(to_delete, client)
+
+def test_create_batch_change_without_owner_group_id_succeeds(shared_zone_test_context):
+    """
+    Test successfully creating a batch change without owner group ID specified
+    """
+    client = shared_zone_test_context.ok_vinyldns_client
+    ok_zone = shared_zone_test_context.ok_zone
+    batch_change_input = {
+        "changes": [
+            get_change_A_AAAA_json("no-owner-group-id.ok.", address="4.3.2.1")
+        ]
+    }
+    to_delete = []
+
+    try:
+        result = client.create_batch_change(batch_change_input, status=202)
+        completed_batch = client.wait_until_batch_change_completed(result)
+        to_delete = [(change['zoneId'], change['recordSetId']) for change in completed_batch['changes']]
+
+        assert_change_success_response_values(result['changes'], zone=ok_zone, index=0, record_name="no-owner-group-id",
+                                              input_name="no-owner-group-id.ok.", record_data="4.3.2.1")
+        assert_that(completed_batch, is_not(has_key('ownerGroupId')))
+
+    finally:
+        clear_zoneid_rsid_tuple_list(to_delete, client)
 
 def test_create_batch_change_partial_failure(shared_zone_test_context):
     """
@@ -2404,3 +2454,123 @@ def test_user_validation_shared(shared_zone_test_context):
                                            error_messages=["User \"ok\" is not authorized."])
     assert_failed_change_in_error_response(response[3], input_name="delete-test-batch.non.test.shared.", change_type="DeleteRecordSet",
                                            error_messages=["User \"ok\" is not authorized."])
+
+def test_create_batch_change_does_not_save_owner_group_id_for_non_shared_zone(shared_zone_test_context):
+    """
+    Test successfully creating a batch change with owner group ID doesn't save value for records in non-shared zone
+    """
+    ok_client = shared_zone_test_context.ok_vinyldns_client
+    ok_zone = shared_zone_test_context.ok_zone
+    ok_group = shared_zone_test_context.ok_group
+
+    update_rs = get_recordset_json(ok_zone, "update", "A", [{"address": "127.0.0.1"}], 300)
+
+    batch_change_input = {
+        "changes": [
+            get_change_A_AAAA_json("no-owner-group-id.ok.", address="4.3.2.1"),
+            get_change_A_AAAA_json("update.ok.", address="1.2.3.4"),
+            get_change_A_AAAA_json("update.ok.", change_type="DeleteRecordSet")
+        ],
+        "ownerGroupId": ok_group['id']
+    }
+    to_delete = []
+
+    try:
+        create_result = ok_client.create_recordset(update_rs, status=202)
+        to_delete.append(ok_client.wait_until_recordset_change_status(create_result, 'Complete'))
+
+        result = ok_client.create_batch_change(batch_change_input, status=202)
+        completed_batch = ok_client.wait_until_batch_change_completed(result)
+        to_delete = [(change['zoneId'], change['recordSetId']) for change in completed_batch['changes']]
+
+        assert_change_success_response_values(result['changes'], zone=ok_zone, index=0, record_name="no-owner-group-id",
+                                              input_name="no-owner-group-id.ok.", record_data="4.3.2.1")
+        assert_change_success_response_values(result['changes'], zone=ok_zone, index=1, record_name="update",
+                                              input_name="update.ok.", record_data="1.2.3.4")
+        assert_change_success_response_values(result['changes'], zone=ok_zone, index=2, record_name="update",
+                                              input_name="update.ok.", change_type="DeleteRecordSet", record_data=None)
+
+        for (zoneId, recordSetId) in to_delete:
+            get_recordset = ok_client.get_recordset(zoneId, recordSetId, status=200)
+            assert_that(get_recordset['recordSet'], is_not(has_key('ownerGroupId')))
+
+    finally:
+        clear_zoneid_rsid_tuple_list(to_delete, ok_client)
+
+def test_create_batch_change_for_shared_zone_updates_recordset_owner_group_id_succeeds(shared_zone_test_context):
+    """
+    Test successfully creating a batch change with owner group ID in shared zone succeeds and sets owner group ID
+    on creates and updates (if existing value does not exist)
+    """
+    shared_client = shared_zone_test_context.shared_zone_vinyldns_client
+    shared_zone = shared_zone_test_context.shared_zone
+
+    update_rs = get_recordset_json(shared_zone, "update", "A", [{"address": "127.0.0.1"}], 300)
+
+    batch_change_input = {
+        "changes": [
+            get_change_A_AAAA_json("no-owner-group-id.shared.", address="4.3.2.1"),
+            get_change_A_AAAA_json("update.shared.", address="1.2.3.4"),
+            get_change_A_AAAA_json("update.shared.", change_type="DeleteRecordSet")
+        ],
+        "ownerGroupId": "shared-zone-group"
+    }
+    to_delete = []
+
+    try:
+        create_result = shared_client.create_recordset(update_rs, status=202)
+        to_delete.append(shared_client.wait_until_recordset_change_status(create_result, 'Complete'))
+
+        create_result = shared_client.get_recordset(create_result['recordSet']['zoneId'], create_result['recordSet']['id'], status=200)
+        assert_that(create_result['recordSet'], is_not(has_key('ownerGroupId')))
+
+        result = shared_client.create_batch_change(batch_change_input, status=202)
+        completed_batch = shared_client.wait_until_batch_change_completed(result)
+        to_delete = [(change['zoneId'], change['recordSetId']) for change in completed_batch['changes']]
+
+        assert_change_success_response_values(result['changes'], zone=shared_zone, index=0, record_name="no-owner-group-id",
+                                              input_name="no-owner-group-id.shared.", record_data="4.3.2.1")
+        assert_change_success_response_values(result['changes'], zone=shared_zone, index=1, record_name="update",
+                                              input_name="update.shared.", record_data="1.2.3.4")
+        assert_change_success_response_values(result['changes'], zone=shared_zone, index=2, record_name="update",
+                                              input_name="update.shared.", change_type="DeleteRecordSet", record_data=None)
+
+        for (zoneId, recordSetId) in to_delete:
+            get_recordset = shared_client.get_recordset(zoneId, recordSetId, status=200)
+            assert_that(get_recordset['recordSet']['ownerGroupId'], is_(batch_change_input['ownerGroupId']))
+
+    finally:
+        clear_zoneid_rsid_tuple_list(to_delete, shared_client)
+
+def test_create_batch_change_for_shared_zone_with_invalid_owner_group_id_fails(shared_zone_test_context):
+    """
+    Test successfully creating a batch change with owner group ID saves value for records in shared zone
+    """
+    shared_client = shared_zone_test_context.shared_zone_vinyldns_client
+
+    batch_change_input = {
+        "changes": [
+            get_change_A_AAAA_json("no-owner-group-id.shared.", address="4.3.2.1")
+        ],
+        "ownerGroupId": "non-existent-owner-group-id"
+    }
+
+    errors = shared_client.create_batch_change(batch_change_input, status=400)
+    assert_that(errors, is_('Group with ID "non-existent-owner-group-id" was not found'))
+
+def test_create_batch_change_for_shared_zone_with_unauthorized_owner_group_id_fails(shared_zone_test_context):
+    """
+    Test successfully creating a batch change with owner group ID saves value for records in shared zone
+    """
+    shared_client = shared_zone_test_context.shared_zone_vinyldns_client
+    ok_group = shared_zone_test_context.ok_group
+
+    batch_change_input = {
+        "changes": [
+            get_change_A_AAAA_json("no-owner-group-id.shared.", address="4.3.2.1")
+        ],
+        "ownerGroupId": ok_group['id']
+    }
+
+    errors = shared_client.create_batch_change(batch_change_input, status=403)
+    assert_that(errors, is_('User "sharedZoneUser." does not belong to group "' + ok_group['id'] + '". Owner group ID is only required for record set changes in shared zones.'))
--- a/modules/api/functional_test/live_tests/batch/create_batch_change_test.py
+++ b/modules/api/functional_test/live_tests/batch/create_batch_change_test.py
@@ -2545,7 +2545,7 @@ def test_create_batch_change_for_shared_zone_updates_recordset_owner_group_id_su
 
 def test_create_batch_change_for_shared_zone_with_invalid_owner_group_id_fails(shared_zone_test_context):
     """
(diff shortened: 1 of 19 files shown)


modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeErrors.scala
```diff
@@ -31,12 +31,19 @@ final case class InvalidBatchChangeResponses(
...
     s"Batch change contained no changes. Batch change must have at least one change, up to a maximum of $limit changes."
 }
-
+final case class GroupDoesNotExist(id: String) extends BatchChangeErrorResponse {
+  def message: String = s"""Group with ID "$id" was not found"""
+}
+final case class UserDoesNotBelongToOwnerGroup(ownerGroupId: String, userName: String)
+    extends BatchChangeErrorResponse {
+  def message: String =
+    s"""User "$userName." does not belong to group "$ownerGroupId". Owner group ID is only required for """ +
```
britneywright: `User "$userName." is not a member of group "$ownerGroupId".`

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeErrors.scala
```diff
@@ -31,12 +31,19 @@ final case class InvalidBatchChangeResponses(
...
-
 final case class BatchChangeIsEmpty(limit: Int) extends BatchChangeErrorResponse {
   def message: String =
     s"Batch change contained no changes. Batch change must have at least one change, up to a maximum of $limit changes."
 }
-
+final case class GroupDoesNotExist(id: String) extends BatchChangeErrorResponse {
+  def message: String = s"""Group with ID "$id" was not found"""
+}
+final case class UserDoesNotBelongToOwnerGroup(ownerGroupId: String, userName: String)
```
britneywright: UserIsNotAMemberOfOwnerGroup

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -153,6 +160,33 @@ class BatchChangeService(
...
+      .getGroup(ownerGroupId)
+      .map {
+        case Some(_) => ().asRight
+        case None => Left(GroupDoesNotExist(ownerGroupId))
+      }
+      .toBatchResult
+
+  def validateCanSeeGroup(ownerGroupId: String, authPrincipal: AuthPrincipal): BatchResult[Unit] =
+    IO {
+      if (canSeeGroup(ownerGroupId, authPrincipal).isRight) ().asRight
```
britneywright: using the wrong function here. canSeeGroup includes support users. Can you use isGroupMember and canEditAll in AuthPrincipal

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -153,6 +159,33 @@ class BatchChangeService(
...
+  def validateOwnerGroupExists(ownerGroupId: String): BatchResult[Unit] =
+    groupRepository
+      .getGroup(ownerGroupId)
+      .map {
+        case Some(_) => ().asRight
+        case None => Left(GroupDoesNotExist(ownerGroupId))
+      }
+      .toBatchResult
+
+  def validateCanSeeGroup(ownerGroupId: String, authPrincipal: AuthPrincipal): BatchResult[Unit] =
```
britneywright: validateGroupMembership or something like that

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeConverter.scala
```diff
@@ -166,15 +170,23 @@ class BatchChangeConverter(batchChangeRepo: BatchChangeRepository, messageQueue:
...
     for {
       deleteChange <- Some(deleteChanges.head)
       zone <- existingZones.getByName(deleteChange.zoneName)
       existingRecordSet <- existingRecordSets.get(
         deleteChange.zoneId,
         deleteChange.recordName,
         deleteChange.typ)
-      newRecordSet = combineAddChanges(addChanges, zone)
+      newRecordSet = {
+        val newOwnerGroupId = if (zone.shared && existingRecordSet.ownerGroupId.isEmpty) {
```
britneywright: i think setOwnerGroupId sounds more accurate here, since it may be a new ID or it could be the existing ID

modules/api/functional_test/live_tests/batch/create_batch_change_test.py
```diff
@@ -2343,3 +2393,123 @@ def test_mx_recordtype_update_delete_checks(shared_zone_test_context):
...
+        to_delete = [(change['zoneId'], change['recordSetId']) for change in completed_batch['changes']]
+
+        assert_change_success_response_values(result['changes'], zone=ok_zone, index=0, record_name="no-owner-group-id",
+                                              input_name="no-owner-group-id.ok.", record_data="4.3.2.1")
+        assert_change_success_response_values(result['changes'], zone=ok_zone, index=1, record_name="update",
+                                              input_name="update.ok.", record_data="1.2.3.4")
+        assert_change_success_response_values(result['changes'], zone=ok_zone, index=2, record_name="update",
+                                              input_name="update.ok.", change_type="DeleteRecordSet", record_data=None)
+
+        for (zoneId, recordSetId) in to_delete:
```
mitruly: Need to perform a `GET` on record sets to check for `ownerGroupId`.

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeConverter.scala
```diff
@@ -206,16 +218,23 @@ class BatchChangeConverter(batchChangeRepo: BatchChangeRepository, messageQueue:
...
       ids = addChanges.map(_.id)
     } yield RecordSetChangeGenerator.forAdd(newRecordSet, zone, userId, ids.toList)
 
   // Combines changes where the RecordData can just be appended to list (A, AAAA, CNAME, PTR)
   // NOTE: CNAME & PTR will only have one data field due to validations, so the combination is fine
-  def combineAddChanges(changes: NonEmptyList[SingleAddChange], zone: Zone): RecordSet = {
+  def combineAddChanges(
+      changes: NonEmptyList[SingleAddChange],
+      zone: Zone,
+      ownerGroupId: Option[String]): RecordSet = {
```
mitruly: `combineAddChanges` is used for only adds and updates (to combine record data for the same record set). Once we're at this point, we know that all validations have gone through successfully, so the `ownerGroupId` to use just needs to get passed in.

modules/api/src/main/scala/vinyldns/api/route/BatchChangeJsonProtocol.scala
```diff
@@ -146,7 +149,8 @@ trait BatchChangeJsonProtocol extends JsonValidation {
         ("createdTimestamp" -> Extraction.decompose(bc.createdTimestamp)) ~
         ("changes" -> Extraction.decompose(bc.changes)) ~
         ("status" -> bc.status.toString) ~
-        ("id" -> bc.id)
+        ("id" -> bc.id) ~
+        ("ownerGroupId" -> bc.ownerGroupId)
```
mitruly: The response from create batch change is still the `BatchChange` entity (and not individual record set changes). `ownerGroupID` on the batch may not necessarily be the same as on the record set.

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeErrors.scala
```diff
@@ -31,12 +31,19 @@ final case class InvalidBatchChangeResponses(
...
 }
-
+final case class GroupDoesNotExist(id: String) extends BatchChangeErrorResponse {
+  def message: String = s"""Group with ID "$id" was not found"""
+}
+final case class NotAMemberOfOwnerGroup(ownerGroupId: String, userName: String)
+    extends BatchChangeErrorResponse {
+  def message: String =
+    s"""User "$userName" is not a member of group "$ownerGroupId". Owner group ID is only required for """ +
+      "record set changes in shared zones."
```
rebstar6: 2nd sentence here feels out of place

rebstar6: maybe s"""User "$userName" must be a member of group "$ownerGroupId" to apply it to batch changes"""

...or something

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -153,6 +159,33 @@ class BatchChangeService(
     allSeq.map(lst => ExistingRecordSets(lst.flatten))
   }
 
+  def validateOwnerGroupId(
```
rebstar6: why BatchService vs BatchValidations?

mitruly: Normally I would put this in `BatchValidations`, but the repo call would have to be in service. Since the repo call is conditional on `ownerGroupId` being passed in, this looked a bit messy trying to split the service and validation. Could split it if we want.

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -153,6 +159,33 @@ class BatchChangeService(
...
+      ownerGroupId: Option[String],
+      authPrincipal: AuthPrincipal): BatchResult[Unit] =
+    ownerGroupId match {
+      case None => IO(().asRight).toBatchResult
```
rebstar6: ().toRightBatchResult

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -153,6 +159,33 @@ class BatchChangeService(
...
+    groupRepository
+      .getGroup(ownerGroupId)
+      .map {
+        case Some(_) => ().asRight
+        case None => Left(GroupDoesNotExist(ownerGroupId))
+      }
+      .toBatchResult
+
+  def validateGroupMembership(ownerGroupId: String, authPrincipal: AuthPrincipal): BatchResult[Unit] =
+    IO {
```
rebstar6: dont wrap in IO, just `toBatchResult` at the end and it will figure it out

modules/api/src/test/scala/vinyldns/api/domain/batch/BatchChangeConverterSpec.scala
```diff
@@ -16,30 +16,33 @@
...
+import vinyldns.core.domain.record.RecordType.{RecordType, _}
+import vinyldns.core.domain.record._
+import vinyldns.core.domain.zone.Zone
 
 class BatchChangeConverterSpec
     extends WordSpec
     with Matchers
     with CatsHelpers
-    with VinylDNSTestData {
+    with BeforeAndAfterEach {
```
rebstar6: why BeforeAndAfterEach?

modules/api/src/test/scala/vinyldns/api/route/BatchChangeRoutingSpec.scala
```diff
@@ -309,6 +315,36 @@ class BatchChangeRoutingSpec
         status shouldBe InternalServerError
       }
     }
+
+    "return a 400 BadRequest for non-existent owner group ID" in {
```
rebstar6: need the 202 with ownerGroupId too

modules/core/src/test/scala/vinyldns/core/TestZoneData.scala
```diff
@@ -40,6 +40,7 @@ object TestZoneData {
   val zoneIp4: Zone = Zone("0.162.198.in-addr.arpa.", "test@test.com", adminGroupId = abcGroup.id)
   val zoneIp6: Zone =
     Zone("1.9.e.f.c.c.7.2.9.6.d.f.ip6.arpa.", "test@test.com", adminGroupId = abcGroup.id)
+  val zoneShared: Zone = okZone.copy(name = "shared-zone.", shared = true)
```
rebstar6: make it new, dont copy on these zones or the IDs wont be unique

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeErrors.scala
mitruly: I'll update to the suggestion. The intention was that we're potentially failing the batch on a bad owner group ID where it might not be needed (eg. if none of the record set changes are for shared zones).

modules/api/src/test/scala/vinyldns/api/domain/batch/BatchChangeConverterSpec.scala
mitruly: Whoops, thought that was always there.

modules/api/src/test/scala/vinyldns/api/route/BatchChangeRoutingSpec.scala
```diff
@@ -255,6 +263,30 @@ class BatchChangeRoutingSpec
...
+        val change = responseAs[JValue]
+
+        print(s"Response from test: $response\n\n")
+
+        print(s"Deserialized change from test:\n${compact(change)}\n\n")
+
+        val validResponseWithOwnerGroupId =
+          createBatchChangeResponse(ownerGroupId = Some("some-group-id"))
+
+        compact(change) shouldBe compact(Extraction.decompose(validResponseWithOwnerGroupId))
```
mitruly: Test is failing because `ownerGroupId` is not being returned as expected in the response from the `POST` request.

modules/api/functional_test/live_tests/batch/create_batch_change_test.py
```diff
@@ -2404,3 +2454,124 @@ def test_user_validation_shared(shared_zone_test_context):
...
+        ],
+        "ownerGroupId": "non-existent-owner-group-id"
+    }
+
+    errors = shared_client.create_batch_change(batch_change_input, status=400)
+    assert_that(errors, is_('Group with ID "non-existent-owner-group-id" was not found'))
+
+def test_create_batch_change_for_shared_zone_with_unauthorized_owner_group_id_fails(shared_zone_test_context):
+    """
+    Test successfully creating a batch change with owner group ID saves value for records in shared zone
```
rebstar6: comment here is off

modules/api/functional_test/live_tests/batch/create_batch_change_test.py
```diff
@@ -2404,3 +2454,124 @@ def test_user_validation_shared(shared_zone_test_context):
...
+        for (zoneId, recordSetId) in to_delete:
+            get_recordset = shared_client.get_recordset(zoneId, recordSetId, status=200)
+            assert_that(get_recordset['recordSet']['ownerGroupId'], is_(batch_change_input['ownerGroupId']))
+
+    finally:
+        clear_zoneid_rsid_tuple_list(to_delete, shared_client)
+
+def test_create_batch_change_for_shared_zone_with_invalid_owner_group_id_fails(shared_zone_test_context):
+    """
+    Test successfully creating a batch change with owner group ID saves value for records in shared zone
```
rebstar6: comment

modules/api/functional_test/live_tests/batch/create_batch_change_test.py
```diff
@@ -451,6 +451,56 @@ def test_create_batch_change_without_comments_succeeds(shared_zone_test_context)
...
+        to_delete = [(change['zoneId'], change['recordSetId']) for change in completed_batch['changes']]
+
+        assert_change_success_response_values(result['changes'], zone=ok_zone, index=0, record_name="owner-group-id",
+                                              input_name="owner-group-id.ok.", record_data="4.3.2.1")
+        assert_that(completed_batch['ownerGroupId'], is_(shared_zone_test_context.ok_group['id']))
+
+    finally:
+        clear_zoneid_rsid_tuple_list(to_delete, client)
+
+def test_create_batch_change_without_owner_group_id_succeeds(shared_zone_test_context):
```
rebstar6: test that batch change fails if the user is not in the owner group specified

rebstar6: oops found it :P

modules/api/functional_test/live_tests/batch/create_batch_change_test.py
```diff
@@ -2404,3 +2454,124 @@ def test_user_validation_shared(shared_zone_test_context):
...
+    on creates and updates (if existing value does not exist)
+    """
+    shared_client = shared_zone_test_context.shared_zone_vinyldns_client
+    shared_zone = shared_zone_test_context.shared_zone
+
+    update_rs = get_recordset_json(shared_zone, "update", "A", [{"address": "127.0.0.1"}], 300)
+
+    batch_change_input = {
+        "changes": [
+            get_change_A_AAAA_json("no-owner-group-id.shared.", address="4.3.2.1"),
```
rebstar6: name these `create-with-owner-group` and `update-with-owner-group`

modules/api/functional_test/live_tests/batch/create_batch_change_test.py
```diff
@@ -2404,3 +2454,124 @@ def test_user_validation_shared(shared_zone_test_context):
...
+        "ownerGroupId": ok_group['id']
+    }
+    to_delete = []
+
+    try:
+        create_result = ok_client.create_recordset(update_rs, status=202)
+        to_delete.append(ok_client.wait_until_recordset_change_status(create_result, 'Complete'))
+
+        result = ok_client.create_batch_change(batch_change_input, status=202)
+        completed_batch = ok_client.wait_until_batch_change_completed(result)
```
rebstar6: on each of these tests, add the validation that the ownerGroupId was saved on the completed_batch. even if not applied to any recordsets, that should be there

modules/api/functional_test/live_tests/batch/create_batch_change_test.py
rebstar6: actaully make `update-with-existing-owner-group` and `update-without-existing-owner-group` -- need to check that update-with-existing-owner-group (current update) has the owner group applied, and update-without-existing-owner-group has its old owner group

can rebrand this test generally to 
`test_create_batch_change_for_shared_zone_ownergroup_applied_logic`

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
rebstar6: got it. a little nitpicky, but i think i would just have a `getOwnerGroup(ownerGroupId: Option[String]): BatchResult[Option[Group]]` in the service, and a `validateBatchInput(batchChangeInput, existingGroup, auth)` where validateBatchInput does the full set of validations on the input item (batch size and owner group stuff)

rebstar6: ^^ follows what we have later where we do the db gets we need in the service, and pass all info needed to each validation in the validations file

modules/api/src/test/scala/vinyldns/api/domain/batch/BatchChangeServiceSpec.scala
```diff
@@ -167,6 +189,31 @@ class BatchChangeServiceSpec
...
+
+    "fail with UserDoesNotBelongToOwnerGroup if normal user does not belong to group specified by owner group ID" in {
+      val ownerGroupId = Some("user-is-not-member")
+      val input = BatchChangeInput(None, List(apexAddA), ownerGroupId)
+      val result = leftResultOf(underTest.applyBatchChange(input, notAuth).value)
+
+      result shouldBe a[NotAMemberOfOwnerGroup]
+    }
+
+    "succeed if owner group ID is provided and user is a super user" in {
```
rebstar6: also succeed if owner group ID is provided and user is in that group :P

modules/api/src/test/scala/vinyldns/api/route/BatchChangeJsonProtocolSpec.scala
```diff
@@ -222,6 +225,16 @@ class BatchChangeJsonProtocolSpec
         List(deleteAChangeInput, addAAAAChangeInput, addCNAMEChangeInput))
     }
 
+    "successfully serialize valid add change when owner group ID is provided" in {
```
rebstar6: do we have a round trip in here? thats the easiest way id think with these json tests

fromJson(toJson(someBatchChange)) should be someBatchChange

can do examples with and without comments/ownerGroupId

nimaeskandary:
@mitruly can you add a test case defining behavior when owner groups do not match up on recordsets? Say you are creating a record with owner group x, and updating another records attribute but it has owner group y, would the outcome be a new record with owner group x, and the existing record still has y?

mitruly:
@nimaeskandary I have a unit test (`not overwrite existing owner group ID for existing record set in shared zone` in `BatchChangeConverterSpec.scala`). Can add a func test, though.

nimaeskandary:
@mitruly more so mean a batch that includes a create and an update, that test doesn't exactly cover my last comment

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
```diff
@@ -50,6 +54,15 @@ class BatchChangeValidations(changeLimit: Int, accessValidation: AccessValidatio
   import RecordType._
   import accessValidation._
 
+  def validateBatchChangeInput(
+      input: BatchChangeInput,
+      existingGroup: Option[Group],
+      authPrincipal: AuthPrincipal): Either[BatchChangeErrorResponse, Unit] =
+    for {
+      _ <- validateBatchChangeInputSize(input)
```
rebstar6: make this validatedNel instead of either

rebstar6: `SingleValidation` type in batch

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
```diff
@@ -50,6 +54,15 @@ class BatchChangeValidations(changeLimit: Int, accessValidation: AccessValidatio
...
+      _ <- validateOwnerGroupId(input.ownerGroupId, existingGroup, authPrincipal)
+    } yield ().asRight
+
   def validateBatchChangeInputSize(
       input: BatchChangeInput): Either[BatchChangeErrorResponse, Unit] =
```
rebstar6: SingleValidation

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
```diff
@@ -60,6 +73,18 @@ class BatchChangeValidations(changeLimit: Int, accessValidation: AccessValidatio
       ().asRight
     }
 
+  def validateOwnerGroupId(
```
rebstar6: SingleValidation

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeErrors.scala
```diff
@@ -16,26 +16,21 @@
...
 package vinyldns.api.domain.batch
 
+import vinyldns.api.domain.DomainValidationError
 import vinyldns.api.domain.batch.BatchChangeInterfaces.ValidatedBatch
 import vinyldns.api.domain.batch.BatchTransformations.ChangeForValidation
 import vinyldns.core.domain.batch.SingleChange
 
 /* Error response options */
 sealed trait BatchChangeErrorResponse
+final case class InvalidBatchChangeInput(errors: List[DomainValidationError])
```
rebstar6: just make errors a NEL here (unless it causes later issues). then you dont have to convert to list

mitruly: I initially had errors as a `NonEmptyList`, but it requires creating a specific serializer so that the response doesn't have `{head: ..., tail: ...}`. Can go that route if we want, though.

PR #442, opened 2019-01-15
feature flag shared zone info

### LINKED ISSUE #349: Display shared zone status (portal)
In Manage Zone, allow users to see a `Zone Type` field indicating if the zone is shared or not (wording tbd - do some brainstorming).

This should be viewable from the main zones list, as well as within the zone itself.

britneywright:
Fixes #349.

Changes in this pull request:
- add `shared_display_enabled` flag, to be set in the portal config, default is false
- display zone type as "Shared" or "Private" in the main Zones table in the portal and also the Manage Zone tab of an individual zone

To test locally: 
- add `shared_display_enabled = true` in portal module local.conf

zone_type_in_zones_table

zone_type_in_manage_zone_tab

### DIFF
--- a/modules/portal/app/models/Meta.scala
+++ b/modules/portal/app/models/Meta.scala
@@ -17,8 +17,10 @@
 package models
 import play.api.Configuration
 
-case class Meta(version: String)
+case class Meta(version: String, shared_display_enabled: Boolean)
 object Meta {
   def apply(config: Configuration): Meta =
-    Meta(config.getOptional[String]("vinyldns.version").getOrElse("unknown"))
+    Meta(
+      config.getOptional[String]("vinyldns.version").getOrElse("unknown"),
+      config.getOptional[Boolean]("shared_display_enabled").getOrElse(false))
 }
--- a/modules/portal/app/models/Meta.scala
+++ b/modules/portal/app/models/Meta.scala
@@ -17,10 +17,10 @@
 package models
 import play.api.Configuration
 
-case class Meta(version: String, shared_display_enabled: Boolean)
+case class Meta(version: String, sharedDisplayEnabled: Boolean)
 object Meta {
   def apply(config: Configuration): Meta =
     Meta(
       config.getOptional[String]("vinyldns.version").getOrElse("unknown"),
-      config.getOptional[Boolean]("shared_display_enabled").getOrElse(false))
+      config.getOptional[Boolean]("shared-display-enabled").getOrElse(false))
 }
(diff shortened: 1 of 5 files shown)


nimaeskandary:
Should we consider adding tooltips for what shared and private mean in this pr? Could be something along the lines of "Shared zones allow other VinylDNS groups to manage RecordSets. These are limited to new RecordSets, or any RecordSets whose owner group attribute is set accordingly."

modules/portal/app/models/Meta.scala
```diff
@@ -17,8 +17,10 @@
...
 import play.api.Configuration
 
-case class Meta(version: String)
+case class Meta(version: String, shared_display_enabled: Boolean)
 object Meta {
   def apply(config: Configuration): Meta =
-    Meta(config.getOptional[String]("vinyldns.version").getOrElse("unknown"))
+    Meta(
+      config.getOptional[String]("vinyldns.version").getOrElse("unknown"),
+      config.getOptional[Boolean]("shared_display_enabled").getOrElse(false))
```
mitruly: I'm honestly not sure what `Meta` is intended for (assuming metadata information like version). I think it would make more sense to store this in `Settings.scala` where we retrieve other stuff (including `portalTestLogin`).

Could also temporarily add `shared_display_enabled = false` to `conf/reference.conf`.

mitruly:
Also agree with @nimaeskandary regarding at least a tooltip addition.

"Type" also kinda feels a bit broad for a boolean classification of shared/non-shared.

britneywright:
@nimaeskandary I'd rather wait until we've got more implemented in the portal to determine what tooltips/messaging we want to put where.

nimaeskandary:
Gotcha im fine holding off on the tooltip then

nimaeskandary:
I think logically this looks good, agree with michael that we may want to brainstorm a little more to see if there are better terms to represent it

modules/portal/app/models/Meta.scala
```diff
@@ -17,8 +17,10 @@
 package models
 import play.api.Configuration
 
-case class Meta(version: String)
+case class Meta(version: String, shared_display_enabled: Boolean)
```
pauljamescleary: do not use snake case in scala, snake_case is reserved for python.  The name should be `sharedDisplayEnabled` (camel case)

modules/portal/app/models/Meta.scala
pauljamescleary: use dashes for the name of the config value...`shared-display-enabled` - dashes are what are used for configuration values instead of underscores

Also, I agree with @mitruly, use the `Settings`

mitruly:
Is the general consensus that `Type` is a good label for classifying a zone as shared or private?

mitruly:
@nimaeskandary That's a fair point that @britneywright also mentioned. Functionally, this is good to me as well.

PR #446, opened 2019-01-16
Consolidate test data (VinylDNSTestData version)

### LINKED ISSUE #440: Cleanup/remove VinylDNSTestData
Card spun off of https://github.com/vinyldns/vinyldns/issues/196

We now have:
VinylDNSTestData (api)
TestZone/RecordSet/MembershipData (core)

We want:
- Base type examples in those core files
- Api tests can import those
- Any objects that are not actually shared between tests should just live in the test using them and removed from the shared files
- Api can have test files too for anything that's not in core, but there should be some logic for what objects are in what file

rebstar6:
Fixes #440 

Changes in this pull request:
- remove all the items in VinylDNSTestData
- switch users over to core files
- kept the few helpers in that file as VinylDNSTestHelpers. Should figure out what to do with them later


modules/api/src/test/scala/vinyldns/api/domain/record/RecordSetChangeSpec.scala
```diff
@@ -17,10 +17,14 @@
...
-import vinyldns.core.domain.record.AData
+import vinyldns.core.TestMembershipData.okAuth
+import vinyldns.core.TestRecordSetData._
+import vinyldns.core.TestZoneData.okZone
+import vinyldns.core.domain.record.{AData, RecordSet}
 
-class RecordSetChangeSpec extends WordSpec with Matchers with VinylDNSTestData {
+class RecordSetChangeSpec extends WordSpec with Matchers {
+
+  val atRs: RecordSet = rsOk.copy(name = "@")
```
mitruly: I think `apexRs` makes more sense for a name here.

modules/api/src/test/scala/vinyldns/api/domain/record/RecordSetValidationsSpec.scala
```diff
@@ -233,6 +234,7 @@ class RecordSetValidationsSpec
     }
 
     "NSValidations" should {
+      val invalidNsRecordToOrigin: RecordSet = ns.copy(name = "@")
```
mitruly: `invalidNsApexRs` feels more concise here.

modules/api/src/test/scala/vinyldns/api/domain/record/RecordSetValidationsSpec.scala
```diff
@@ -266,6 +268,7 @@ class RecordSetValidationsSpec
     }
 
     "CnameValidations" should {
+      val invalidCnameToOrigin: RecordSet = cname.copy(name = "@")
```
mitruly: `invalidCnameApexRs` feels more concise here.

PR #448, opened 2019-01-17
add ownerGroupName to getRecordSets response

### LINKED ISSUE #438: Incorporate owner group into getRecordSets endpoint, logic
Issue spun off from #350. getRecordSet endpoint and logic is handled.

To-do:
Return the owner group ID AND name for each record that has one
no access changes

Tricky part is figuring out at what step to get the owner group name. getRecordSets adds access rules during access validations and returns the RecordSetListInfo type to the RecordSetService.

britneywright:
Fixes #438.

Changes in this pull request:
- add ownerGroupName to getRecordsets response

### DIFF
--- a/modules/api/functional_test/live_tests/recordsets/list_recordsets_test.py
+++ b/modules/api/functional_test/live_tests/recordsets/list_recordsets_test.py
@@ -104,7 +104,7 @@ def test_list_recordsets_no_start(rs_fixture):
     list_results = client.list_recordsets(ok_zone['id'], status=200)
     rs_fixture.check_recordsets_page_accuracy(list_results, size=17, offset=0)
 
-def test_list_recordsets_with_ownergroup_id_and_ownergroup_name(rs_fixture):
+def test_list_recordsets_with_owner_group_id_and_owner_group_name(rs_fixture):
     """
     Test that record sets with an owner group return the owner group ID and name
     """
--- a/modules/api/functional_test/live_tests/recordsets/list_recordsets_test.py
+++ b/modules/api/functional_test/live_tests/recordsets/list_recordsets_test.py
@@ -11,6 +11,7 @@ class ListRecordSetsFixture():
     def __init__(self, shared_zone_test_context):
         self.test_context = shared_zone_test_context.ok_zone
         self.client = shared_zone_test_context.ok_vinyldns_client
+        self.shared_group = shared_zone_test_context.shared_record_group
         self.new_rs = {}
         existing_records = self.client.list_recordsets(self.test_context['id'])['recordSets']
         assert_that(existing_records, has_length(7))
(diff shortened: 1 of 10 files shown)


modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetService.scala
```diff
@@ -217,6 +220,17 @@ class RecordSetService(
     } yield recordSetChangesInfo
   }
 
+  def getListWithGroupNames(
+      recordsets: List[RecordSet],
+      groups: Set[Group]): Result[List[RecordSetInfo]] =
+    recordsets.map { rs =>
+      val ownerGroupName = groups.find(_.id == rs.ownerGroupId.getOrElse(None)) match {
```
mitruly: The following achieves the same effect:
```
  def getListWithGroupNames(
      recordsets: List[RecordSet],
      groups: Set[Group]): Result[List[RecordSetInfo]] =
    recordsets.map { rs =>
      val ownerGroupName =
        rs.ownerGroupId.flatMap(groupId => groups.find(_.id == groupId).map(_.name))
      RecordSetInfo(rs, ownerGroupName)
    }.toResult
```

modules/api/src/test/scala/vinyldns/api/domain/record/RecordSetServiceSpec.scala
```diff
@@ -506,7 +518,8 @@ class RecordSetServiceSpec
             authPrincipal = AuthPrincipal(okAuth.signedInUser.copy(isSupport = true), Seq.empty)
           )
           .value)
-      result.recordSets shouldBe List(RecordSetListInfo(aaaa, AccessLevel.Read))
+      result.recordSets shouldBe List(
```
mitruly: I think there should also be a unit test for owner group ID is returned without group name passes if the group is not found.

modules/api/functional_test/live_tests/recordsets/list_recordsets_test.py
```diff
@@ -103,6 +104,46 @@ def test_list_recordsets_no_start(rs_fixture):
     list_results = client.list_recordsets(ok_zone['id'], status=200)
     rs_fixture.check_recordsets_page_accuracy(list_results, size=17, offset=0)
 
+def test_list_recordsets_with_ownergroup_id_and_ownergroup_name(rs_fixture):
```
mitruly: Can an underscore be added between "owner" and "group" in the test name?

modules/api/functional_test/live_tests/recordsets/list_recordsets_test.py
```diff
@@ -103,6 +104,46 @@ def test_list_recordsets_no_start(rs_fixture):
...
+    Test that record sets with an owner group return the owner group ID and name
+    """
+    client = rs_fixture.client
+    ok_zone = rs_fixture.test_context
+    shared_group = rs_fixture.shared_group
+    created_rs_id = None
+    try:
+        # create a record in the zone with an owner group ID
+        new_rs = get_recordset_json(ok_zone,
+                                    "test_get_recordset", "TXT", [{'text':'should-work'}],
```
mitruly: Indent formatting is funky here.

britneywright: that's how the attributes indent naturally when it's multi-line

pauljamescleary: @mitruly this is ok as it aligns with the line above

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneProtocol.scala
```diff
@@ -112,23 +112,26 @@ case class RecordSetListInfo(
...
+      name = recordSet.name,
+      typ = recordSet.typ,
+      ttl = recordSet.ttl,
+      status = recordSet.status,
+      created = recordSet.created,
+      updated = recordSet.updated,
+      records = recordSet.records,
+      id = recordSet.id,
+      account = recordSet.account,
+      accessLevel = accessLevel,
```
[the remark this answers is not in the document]
pauljamescleary: we also have owner group name.  Using these "Info" things allows us to build essentally a "View" that is comprised of more than just our models.  This is actually a good pattern (for reference, look into the MVC pattern)

modules/api/functional_test/live_tests/recordsets/list_recordsets_test.py
```diff
@@ -103,6 +104,46 @@ def test_list_recordsets_no_start(rs_fixture):
...
+        assert_that(any(r['id'] == result_rs['id'] for r in list_results['recordSets']), is_(True))
+
+        # confirm the created recordset has an ownerGroupId and ownerGroupName
+        for rs in list_results['recordSets']:
+            if rs['id'] == result_rs['id']:
+                assert_that(rs['ownerGroupId'], is_(shared_group['id']))
+                assert_that(rs['ownerGroupName'], is_("record-ownergroup"))
+
+        # set created_rs_id after confirming record set creation so it can be deleted in test cleanup
+        created_rs_id = result_rs['id']
```
nimaeskandary: for the delete I think just using result_rs is better, if this tests fails anywhere before this line the delte wouldn't work because created_rs_id wouldn't be set

britneywright: updated

modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetService.scala
```diff
@@ -145,7 +145,10 @@ class RecordSetService(
       recordSetResults <- recordSetRepository
         .listRecordSets(zoneId, startFrom, maxItems, recordNameFilter)
         .toResult[ListRecordSetResults]
-      setsWithAccess <- getListAccessLevels(authPrincipal, recordSetResults.recordSets, zone).toResult
+      rsOwnerGroupIds = recordSetResults.recordSets.flatMap(_.ownerGroupId).toSet
+      rsGroups <- groupRepository.getGroups(rsOwnerGroupIds).toResult[Set[Group]]
+      setsWithGroupName <- getListWithGroupNames(recordSetResults.recordSets, rsGroups)
```
rebstar6: this is another `=` not `<=`. getListWithGroupNames can just return List[RecordSetInfo], theres no IO there

Issue #421, opened 2019-01-03
Support DS records


pauljamescleary:
**Is your feature request related to a problem? Please describe.**
DS records are required in order to support DNS Sec.  Currently, VinylDNS does not support DS as a record type.

**Describe the solution you'd like**
Support DS record types

**Additional context**
[DS Record Type RFC](https://tools.ietf.org/html/rfc4034)

Also, **dnsjava** already supports the DS record type (woot!)

**Questions**
There are multiple record types related to DNS SEC: DNSKEY, RRSIG, NSEC.  It is unclear if we need to support all of these, or what the order of priority would be across these record types.

**Update: NSEC, NSEC3, RRSIG, etc. records currently should not even be imported into VinylDNS.  Presently we _only_ require DS record support**

Also, as far as a "larger" set of DNSSEC features, those are outside of the scope of this issue but require further discussion.

----
Edit: 
- only need in zone/recordset path (dont worry about batch for now)
- this card is api only

make sure we can:
- sync these records in appropriately from the backend (sync)
- apply these records in the backend (create/update)
- delete them
- standard protobuf serialization things

PR #449, opened 2019-01-18
DS record support: DNS connection side

rebstar6:
Part of #421

### DIFF
--- a/modules/core/src/main/scala/vinyldns/core/domain/record/RecordData.scala
+++ b/modules/core/src/main/scala/vinyldns/core/domain/record/RecordData.scala
@@ -78,8 +78,8 @@ final case class SSHFPData(algorithm: Integer, typ: Integer, fingerprint: String
 final case class TXTData(text: String) extends RecordData
 
 final case class DSData(
-    keyTag: Integer, // footprint - The original KEY record's footprint (keyid).
+    keyTag: Integer, // footprint in DNSJava
     algorithm: Integer,
-    digestType: Integer, //digestid - The digest id code; algorithm used to construct the digest
-    digest: ByteVector) //digest - A hash of the original key
+    digestType: Integer, //digestid in DNSJava
+    digest: ByteVector)
     extends RecordData
--- a/modules/core/src/main/scala/vinyldns/core/domain/record/RecordData.scala
+++ b/modules/core/src/main/scala/vinyldns/core/domain/record/RecordData.scala
@@ -77,9 +77,47 @@ final case class SSHFPData(algorithm: Integer, typ: Integer, fingerprint: String
 
 final case class TXTData(text: String) extends RecordData
 
+sealed abstract class DigestType(val value: Int)
+object DigestType {
+  // see https://www.iana.org/assignments/ds-rr-types/ds-rr-types.xhtml
+  case object SHA1 extends DigestType(1)
+  case object SHA256 extends DigestType(2)
+  case object GOSTR341194 extends DigestType(3)
+  case object SHA384 extends DigestType(4)
+  final case class UnknownDigestType(x: Int) extends DigestType(x)
+
+  def fromInt(value: Int): DigestType =
+    value match {
+      case 1 => SHA1
+      case 2 => SHA256
+      case 3 => GOSTR341194
+      case 4 => SHA384
+      case other => UnknownDigestType(other)
+    }
+}
+
+sealed abstract class DnsSecAlgorithm(val value: Int)
+object DnsSecAlgorithm {
+  // see https://www.iana.org/assignments/dns-sec-alg-numbers/dns-sec-alg-numbers.xhtml
+  case object DSA extends DnsSecAlgorithm(3)
+  case object RSASHA1 extends DnsSecAlgorithm(5)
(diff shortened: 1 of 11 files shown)


modules/core/src/main/scala/vinyldns/core/domain/record/RecordData.scala
```diff
@@ -75,3 +76,10 @@ object SRVData {
 final case class SSHFPData(algorithm: Integer, typ: Integer, fingerprint: String) extends RecordData
 
 final case class TXTData(text: String) extends RecordData
+
+final case class DSData(
+    keyTag: Integer, // footprint in DNSJava
+    algorithm: Integer,
```
mitruly: Would be preferable to have enumerations for `algorithm` and `digestType` (as there are only a few valid values for both), but we don't do it for `SSHFPData`, either.

modules/core/src/test/scala/vinyldns/core/TestRecordSetData.scala
```diff
@@ -137,6 +138,27 @@ object TestRecordSetData {
...
 
+  // example at https://tools.ietf.org/html/rfc4034#page-18
+  val dSDataSha1 =
+    DSData(60485, 5, 1, ByteVector.fromValidHex("2BB183AF5F22588179A53B0A98631FAD1A292118"))
+
+  // example at https://tools.ietf.org/html/rfc4509#page-3
+  val dSDataSha256 =
+    DSData(
+      60485,
+      5,
```
pauljamescleary: when you see this "in action", I definitely like using an `enum`

modules/core/src/main/scala/vinyldns/core/protobuf/ProtobufConversions.scala
```diff
@@ -264,7 +264,7 @@ trait ProtobufConversions {
 
   /* This cannot be called toPB because RecordData is the base type for things like AData, cannot overload */
   def toRecordData(data: RecordData): VinylDNSProto.RecordData = {
-    val d = data match {
+    val d = (data: @unchecked) match {
```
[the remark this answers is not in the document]
rebstar6: oh yeah thanks, meant to make that note myself :P

modules/core/src/test/scala/vinyldns/core/TestRecordSetData.scala
rebstar6: for storage or for input? im assuming anyone working with these will know what they want for these values, so im inclined to leave the input at numbers.
we can definitely pass around as an enum though, or at a minimum make sure the up front validations ensure the input number is valid (seems to be 1 or 2 for digest, theres a few more for algorithm, i have to do some spec jumping around.


the only argument for int (i think) is that if we enum id want to be 100% certain we have all the options, included not recommended ones. i dont want to be unable to sync a zone in because of not recommended algorithm. Will double check on the valid values

modules/core/src/main/scala/vinyldns/core/domain/record/RecordData.scala
```diff
@@ -75,3 +76,48 @@ object SRVData {
...
 final case class TXTData(text: String) extends RecordData
+
+sealed abstract class DigestType(val value: Int)
+object DigestType {
+  // see https://www.iana.org/assignments/ds-rr-types/ds-rr-types.xhtml
+  case object SHA1 extends DigestType(1)
+  case object SHA256 extends DigestType(2)
+  case object GOSTR341194 extends DigestType(3)
+  case object SHA384 extends DigestType(4)
+  final case class UnknownDigestType(x: Int) extends DigestType(x)
```
rebstar6: this unknown is for the `fromDSRecord` function - that returns just a RecordSet (not an Either or Validated), so if I dont have a fallback, it could throw.

Proposing this as a solution where input validations would disallow the "Unknown" type. Open to suggestions though

modules/core/src/main/scala/vinyldns/core/domain/record/RecordData.scala
```diff
@@ -75,3 +76,48 @@ object SRVData {
...
+      case 2 => SHA256
+      case 3 => GOSTR341194
+      case 4 => SHA384
+      case other => UnknownDigestType(other)
+    }
+}
+
+sealed abstract class DnsSecAlgorithm(val value: Int)
+object DnsSecAlgorithm {
+  // see https://www.iana.org/assignments/dns-sec-alg-numbers/dns-sec-alg-numbers.xhtml
```
rebstar6: need to get validation on these options here, wasnt 100% on which ones really should be pulled in

modules/core/src/main/scala/vinyldns/core/domain/record/RecordData.scala
mitruly: I like this approach.

modules/core/src/main/scala/vinyldns/core/domain/record/RecordData.scala
mitruly: Not entirely sure, either. I'd be comfortable with supporting the ones that have a mnemonic and are indicated as a proposed standard, though we could be safe and implement all of the ones with a mnemonic.

Arguably we can try using different values and seeing which ones are allowed by the DNS backend.

PR #452, opened 2019-01-24
DS protobuf conversions

rebstar6:
another part of #421


modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlRecordSetRepositoryIntegrationSpec.scala
```diff
@@ -71,6 +71,22 @@ class MySqlRecordSetRepositoryIntegrationSpec
   }
 
   "inserting record sets" should {
+    "properly add and delete DS records" in {
+      val addChange = makeTestAddChange(ds, okZone)
+      val testRecord = addChange.recordSet
+      val deleteChange = makeTestDeleteChange(testRecord, okZone)
+
+      val dbCalls = for {
```
mitruly: Snazzy ✨

PR #456, opened 2019-01-25
DS routing/serialization

rebstar6:
should be 2nd to last PR; get that serialization!

### DIFF
--- a/modules/api/functional_test/live_tests/shared_zone_test_context.py
+++ b/modules/api/functional_test/live_tests/shared_zone_test_context.py
@@ -282,6 +282,7 @@ class SharedZoneTestContext(object):
             self.ok_vinyldns_client.wait_until_zone_exists(classless_zone_delegation_change)
             self.ok_vinyldns_client.wait_until_zone_exists(system_test_zone_change)
             self.ok_vinyldns_client.wait_until_zone_exists(parent_zone_change)
+            self.ok_vinyldns_client.wait_until_zone_exists(ds_zone_change)
             self.shared_zone_vinyldns_client.wait_until_zone_change_status_synced(shared_zone_change)
             shared_sync_change = self.shared_zone_vinyldns_client.sync_zone(self.shared_zone['id'])
             self.shared_zone_vinyldns_client.wait_until_zone_change_status_synced(non_test_shared_zone_change)
--- a/modules/api/functional_test/live_tests/shared_zone_test_context.py
+++ b/modules/api/functional_test/live_tests/shared_zone_test_context.py
@@ -243,6 +243,29 @@ class SharedZoneTestContext(object):
                 }, status=202)
             self.parent_zone = parent_zone_change['zone']
 
+            # mimicking the spec example
+            ds_zone_change = self.ok_vinyldns_client.create_zone(
+                {
+                    'name': 'example.com.',
+                    'email': 'test@test.com',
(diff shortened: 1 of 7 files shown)


modules/api/functional_test/live_tests/shared_zone_test_context.py
```diff
@@ -243,6 +243,29 @@ def __init__(self):
...
+                    },
+                    'transferConnection': {
+                        'name': 'example.',
+                        'keyName': VinylDNSTestContext.dns_key_name,
+                        'key': VinylDNSTestContext.dns_key,
+                        'primaryServer': VinylDNSTestContext.dns_ip
+                    }
+                }, status=202)
+            self.ds_zone = ds_zone_change['zone']
+
```
nimaeskandary: should we have the extra validation on line 275? `wait until our zones are created`

modules/api/src/test/scala/vinyldns/api/route/VinylDNSJsonProtocolSpec.scala
```diff
@@ -269,15 +270,15 @@ class VinylDNSJsonProtocolSpec
     "parse a record set with an absolute CNAME record passes" in {
       val recordSetJValue: JValue =
         ("zoneId" -> "1") ~~
-          ("name" -> "Comcast") ~~
```
nimaeskandary: 🦚

modules/core/src/main/scala/vinyldns/core/domain/record/RecordData.scala
```diff
@@ -102,15 +102,33 @@ object DnsSecAlgorithm {
...
+  case object PRIVATEOID extends DnsSecAlgorithm(254)
   final case class UnknownAlgorithm private (x: Int) extends DnsSecAlgorithm(x)
 
   def apply(value: Int): DnsSecAlgorithm =
     value match {
       case 3 => DSA
       case 5 => RSASHA1
       case 6 => DSA_NSEC3_SHA1
-      case 7 => RSA_NSEC3_SHA1
+      case 7 => RSASHA1_NSEC3_SHA1
```
nimaeskandary: Checked against https://www.iana.org/assignments/dns-sec-alg-numbers/dns-sec-alg-numbers.xhtml and all look right

modules/api/functional_test/live_tests/shared_zone_test_context.py
rebstar6: added

modules/api/src/main/scala/vinyldns/api/route/DnsJsonProtocol.scala
```diff
@@ -435,6 +438,38 @@ trait DnsJsonProtocol extends JsonValidation {
         ("fingerprint" -> rr.fingerprint)
   }
 
+  case object DSSerializer extends ValidationSerializer[DSData] {
+    override def fromJson(js: JValue): ValidatedNel[String, DSData] =
+      (
+        (js \ "keytag")
```
mitruly: Should probably use camel-casing here for consistency (`keyTag` and `digestType`).

mitruly: Actually nvm, this is for the actual DNS record data. Lowercase all the way~

PR #457, opened 2019-01-25
Karma config updates

### LINKED ISSUE #460: Portal frontend library updates
**Motivation**
We want to add more tests for the portal JavaScript. We're currently using PhantomJS in our testing and that uses Internet Explorer which doesn't support some of the ECMAScript 2015 functions we use that are supported by all other browsers.

**Scope of change**
- Change from PhantomJS to ChromeHeadless
- upgrade/downgrade related libraries as needed

britneywright:
fixes #460 

Was working on some of the front end unit tests and realized PhantomJS doesn't support some functions we use like `find` that are part of ECMAScript2015 because it uses the IE browser https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find#Browser_compatibility. A long rabbit hole later:
 - I'm using ChromeHeadless instead of PhantomJS 
- had to rearrange the javascript files we pull in 
- had to downgrade the jquery version we use because the Gentellela template we use doesn't support the version we've been using. That weirdly only causes failures in tests 🤷🏽‍♀️ 

Testing steps:

1.  run `npm install`
2. run `grunt unit` from the `modules/portal` directory to run the unit tests

### DIFF
--- a/modules/portal/public/lib/controllers/controller.records.spec.js
+++ b/modules/portal/public/lib/controllers/controller.records.spec.js
@@ -250,7 +250,7 @@ describe('Controller: RecordsController', function () {
     });
 
     it('isGroupMember returns true if user is a member of a given group', function() {
-        mockGroups = {data: { groups: [
+        this.scope.myGroups = [
             {id: "c8234503-bfda-4b80-897f-d74129051eaa",
                 name: "test",
                 email: "test@test.com",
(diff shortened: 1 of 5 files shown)


modules/portal/public/lib/controllers/controller.records.spec.js
```diff
@@ -249,4 +249,37 @@ describe('Controller: RecordsController', function () {
...
+                name: "test",
+                email: "test@test.com",
+                admins: [{id: "7096b806-c12a-4171-ba13-7fabb523acee"}],
+                created: "2017-02-15T14:58:31Z",
+                members: [{id: "7096b806-c12a-4171-ba13-7fabb523acee"}],
+                status: "Active"}
+                ],
+            maxItems: 100}};
+
+        this.scope.myGroups = mockGroups.data.groups;
```
mitruly: We're not stubbing with the mock object (`mockGroups`) directly here or simulating a refresh; just testing the `isGroupMember` function directly on `myGroups` so I think it makes more sense to just set `this.scope.myGroups` directly:
```
        this.scope.myGroups = [
            {id: "c8234503-bfda-4b80-897f-d74129051eaa",
                name: "test",
                email: "test@test.com",
                admins: [{id: "7096b806-c12a-4171-ba13-7fabb523acee"}],
                created: "2017-02-15T14:58:31Z",
                members: [{id: "7096b806-c12a-4171-ba13-7fabb523acee"}],
                status: "Active"
            }
        ];
```

Same with the next test.

mitruly:
Note to reviewers: `npm install` will be needed to download/update the dependencies.

britneywright:
also phantomJS is no longer an active project https://github.com/ariya/phantomjs/issues/15344

PR #458, opened 2019-01-25
portal: add record owner group name to batch form

### LINKED ISSUE #343: Select owner group on submit batch (portal)
`batchchanges/new` page:
On this screen, the user should have a field for the owner group name. This field should allow a user select any group they are in using the existing groups dropdown logic (see #415 for update) . The user can leave this field blank.

It should be clear on the screen to the user that this field is only required for changes in shared zones, and that it will apply that group ownership only to new or unowned records.

On submit, the groupId associated with the name the user has selected should be submitted with the batch request. If the group ID cannot be determined, the portal should display an error on the field.

britneywright:
Fixes #343 .

Changes in this pull request:
- add drop down menu in the batch change form to select the record owner group that will be applied to new and unowned records in shared zones 
- this is in the feature flag so `shared-display-enabled = true` should be added to local.conf or reference.conf to view
- we aren't adding an messaging around the field at this time, we'll do all the portal messaging in a separate issue/pr.
- batch change validations aren't merged in yet so the batch change will be accepted whether the owner group is set or not

### DIFF
--- a/modules/portal/app/views/batchChanges/batchChangeNew.scala.html
+++ b/modules/portal/app/views/batchChanges/batchChangeNew.scala.html
@@ -32,9 +32,22 @@
                     <div class="panel panel-default">
                         <div class="panel-heading">
                             <div class="form-group">
-                                <label class="h4">Description: </label>
+                                <label class="h5">Description: </label>
                                 <input type="text" class="form-control" ng-model="newBatch.comments">
                             </div>
+                            @if(meta.sharedDisplayEnabled) {
+                            <div class="form-group row">
+                                <div class="col-md-6">
+                                    <label class="h5" for="recordOwnerGroup">Record Owner Group</label>
+                                    <select class="form-control" id="recordOwnerGroup" ng-model="newBatch.ownerGroupId">
+                                        <option></option>
+                                        <option ng-repeat="group in myGroups | orderBy: 'name'" value="{{ group.id }}"
+                                                ng-selected="updateZoneInfo.adminGroupName == group.name">
+                                            {{group.name}}</option>
+                                    </select>
+                                </div>
+                            </div>
+                            }
                         </div>
                         <div class="panel-body">
                             <h4>Changes</h4>
@@ -156,20 +169,6 @@
                             </table>
                             <button type="button" id="addChange" class="btn btn-default" ng-click="addSingleChange()" ng-disabled="newBatch.changes.length == batchChangeLimit"><span class="fa fa-plus"></span> Add a Change</button>
                             <span ng-if="newBatch.changes.length == batchChangeLimit">Limit reached. Cannot add more than {{batchChangeLimit}} records per batch change.</span>
-                            @if(meta.sharedDisplayEnabled) {
-                                <div class="form-group">
-                                    &nbsp;
-                                </div>
-                                <div class="row col-md-6">
-                                    <label for="recordOwnerGroup">Record Owner Group</label>
-                                    <select class="form-control" id="recordOwnerGroup" ng-model="newBatch.ownerGroupId">
-                                        <option></option>
-                                        <option ng-repeat="group in myGroups | orderBy: 'name'" value="{{ group.id }}"
-                                                ng-selected="updateZoneInfo.adminGroupName == group.name">
-                                            {{group.name}}</option>
-                                    </select>
-                                </div>
-                            }
                         </div>
                         <div class="panel-footer clearfix">
                             <div ng-if="formStatus=='pendingSubmit'" class="pull-right">
--- a/modules/portal/app/views/batchChanges/batchChangeNew.scala.html
+++ b/modules/portal/app/views/batchChanges/batchChangeNew.scala.html
@@ -1,4 +1,4 @@
-@(rootAccountName: String)(implicit request: play.api.mvc.Request[Any], customLinks: models.CustomLinks)
+@(rootAccountName: String)(implicit request: play.api.mvc.Request[Any], customLinks: models.CustomLinks, meta: models.Meta)
 
 @content = {
 <!-- PAGE CONTENT -->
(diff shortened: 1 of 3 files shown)


modules/portal/public/lib/batch-change/batch-change-new.controller.js
```diff
@@ -18,7 +18,14 @@
     'use strict';
 
     angular.module('batch-change')
-        .controller('BatchChangeNewController', function($scope, jsConfig, $log, $location, $timeout, batchChangeService, utilityService){
+        .controller('BatchChangeNewController', function($scope, jsConfig, $log, $location, $timeout, batchChangeService, utilityService, groupsService){
+
```
mitruly: New line here isn't needed.

modules/portal/public/lib/batch-change/batch-change-new.controller.js
```diff
@@ -18,7 +18,14 @@
...
+            $scope.load = function(){
+                groupsService.getMyGroupsStored().then(
```
mitruly: Should also add:
```
.catch(function (error) {
    handleError(error, 'groupsService::getMyGroupsStored-failure');
});
```

modules/portal/app/views/batchChanges/batchChangeNew.scala.html
```diff
@@ -156,6 +156,20 @@ <h4>Changes</h4>
                             </table>
                             <button type="button" id="addChange" class="btn btn-default" ng-click="addSingleChange()" ng-disabled="newBatch.changes.length == batchChangeLimit"><span class="fa fa-plus"></span> Add a Change</button>
                             <span ng-if="newBatch.changes.length == batchChangeLimit">Limit reached. Cannot add more than {{batchChangeLimit}} records per batch change.</span>
+                            @if(meta.sharedDisplayEnabled) {
+                                <div class="form-group">
+                                    &nbsp;
```
mitruly: Should we be using `<br />` instead of `&nbsp;` here for consistency?

britneywright:
Here's a new screenshot with record owner group moved to the top of the form, just below Description. I also made the headings for both a bit smaller, from h4 to h5.

screen shot 2019-01-28 at 5 05 34 pm

modules/portal/public/lib/batch-change/batch-change-new.controller.js
```diff
@@ -18,7 +18,16 @@
     'use strict';
 
     angular.module('batch-change')
-        .controller('BatchChangeNewController', function($scope, jsConfig, $log, $location, $timeout, batchChangeService, utilityService){
+        .controller('BatchChangeNewController', function($scope, jsConfig, $log, $location, $timeout, batchChangeService, utilityService, groupsService){
+            $scope.load = function(){
+                groupsService.getMyGroupsStored()
+                    .then(function (results) {
+                        $scope.myGroups = results['groups'];
```
rebstar6: hmm this is interesting. really we only need this load to happen in the case that we have shared configured on. does this slow the load down?

ideally if getMyGroupsStored works as planned it shouldnt, but in practice im seeing a lot of groups loading time across pages. can you inspect (on firefox :P) and see if you can figure out why? time box it to an hour or two, im fine with the extra load if we cant get something better in that timeframe.

modules/portal/public/lib/batch-change/batch-change-new.controller.js
```diff
@@ -18,7 +18,14 @@
...
 
     angular.module('batch-change')
-        .controller('BatchChangeNewController', function($scope, jsConfig, $log, $location, $timeout, batchChangeService, utilityService){
+        .controller('BatchChangeNewController', function($scope, jsConfig, $log, $location, $timeout, batchChangeService, utilityService, groupsService){
+            groupsService.getMyGroups()
+                .then(function (results) {
+                    $scope.myGroups = results['groups'];
+                })
+                .catch(function (error) {
+                    handleError(error, 'groupsService::getMyGroupsStored-failure');
```
mitruly: The error message should be updated to `groupsService::getMyGroups-failure` if we are switching this.

modules/portal/public/lib/batch-change/batch-change.spec.js
```diff
@@ -29,7 +30,7 @@ describe('BatchChange', function(){
             module('ngMock')
         });
 
-        beforeEach(inject(function ($rootScope, $controller, $q, batchChangeService, pagingService, utilityService) {
+        beforeEach(inject(function ($rootScope, $controller, $q, batchChangeService, pagingService, utilityService, groupsService) {
```
mitruly: There's an initial failure due to `getMyGroups` having an invalid `maxItems` value:
```
        this.getMyGroups = function () {
            var url = '/api/groups';
            url = this.urlBuilder(url, { maxItems: 1000});
            return $http.get(url);
        };
```

This should either be changed to `{ maxItems: 100}` or delegate to the `maxItems` default in the API (which is also 100).

With this change, there is still an issue due to an unexpected request, indicating that we should probably be spying and expecting the request to `groupsService` here:
`Unexpected request: GET /api/groups?maxItems=100`

modules/portal/app/views/batchChanges/batchChangeNew.scala.html
```diff
@@ -32,9 +32,22 @@
...
-                                <label class="h4">Description: </label>
+                                <label class="h5">Description: </label>
                                 <input type="text" class="form-control" ng-model="newBatch.comments">
                             </div>
+                            @if(meta.sharedDisplayEnabled) {
+                            <div class="form-group row">
+                                <div class="col-md-6">
+                                    <label class="h5" for="recordOwnerGroup">Record Owner Group</label>
+                                    <select class="form-control" id="recordOwnerGroup" ng-model="newBatch.ownerGroupId">
+                                        <option></option>
```
nimaeskandary: Could this be `<option value=''>No Group</option>`

britneywright: I prefer it empty but I'll leave it to @rebstar6 to decide

modules/portal/public/lib/batch-change/batch-change-new.controller.js
```diff
@@ -86,5 +93,7 @@
                 var alert = utilityService.failure(error, type);
                 $scope.alerts.push(alert);
             }
+
+            $timeout($scope.load, 0);
```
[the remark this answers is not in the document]
britneywright: oops thought i removed this.

modules/portal/public/lib/batch-change/batch-change-new.controller.js
```diff
@@ -18,7 +18,14 @@
     'use strict';
 
     angular.module('batch-change')
-        .controller('BatchChangeNewController', function($scope, jsConfig, $log, $location, $timeout, batchChangeService, utilityService){
+        .controller('BatchChangeNewController', function($scope, jsConfig, $log, $location, $timeout, batchChangeService, utilityService, groupsService){
+            groupsService.getMyGroups()
```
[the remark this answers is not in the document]
britneywright: this is what i mentioned in another comment https://github.com/vinyldns/vinyldns/pull/458#discussion_r252023690 getMyGroupsStored doesn't provide any value since we aren't doing an reloading in the new batch change page.

modules/portal/app/views/batchChanges/batchChangeNew.scala.html
nimaeskandary: @britneywright you can ignore my X

PR #459, opened 2019-01-28
return owner group name in get and list batch

### LINKED ISSUE #352: Incorporate owner group name into batch endpoints
- Return the owner group ID AND name in the getBatchChange and listBatchChangeSummaries endpoints
~~- In the create batch endpoint, allow users to input an ownerGroupId on the batch change (optional)~~

~~Users can enter only IDs for groups they are in ownerGroupId for create batch; super users can enter any group~~

> Note: `In the create batch endpoint, allow users to input an ownerGroupId on the batch change (optional)` is DONE

nimaeskandary:
Fixes #352.

Changes in this pull request:
- return owner group name in get batch endpoint
- return owner group name in list batch summaries

### DIFF
--- a/modules/api/functional_test/live_tests/batch/get_batch_change_test.py
+++ b/modules/api/functional_test/live_tests/batch/get_batch_change_test.py
@@ -32,6 +32,92 @@ def test_get_batch_change_success(shared_zone_test_context):
                 pass
 
 
+def test_get_batch_change_with_record_owner_group_success(shared_zone_test_context):
+    """
+    Test successfully getting a batch change with an ownerGroupId and ownerGroupName
+    """
+    client = shared_zone_test_context.shared_zone_vinyldns_client
+    group = shared_zone_test_context.shared_record_group
+    batch_change_input = {
+        "comments": "this is optional",
+        "changes": [
+            get_change_A_AAAA_json("testing-get-batch-with-owner-group.shared.", address="1.1.1.1")
+        ],
+        "ownerGroupId": group['id']
+    }
+    to_delete = []
+    try:
+        batch_change = client.create_batch_change(batch_change_input, status=202)
+        completed_batch = client.wait_until_batch_change_completed(batch_change)
+
+        record_set_list = [(change['zoneId'], change['recordSetId']) for change in completed_batch['changes']]
+        to_delete = set(record_set_list)
+
+        result = client.get_batch_change(batch_change['id'], status=200)
+        assert_that(result, is_(completed_batch))
+        assert_that(result['ownerGroupId'], is_(group['id']))
+        assert_that(result['ownerGroupName'], is_(group['name']))
+
+    finally:
+        for result_rs in to_delete:
+            try:
+                delete_result = client.delete_recordset(result_rs[0], result_rs[1], status=202)
+                client.wait_until_recordset_change_status(delete_result, 'Complete')
+            except:
+                pass
+
+
+def test_get_batch_change_with_deleted_record_owner_group_success(shared_zone_test_context):
+    """
+    Test that if the owner group no longer exists that getting a batch change will still succeed,
+    with the ownerGroupName attribute set to None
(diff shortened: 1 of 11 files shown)


modules/api/functional_test/live_tests/batch/get_batch_change_test.py
```diff
@@ -32,6 +32,92 @@ def test_get_batch_change_success(shared_zone_test_context):
...
+    finally:
+        for result_rs in to_delete:
+            try:
+                delete_result = client.delete_recordset(result_rs[0], result_rs[1], status=202)
+                client.wait_until_recordset_change_status(delete_result, 'Complete')
+            except:
+                pass
+
+
+def test_get_batch_change_with_deleted_record_owner_group_success(shared_zone_test_context):
```
mitruly: :fire:

modules/api/functional_test/live_tests/batch/list_batch_change_summaries_test.py
```diff
@@ -153,3 +154,94 @@ def test_list_batch_change_summaries_with_list_batch_change_summaries_with_no_ch
...
+    try:
+        batch_change = client.create_batch_change(batch_change_input, status=202)
+        completed_batch = client.wait_until_batch_change_completed(batch_change)
+
+        record_set_list = [(change['zoneId'], change['recordSetId']) for change in completed_batch['changes']]
+        record_to_delete = set(record_set_list)
+
+        batch_change_summaries_result = client.list_batch_change_summaries(status=200)["batchChanges"]
+
+        under_test = [item for item in batch_change_summaries_result if 'comments' in item and item['comments'] == comments]
```
rebstar6: each batch change has an ID...filter by that, we shouldnt need to do this comments thing (pull from `batch_change`)

modules/api/src/main/scala/vinyldns/api/route/BatchChangeJsonProtocol.scala
```diff
@@ -153,6 +148,19 @@ trait BatchChangeJsonProtocol extends JsonValidation {
         ("ownerGroupId" -> bc.ownerGroupId)
   }
 
+  case object BatchChangeInfoSerializer extends ValidationSerializer[BatchChangeInfo] {
```
rebstar6: do we still need BatchChangeSerializer? may be able to delete now that this is here

modules/api/functional_test/live_tests/batch/list_batch_change_summaries_test.py
nimaeskandary: ooph didn't think of that

modules/api/functional_test/live_tests/batch/list_batch_change_summaries_test.py
```diff
@@ -153,3 +154,100 @@ def test_list_batch_change_summaries_with_list_batch_change_summaries_with_no_ch
...
     assert_that(batch_change_summaries_result, has_length(0))
+
+
+def test_list_batch_change_summaries_with_record_owner_group_passes(shared_zone_test_context):
+    """
+    Test that getting a batch change summary with an record owner group set returns the record owner group name and id
+    """
+    client = shared_zone_test_context.shared_zone_vinyldns_client
+    group = shared_zone_test_context.shared_record_group
+    comments = str(uuid.uuid4())
```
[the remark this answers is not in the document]
nimaeskandary: really any zone would have worked except the list batch change summaries zone because that one is all about known page sizes, picked the shared one cuz the test is about record owner groups

modules/api/src/main/scala/vinyldns/api/route/BatchChangeJsonProtocol.scala
nimaeskandary: yeah needed for now since apply batch change returns a batch change object not a batch change info object

PR #454, opened 2019-01-24
Incorporate owner group validations into delete record set endpoint 

mitruly:
Pre-req for #342.

Changes in this pull request:
- Update `deleteRecordSet` validation to pass in owner group ID
- Update tests

PR #461, opened 2019-01-28
Batch change owner group validations

### LINKED ISSUE #342: Incorporate shared zone logic into batch

mitruly:
Fixes #342.
Add context validations for batch change. Branched off of #454 (from commit `c789fac`).

Changes in this pull request:
- Implement batch change validations
- Update access validations to accept requests when `ownerGroupId` is not defined in existing record set
- Added functional and unit tests

### DIFF
--- a/modules/api/functional_test/live_tests/batch/create_batch_change_test.py
+++ b/modules/api/functional_test/live_tests/batch/create_batch_change_test.py
@@ -2777,9 +2777,9 @@ def test_create_batch_change_validation_without_owner_group_id(shared_zone_test_
             delete_result = shared_client.delete_recordset(shared_zone['id'], rsId, status=202)
             shared_client.wait_until_recordset_change_status(delete_result, 'Complete')
 
-def test_create_batch_delete_record_for_normal_user_in_owner_group_succeeds(shared_zone_test_context):
+def test_create_batch_delete_record_for_unassociated_user_in_owner_group_succeeds(shared_zone_test_context):
     """
-    Test delete change in batch for a record in a shared zone for a normal user belonging to the record owner group succeeds
+    Test delete change in batch for a record in a shared zone for an unassociated user belonging to the record owner group succeeds
     """
     shared_client = shared_zone_test_context.shared_zone_vinyldns_client
     ok_client = shared_zone_test_context.ok_vinyldns_client
@@ -2803,9 +2803,9 @@ def test_create_batch_delete_record_for_normal_user_in_owner_group_succeeds(shar
     assert_change_success_response_values(completed_batch['changes'], zone=shared_zone, index=0, record_name="shared-delete",
                                           input_name="shared-delete.shared.", record_data=None, change_type="DeleteRecordSet")
 
-def test_create_batch_delete_record_for_normal_user_not_in_owner_group_fails(shared_zone_test_context):
+def test_create_batch_delete_record_for_unassociated_user_not_in_owner_group_fails(shared_zone_test_context):
     """
-    Test delete change in batch for a record in a shared zone for a normal user not belonging to the record owner group fails
+    Test delete change in batch for a record in a shared zone for an unassociated user not belonging to the record owner group fails
     """
     shared_client = shared_zone_test_context.shared_zone_vinyldns_client
     dummy_client = shared_zone_test_context.dummy_vinyldns_client
--- a/modules/api/functional_test/live_tests/batch/create_batch_change_test.py
+++ b/modules/api/functional_test/live_tests/batch/create_batch_change_test.py
@@ -2410,7 +2410,8 @@ def test_user_validation_ownership(shared_zone_test_context):
             get_change_A_AAAA_json("update-test-batch.shared.", change_type="DeleteRecordSet"),
             get_change_A_AAAA_json("update-test-batch.shared.", address="1.1.1.1"),
             get_change_A_AAAA_json("delete-test-batch.shared.", change_type="DeleteRecordSet"),
-        ]
+        ],
+        "ownerGroupId": "shared-zone-group"
     }
 
     response = client.create_batch_change(batch_change_input, status=400)
@@ -2441,7 +2442,8 @@ def test_user_validation_shared(shared_zone_test_context):
             get_change_A_AAAA_json("update-test-batch.non.test.shared.", change_type="DeleteRecordSet"),
             get_change_A_AAAA_json("update-test-batch.non.test.shared.", address="1.1.1.1"),
             get_change_A_AAAA_json("delete-test-batch.non.test.shared.", change_type="DeleteRecordSet")
-        ]
+        ],
+        "ownerGroupId": shared_zone_test_context.ok_group['id']
     }
 
     response = client.create_batch_change(batch_change_input, status=400)
--- a/modules/api/functional_test/live_tests/batch/create_batch_change_test.py
+++ b/modules/api/functional_test/live_tests/batch/create_batch_change_test.py
@@ -2602,3 +2602,260 @@ def test_create_batch_change_for_shared_zone_with_unauthorized_owner_group_id_fa
 
     errors = shared_client.create_batch_change(batch_change_input, status=400)['errors']
     assert_that(errors, contains('User "sharedZoneUser" must be a member of group "' + ok_group['id'] + '" to apply this group to batch changes.'))
+
+def test_create_batch_change_validation_with_owner_group_id(shared_zone_test_context):
+    """
+    Test creating a batch change should properly set owner group ID in the following circumstances:
+    - create in shared zone
+    - update in shared zone without existing owner group ID
+
+    Owner group ID will be ignored in the following circumstances:
+    - create in private zone
+    - update in private zone
+    - update in shared zone with pre-existing owner group ID
+    - delete in either private or shared zone
+    """
+    shared_client = shared_zone_test_context.shared_zone_vinyldns_client
+    ok_client = shared_zone_test_context.ok_vinyldns_client
+    shared_group = shared_zone_test_context.shared_record_group
+    ok_group = shared_zone_test_context.ok_group
+    shared_zone = shared_zone_test_context.shared_zone
+    ok_zone = shared_zone_test_context.ok_zone
+
+    # record sets to setup
+    private_update = get_recordset_json(ok_zone, "private-update", "A", [{ "address": "1.1.1.1" }], 200)
+    shared_update_no_owner_group = get_recordset_json(shared_zone, "shared-update-no-owner-group", "A", [{ "address": "1.1.1.1" }], 200)
+    shared_update_existing_owner_group = get_recordset_json(shared_zone, "shared-update-existing-owner-group", "A", [{ "address": "1.1.1.1" }], 200, shared_group['id'])
+    private_delete = get_recordset_json(ok_zone, "private-delete", "A", [{ "address": "1.1.1.1" }], 200)
+    shared_delete = get_recordset_json(shared_zone, "shared-delete", "A", [{ "address": "1.1.1.1" }], 200)
+
+    to_delete_ok = {}
+    to_delete_shared = {}
+
+    batch_change_input = {
+        "changes": [
+            get_change_A_AAAA_json("private-create.ok."),
+            get_change_A_AAAA_json("shared-create.shared."),
+            get_change_A_AAAA_json("private-update.ok.", ttl=300),
+            get_change_A_AAAA_json("private-update.ok.", change_type="DeleteRecordSet"),
+            get_change_A_AAAA_json("shared-update-no-owner-group.shared.", ttl=300),
+            get_change_A_AAAA_json("shared-update-no-owner-group.shared.", change_type="DeleteRecordSet"),
+            get_change_A_AAAA_json("shared-update-existing-owner-group.shared.", ttl=300),
+            get_change_A_AAAA_json("shared-update-existing-owner-group.shared.", change_type="DeleteRecordSet"),
+            get_change_A_AAAA_json("private-delete.ok.", change_type="DeleteRecordSet"),
+            get_change_A_AAAA_json("shared-delete.shared.", change_type="DeleteRecordSet")
+        ],
+        "ownerGroupId": ok_group['id']
+    }
+
+    try:
+        for rs in [private_update, private_delete]:
+            create_rs = ok_client.create_recordset(rs, status=202)
+            ok_client.wait_until_recordset_change_status(create_rs, 'Complete')
+
+        for rs in [shared_update_no_owner_group, shared_update_existing_owner_group, shared_delete]:
+            create_rs = shared_client.create_recordset(rs, status=202)
+            shared_client.wait_until_recordset_change_status(create_rs, 'Complete')
+
+        result = ok_client.create_batch_change(batch_change_input, status=202)
+        completed_batch = ok_client.wait_until_batch_change_completed(result)
+
+        assert_that(completed_batch['ownerGroupId'], is_(ok_group['id']))
+
+        # set here because multiple items in the batch combine to one RS
+        record_set_list = [(change['zoneId'], change['recordSetId']) for change in completed_batch['changes'] if "delete" not in change['recordName'] and change['zoneId'] == ok_zone['id']]
+        to_delete_ok = set(record_set_list)
+
+        record_set_list = [(change['zoneId'], change['recordSetId']) for change in completed_batch['changes'] if "delete" not in change['recordName'] and change['zoneId'] == shared_zone['id']]
+        to_delete_shared = set(record_set_list)
+
+        assert_change_success_response_values(completed_batch['changes'], zone=ok_zone, index=0, record_name="private-create",
+                                              input_name="private-create.ok.", record_data="1.1.1.1")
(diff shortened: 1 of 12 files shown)


modules/api/src/main/scala/vinyldns/api/domain/AccessValidations.scala
```diff
@@ -183,7 +184,7 @@ object AccessValidations extends AccessValidationAlgebra {
     case testUser if testUser.isTestUser && !zone.isTest => AccessLevel.NoAccess
     case admin if admin.canEditAll || admin.isGroupMember(zone.adminGroupId) =>
       AccessLevel.Delete
-    case recordOwner if zone.shared && recordOwnerGroupId.exists(recordOwner.isGroupMember) =>
+    case recordOwner if zone.shared && recordOwnerGroupId.forall(recordOwner.isGroupMember) =>
```
mitruly: `exists` means that `recordOwnerGroupId` _must_ exist _and_ the user has to be a member of the record owner group. `forall` means that either the `recordOwnerGroupId` must exist and the user is a member _or_ the record has no owner group.

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
```diff
@@ -376,4 +390,29 @@ class BatchChangeValidations(changeLimit: Int, accessValidation: AccessValidatio
...
+
+  def updateOwnerGroupIsValid(
+      change: ChangeForValidation,
+      existingRecordSets: ExistingRecordSets,
+      ownerGroupId: Option[String]): SingleValidation[Unit] =
+    if (!change.zone.shared) {
+      ().validNel
+    } else {
+      existingRecordSets.get(change.zone.id, change.recordName, change.inputChange.typ) match {
+        case None => RecordDoesNotExist(change.inputChange.inputName).invalidNel
```
mitruly: Would definitely like a better approach to this, if possible. `RecordDoesNotExist` should never actually get hit, since we build `ExistingRecordSets` from the records we pull from the service.

modules/api/functional_test/live_tests/batch/create_batch_change_test.py
```diff
@@ -2600,3 +2602,260 @@ def test_create_batch_change_for_shared_zone_with_unauthorized_owner_group_id_fa
 
     errors = shared_client.create_batch_change(batch_change_input, status=400)['errors']
     assert_that(errors, contains('User "sharedZoneUser" must be a member of group "' + ok_group['id'] + '" to apply this group to batch changes.'))
+
+def test_create_batch_change_validation_with_owner_group_id(shared_zone_test_context):
```
nimaeskandary: this may be our biggest test yet 😛

modules/api/functional_test/live_tests/recordsets/create_recordset_test.py
```diff
@@ -1854,20 +1854,28 @@ def test_create_with_owner_group_in_shared_zone_by_acl_passes(shared_zone_test_c
             shared_zone_test_context.shared_zone_vinyldns_client.wait_until_recordset_change_status(delete_result, 'Complete')
 
 
-def test_create_in_shared_zone_by_unassociated_user_fails(shared_zone_test_context):
+def test_create_in_shared_zone_by_unassociated_user_succeeds(shared_zone_test_context):
```
nimaeskandary: Can we have a test on what happens when they have no owner group set? Not to worried about whether it passes or succeeds, but a test will track the current behavior if it changes

modules/api/functional_test/live_tests/batch/create_batch_change_test.py
```diff
@@ -2600,3 +2602,260 @@ def test_create_batch_change_for_shared_zone_with_unauthorized_owner_group_id_fa
...
+        assert_change_success_response_values(completed_batch['changes'], zone=shared_zone, index=7, record_name="shared-update-existing-owner-group",
+                                              input_name="shared-update-existing-owner-group.shared.", record_data=None, change_type="DeleteRecordSet")
+        assert_change_success_response_values(completed_batch['changes'], zone=ok_zone, index=8, record_name="private-delete",
+                                              input_name="private-delete.ok.", record_data=None, change_type="DeleteRecordSet")
+        assert_change_success_response_values(completed_batch['changes'], zone=shared_zone, index=9, record_name="shared-delete",
+                                              input_name="shared-delete.shared.", record_data=None, change_type="DeleteRecordSet")
+
+        # verify record set owner group
+        for result_rs in to_delete_ok:
+            rs_result = ok_client.get_recordset(result_rs[0], result_rs[1], status=200)
```
nimaeskandary: confused on the `result_rs[0], result_rs[1]`, aren't these zone id and record set id?

nimaeskandary: oh nvm, see how you make the tuples

modules/api/functional_test/live_tests/batch/create_batch_change_test.py
```diff
@@ -2600,3 +2602,260 @@ def test_create_batch_change_for_shared_zone_with_unauthorized_owner_group_id_fa
...
+                                              input_name="private-update.ok.", record_data=None, change_type="DeleteRecordSet")
+        assert_change_success_response_values(completed_batch['changes'], zone=shared_zone, index=4, record_name="shared-update-no-owner-group",
+                                              input_name="shared-update-no-owner-group.shared.", record_data="1.1.1.1", ttl=300)
+        assert_change_success_response_values(completed_batch['changes'], zone=shared_zone, index=5, record_name="shared-update-no-owner-group",
+                                              input_name="shared-update-no-owner-group.shared.", record_data=None, change_type="DeleteRecordSet")
+        assert_change_success_response_values(completed_batch['changes'], zone=shared_zone, index=6, record_name="shared-update-existing-owner-group",
+                                              input_name="shared-update-existing-owner-group.shared.", record_data="1.1.1.1", ttl=300)
+        assert_change_success_response_values(completed_batch['changes'], zone=shared_zone, index=7, record_name="shared-update-existing-owner-group",
```
nimaeskandary: are validating that records like this have their owner group id unchanged in other tests?

modules/api/functional_test/live_tests/recordsets/create_recordset_test.py
mitruly: Good suggestion - I'll add one.

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
```diff
@@ -376,4 +390,29 @@ class BatchChangeValidations(changeLimit: Int, accessValidation: AccessValidatio
           VinylDNSConfig.highValueRegexList,
           change.inputName)
     }
+
+  def createOwnerGroupIsValid(
```
nimaeskandary: we already have a `validateOwnerGroupId` in this file, can we change the wording of the functions or add comments to the top of all these explaining their intended function?

modules/api/functional_test/live_tests/batch/create_batch_change_test.py
mitruly: This is a scenario that is strictly for batch update (where specifying a valid owner group ID is ignored if there is an existing owner group ID on the record set). For the update record set endpoint, update to owner group would succeed if the user has access. I'll double-check that there is a test for this, though I think there already is one.

modules/api/src/main/scala/vinyldns/api/domain/DomainValidationErrors.scala
```diff
@@ -123,8 +123,8 @@ final case class CnameIsNotUniqueError(name: String, typ: RecordType)
       s"""Existing record with name "$name" and type "$typ" conflicts with this record."""
 }
 
-final case class UserIsNotAuthorized(user: String) extends DomainValidationError {
-  def message: String = s"""User "$user" is not authorized."""
+final case class UserIsNotAuthorized(userId: String) extends DomainValidationError {
```
rebstar6: username is actually probably more helpful here

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
```diff
@@ -148,16 +150,17 @@ class BatchChangeValidations(changeLimit: Int, accessValidation: AccessValidatio
...
 
     // Updates are a combination of an add and delete for a record with the same name and type in a zone.
     changes.mapValid {
       case addUpdate: AddChangeForValidation
           if changeGroups.containsDeleteChangeForValidation(addUpdate.recordKey) =>
-        validateAddUpdateWithContext(addUpdate, changeGroups, auth)
+        validateAddUpdateWithContext(addUpdate, changeGroups, existingRecords, auth, ownerGroupId)
       case add: AddChangeForValidation =>
-        validateAddWithContext(add, changeGroups, existingRecords, auth)
+        validateAddWithContext(add, changeGroups, existingRecords, auth, ownerGroupId)
```
rebstar6: when passing this around, can you call it `inputOwnerGroupId` or `batchOwnerGroupId`? it gets confusing later when you are using the ownerGroupIds on the existing records

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
```diff
@@ -192,7 +195,8 @@ class BatchChangeValidations(changeLimit: Int, accessValidation: AccessValidatio
     }
 
     val validations = typedValidations |+|
-      userCanAddUpdateRecordSet(change, auth)
+      userCanAddUpdateRecordSet(change, auth) |+|
```
rebstar6: kill the `userCanAddUpdateRecordSet` function, it was built on the assumption that adds and updates follow the same logic, which they no longer do.

need to split out so update passes in the existing owner group id

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
rebstar6: merge these two functions into:
```
  def ownerGroupProvidedIfNeeded(change: AddChangeForValidation,
                                 existingRecord: Option[RecordSet],
                                 ownerGroupId: Option[String]): SingleValidation[Unit] = {
    if (!change.zone.shared || ownerGroupId.isDefined) {
      ().validNel
    } else {
      existingRecord match {
        case Some(rs) if rs.ownerGroupId.isDefined => ().validNel
        case _ =>
          // rs exists without owner group, or this is a create
          MissingOwnerGroupId(change.recordName, change.zone.name).invalidNel
      }
    }
  }
```

modules/api/functional_test/live_tests/recordsets/create_recordset_test.py
```diff
@@ -1853,21 +1853,49 @@ def test_create_with_owner_group_in_shared_zone_by_acl_passes(shared_zone_test_c
             delete_result = shared_zone_test_context.shared_zone_vinyldns_client.delete_recordset(zone['id'], create_rs['id'], status=202)
             shared_zone_test_context.shared_zone_vinyldns_client.wait_until_recordset_change_status(delete_result, 'Complete')
 
+def test_create_in_shared_zone_without_owner_group_id_succeeds(shared_zone_test_context):
+    """
+    Test that creating a record in a shared zone without and owner group ID specified succeeds
```
britneywright: "an" not "and"

modules/api/functional_test/live_tests/recordsets/create_recordset_test.py
```diff
@@ -1853,21 +1853,49 @@ def test_create_with_owner_group_in_shared_zone_by_acl_passes(shared_zone_test_c
...
+        create_rs = shared_client.wait_until_recordset_change_status(create_response, 'Complete')['recordSet']
+        assert_that(create_rs, is_not(has_key('ownerGroupId')))
+
+    finally:
+        if create_rs:
+            delete_result = dummy_client.delete_recordset(create_rs['zoneId'], create_rs['id'], status=202)
+            shared_client.wait_until_recordset_change_status(delete_result, 'Complete')
 
-def test_create_in_shared_zone_by_unassociated_user_fails(shared_zone_test_context):
+def test_create_in_shared_zone_by_unassociated_user_succeeds(shared_zone_test_context):
```
britneywright: what happens if a user has an ACL rule that only gives them read access to a zone, or a support user?

britneywright: so support users can't create records in shared zones?

britneywright: Cool, yeah just wrapping my mind the implications of the validation changes.

modules/api/src/main/scala/vinyldns/api/domain/AccessValidations.scala
britneywright: ah i see it now, since recordOwner comes before supportUser and aclAccess we match on it first. I would like to see tests for support users and acl users for unowned records in shared zones so we know if the behavior changes later. I'm fine if it's in a separate PR, but if that's the case we should make an issue before this is closed.

mitruly: @britneywright Added a test for ACL. Can't add support users as we don't test either support or super users in functional tests as we can't add them.

modules/api/src/test/scala/vinyldns/api/domain/AccessValidationsSpec.scala
```diff
@@ -490,6 +490,18 @@ class AccessValidationsSpec
       result shouldBe AccessLevel.Delete
     }
 
+    "return AccessLevel.Delete if the record is unowned and the zone is shared" in {
```
britneywright: 👍

modules/api/functional_test/live_tests/batch/create_batch_change_test.py
rebstar6: wow no kidding

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
```diff
@@ -192,7 +199,11 @@ class BatchChangeValidations(changeLimit: Int, accessValidation: AccessValidatio
     }
 
     val validations = typedValidations |+|
-      userCanAddUpdateRecordSet(change, auth)
+      userCanUpdateRecordSet(change, auth, batchOwnerGroupId) |+|
```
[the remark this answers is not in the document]
mitruly: Good catch - this one should have been the normal create validation (without owner group) since the delete part of the update validation should take care of whether the user can delete the record.

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
```diff
@@ -169,30 +176,33 @@ class BatchChangeValidations(changeLimit: Int, accessValidation: AccessValidatio
...
     // Updates require checking against other batch changes since multiple adds
     // could potentially be grouped with a single delete
     val typedValidations = change.inputChange.typ match {
       case CNAME | TXT => recordIsUniqueInBatch(change, changeGroups)
       case _ => ().validNel
     }
 
-    val validations = typedValidations |+|
-      userCanAddUpdateRecordSet(change, auth)
+    val validations = typedValidations |+| userCanAddRecordSet(change, auth) |+|
```
rebstar6: this still feels like it should be split out, no? that said, are we just saying were relying on `userCanUpdateRecordSet` being called from the delete side (?)

rebstar6: its a little fishy, the add validation is no longer really valid. the user may have access because they are in the owner group, but this validation wouldnt know it. makes more sense to me to have the same validation on the add and delete side at this point

modules/api/functional_test/live_tests/batch/create_batch_change_test.py
```diff
@@ -2859,3 +2865,39 @@ def test_create_batch_delete_record_for_zone_admin_not_in_owner_group_succeeds(s
 
     assert_change_success_response_values(completed_batch['changes'], zone=shared_zone, index=0, record_name="shared-delete",
                                           input_name="shared-delete.shared.", record_data=None, change_type="DeleteRecordSet")
+
+def test_create_batch_update_record_in_shared_zone_for_unassociated_user_in_owner_group_succeeds(shared_zone_test_context):
```
mitruly: Proactively adding an update test for when record filters get merged in.

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
```diff
@@ -198,11 +198,18 @@ class BatchChangeValidations(changeLimit: Int, accessValidation: AccessValidatio
       case _ => ().validNel
     }
 
-    val validations = typedValidations |+| userCanAddRecordSet(change, auth) |+|
```
mitruly: Converting from `userCanAddRecordSet` to `userCanUpdateRecordSet` based on @rebstar6's proposal in the event that a user is updating an assigned record (which could not otherwise be added by the user outright).

PR #465, opened 2019-01-29
Update getRecordSetChange access validation

### LINKED ISSUE #453: Update getRecordSetChange access validation
**Motivation**
Currently `getRecordSetChange` in `RecordSetService` uses `canSeeZone(...)` to validate whether a user can access the change. Ideally, the user should be able to access the record set change if the user has access to the record set itself.

**Scope of change**
Update `RecordSetService` to use `canViewRecordSet` validation and corresponding tests.

britneywright:
Fixes #453.

Changes in this pull request:
- use canViewRecordSet validation instead of canSeeZone validation
- add unit tests in RecordSetServiceSpec
- ~~updated one func test to illustrate that a user in the record owner group of a record in a shared zone can view a related record set change. put that in a separate commit that doesn't need to be merged since i know @mitruly intends to test that in #454 https://github.com/vinyldns/vinyldns/pull/454/files#diff-8f6bb5f233fd1e5366e4831da659d5d8R727~~
- update `test_user_cannot_delete_record_with_write_txt_read_all` func test to confirm a user without admin rights can still get a record set change if they have other valid record access.

### DIFF
--- a/modules/api/functional_test/live_tests/recordsets/delete_recordset_test.py
+++ b/modules/api/functional_test/live_tests/recordsets/delete_recordset_test.py
@@ -494,7 +494,7 @@ def test_user_cannot_delete_record_with_write_txt_read_all(shared_zone_test_cont
         # dummy should be able to add the RS
         new_rs = get_recordset_json(ok_zone, "www-user-cant-delete", "TXT", [{'text':'should-work'}])
         rs_change = dummy_client.create_recordset(new_rs, status=202)
-        created_rs = client.wait_until_recordset_change_status(rs_change, 'Complete')['recordSet']
+        created_rs = dummy_client.wait_until_recordset_change_status(rs_change, 'Complete')['recordSet']
         verify_recordset(created_rs, new_rs)
 
         #dummy cannot delete the RS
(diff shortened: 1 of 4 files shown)


britneywright:
@mitruly I updated the func test you mentioned and removed the commit with the other func test update that isn't needed since we have this one in place now.

mitruly:
With the merge of https://github.com/vinyldns/vinyldns/pull/454, `test_delete_for_user_in_record_owner_group_in_shared_zone_succeeds` in `delete_recordset_test.py` will also need to get updated.

britneywright:
Func test updated @mitruly

nimaeskandary:
hm can we have a func test for this too the get record set change route and the list record set changes routes? For listing you could do something similar to `https://github.com/vinyldns/vinyldns/pull/459/files#diff-3a8bdff60580cf5efe1357597232ce5eR159` and not pollute the zone that other listing functions utilize

britneywright:
@nimaeskandary I didn't change the behavior of listRecordSetChanges, this PR shouldn't impact that endpoint

PR #467, opened 2019-01-30
make shared zone ids unique across runs

rebstar6:
maintenance thing - because we are using the same ids always for the shared zones, one failed run can corrupt the database. pulled that out, just getting those zones by a list

### DIFF
--- a/modules/api/functional_test/live_tests/shared_zone_test_context.py
+++ b/modules/api/functional_test/live_tests/shared_zone_test_context.py
@@ -266,10 +266,14 @@ class SharedZoneTestContext(object):
                 }, status=202)
             self.ds_zone = ds_zone_change['zone']
 
-            shared_zone_change = self.set_up_shared_zone('shared-zone')
+            get_shared_zones = self.shared_zone_vinyldns_client.list_zones(status=200)['zones']
+            shared_zone = [zone for zone in get_shared_zones if zone["name"] == "shared."]
+            non_test_shared_zone = [zone for zone in get_shared_zones if zone["name"] == "non.test.shared."]
+
+            shared_zone_change = self.set_up_shared_zone(shared_zone[0]["id"])
             self.shared_zone = shared_zone_change['zone']
 
-            non_test_shared_zone_change = self.set_up_shared_zone('non-test-shared-zone')
+            non_test_shared_zone_change = self.set_up_shared_zone(non_test_shared_zone[0]["id"])
             self.non_test_shared_zone = non_test_shared_zone_change['zone']
 
             # wait until our zones are created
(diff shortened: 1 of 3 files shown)


modules/api/functional_test/live_tests/shared_zone_test_context.py
```diff
@@ -266,10 +266,14 @@ def __init__(self):
                 }, status=202)
             self.ds_zone = ds_zone_change['zone']
 
-            shared_zone_change = self.set_up_shared_zone('shared-zone')
+            get_shared_zones = self.shared_zone_vinyldns_client.list_zones(status=200)['zones']
+            shared_zone = [zone for zone in get_shared_zones if zone["name"] == "shared."]
```
mitruly: Should probably used single quotes for accessing JSON attributes to stay consistent. Strings are fine in double quotes.

modules/api/src/main/scala/vinyldns/api/repository/TestDataLoader.scala
```diff
@@ -161,7 +161,6 @@ object TestDataLoader {
   final val sharedZone = Zone(
     name = "shared.",
     email = "email",
-    id = "shared-zone",
```
britneywright: we use these ID's in shared_zone_test_context.py to retrieve the shared zone and set the zone connection info. We could change that to look up the zones shared_test_user has access to in order to get whatever zone ID get's assigned the sharedZone

modules/api/functional_test/live_tests/shared_zone_test_context.py
```diff
@@ -266,10 +266,14 @@ def __init__(self):
...
+            non_test_shared_zone = [zone for zone in get_shared_zones if zone["name"] == "non.test.shared."]
+
+            shared_zone_change = self.set_up_shared_zone(shared_zone[0]["id"])
```
mitruly: Optional: Might be good to check if `shared_zone[0]` exists before accessing here (since someone could change a shared zone name and this would fail). On the other hand, it would fail here rather than later so 🤷‍♂️

modules/api/src/main/scala/vinyldns/api/repository/TestDataLoader.scala
mitruly: This is still handled in `shared_zone_test_context` with the same way; we're basically identifying the shared zone by name, taking its corresponding ID and then updating the connection info using the randomly populated ID.

modules/api/functional_test/live_tests/shared_zone_test_context.py
rebstar6: yeah thats what i figure, dont need the assertion, will fail anyway

modules/api/src/main/scala/vinyldns/api/repository/TestDataLoader.scala
rebstar6: i pull it from the shared users list_zones() by name

britneywright: missed it :)

PR #469, opened 2019-01-30
portal: add record owner group to record modals

### LINKED ISSUE #468: Portal: add record owner group to records modal
Update records modal in the Manage Records tab in the Zone section of the portal. 
- Should default to no owner group in the create modal.
- Should default to the owner group in the update modal if the record already has an owner group.
- Should default to no owner group in the update modal if the record is unowned.
- In the old and new recordset modals the record owner group field should be read-only with whatever values the record had/has

britneywright:
Fixes #468

Changes in this pull request:
- added a jsConfig value to hide the portal additions. since the modal file isn't scala it can't use the scala feature flag we set up.
- Adds record owner group selection in all of the record modals.

### DIFF
--- a/modules/portal/public/lib/constants.js
+++ b/modules/portal/public/lib/constants.js
@@ -17,6 +17,5 @@
 angular
     .module('constants', [])
     .constant('jsConfig', {
-        batchChangeLimit: 20,
-        sharedDisplayEnabled: false
+        batchChangeLimit: 20
     });
--- a/modules/portal/public/lib/constants.js
+++ b/modules/portal/public/lib/constants.js
@@ -17,5 +17,6 @@
 angular
     .module('batch-change')
     .constant('jsConfig', {
-        batchChangeLimit: 20
+        batchChangeLimit: 20,
+        sharedDisplayEnabled: false
     });
--- a/modules/portal/public/lib/constants.js
+++ b/modules/portal/public/lib/constants.js
@@ -15,7 +15,7 @@
  */
 
 angular
-    .module('batch-change')
+    .module('constants', [])
     .constant('jsConfig', {
         batchChangeLimit: 20,
         sharedDisplayEnabled: false
--- a/modules/portal/public/lib/controllers/controller.records.js
+++ b/modules/portal/public/lib/controllers/controller.records.js
@@ -261,7 +261,6 @@ angular.module('controller.records', [])
     function createRecordSet(record) {
         var payload = recordsService.toVinylRecord(record);
         payload.zoneId = $scope.zoneId;
-        payload.ownerGroupId = record.ownerGroupId;
         return recordsService
             .createRecordSet($scope.zoneId, payload)
             .then(recordSetSuccess("Create Record"))
--- a/modules/portal/public/lib/controllers/controller.records.js
+++ b/modules/portal/public/lib/controllers/controller.records.js
@@ -15,7 +15,7 @@
  */
 
 angular.module('controller.records', [])
-    .controller('RecordsController', function ($scope, $timeout, $log, recordsService, groupsService, pagingService, utilityService, $q) {
+    .controller('RecordsController', function ($scope, jsConfig, $timeout, $log, recordsService, groupsService, pagingService, utilityService, $q) {
 
     /**
       * Scope data initial setup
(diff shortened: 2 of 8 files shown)


modules/portal/public/templates/record-modal.html
```diff
@@ -314,6 +314,16 @@
                 </modal-invalid>
             </modal-element>
 
+            <modal-element label="Record Owner Group" ng-if="sharedDisplayEnabled">
+                <select class="form-control"
+                        ng-model="currentRecord.ownerGroupId"
+                        ng-disabled="recordModal.details.readOnly">
+                    <option ng-repeat="group in myGroups | orderBy: 'name'" value="{{ group.id }}"
```
mitruly: Encountered a couple of issues:
- There's no way to de-select a group that is selected despite no owner group being valid sometimes. The `Clear Form` button doesn't reset this, either; currently have to close and re-open modal.

- Only groups user is a member of populate here, even  if user is a super user.

mitruly: `clearRecord` (in `controller.records.js`) should be updated to clear out the record owner group:
```
    $scope.clearRecord = function(record) {
        record.ttl = undefined;
        record.data = undefined;
    };
```

modules/portal/public/templates/record-modal.html
```diff
@@ -314,6 +314,17 @@
                 </modal-invalid>
             </modal-element>
 
+            <modal-element label="Record Owner Group" ng-if="sharedDisplayEnabled">
```
mitruly: Note to reviewers: `sharedDisplayEnabled` must be set to `true` when testing to access this functionality.

modules/portal/public/lib/constants.js
```diff
@@ -15,7 +15,8 @@
  */
 
 angular
-    .module('batch-change')
+    .module('constants', [])
     .constant('jsConfig', {
-        batchChangeLimit: 20
+        batchChangeLimit: 20,
+        sharedDisplayEnabled: false
```
mitruly: Note to reviewers: `sharedDisplayEnabled` must be set to `true` when testing to access this functionality.

modules/portal/public/templates/record-modal.html
britneywright: added, if the feature flag is enabled. it's interesting the record data doesn't actually seem to clear like TTL does.

modules/portal/public/lib/controllers/controller.records.js
```diff
@@ -267,6 +273,9 @@ angular.module('controller.records', [])
     function updateRecordSet(record) {
         var payload = recordsService.toVinylRecord(record);
         payload.zoneId = $scope.zoneId;
+        if (record.ownerGroupId) {
```
nimaeskandary: curious why toVinylRecord isn't being updated instead

britneywright: i'll try it out.

britneywright: updated

modules/portal/public/lib/constants.js
rebstar6: is this pulling from meta, or its a second config for the same thing?

britneywright: It's a second config for the same thing. Because the records modal is purely javascript it can't pick up the scala config setting in meta.

rebstar6: yeesh, not ideal. is that something we inject in when we run, or it has to be hard coded in config like that? for example, we can build a version of the portal, and run with different application.conf's with different config values. If we change a value in here, do we have to do a whole re-build and deploy? or this file is passed in just like the other config?

britneywright: let me take another look and see how it can be passed in.

rebstar6: sounds good. if this is how it has to be, it's fine

rebstar6:
ran it, does what we want. look into the config thing, if within the day we're not at a better solution for that, im fine with this being merged

modules/portal/public/lib/constants.js
britneywright: figured it out and updated

britneywright:
updated so it's now only for shared zones and only when the shareDisplayEnabled flag is set to true @rebstar6

PR #472, opened 2019-01-30
Make DS records work

rebstar6:
the final

### DIFF
--- a/modules/api/functional_test/live_tests/recordsets/create_recordset_test.py
+++ b/modules/api/functional_test/live_tests/recordsets/create_recordset_test.py
@@ -1898,17 +1898,23 @@ def test_create_with_owner_group_when_not_member_fails(shared_zone_test_context)
     assert_that(error, is_('User not in record owner group with id "' + group['id'] + '"'))
 
 
-def test_create_ds_valid_input(shared_zone_test_context):
+def test_create_ds_success(shared_zone_test_context):
     """
-    Test that creating a DS record passes input validations
+    Test that creating a valid DS record succeeds
     """
 
     client = shared_zone_test_context.ok_vinyldns_client
     zone = shared_zone_test_context.ds_zone
     record_data = [{'keytag': 60485, 'algorithm': 5, 'digesttype': 1, 'digest': '2BB183AF5F22588179A53B0A98631FAD1A292118'}]
-    record_json = get_recordset_json(zone, 'dskey', 'DS', record_data)
-    # input was successful if it gets past to determine auth; this will be updated to success test
-    client.create_recordset(record_json, status=403)
+    record_json = get_recordset_json(zone, 'dskey', 'DS', record_data, ttl=3600)
+    result_rs = None
+    try:
+        result = client.create_recordset(record_json, status=202)
+        result_rs = client.wait_until_recordset_change_status(result, 'Complete')['recordSet']
+    finally:
+        if result_rs:
+            client.delete_recordset(result_rs['zoneId'], result_rs['id'], status=(202,404))
+            client.wait_until_recordset_deleted(result_rs['zoneId'], result_rs['id'])
 
 
 def test_create_ds_non_hex_digest(shared_zone_test_context):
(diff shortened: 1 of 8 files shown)


modules/api/functional_test/live_tests/recordsets/create_recordset_test.py
```diff
@@ -1948,3 +1968,56 @@ def test_create_ds_unknown_digest_type(shared_zone_test_context):
     record_json = get_recordset_json(zone, 'dskey', 'DS', record_data)
     errors = client.create_recordset(record_json, status=400)['errors']
     assert_that(errors, contains_inanyorder("Digest Type 0 is not a supported DS record digest type"))
+
+
+def test_create_ds_bad_ttl(shared_zone_test_context):
```
mitruly: Should probably add `_fails` to test name for consistency.

modules/api/functional_test/live_tests/recordsets/update_recordset_test.py
```diff
@@ -2223,3 +2223,105 @@ def test_update_to_group_a_user_is_notin_fails(shared_zone_test_context):
         if create_rs:
             delete_result = shared_client.delete_recordset(zone['id'], create_rs['id'], status=202)
             shared_client.wait_until_recordset_change_status(delete_result, 'Complete')
+
```
mitruly: Extra new line here (3 vs 2).

modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetValidations.scala
```diff
@@ -149,6 +149,32 @@ object RecordSetValidations {
...
+  def dsValidations(
+      newRecordSet: RecordSet,
+      existingRecordsWithName: List[RecordSet],
+      zone: Zone): Either[Throwable, Unit] = {
+    // see https://tools.ietf.org/html/rfc4035#section-2.4
+    val nsChecks = existingRecordsWithName.find(_.typ == NS) match {
+      case Some(ns) if ns.ttl == newRecordSet.ttl => ().asRight
+      case Some(ns) =>
+        InvalidRequest(
+          s"DS record ${newRecordSet.name} must have TTL matching its linked NS (${ns.ttl})").asLeft
```
mitruly: Can we wrap the record name and zone name in quotes to make it more obvious? One of the func test messages makes it not super obvious: `DS record dskey must have TTL matching its linked NS (3600)`

modules/api/functional_test/live_tests/recordsets/create_recordset_test.py
```diff
@@ -1898,17 +1898,37 @@ def test_create_with_owner_group_when_not_member_fails(shared_zone_test_context)
...
+
+        # verifying recordset in dns backend
+        answers = dns_resolve(zone, result_rs['name'], result_rs['type'])
+        assert_that(answers, has_length(2))
+        rdata_strings = [x.upper() for x in rdata(answers)]
+        assert_that('60485 5 1 2BB183AF5F22588179A53B0A98631FAD1A292118', is_in(rdata_strings))
+        assert_that('60485 5 2 D4B7D520E7BB5F0F67674A0CCEB1E3E0614B93C4F9E99B8383F6A1E4469DA50A', is_in(rdata_strings))
+    finally:
+        if result_rs:
+            client.delete_recordset(result_rs['zoneId'], result_rs['id'], status=(202,404))
```
mitruly: Is there a reason a reason a DS record might not exist after successful creation? I changed all of the `delete_recordset` calls to `status=202` and they passed locally.

modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetValidations.scala
```diff
@@ -149,6 +149,32 @@ object RecordSetValidations {
 
   }
 
+  def dsValidations(
+      newRecordSet: RecordSet,
+      existingRecordsWithName: List[RecordSet],
+      zone: Zone): Either[Throwable, Unit] = {
+    // see https://tools.ietf.org/html/rfc4035#section-2.4
+    val nsChecks = existingRecordsWithName.find(_.typ == NS) match {
+      case Some(ns) if ns.ttl == newRecordSet.ttl => ().asRight
```
mitruly: The `NS` record `ttl` check is interesting here because we are proposing that `NS` gets created first (which is fine). However, the following scenario would also make the `DS` `digest` invalid, which I wonder if we should try to synchronize the `DS` and `NS` records if they are pointing to the same name/child but have different `ttl`s:
- `NS` record `foo` gets created in `example.com.` with `ttl` 200
- `DS` record `foo` gets created in `example.com.` with `ttl` 200
- `NS` record `foo` gets updated in `example.com.` to `ttl` 300

We definitely don't want to get stuck in a loop where we can't delete/update records, but in the third step above, it might be good to synchronize an update to `DS` as well. This does complicate things, though.

modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetValidations.scala
```diff
@@ -149,6 +149,32 @@ object RecordSetValidations {
...
+            s"name in the zone ${zone.name}").asLeft
+    }
+
+    for {
+      _ <- isNotDotted(newRecordSet, zone)
+      _ <- isNotOrigin(
+        newRecordSet,
+        zone,
+        s"Record with name ${newRecordSet.name} is an DS record at apex and cannot be added")
+      _ <- nsChecks
```
mitruly: This is more of a discussion bit.

Also in RFC 4035, there is a blurb:
```
A DS RR SHOULD point to a DNSKEY RR that is present in the child's
   apex DNSKEY RRset, and the child's apex DNSKEY RRset SHOULD be signed
   by the corresponding private key.  DS RRs that fail to meet these
   conditions are not useful for validation, but because the DS RR and
   its corresponding DNSKEY RR are in different zones, and because the
   DNS is only loosely consistent, temporary mismatches can occur.
```

Similar with the NS statement, we don't want to get stuck in a loop during an update attempt, but I'm wondering if we're adding a `DS` record if we should also check/warn of a mismatched `DNSKEY`.

modules/api/functional_test/live_tests/recordsets/create_recordset_test.py
rebstar6: this is what we do in general in cleanup. shouldnt happen, but we consider either an acceptable outcome, as long as the thing isnt there

modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetValidations.scala
rebstar6: we dont sync in DNSKEYs, so we cant check them :P

mitruly: Fair enough :P

modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetValidations.scala
rebstar6: the NS has to be created first. You can certainly get in weird situations if you edit the NS after the DS has been created, though I dont want to go crazy with it.

I *think* it's fine, DS record management should really only be done by people who know what theyre doing

modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetValidations.scala
rebstar6: eventually we will discuss more comprehensive dnssec management - i think this idea falls into that

modules/portal/public/lib/services/records/service.records.js
```diff
@@ -167,6 +167,7 @@ angular.module('service.records', [])
                     newRecord.onlyFour = true;
                     break;
                 default:
+                    newRecord.canBeEdited = false;
```
rebstar6: quick add here, we can properly add DS to the portal later, this just defaults editability for types that we dont know to false rather than true. means the buttons dont show up until we've handled the type

PR #475, opened 2019-02-05
pac4j oidc implementation

rebstar6:
Fixes #.

Changes in this pull request:
-
-


modules/portal/app/actions/VinylDnsAction.scala
```diff
@@ -16,32 +16,94 @@
...
-  def invokeBlock[A](request: Request[A], block: UserRequest[A] => Future[Result]): Future[Result] =
-    // if the user name is not in session, reject
-    request.session.get("username") match {
+  def createUser(
+      un: String,
+      fname: Option[String],
+      lname: Option[String],
+      email: Option[String]): Future[Option[User]] = Future.successful(None)
+
+  def invokeBlock[A](
```
nimaeskandary: wonder if the two actions should override this as well?

modules/portal/app/modules/SecurityModule.scala
```diff
@@ -0,0 +1,96 @@
...
+  SecurityComponents
+}
+
+/**
+  * Guice DI module to be included in application.conf
+  */
+class SecurityModule(environment: Environment, configuration: Configuration)
+    extends AbstractModule {
+
+  lazy val discoveryUrl: String = configuration.get[String]("oidc.oidc-metadata")
```
nimaeskandary: this config value could be named just `metadata` since it's under oidc

modules/portal/app/controllers/FrontendController.scala
```diff
@@ -16,50 +16,78 @@
...
+    extends Security[CommonProfile] {
+
+  val oidcEnabled: Boolean = configuration.getOptional[Boolean]("oidc.enabled").getOrElse(false)
+  lazy val oidcUsernameField: String =
+    configuration.getOptional[String]("oidc.jwt-username-field").getOrElse("username")
+
+  private val vinyldnsServiceBackend =
+    configuration
+      .getOptional[String]("portal.vinyldns.backend.url")
+      .getOrElse("http://localhost:9000")
```
pauljamescleary: we probably shouldn't do optional here.  I don't know how destructive things will be if someone accidentally omitted the `portal.vinyldns.backend.url` parameter in a real production environment.  Better to assume it is there, make sure it is in the local.conf (which it should be).  If it isn't there let the app blow up

modules/portal/app/controllers/FrontendController.scala
```diff
@@ -16,50 +16,78 @@
...
+            case _ =>
+              Ok(views.html.login())
+          }
+      }
     }
   }
 
   def logout(): Action[AnyContent] = Action { implicit request =>
-    Redirect("/login").withNewSession
+    if (oidcEnabled) {
```
pauljamescleary: why redirect to `/` if oidcEnabled?

modules/portal/app/controllers/FrontendController.scala
```diff
@@ -16,50 +16,78 @@
...
+      Redirect("/odic-logout")
```
mitruly: This should be `/oidc-logout` :P

rebstar6: gooood call

modules/portal/app/actions/LdapFrontendAction.scala
```diff
@@ -31,11 +33,15 @@ import scala.concurrent.{ExecutionContext, Future}
   * If the user is locked out, redirect to login screen
```
britneywright:
```suggestion
  * If the user is locked out, redirect to no access screen
```

modules/portal/app/actions/OidcFrontendAction.scala
```diff
@@ -0,0 +1,81 @@
...
+import play.api.mvc.Results.Redirect
+import vinyldns.core.domain.membership.User
+
+import scala.concurrent.{ExecutionContext, Future}
+
+/**
+  * Custom action that will take a request and do an account lookup for the user
+  * If the user is not in session, redirect to the login screen
+  * If the user is in session, but the account is not found, redirect to the login screen with a different message
+  * If the user is locked out, redirect to login screen
```
britneywright:
```suggestion
  * If the user is locked out, redirect to no access screen
```

modules/portal/app/actions/OidcFrontendAction.scala
```diff
@@ -0,0 +1,81 @@
...
+    Future.successful(
+      Redirect("/").withNewSession
+        .withHeaders(cacheHeaders: _*))
+
+  def cantFindAccountResult(un: String): Future[Result] =
+    Future.successful(
+      Redirect("/")
+        .withHeaders(cacheHeaders: _*))
+
+  // TODO need new screen for this
```
britneywright: can remove this comment

modules/portal/app/actions/LdapFrontendAction.scala
rebstar6: leaving this one as is, this is still true if ldap is enabled

rebstar6: oh wait, wrong line. ignore that

modules/portal/app/actions/VinylDnsAction.scala
rebstar6: defeats the point of having this - at that point would just not share the trait

modules/portal/app/controllers/FrontendController.scala
rebstar6: this is how weve been doing it historically. can move to non-optional though

rebstar6:
closing for https://github.com/vinyldns/vinyldns/pull/488

Issue #445, opened 2019-01-16
MFA


pauljamescleary:
**Describe the solution you'd like**
Add support to integrate with a Multi-Factor Authentication system when logging into the portal.

The integration should be configuration driven (like other external dependencies).  An example integration would be Microsoft Authenticator.

PR #476, opened 2019-02-05
portal improvements

britneywright:
part of #445.

Changes in this pull request:
- create a dedicated view for locked out users
- add VinylDNS version to footer of all pages in the portal


modules/portal/app/views/main.scala.html
```diff
@@ -88,7 +88,7 @@
                 @header(rootAccountName)(request)
                 @pageContent
 
-                <footer/></footer>
+                <footer/><p class="main-footer-text text-right">VinylDNS (version @{meta.version})</p></footer>
```
rebstar6: if this has meta do all the other pages need them?

britneywright: yes, main gets the customLinks and Meta models from the pages, they're passed through. It needs to be included in both main and the pages that use main.

nimaeskandary:
tested this out, looks good to me

PR #479, opened 2019-02-08
Record type filter for shared zones

### LINKED ISSUE #455: Add Record Type filter into shared zones logic
This card adds the concept of "shared-approved" types. For this, we should include a configuration option that takes in a list of record types. If excluded, the system will assume all types are approved.
This configuration will be a filter on the shared zones logic as implemented by cards tied to the epic https://github.com/vinyldns/vinyldns/issues/346. 
in summary, for a new or unowned recordset, we will only allow changes if the record type is in that configured list. For an owned recordset, this filter does not apply.

for example, in a shared zone, if only A type is in the approved list, a user outside of the admin group/ACLs may not create a new MX record in a shared zone, or pickup and edit an unowned one. However, zone admins may assign an existing MX record to a group, at which point members of that group may edit it.

---
Below info for context, includes overview of permissions to this point:

*Non-shared zones:*
- Current logic applies; anyone in the Zone Admin group can create/update/delete our supported types: A, AAAA, CNAME, PTR, MX, NS, SRV, TXT, SSHFP, SPF (+ soon to be DS)
- ACL rules can grant other groups or individuals access to specific regex's and/or record types

*Shared zones*
- all the logic above still applies for zone admins/ACL
...additionally:
- For CREATE recordset
    - Anyone can create it, so long as it is valid and within the "shared-approved" types
- For UPDATE recordset
    - If the recordset is *unowned*, anyone can update it, so long as the update is valid and it is one of the "shared-approved" types
    - If the recordset is *unowned* and some other, non approved type, old logic applies, the zone admins are in charge of making changes or explicitly granting access
    - If the recordset is *owned* by some group, *regardless of type*, members of that group can edit it, and nobody else (except Zone Admins, ACL folks)
- For DELETE recordset
    - If the recordset is *unowned*, anyone can delete it, so long as it is one of the "shared-approved" types
    - If the recordset is *unowned* and some other, non approved type, old logic applies, the zone admins are in charge of deleting or explicitly granting access
    - If the recordset is *owned* by some group, *regardless of type*, members of that group can delete it, and nobody else (except Zone Admins, ACL folks)

2 notes on locking down permissions:
- high value domains is in. any names in that list will be locked down so that *nobody* (not even super users) can make changes
- for flow purposes, a super user can load in a will-be shared zone 1st as non-shared. At that point, supers (or zone admins) can pre-assign a group to any recordsets that are a shared-approved type, but that they dont actually want everyone in the system to be able to access

britneywright:
Fixes #455.

Changes in this pull request:
- create a 'shared-approved-types' config value to whitelist the types of unowned records that any users can access and modify in shared zones


project/Dependencies.scala
```diff
@@ -48,7 +48,8 @@ object Dependencies {
     "org.typelevel"             %% "cats-effect"                    % catsEffectV,
     "com.47deg"                 %% "github4s"                       % "0.18.6",
     "com.comcast"               %% "ip4s-core"                      % ip4sV,
-    "com.comcast"               %% "ip4s-cats"                      % ip4sV
+    "com.comcast"               %% "ip4s-cats"                      % ip4sV,
+    "com.iheart"                %% "ficus"                          % "1.4.3"
```
rebstar6: 1.4.4 is the newest version (1.4.3 is from 2017). should use that one

modules/portal/public/test_frameworks/jasmine-jquery.js
```diff
@@ -0,0 +1,838 @@
...
+WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+*/
+
+(function (root, factory) {
+     if (typeof module !== 'undefined' && module.exports) {
+        factory(root, root.jasmine, require('jquery'));
+    } else {
+        factory(root, root.jasmine, root.jQuery);
+    }
+}((function() {return this; })(), function (window, jasmine, $) { "use strict";
```
rebstar6: did these get pulled in on rebase? this is an api card, shouldnt see these portal changes in here

docker/api/docker.conf
```diff
@@ -188,6 +188,8 @@ vinyldns {
       "fd69:27cc:fe91:0:0:0:ffff:0"
     ]
   }
+
+  shared-approved-types = ["A", "AAAA", "CNAME", "PTR"] # types of unowned records that users can access in shared zones
```
rebstar6: nitpick - can you move the comment to the line above for these config changes?

modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
```diff
@@ -63,6 +66,11 @@ object VinylDNSConfig {
   lazy val highValueIpList: List[IpAddress] =
     getOptionalStringList("high-value-domains.ip-list").flatMap(ip => IpAddress(ip))
 
+  lazy val sharedApprovedTypes: Set[RecordType.Value] =
```
rebstar6: a note - this logic is the same as ```vinyldnsConfig.as[Option[Set[RecordType.Value]]]("shared-approved-types").getOrElse(Set())```

im fine with either

project/Dependencies.scala
rebstar6: scratch that. maven has 1.4.4 but elsewhere communicated 1.4.3. oh well, stick with 3.

PR #484, opened 2019-02-12
Portal support for DS records

### LINKED ISSUE #443: Add DS to acl logic
Portal and api
- Add it to the list in the portal when making an ACL rule
- Verify that you can add a DS ACL rule in the api...may not require changes

Related to #421

mitruly:
Adds portal modal and service support for DS records. Fixes #443.

Changes in this pull request:
- Add support for DS records in ACL rule modal
- Add conversion for DS record to display record in record set info

### DIFF
--- a/modules/portal/app/views/zones/zoneTabs/manageRecords.scala.html
+++ b/modules/portal/app/views/zones/zoneTabs/manageRecords.scala.html
@@ -147,19 +147,27 @@
                                 </li>
                             </ul>
                         </span>
-                        <span ng-if="record.type == 'SPF'">
-                            <ul class="table-cell-list">
-                                <li ng-repeat="item in record.spfRecordData track by $index"
-                                    ng-if="!record.onlyFour || $index <= 3">{{item}}</li>
-                                <li ng-if="record.onlyFour && record.spfRecordData.length > 4">
-                                    <a href ng-click="record.onlyFour=false">more...</a>
-                                </li>
-                                <li ng-if="!record.onlyFour">
-                                    <a href ng-click="record.onlyFour=true">less...</a>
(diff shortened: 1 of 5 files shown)


mitruly:
Portal images with changes for DS records:

image
image
image
image

Note: View record set info was not updated as `SOA` is also not supported.

modules/portal/public/lib/controllers/controller.manageZones.js
```diff
@@ -63,6 +63,7 @@ angular.module('controller.manageZones', [])
             readOnly: false
         }
     };
+    $scope.aclRecordTypes = ['A', 'AAAA', 'CNAME', 'DS', 'MX', 'NS', 'TXT', 'SRV', 'PTR', 'SPF', 'SSHFP'];
```
mitruly: Needed to define a separate list for record types that ACL can be created for, as this differs from the record create/view list.

modules/portal/app/views/zones/zoneTabs/manageRecords.scala.html
```diff
@@ -147,19 +147,27 @@ <h3 class="panel-title">Records {{ getRecordPageTitle() }}</h3>
                                 </li>
                             </ul>
                         </span>
-                        <span ng-if="record.type == 'SPF'">
```
mitruly: Reordered types for readability.

britneywright:
should reorder SRV and SPF record types here in keeping with the alphabetized ordering. https://github.com/vinyldns/vinyldns/blob/de5bf17676df3ad5a42cb3f401f2ea93b28ad994/modules/portal/public/lib/services/records/service.records.js#L153-L166

PR #485, opened 2019-02-13
add tests to make sure username search is case insensitive

nimaeskandary:
Changes in this pull request:

- in general sql collations are case insensitive by default
- mariadb and mysql 5.7 have the default `latin_swedish_ci`, in which the `ci` stands for `case insensitive` : https://mariadb.com/kb/en/library/setting-character-sets-and-collations/, https://dev.mysql.com/doc/refman/5.7/en/charset-applications.html


rebstar6:
cool. ez.

PR #488, opened 2019-02-14
mfa though

rebstar6:
3rd times the charm?


modules/portal/app/controllers/OidcAuthenticator.scala
```diff
@@ -0,0 +1,189 @@
...
+        "response_type" -> responseType,
+        "redirect_uri" -> redirectUri,
+        "scope" -> "openid profile email")
+
+    (call.url, call.queryString)
+  }
+
+  def isNotExpired(claimsSet: JWTClaimsSet): Boolean = {
+    val now = new Date()
+    val expirationTime = claimsSet.getExpirationTime
```
nimaeskandary: is this an optional? according to https://tools.ietf.org/html/rfc7519#section-4.1 seems the claimset doesn't need to include it

modules/portal/app/actions/FrontendAction.scala
```diff
@@ -50,6 +55,6 @@ class FrontendAction(val userLookup: String => IO[Option[User]])(
 
   def lockedUserResult(un: String): Future[Result] =
     Future.successful(
-      Redirect("/noaccess")
+      Redirect("/noaccess").withNewSession
```
britneywright: if we add `withNewSession` here they'll be sent back through the authentication process if they try to go to a another page in the portal, vs continuing to be sent to the noaccess page.

modules/portal/app/actions/VinylDnsAction.scala
```diff
@@ -17,31 +17,68 @@
...
+        logger.info("User is not logged in or token expired; redirecting to login screen")
+        notLoggedInResult
 
       case Some(un) =>
         // user name in session, let's get it from the repo
         userLookup(un).unsafeToFuture().flatMap {
           // Odd case, but let's handle with a different error message
-          case None => cantFindAccountResult(un)
+          case None =>
+            logger.error(s"Cant find account for user  with username $un")
```
britneywright: there's an extra space here between "user" and "with"

modules/portal/app/actions/FrontendAction.scala
nimaeskandary: hm, now that I think about it the no access page breaks the logout flow no? Could we potentially make an issue to add a link to `/logout` on the no access screen and then this `.withNewSession` could be removed?

nimaeskandary: Or maybe add that link before this pr goes in @rebstar6 😛

britneywright: I don't think we need that. it's fine for the user to remain logged in and locked out. they'll always wind up on the no access screen so we're saving them the effort of logging in over and over to only be redirected to the no access page. If their account were unlocked they'd have access again.

nimaeskandary: Hm, I don't see a reason to not let the logged in user log out at all if they want, keeping them locked into their account in their browser seems like an issue

rebstar6: i think its ok as is, logout was an extra add anyway

PR #490, opened 2019-02-15
Fix IPv4 PTR zone discovery

### LINKED ISSUE #480: Batch change IPv4 zone discovery fails on edge case
**Describe the bug**
User will encounter an internal server error in `BatchChangeService` if the initial IPv4 zone match succeeds but the PTR does not find any zones that the PTR can fall into.

**VinylDNS Version**

**To Reproduce**
1. Run VinylDNS on localhost
1. Create a group
1. Connect to zone `192/30.2.0.192.in-addr.arpa.`
1. Create a batch change to create a PTR record with name `192.0.2.1`

An error will be generated by the API.

**Expected behavior**
User should fail with a `ZoneDiscoveryError`.

mitruly:
Fixes #480. Properly handle internal server error when PTRv4 zone discovery fails.

Changes in this pull request:
- Cover scenario where initial zone discovery finds potentially matching zones but fails on finding a record name match for IPv4 record

### DIFF
--- a/modules/api/functional_test/live_tests/batch/create_batch_change_test.py
+++ b/modules/api/functional_test/live_tests/batch/create_batch_change_test.py
@@ -1819,6 +1819,7 @@ def test_ipv4_ptr_record_when_zone_discovery_only_finds_mismatched_delegated_zon
     }
 
     try:
+        # delete classless base zone (2.0.192.in-addr.arpa); only remaining zone is delegated zone (192/30.2.0.192.in-addr.arpa)
         ok_client.delete_zone(classless_base_zone['id'], status=202)
         ok_client.wait_until_zone_deleted(classless_base_zone['id'])
         response = ok_client.create_batch_change(batch_change_input, status=400)
@@ -1826,6 +1827,7 @@ def test_ipv4_ptr_record_when_zone_discovery_only_finds_mismatched_delegated_zon
                                                error_messages=['Zone Discovery Failed: zone for "192.0.2.1" does not exist in VinylDNS. If zone exists, then it must be created in VinylDNS.'])
 
     finally:
+        # re-create classless base zone and update zone info in shared_zone_test_context for use in future tests
          zone_create_change = ok_client.create_zone(shared_zone_test_context.classless_base_zone_json, status=202)
          shared_zone_test_context.classless_base_zone = zone_create_change['zone']
          ok_client.wait_until_zone_exists(zone_create_change)
--- a/modules/api/functional_test/live_tests/batch/create_batch_change_test.py
+++ b/modules/api/functional_test/live_tests/batch/create_batch_change_test.py
@@ -1805,6 +1805,31 @@ def test_ipv4_ptr_recordtype_add_checks(shared_zone_test_context):
         clear_recordset_list(to_delete, client)
 
 
+def test_ipv4_ptr_record_when_zone_discovery_only_finds_mismatched_delegated_zone_fails(shared_zone_test_context):
+    """
+    Test IPv4 PTR record discovery for only delegated zones that do not match the record name fails
+    """
+    ok_client = shared_zone_test_context.ok_vinyldns_client
+    classless_base_zone = shared_zone_test_context.classless_base_zone
+
+    batch_change_input = {
+        "changes": [
+            get_change_PTR_json("192.0.2.1")
+        ]
+    }
+
+    try:
+        ok_client.delete_zone(classless_base_zone['id'], status=202)
+        ok_client.wait_until_zone_deleted(classless_base_zone['id'])
+        response = ok_client.create_batch_change(batch_change_input, status=400)
+        assert_failed_change_in_error_response(response[0], input_name="192.0.2.1", record_type="PTR", record_data="test.com.",
+                                               error_messages=['Zone Discovery Failed: zone for "192.0.2.1" does not exist in VinylDNS. If zone exists, then it must be created in VinylDNS.'])
+
+    finally:
+         zone_create_change = ok_client.create_zone(shared_zone_test_context.classless_base_zone_json, status=202)
+         shared_zone_test_context.classless_base_zone = zone_create_change['zone']
+         ok_client.wait_until_zone_exists(zone_create_change)
+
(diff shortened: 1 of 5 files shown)


modules/api/functional_test/live_tests/shared_zone_test_context.py
```diff
@@ -142,26 +142,28 @@ def __init__(self):
             )
             self.ip4_reverse_zone = ip4_reverse_zone_change['zone']
 
+            self.classless_base_zone_json = {
```
mitruly: Converted to a JSON since we need to recreate this zone after one of the tests.

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -222,12 +222,12 @@ class BatchChangeService(
       change: ChangeInput,
       zoneMap: ExistingZones): SingleValidation[ChangeForValidation] = {
     val zones = zoneMap.getipv4PTRMatches(change.inputName)
+    val recordName = change.inputName.split('.').takeRight(1).mkString
+    val validZones = zones.filter(zn => ptrIsInZone(zn, recordName, PTR).isRight)
```
mitruly: We can do the `.filter` before the checks since an empty `zones` collection will remain empty.

modules/api/src/test/scala/vinyldns/api/domain/batch/BatchChangeServiceSpec.scala
```diff
@@ -86,7 +86,7 @@ class BatchChangeServiceSpec
   private val onlyApexZone = Zone("only.apex.exists.", "email", id = "onlyApex")
   private val onlyBaseZone = Zone("only.base.", "email", id = "onlyBase")
   private val ptrZone = Zone("55.144.10.in-addr.arpa.", "email", id = "nonDelegatedPTR")
-  private val delegatedPTRZone = Zone("64/25.55.144.10.in-addr.arpa.", "email", id = "delegatedPTR")
+  private val delegatedPTRZone = Zone("64/25.55.144.10.in-addr.arpa.", "email", id = "delegatedPTR") /**/
```
britneywright: what's the `/**/` for?

mitruly: This shouldn't be here - removing it.

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
britneywright: but if `zones` is empty you wouldn't even need `recordName` or `validZones`

mitruly: Reworking it like this is cleaner and more concise because it collapses the following scenario:
```
if (zones.isEmpty) ZoneDiscoveryError(...).invalidNel
else if (validZones.isEmpty) ZoneDiscoveryError(...).invalidNel
else {
     # return some zone.validNel
}
```

britneywright: ok so do you need zones and validZones then?

```
val recordName = change.inputName.split('.').takeRight(1).mkString
val validZones = zoneMap.getipv4PTRMatches(change.inputName).filter(zn => ptrIsInZone(zn, recordName, PTR).isRight)
```

mitruly: Sure, they could be combined.

modules/api/functional_test/live_tests/batch/create_batch_change_test.py
```diff
@@ -1805,6 +1805,33 @@ def test_ipv4_ptr_recordtype_add_checks(shared_zone_test_context):
...
+        ]
+    }
+
+    try:
+        # delete classless base zone (2.0.192.in-addr.arpa); only remaining zone is delegated zone (192/30.2.0.192.in-addr.arpa)
+        ok_client.delete_zone(classless_base_zone['id'], status=202)
+        ok_client.wait_until_zone_deleted(classless_base_zone['id'])
+        response = ok_client.create_batch_change(batch_change_input, status=400)
+        assert_failed_change_in_error_response(response[0], input_name="192.0.2.1", record_type="PTR", record_data="test.com.",
+                                               error_messages=['Zone Discovery Failed: zone for "192.0.2.1" does not exist in VinylDNS. If zone exists, then it must be created in VinylDNS.'])
```
britneywright: I created an issue to update the ZoneDiscoveryError messaging https://github.com/vinyldns/vinyldns/issues/491. not a priority.

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -221,13 +221,12 @@ class BatchChangeService(
...
+    val recordName = change.inputName.split('.').takeRight(1).mkString
+    val validZones = zoneMap.getipv4PTRMatches(change.inputName).filter(ptrIsInZone(_, recordName, PTR).isRight)
 
-    if (zones.isEmpty)
+    if (validZones.isEmpty) {
       ZoneDiscoveryError(change.inputName).invalidNel
-    else {
-      val recordName = change.inputName.split('.').takeRight(1).mkString
-      val validZones = zones.filter(zn => ptrIsInZone(zn, recordName, PTR).isRight)
+    } else {
```
rebstar6: the option.get's here are gross.. i think this is the same but less error prone:

```
    val recordName = change.inputName.split('.').takeRight(1).mkString
    val validZones = zoneMap.getipv4PTRMatches(change.inputName).filter(ptrIsInZone(_, recordName, PTR).isRight)

    val zone = {
      if (validZones.size > 1) validZones.find(zn => zn.name.contains("/"))
      else validZones.headOption
    }
    zone match {
      case Some(z) => ChangeForValidation(z, recordName, change).validNel
      case None => ZoneDiscoveryError(change.inputName).invalidNel
    }
```

PR #495, opened 2019-02-19
Add fqdn col to mysql record set repo with a search function

nimaeskandary:
Follows
```
RecordSet table needs to support a FQDN index:

Update RecordSetRepository trait with FQDN query filter method that throws an error
Add FQDN index to RecordSet table
Provide migration script
Add a query filter method (similar to listRecordSets) for filtering record sets by FQDN
```

Changes in this pull request:
- added fqdn col
- added `getRecordSetsByFQDNs` that gets recordsets from a set of fqdns


modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlRecordSetRepositoryIntegrationSpec.scala
```diff
@@ -128,12 +128,32 @@ class MySqlRecordSetRepositoryIntegrationSpec
       repo.getRecordSet(okZone.id, deletes(1).recordSet.id).unsafeRunSync() shouldBe None
 
       // make sure the updates are updated
-      repo.getRecordSet(okZone.id, updates(0).recordSet.id).unsafeRunSync().map(_.name) shouldBe Some(updates(0).recordSet.name)
-      repo.getRecordSet(okZone.id, updates(1).recordSet.id).unsafeRunSync().map(_.name) shouldBe Some(updates(1).recordSet.name)
+      repo.getRecordSet(okZone.id, updates(0).recordSet.id).unsafeRunSync().map(_.name) shouldBe
```
nimaeskandary: the `save record with fqdn` is new, these other lines are because intellij wanted to split them up for being too long

modules/mysql/src/main/resources/db/migration/V3.11__RecordSetFqdn.sql
```diff
@@ -0,0 +1,6 @@
+CREATE SCHEMA IF NOT EXISTS ${dbName};
+
+USE ${dbName};
+
+ALTER TABLE recordset ADD COLUMN fqdn VARCHAR(512);
```
nimaeskandary: did max record name plus max zone name as defined by their sql col size

pauljamescleary:
@nimaeskandary did you run a query plan to make sure you were hitting the index?  If so, could you post the results of the query plan here?

nimaeskandary:
@pauljamescleary can def do that, will post today or tomorrow but heading out for now

nimaeskandary:
```
EXPLAIN SELECT data FROM recordset WHERE fqdn IN ("name1.zone1.")

[
  {
    "id": "1",
    "select_type": "SIMPLE",
    "table": "recordset",
    "partitions": null,
    "type": "ref",
    "possible_keys": "fqdn_index",
    "key": "fqdn_index",
    "key_len": "515",
    "ref": "const",
    "rows": "1",
    "filtered": 100,
    "Extra": null
  }
]
```

@pauljamescleary seems to be using the index

modules/dynamodb/src/main/scala/vinyldns/dynamodb/repository/DynamoDBRecordSetRepository.scala
```diff
@@ -234,4 +234,13 @@ class DynamoDBRecordSetRepository private[repository] (
 
       responseFuture.map(resp => resp.asInstanceOf[QueryResponseCount].count)
     }
+
+  def getRecordSetsByFQDNs(names: Set[String]): IO[List[RecordSet]] =
+    monitor("repo.RecordSet.getRecordSetsByFQDNs") {
+      IO(
```
rebstar6: IO.pure i think

nimaeskandary: hm I tried that at first but couldn't get it to compile

rebstar6: jk IO.raiseError()

modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlRecordSetRepositoryIntegrationSpec.scala
```diff
@@ -247,4 +251,37 @@ class MySqlRecordSetRepositoryIntegrationSpec
       results shouldBe empty
     }
   }
+  "getRecordSetsByFQDNs" should {
```
rebstar6: since this is for searching, should also be case insensitive - can you add tests confirming that?

nimaeskandary: Yeah can do, should be already will confirm with tests

modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlRecordSetRepositoryIntegrationSpec.scala
rebstar6: classic intellij

modules/mysql/src/main/resources/db/migration/V3.11__RecordSetFqdn.sql
```diff
@@ -0,0 +1,6 @@
+CREATE SCHEMA IF NOT EXISTS ${dbName};
+
+USE ${dbName};
+
+ALTER TABLE recordset ADD COLUMN fqdn VARCHAR(512) NOT NULL;
```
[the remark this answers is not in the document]
mitruly: I'm alright with keeping this larger if that's the consensus, since we _do_ have validations for the name length.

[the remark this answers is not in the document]
nimaeskandary: @mitruly if that is the max then i am down to bring it down to 255, think its more efficient for the table anyway

PR #497, opened 2019-02-19
Update getUsers in MySQL user repository implementation to honor optional parameters

### LINKED ISSUE #496: MySQL user repository's getUser method doesn't support optional parameters
**Describe the bug**
The MySQL user repo's implementation of `getUsers` doesn't respect `maxItems` or `startFrom` conditional parameters.

mitruly:
Fixes #496.

Changes in this pull request:
- Update SQL query for conditional building based on optional parameters
- Add integration tests

### DIFF
--- a/modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlUserRepository.scala
+++ b/modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlUserRepository.scala
@@ -101,6 +101,7 @@ class MySqlUserRepository(cryptoAlgebra: CryptoAlgebra)
             sb.append(" IN (" + userIds.toList.as("?").mkString(",") + ")")
             startFrom.foreach(start => sb.append(s" AND id > '$start'"))
             sb.append(" ORDER BY id ASC")
+            // Grab one more than the maxItem limit, if provided, to determine whether nextId should be returned
             maxItems.foreach(limit => sb.append(s" LIMIT ${limit + 1}"))
             val query = sb.toString
             SQL(query)
--- a/modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlUserRepository.scala
+++ b/modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlUserRepository.scala
@@ -81,10 +81,6 @@ class MySqlUserRepository(cryptoAlgebra: CryptoAlgebra)
       }
     }
 
-  /*
-   * startFrom and maxItems were originally made to batch the search in the dynamodb implementation,
-   * not needed here with the current sql statement. This function is not exposed by an API route
-   */
   def getUsers(
       userIds: Set[String],
       startFrom: Option[String],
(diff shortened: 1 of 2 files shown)


modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlUserRepository.scala
```diff
@@ -96,15 +96,27 @@ class MySqlUserRepository(cryptoAlgebra: CryptoAlgebra)
           ListUsersResults(List[User](), None)
         else {
           val users = DB.readOnly { implicit s =>
-            val inClause = " IN (" + userIds.toList.as("?").mkString(",") + ")"
-            val query = BASE_GET_USERS + inClause
+            val sb = new StringBuilder
```
mitruly: Opting to use string builder since it's cleaner than using pattern matching to create SQL clauses.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlUserRepository.scala
```diff
@@ -96,15 +96,27 @@ class MySqlUserRepository(cryptoAlgebra: CryptoAlgebra)
...
+            sb.append(BASE_GET_USERS)
+            sb.append(" IN (" + userIds.toList.as("?").mkString(",") + ")")
+            startFrom.foreach(start => sb.append(s" AND id > '$start'"))
+            sb.append(" ORDER BY id ASC")
+            maxItems.foreach(limit => sb.append(s" LIMIT ${limit + 1}"))
```
mitruly: This is doing something similar to DynamoDB where we grab more than the limit and see if we should return a `nextId`.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlUserRepository.scala
```diff
@@ -96,15 +96,27 @@ class MySqlUserRepository(cryptoAlgebra: CryptoAlgebra)
...
               .map(toUser(1))
               .list()
               .apply()
           }
-          ListUsersResults(users, None)
+
+          maxItems match {
+            case Some(limit) =>
+              val returnUsers = users.take(limit)
+              if (users.size == limit + 1) ListUsersResults(returnUsers, Some(returnUsers.last.id))
```
mitruly: Only need to return a `nextId` if there are more than `limit` items.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlUserRepository.scala
```diff
@@ -96,15 +96,28 @@ class MySqlUserRepository(cryptoAlgebra: CryptoAlgebra)
           ListUsersResults(List[User](), None)
         else {
           val users = DB.readOnly { implicit s =>
-            val inClause = " IN (" + userIds.toList.as("?").mkString(",") + ")"
```
rebstar6: the comment above this function seems no longer relevant

PR #498, opened 2019-02-19
remove default repos from portal conf

rebstar6:
allows configuration at deploy time for the enabled repos


nimaeskandary:
@rebstar6 Wondering what the difference is to make I sure I get how these confs are loaded, wouldn't the ref conf be loaded anyway before app conf and have those env vars override anyway?

rebstar6:
right but we cant remove a conf. to enable mysql, for example, you need to remove `user {}` and `user-change {}` configs from ref.conf so that they can be put in the mysql config.

nimaeskandary:
Ah I see, okay cool

PR #503, opened 2019-02-21
Expand input_name column in single_change table

### LINKED ISSUE #462: batch create endpoint doesn't like long record names
**Describe the bug**
Was writing a func test and `testing-listing-batch-summaries-with-owner-group.shared.` failed due to the name being too long. Upon further inspection using a dev environment not on my local, and going through the portal, a 137 character record name on the batch screen does not get past our validations and fails with `Invalid domain name: "long-name-lalalalalalalalalalalalalalalalalalalalalalalalalalalalalalalalalalalalalalalalalalalalalalalalalalalalalalalala.old-vinyldns2.", valid domain names must be letters, numbers, and hyphens, joined by dots, and terminated with a dot.`

This is strange because my `testing-listing-batch-summaries-with-owner-group.shared.` seemed to go through the validations and failed with a 500 on a sql error on the column length for batch name. So could be two separate issues with record name length: one being the validation stops records prematurely when the name gets past a certain point, and the other being our batch mysql repo column size does not support names past a certain point 

**VinylDNS Version**
0.8.0

**To Reproduce**
1. attempt to create a batch record with varying record name lengths

mitruly:
Expand `input_name` column from `VARCHAR(45)` to `VARCHAR(255)`. Fixes #462.

Changes in this pull request:
- Add migration script for `input_name` column expansion
- Add functional test for max label length creation in batch change

### DIFF
--- a/modules/api/functional_test/live_tests/batch/create_batch_change_test.py
+++ b/modules/api/functional_test/live_tests/batch/create_batch_change_test.py
@@ -2928,3 +2928,31 @@ def test_create_batch_update_record_in_shared_zone_for_unassociated_user_in_owne
         if create_rs:
             delete_rs = shared_client.delete_recordset(shared_zone['id'], create_rs['recordSet']['id'], status=202)
             shared_client.wait_until_recordset_change_status(delete_rs, 'Complete')
+
+def test_create_batch_record_with_max_label_length_succeeds(shared_zone_test_context):
+    """
+    Test creating a change in batch with a record containing the max label length (63) should succeed
+    """
+    parent_zone = shared_zone_test_context.parent_zone
+    ok_client = shared_zone_test_context.ok_vinyldns_client
+    rs_name = "a"*63
+    to_delete = []
+
+    batch_change_input = {
+        "changes": [
+            get_change_A_AAAA_json(rs_name + ".parent.com.")
+        ]
+    }
+
+    try:
+        result = ok_client.create_batch_change(batch_change_input, status=202)
+        completed_batch = ok_client.wait_until_batch_change_completed(result)
+
+        record_set_list = [(change['zoneId'], change['recordSetId']) for change in completed_batch['changes']]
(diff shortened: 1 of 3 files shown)


modules/mysql/src/main/resources/db/migration/V3.12__BatchChangeExpandInputName.sql
```diff
@@ -0,0 +1,5 @@
+CREATE SCHEMA IF NOT EXISTS ${dbName};
+
+USE ${dbName};
+
+ALTER TABLE single_change MODIFY input_name VARCHAR(255) NOT NULL;
```
mitruly: Using `255` value and `V3.12` prefix to maintain consistency with record set FQDN PR (https://github.com/vinyldns/vinyldns/pull/495/files#diff-083908ccc83ddaf84ec7879a412c7aa9).

modules/api/functional_test/live_tests/batch/create_batch_change_test.py
```diff
@@ -2928,3 +2928,31 @@ def test_create_batch_update_record_in_shared_zone_for_unassociated_user_in_owne
         if create_rs:
             delete_rs = shared_client.delete_recordset(shared_zone['id'], create_rs['recordSet']['id'], status=202)
             shared_client.wait_until_recordset_change_status(delete_rs, 'Complete')
+
+def test_create_batch_record_with_max_label_length_succeeds(shared_zone_test_context):
```
mitruly: Couldn't think of a better way to quickly test this, as our host names are pretty short. Could arguably add a batch change integration test, but didn't want to overkill.

britneywright: should we add a failure test case as well?

mitruly: I don't think so, as we have tests in `create_batch_change_test.py` that already exhaustively test the validation. I will change this to a WIP since the most appropriate place is an integration test (as it would be repository-agnostic as well).

PR #506, opened 2019-02-22
portal: add AccessLogging filter

### LINKED ISSUE #505: Add access logs to portal
Look at the AccessLoggingFilter at https://www.playframework.com/documentation/2.6.x/ScalaLogging.

We want to log: remote ip address, user-agent, and response status code on every request

britneywright:
Fixes #505.
Changes in this pull request:
- Create AccessLogging Filter, includes:
  - request method
  - URI
  - response status
  - remote-address (user's IP address)
  - user-agent (not parsed, includes operating system and browser information)

example of access log message

Follows the filter pattern here: https://www.playframework.com/documentation/2.6.x/ScalaLogging#Logging-patterns

Note: Filters apply to all requests, including intermediary requests like favicon.ico and API calls. To be more selective we'd need to use a custom action instead. The benefit of the filter is it's called after the Future completes so we're able to include the response status.

### DIFF
--- a/modules/portal/app/actions/VinylDnsAction.scala
+++ b/modules/portal/app/actions/VinylDnsAction.scala
@@ -18,8 +18,9 @@ package actions
 
 import cats.effect.IO
 import controllers.{OidcAuthenticator, VinylDNS}
-import org.slf4j.{Logger}
+import org.slf4j.Logger
 import play.api.mvc.{ActionFunction, Request, Result, Session}
+import play.mvc.Http
 import vinyldns.core.domain.membership.{LockStatus, User}
 
 import scala.concurrent.{ExecutionContext, Future}
--- a/modules/portal/app/actions/VinylDnsAction.scala
+++ b/modules/portal/app/actions/VinylDnsAction.scala
@@ -20,7 +20,6 @@ import cats.effect.IO
 import controllers.{OidcAuthenticator, VinylDNS}
 import org.slf4j.Logger
 import play.api.mvc.{ActionFunction, Request, Result, Session}
-import play.mvc.Http
 import vinyldns.core.domain.membership.{LockStatus, User}
 
 import scala.concurrent.{ExecutionContext, Future}
--- a/modules/portal/app/filters/AccessLoggingFilter.scala
+++ b/modules/portal/app/filters/AccessLoggingFilter.scala
@@ -31,7 +31,7 @@ class AccessLoggingFilter @Inject()(
 
   private val logger = LoggerFactory.getLogger(classOf[AccessLoggingFilter])
 
-  def apply(next: (RequestHeader) => Future[Result])(request: RequestHeader): Future[Result] = {
+  def apply(next: RequestHeader => Future[Result])(request: RequestHeader): Future[Result] = {
     val resultFuture = next(request)
 
     resultFuture.foreach(result => {
--- a/modules/portal/app/filters/AccessLoggingFilter.scala
+++ b/modules/portal/app/filters/AccessLoggingFilter.scala
@@ -36,9 +36,10 @@ class AccessLoggingFilter @Inject()(
 
     resultFuture.foreach(result => {
       if (!request.uri.contains("/public") && !request.uri.contains("/assets")) {
-        val msg = s"method=${request.method} uri=${request.uri} status=${result.header.status} " +
-          s"remote-address=${request.remoteAddress} " +
-          s"user-agent=${request.headers.get(Http.HeaderNames.USER_AGENT).getOrElse("unknown")}"
+        val msg = s"Request: method=${request.method}, path=${request.uri}, " +
+          s"remote_address=${request.remoteAddress}, " +
+          s"user_agent=${request.headers.get(Http.HeaderNames.USER_AGENT).getOrElse("unknown")} " +
+          s"| Response: status_code=${result.header.status} "
         logger.info(msg)
       }
     })
--- a/modules/portal/app/filters/AccessLoggingFilter.scala
+++ b/modules/portal/app/filters/AccessLoggingFilter.scala
@@ -20,12 +20,9 @@ import akka.stream.Materializer
 import javax.inject.Inject
 import org.slf4j.LoggerFactory
 import play.api.mvc.{Filter, RequestHeader, Result}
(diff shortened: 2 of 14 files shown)


modules/portal/app/filters/Filters.scala
```diff
@@ -0,0 +1,50 @@
...
+import org.slf4j.LoggerFactory
+import play.api.mvc.{Filter, RequestHeader, Result}
+import play.filters.csrf.CSRFFilter
+import play.http.DefaultHttpFilters
+import play.mvc.Http
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class Filters @Inject()(access: AccessLoggingFilter, csrf: CSRFFilter)
+    extends DefaultHttpFilters(access, csrf)
```
pauljamescleary: don't think we need csrf right?  Perhaps we just include enabled filters, as CSRF and a bunch of others will come through... https://www.playframework.com/documentation/2.6.x/ScalaHttpFilters#Using-filters

pauljamescleary:
This looks great!  A few comments:

1. For the output, instead of `Some(user-agent-here)`, do something like `userAgent.getOrElse("unknown"), this way we do not need to parse `Some` from the logs
2. For favicon and other useless stuff, we can filter those out at the filter itself.  Like `if path contains x` for example.

modules/portal/app/filters/Filters.scala
britneywright: I thought that should've been the case but when it's not included I get an Execution Exception `RuntimeException: No CSRF token was generated for this request! Is the CSRF filter installed?`

pauljamescleary: Ewww, realistically what we want is perhaps just to have this filter be "standalone", and then do some kind of `play.filters.enabled += my.LoggingFilter`?

Not sure if that makes sense, but we don't want to override the whole logging chain but just add our LoggingFilter to the existing logging chain through config.  

If you look at the source code for the CSRF filter you can see how they did that.

The docs are admittedly not the best

britneywright: Yup that worked!

modules/portal/app/filters/Filters.scala
```diff
@@ -0,0 +1,51 @@
...
+    executionContext: ExecutionContext)
+    extends Filter {
+
+  private val logger = LoggerFactory.getLogger(classOf[AccessLoggingFilter])
+
+  def apply(next: (RequestHeader) => Future[Result])(request: RequestHeader): Future[Result] = {
+    val resultFuture = next(request)
+
+    resultFuture.foreach(result => {
+      if (!request.uri.contains("/public") && !request.uri.contains("/assets")) {
```
britneywright: This skips logging for css, javascript and images that are requested when the portal pages load.

modules/portal/app/filters/Filters.scala
```diff
@@ -0,0 +1,51 @@
...
+import akka.stream.Materializer
+import javax.inject.Inject
+import org.slf4j.LoggerFactory
+import play.api.mvc.{Filter, RequestHeader, Result}
+import play.http.DefaultHttpFilters
+import play.mvc.Http
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class Filters @Inject()(access: AccessLoggingFilter) extends DefaultHttpFilters(access)
```
rebstar6: not sure we need class Filters. for me, if i remove it, rename this file to AccessLoggingFilter, and have both these in conf:
```
play.filters.enabled += "play.filters.csrf.CSRFFilter"
play.filters.enabled += "filters.AccessLoggingFilter"```

it seems to work?

File: modules/portal/app/filters/Filters.scala (hunk 2 above)
Comment (rebstar6) [reply modules/portal/app/filters/Filters.scala]: cool, yeah definitely filter those

File: modules/portal/app/filters/Filters.scala (hunk 3 above)
Comment (britneywright) [reply modules/portal/app/filters/Filters.scala]: we don't need `play.filters.enabled += "play.filters.csrf.CSRFFilter"` because it's CSRF is one of the filters enabled by default.

File: modules/portal/conf/reference.conf [hunk 4]
```diff
@@ -99,7 +99,7 @@ LDAP {
 
 }
 
-play.filters.enabled += "play.filters.csrf.CSRFFilter"
```
rebstar6: keep the csrf filter too, we still want that

modules/portal/app/actions/VinylDnsAction.scala
```diff
@@ -18,8 +18,9 @@ package actions
 
 import cats.effect.IO
 import controllers.{OidcAuthenticator, VinylDNS}
-import org.slf4j.{Logger}
+import org.slf4j.Logger
 import play.api.mvc.{ActionFunction, Request, Result, Session}
+import play.mvc.Http
```
rebstar6: small thing - `play.mvc.Http ` not needed here i think. same for the other file

britneywright: removed them!

modules/portal/app/filters/AccessLoggingFilter.scala
```diff
@@ -0,0 +1,48 @@
...
+    extends Filter {
+
+  private val logger = LoggerFactory.getLogger(classOf[AccessLoggingFilter])
+
+  def apply(next: (RequestHeader) => Future[Result])(request: RequestHeader): Future[Result] = {
+    val resultFuture = next(request)
+
+    resultFuture.foreach(result => {
+      if (!request.uri.contains("/public") && !request.uri.contains("/assets")) {
+        val msg = s"method=${request.method} uri=${request.uri} status=${result.header.status} " +
```
nimaeskandary: I'm pretty good with this, just a nitpick that we could look at `logMessage` in `api.route.VinylDNSService` since it makes a similar log for the api and we could format it to look the same, may make reusing splunk queries easier on us in the future, those are formatted like `Headers: ... | Request: protocol=... , method=... , path =... | Response: status_code=... | Duration: request_duration=...`

britneywright: we have some different needs from logMessage, but i did update it to follow the same pattern, includes Request and Response, comma separated attribute in each, with the pipe between the two groupings.

PR #508, opened 2019-02-25
log rejections

### LINKED ISSUE #281: Add logging layer in all the route handlers
When we receive errors in the route handler, we map them to responses without logging the root issue. We need to add logging so that we know whats actually happening behind the scenes when we get something like a 400

nimaeskandary:
Fixes #281 .

Changes in this pull request:
- previously rejections were un logged, so something like an error from a non existent route or the user not providing a body in a post would be ignored by our logger
- logs method type (get, post, etc) and path for rejections

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/route/VinylDNSService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/route/VinylDNSService.scala
@@ -43,18 +43,25 @@ object VinylDNSService {
     "(?i)(/?zones/)(?:[0-9a-f]-?)+(/recordsets/)(?:[0-9a-f]-?)+(/changes/)(?:[0-9a-f]-?)+(.*)".r
   val TsigKeyRegex: Regex = """(?is)(.*?\"key\"\s*:\s*\")(?:.+?)(\".*)""".r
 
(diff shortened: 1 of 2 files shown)


modules/api/src/main/scala/vinyldns/api/route/VinylDNSService.scala
```diff
@@ -43,18 +43,25 @@ object VinylDNSService {
...
+            s", entity=${res.entity}"
+          } else ""
+          s"Response: status_code=${res.status.intValue}$errorResponse"
+        case None => ""
+      }
     Seq(
       s"Headers: [$requestHeadersNoAuth]",
       s"Request: protocol=${req.protocol.value}, method=${req.method.value}, path=${sanitizePath(req.uri)}",
-      s"Response: status_code=${res.status.intValue}$errorResponse",
+      s"$response",
```
mitruly: Tested this out locally and it works for me when hitting an unhandled route. One question (might be targeted at @pauljamescleary as well) is do we want to include the request entity with rejections as well? With `PUT`/`POST`, it might be better if we can see the payload that is being sent as well.

rebstar6: i think this is fine for now

PR #513, opened 2019-02-28
fix zone connection validation

britneywright:
Changes in this pull request:
- Filter zone record list for apex records only before checking against approved name server list.

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneConnectionValidator.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneConnectionValidator.scala
@@ -45,8 +45,9 @@ class ZoneConnectionValidator(defaultConnection: ZoneConnection)
 
   def loadDns(zone: Zone): IO[ZoneView] = DnsZoneViewLoader(zone).load()
 
-  def runZoneChecks(zoneView: ZoneView): Result[ZoneView] =
-    validateDnsZone(VinylDNSConfig.approvedNameServers, zoneView.recordSetsMap.values.toList)
+  def runZoneChecks(zoneView: ZoneView): Result[ZoneView] = {
+    val apexRecords = zoneView.recordSetsMap.values.filter(_.name == zoneView.zone.name).toList
+    validateDnsZone(VinylDNSConfig.approvedNameServers, apexRecords)
       .map(_ => zoneView)
       .leftMap(
         nel =>
@@ -56,6 +57,7 @@ class ZoneConnectionValidator(defaultConnection: ZoneConnection)
             "Zone could not be loaded due to validation errors."))
       .toEither
       .toResult
+  }
 
   def getDnsConnection(zone: Zone): Result[DnsConnection] =
     Either.catchNonFatal(dnsConnection(zone.connection.getOrElse(defaultConnection))).toResult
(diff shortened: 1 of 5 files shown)


modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneConnectionValidator.scala
```diff
@@ -45,8 +45,9 @@ class ZoneConnectionValidator(defaultConnection: ZoneConnection)
 
   def loadDns(zone: Zone): IO[ZoneView] = DnsZoneViewLoader(zone).load()
 
-  def runZoneChecks(zoneView: ZoneView): Result[ZoneView] =
-    validateDnsZone(VinylDNSConfig.approvedNameServers, zoneView.recordSetsMap.values.toList)
+  def runZoneChecks(zoneView: ZoneView): Result[ZoneView] = {
```
rebstar6: hah, easy. if you throw a non-approved NS in `dummy.hosts` (call it `non-approved-delegation` or something), do all the tests continue to pass.

dont need an explicit func test, but adding that record will validate

britneywright: everything passed locally, we'll see about Travis.

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneConnectionValidator.scala
```diff
@@ -45,8 +45,9 @@ class ZoneConnectionValidator(defaultConnection: ZoneConnection)
...
+    val apexRecords = zoneView.recordSetsMap.values.filter(_.name == zoneView.zone.name).toList
```
mitruly: Digging a bit deeper into what `validateDnsZone` is doing and what we are trying to do, we can clean this up a bit more. `validateDnsZone` is only doing anything with `NS` type records and we are filtering by name; `recordSetsMap` is a `Map` which has a `(recordName, recordType)` tuple as a key, so we can actually do a direct lookup to the NS record.

Something like the following should work (though other tests potentially get updated and unused code cleaned up):
```
  def runZoneChecks(zoneView: ZoneView): Result[ZoneView] = {
    val apexNsCheck = zoneView.recordSetsMap.get((zoneView.zone.name, RecordType.NS)) match {
      case Some(ns) => containsApprovedNameServers(VinylDNSConfig.approvedNameServers, ns)
      case None => "Missing an apex NS record".invalidNel
    }

    apexNsCheck
      .map(_ => zoneView)
      .leftMap(
        nel =>
          ZoneValidationFailed(
            zoneView.zone,
            nel.toList,
            "Zone could not be loaded due to validation errors."))
      .toEither
      .toResult
  }
```

The above is more optimized as there's no need to traverse and filter a collection if we know the exact item we are looking for. Also passed all func tests locally.

mitruly: Unused code would include `validNameServer` and `validateDnsZone` in `ZoneRecordValidations`.

nimaeskandary: would recommend overloading `ZoneRecordValidations.validateDnsZone` to take in the map so it can have tests of its own if going this route, the existing `validateDnsZone` already has tests

rebstar6: if youre going to go this way, id go towards what michael is indicating. at least make `validateDnsZone` instead `apexNsValidation` or something since thats all its doing. and if kept separate, it should return `ValidatedNel[String, Unit]` since the response is unused

PR #511, opened 2019-02-27
portal ptr correspondence

britneywright:
Fixes #510.

Changes in this pull request:
- Add "A+PTR" and "AAAA+PTR" record types in the new batch change form, "A+PTR" is now the default option, and "AAA+PTR" is second.
- On submit, re-format the batch change data to split the "A+PTR" or "AAAA+PTR" entries into two entries, one for the forward record and one for the reverse record.

Screenshot: new record types in record type menu
record type menu

Screenshot: data entered in form for A+PTR record type
data entry

Screenshot: on submission attempt the entry is split into 2 entries, if there are any errors they appear on the relevant entry
screen shot 2019-02-27 at 2 49 01 pm

Screenshot: batch change detail, shows the split forward and reverse records
batch change detail

Some testing suggestions:
Load zones: `1.9.e.f.c.c.7.2.9.6.d.f.ip6.arpa.`, `2.0.192.in-addr.arpa.`, and `ok.`

In a new batch change form:
"Add" "A+PTR" "a.ok." "200" "192.0.2.4"
"Add" "AAAA+PTR" "aaaa.ok." "200" "fd69:27cc:fe91::2"
"DeleteRecordSet" "AAAA+PTR" "a.ok." "200" "192.0.2.4"
"DeleteRecordSet" "A+PTR" "aaaa.ok." "200" "fd69:27cc:fe91::2"

To test batch change record limit you'll need to add 20 entries (or whatever your limit is set at) with at least one of them being an "A+PTR" or "AAAA+PTR", when you confirm the submission it will convert to 21 records and report an error that the limit is exceeded.

PR #518, opened 2019-03-05
add ptr correspondence types to portal batch change docs

britneywright:
doc update in support of #511

### DIFF
--- a/modules/docs/src/main/tut/portal/batch-changes.md
+++ b/modules/docs/src/main/tut/portal/batch-changes.md
@@ -13,7 +13,8 @@ Batch change is an alternative to submitting individual RecordSet changes and pr
 To update an existing record, you must delete the record first and add all expected records within the same batch; a delete and add of the same recordset within a batch change request will be processed as an update.
 
 Batch changes are also susceptible to the following restrictions:
-* Current supported record types for batch change are: **A**, **AAAA**, **CNAME**, **PTR**, **TXT**, and **MX**. Additionally, there are **A+PTR** and **AAAA+PTR** types that, behind the scenes, will be processed as separate A (or AAAA) and PTR changes, and will appear in the completed batch change summary as separate items.
+* Current supported record types for batch change are: **A**, **AAAA**, **CNAME**, **PTR**, **TXT**, and **MX**.
+* Additionally, there are **A+PTR** and **AAAA+PTR** types that will be processed as separate A (or AAAA) and PTR changes in the VinylDNS backend. Deletes for **A+PTR** and **AAAA+PTR** require Input Name and Record Data.
 * Batch change requests must contain at least one change.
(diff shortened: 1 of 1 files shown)


modules/docs/src/main/tut/portal/batch-changes.md
```diff
@@ -13,7 +13,7 @@ Batch change is an alternative to submitting individual RecordSet changes and pr
 To update an existing record, you must delete the record first and add all expected records within the same batch; a delete and add of the same recordset within a batch change request will be processed as an update.
 
 Batch changes are also susceptible to the following restrictions:
-* Current supported record types for batch change are: **A**, **AAAA**, **CNAME**, **PTR**, **TXT**, and **MX**.
+* Current supported record types for batch change are: **A**, **A+PTR**, **AAAA**, **AAAA+PTR**, **CNAME**, **PTR**, **TXT**, and **MX**.
```
rebstar6: actually, split out to new sentence, something like:

Current supported record types for batch change are: **A**, **AAAA**, **CNAME**, **PTR**, **TXT**, and **MX**.
In addition, there are  **A+PTR** and **AAAA+PTR** types; note that behind the scenes, these will be processed as separate A (or AAAA) and PTR changes, and will appear in the final change as separate items.

modules/docs/src/main/tut/portal/batch-changes.md
```diff
@@ -13,7 +13,7 @@ Batch change is an alternative to submitting individual RecordSet changes and pr
 To update an existing record, you must delete the record first and add all expected records within the same batch; a delete and add of the same recordset within a batch change request will be processed as an update.
 
 Batch changes are also susceptible to the following restrictions:
-* Current supported record types for batch change are: **A**, **AAAA**, **CNAME**, **PTR**, **TXT**, and **MX**.
+* Current supported record types for batch change are: **A**, **AAAA**, **CNAME**, **PTR**, **TXT**, and **MX**. Additionally, there are **A+PTR** and **AAAA+PTR** types that, behind the scenes, will be processed as separate A (or AAAA) and PTR changes, and will appear in the completed batch change summary as separate items.
```
mitruly: I feel like this seems a bit verbose. Can we include `A+PTR` and `AAAA+PTR` with the other supported record types and link them to a footnote that talks about how they will be processed?

britneywright: we don't do footnotes and I don't think this is the place to start. I don't have strong opinions on how much or how little we explain it though.

mitruly: If that's the case, then how about something like:
```
Additionally, there are **A+PTR** and **AAAA+PTR** types that will be processed as a separate A (or AAAA) and PTR change, and will also appear separately in the completed batch change summary.
```

PR #519, opened 2019-03-05
add help text to batch change form

britneywright:
part of #346.

Changes in this pull request:
- add tooltips in New Batch Change form
- add link to create a group under the record owner group selection

new batch change form screenshot


mitruly:
I like this, but wonder if there should be something clearer for `A+PTR`/`AAAA+PTR` types (since both types require both, but need to have the FQDN for input name and IP for record data). If not here, there should be an indication in the portal docs.

britneywright:
> I like this, but wonder if there should be something clearer for `A+PTR`/`AAAA+PTR` types (since both types require both, but need to have the FQDN for input name and IP for record data). If not here, there should be an indication in the portal docs.

you mean for deletes? forgot about that, just added a note about it in the portal docs.

PR #520, opened 2019-03-06
mysql user repo in tests

rebstar6:
switching all the configs to be mysql for the user repo. this means our func tests and integration tests will run with that instead of dynamo

### DIFF
--- a/docker/api/docker.conf
+++ b/docker/api/docker.conf
@@ -87,6 +87,9 @@ vinyldns {
       },
       batch-change {
         # no additional settings for now
+      },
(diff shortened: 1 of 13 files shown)


docker/api/docker.conf
```diff
@@ -87,6 +87,9 @@ vinyldns {
       },
       batch-change {
         # no additional settings for now
+      },
```
mitruly: I recently found out that we don't need commas here (possibly since each one is on a new line, similar to dynamo settings), but can leave them here.

PR #517, opened 2019-03-01
[WIP/POC - DO NOT MERGE] Named connections 

rebstar6:
POC for alternate defaults (by name)

PR #522, opened 2019-03-06
BackendID for connection info

rebstar6:
alternate to https://github.com/vinyldns/vinyldns/pull/517

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
+++ b/modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
@@ -68,20 +68,7 @@ object VinylDNSConfig {
 
   lazy val sharedApprovedTypes: Set[RecordType.Value] =
     vinyldnsConfig.as[Option[Set[RecordType.Value]]]("shared-approved-types").getOrElse(Set())
-
-  lazy val dnsBackends: List[DnsBackend] = {
-    if (vinyldnsConfig.hasPath("backends")) {
-      vinyldnsConfig
-        .getConfigList("backends")
-        .asScala
-        .map {
-          pureconfig.loadConfigOrThrow[DnsBackend]
-        }
-        .toList
-        .map(_.encrypted(Crypto.instance))
-    } else List.empty
-  }
-
+  
   lazy val configuredDnsConnections: ConfiguredDnsConnections = {
 
     val defaultZoneConnection = {
--- a/modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
+++ b/modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
@@ -68,7 +68,7 @@ object VinylDNSConfig {
 
   lazy val sharedApprovedTypes: Set[RecordType.Value] =
     vinyldnsConfig.as[Option[Set[RecordType.Value]]]("shared-approved-types").getOrElse(Set())
-  
+
   lazy val configuredDnsConnections: ConfiguredDnsConnections = {
 
     val defaultZoneConnection = {
--- a/modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
+++ b/modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
@@ -115,7 +115,7 @@ object VinylDNSConfig {
       } else List.empty
     }
 
-    new ConfiguredDnsConnections(defaultZoneConnection, defaultTransferConnection, dnsBackends)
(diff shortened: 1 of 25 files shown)


modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
```diff
@@ -69,22 +69,53 @@ object VinylDNSConfig {
...
+    val defaultTransferConnection = {
+      val connectionConfig = VinylDNSConfig.vinyldnsConfig.getConfig("defaultTransferConnection")
+      val name = connectionConfig.getString("name")
+      val keyName = connectionConfig.getString("keyName")
+      val key = connectionConfig.getString("key")
+      val primaryServer = connectionConfig.getString("primaryServer")
+      ZoneConnection(name, keyName, key, primaryServer).encrypted(Crypto.instance)
+    }
+
+    val dnsBackends = {
```
pauljamescleary: why do we parse all of these twice from config?  Can the ConfiguredDnsConnections just reference the other lazy vals already here?

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneConnectionValidator.scala
```diff
@@ -24,24 +24,60 @@ import vinyldns.api.Interfaces._
...
+object ZoneConnectionValidator {
+
+  def getZoneConnection(
+      zone: Zone,
+      configuredDnsConnections: ConfiguredDnsConnections): ZoneConnection = {
+    val backendConnection = zone.backendId
+      .flatMap(getBackend(_, configuredDnsConnections))
+      .map(_.zoneConnection)
+
+    backendConnection
```
pauljamescleary: hmm, so the backend id would override the connection actually on the zone, are we sure about that?

Zone connection on a zone seems more specific.  Idk

modules/core/src/main/scala/vinyldns/core/domain/zone/Zone.scala
```diff
@@ -138,3 +144,19 @@ case class ZoneConnection(name: String, keyName: String, key: String, primarySer
...
+    zoneConnection: ZoneConnection,
+    transferConnection: ZoneConnection) {
+
+  def encrypted(crypto: CryptoAlgebra): DnsBackend = copy(
+    zoneConnection = zoneConnection.encrypted(crypto),
+    transferConnection = transferConnection.encrypted(crypto)
+  )
+}
+
+class ConfiguredDnsConnections(
```
pauljamescleary: Is there a reason this is not a `case class` since it only has data?

modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
rebstar6: it should only be parsing once, i removed them earlier

modules/api/src/test/scala/vinyldns/api/domain/zone/ZoneServiceSpec.scala
```diff
@@ -268,6 +280,12 @@ class ZoneServiceSpec
       val result = rightResultOf(underTest.updateZone(newZone, okAuth).value)
       result shouldBe a[ZoneChange]
     }
+    "return a InvalidRequest if zone has a specified backend ID that is invalid" in {
```
britneywright: return "an" instead of "a"

modules/api/src/test/scala/vinyldns/api/domain/zone/ZoneServiceSpec.scala
```diff
@@ -166,6 +171,13 @@ class ZoneServiceSpec
       val error = leftResultOf(underTest.connectToZone(newZone, okAuth).value)
       error shouldBe a[NotAuthorizedError]
     }
+
+    "return a InvalidRequest if zone has a specified backend ID that is invalid" in {
```
britneywright: return "an" instead of "a"

modules/core/src/main/scala/vinyldns/core/domain/zone/Zone.scala
```diff
@@ -138,3 +144,19 @@ case class ZoneConnection(name: String, keyName: String, key: String, primarySer
   def decrypted(crypto: CryptoAlgebra): ZoneConnection =
     copy(key = crypto.decrypt(key))
 }
+
+final case class DnsBackend(
+    id: String,
+    zoneConnection: ZoneConnection,
```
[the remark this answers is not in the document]
rebstar6: yup

PR #524, opened 2019-03-14
Sync users against LDAP (in the portal)

### LINKED ISSUE #523: Sync users against LDAP
Due to VinylDNS maintaining its own users separate from LDAP, it is necessary to perform a check for currently `Unlocked` users against LDAP and lock any users that are deactivated.

mitruly:
Manage user locking via LDAP syncing. Fixes #523.

Changes in this pull request:
- Create `TaskHandler` to handle process flow for tasks (implemented via FS2 `Stream`)
- Create simple table to handle task contention for distributed portal setup

### DIFF
--- a/modules/portal/app/controllers/LdapAuthenticator.scala
+++ b/modules/portal/app/controllers/LdapAuthenticator.scala
@@ -18,12 +18,13 @@ package controllers
 
 import java.util
 
-import cats.effect.IO
+import cats.effect.{ContextShift, IO}
 import cats.implicits._
 import controllers.LdapAuthenticator.LdapByDomainAuthenticator
 import controllers.VinylDNS.UserDetails
 import javax.naming.directory._
 import javax.naming.Context
+import vinyldns.core.domain.membership.User
 import vinyldns.core.health.HealthCheck._
 
 import scala.collection.JavaConverters._
@@ -178,6 +179,8 @@ final case class LdapServiceException(errorMessage: String)
     extends LdapException(s"Encountered error communicating with LDAP service: $errorMessage")
 final case class InvalidCredentials(username: String)
     extends LdapException(s"Provided credentials were invalid for user [$username].")
+final case class NoLdapSearchDomainsConfigured()
+    extends LdapException("No LDAP search domains were configured so user lookup is impossible.")
 
 /**
   * Top level ldap authenticator that tries authenticating on multiple domains. Authentication is
@@ -190,6 +193,9 @@ class LdapAuthenticator(
     authenticator: LdapByDomainAuthenticator,
     serviceAccount: ServiceAccount)
     extends Authenticator {
+
+  private implicit val cs: ContextShift[IO] =
+    IO.contextShift(scala.concurrent.ExecutionContext.global)
 
   /**
     * Attempts to search for user in specified LDAP domains. Attempts all LDAP domains that are specified in order; in
@@ -226,10 +232,16 @@ class LdapAuthenticator(
     }
 
   def authenticate(username: String, password: String): Either[LdapException, LdapUserDetails] =
-    findUserDetails(searchBase, username, authenticator.authenticate(_, username, password), true)
+    // Need to check domains here due to recursive nature of findUserDetails
+    if (searchBase.isEmpty) Left(NoLdapSearchDomainsConfigured())
+    else
+      findUserDetails(searchBase, username, authenticator.authenticate(_, username, password), true)
 
   def lookup(username: String): Either[LdapException, LdapUserDetails] =
-    findUserDetails(searchBase, username, authenticator.lookup(_, username, serviceAccount), true)
+    // Need to check domains here due to recursive nature of findUserDetails
+    if (searchBase.isEmpty) Left(NoLdapSearchDomainsConfigured())
+    else
+      findUserDetails(searchBase, username, authenticator.lookup(_, username, serviceAccount), true)
 
   def healthCheck(): HealthCheck =
     IO {
@@ -240,11 +252,24 @@ class LdapAuthenticator(
         case _ => ().asRight
       }
     }.asHealthCheck
+
+  // List[User] => List[Either[LdapException, LdapUserDetails]] => List[User]
+  def getUsersNotInLdap(users: List[User]): IO[List[User]] =
+    users
+      .map { u =>
+        IO(lookup(u.userName)).map {
+          case Left(_: UserDoesNotExistException) => Some(u) // Only grab users that do not exist
+          case _ => None
+        }
+      }
+      .parSequence
+      .map(_.flatten)
 }
 
 trait Authenticator {
   def authenticate(username: String, password: String): Either[LdapException, LdapUserDetails]
   def lookup(username: String): Either[LdapException, LdapUserDetails]
+  def getUsersNotInLdap(usernames: List[User]): IO[List[User]]
   def healthCheck(): HealthCheck
 }
 
@@ -282,6 +307,9 @@ class TestAuthenticator(authenticator: Authenticator) extends Authenticator {
     }
 
   def healthCheck(): HealthCheck = authenticator.healthCheck()
+
+  def getUsersNotInLdap(users: List[User]): IO[List[User]] =
+    authenticator.getUsersNotInLdap(users)
 }
 
 case class LdapSearchDomain(organization: String, domainName: String)
--- a/modules/portal/app/controllers/LdapAuthenticator.scala
+++ b/modules/portal/app/controllers/LdapAuthenticator.scala
@@ -93,8 +93,6 @@ object LdapAuthenticator {
   private[controllers] object LdapByDomainAuthenticator {
     def apply(settings: Settings): LdapByDomainAuthenticator =
       new LdapByDomainAuthenticator(settings, createContext(settings))
-
-    def apply(): LdapByDomainAuthenticator = LdapByDomainAuthenticator(Settings)
   }
 
   /**
--- a/modules/portal/app/controllers/UserAccountAccessor.scala
+++ b/modules/portal/app/controllers/UserAccountAccessor.scala
@@ -16,13 +16,16 @@
 
 package controllers
 
-import cats.effect.IO
+import cats.effect.{ContextShift, IO}
+import cats.implicits._
 import javax.inject.{Inject, Singleton}
 import org.joda.time.DateTime
-import vinyldns.core.domain.membership.{User, UserChange, UserChangeRepository, UserRepository}
+import vinyldns.core.domain.membership._
 
 @Singleton
 class UserAccountAccessor @Inject()(users: UserRepository, changes: UserChangeRepository) {
+
+  implicit val cs: ContextShift[IO] = IO.contextShift(scala.concurrent.ExecutionContext.global)
 
   /**
     * Lookup a user in the store. Using identifier as the user id and/or name
@@ -53,4 +56,18 @@ class UserAccountAccessor @Inject()(users: UserRepository, changes: UserChangeRe
 
   def getUserByKey(key: String): IO[Option[User]] =
     users.getUserByAccessKey(key)
+
+  def getAllUsers: IO[List[User]] =
+    users.getAllUsers
+
+  def lockUsers(usersToLock: List[User]): IO[List[User]] =
+    for {
+      lockedUsers <- users.save(usersToLock.map(_.copy(lockStatus = LockStatus.Locked)))
+      _ <- usersToLock
+        .map(
+          u =>
+            changes.save(UserChange
+              .UpdateUser(u.copy(lockStatus = LockStatus.Locked), "system", DateTime.now, u)))
+        .parSequence
+    } yield lockedUsers
 }
--- a/modules/portal/app/controllers/UserAccountAccessor.scala
+++ b/modules/portal/app/controllers/UserAccountAccessor.scala
@@ -60,14 +60,17 @@ class UserAccountAccessor @Inject()(users: UserRepository, changes: UserChangeRe
   def getAllUsers: IO[List[User]] =
     users.getAllUsers
 
-  def lockUsers(usersToLock: List[User]): IO[List[User]] =
+  def lockUsers(usersToLock: List[User]): IO[List[User]] = {
+    val currentTime = DateTime.now
     for {
       lockedUsers <- users.save(usersToLock.map(_.copy(lockStatus = LockStatus.Locked)))
       _ <- usersToLock
-        .map(
-          u =>
-            changes.save(UserChange
-              .UpdateUser(u.copy(lockStatus = LockStatus.Locked), "system", DateTime.now, u)))
+        .zip(lockedUsers)
+        .map {
+          case (oldUser, newUser) =>
+            changes.save(UserChange.UpdateUser(newUser, "system", currentTime, oldUser))
+        }
         .parSequence
     } yield lockedUsers
+  }
 }
--- a/modules/portal/app/modules/VinylDNSModule.scala
+++ b/modules/portal/app/modules/VinylDNSModule.scala
@@ -32,20 +32,28 @@ package modules
  * limitations under the License.
  */
 
+import cats.effect.{ContextShift, IO, Timer}
 import com.google.inject.AbstractModule
 import controllers._
 import controllers.repository.{PortalDataAccessor, PortalDataAccessorProvider}
+import org.slf4j.LoggerFactory
 import play.api.{Configuration, Environment}
+import tasks.TaskHandler
 import vinyldns.core.crypto.CryptoAlgebra
 import vinyldns.core.domain.membership.{UserChangeRepository, UserRepository}
 import vinyldns.core.health.HealthService
 import vinyldns.core.repository.DataStoreLoader
+import vinyldns.core.task.TaskRepository
 
 // $COVERAGE-OFF$
 class VinylDNSModule(environment: Environment, configuration: Configuration)
     extends AbstractModule {
 
   val settings = new Settings(configuration)
+  implicit val t: Timer[IO] = IO.timer(scala.concurrent.ExecutionContext.global)
+  implicit val cs: ContextShift[IO] =
+    IO.contextShift(scala.concurrent.ExecutionContext.Implicits.global)
+  private val logger = LoggerFactory.getLogger("vinyldns.portal.modules.VinylDNSModule")
 
   def configure(): Unit = {
     val startApp = for {
@@ -56,12 +64,21 @@ class VinylDNSModule(environment: Environment, configuration: Configuration)
         .loadAll[PortalDataAccessor](repoConfigs, crypto, PortalDataAccessorProvider)
       auth = authenticator()
       healthService = new HealthService(auth.healthCheck() :: loaderResponse.healthChecks)
+      repositories = loaderResponse.accessor
+      _ <- if (settings.ldapSyncEnabled) {
+        new TaskHandler(repositories.taskRepository)
+          .run(
+            settings,
+            new UserAccountAccessor(repositories.userRepository, repositories.userChangeRepository),
+            auth)
+          .start
+      } else IO.pure(()).start
     } yield {
-      val repositories = loaderResponse.accessor
       bind(classOf[CryptoAlgebra]).toInstance(crypto)
       bind(classOf[Authenticator]).toInstance(auth)
       bind(classOf[UserRepository]).toInstance(repositories.userRepository)
       bind(classOf[UserChangeRepository]).toInstance(repositories.userChangeRepository)
+      bind(classOf[TaskRepository]).toInstance(repositories.taskRepository)
       bind(classOf[HealthService]).toInstance(healthService)
     }
 
--- a/modules/portal/app/modules/VinylDNSModule.scala
+++ b/modules/portal/app/modules/VinylDNSModule.scala
@@ -72,7 +72,7 @@ class VinylDNSModule(environment: Environment, configuration: Configuration)
             new UserAccountAccessor(repositories.userRepository, repositories.userChangeRepository),
             auth)
           .start
-      } else IO.pure(()).start
+      } else IO.unit
     } yield {
       bind(classOf[CryptoAlgebra]).toInstance(crypto)
       bind(classOf[Authenticator]).toInstance(auth)
--- a/modules/portal/app/tasks/TaskHandler.scala
+++ b/modules/portal/app/tasks/TaskHandler.scala
@@ -35,17 +35,14 @@ class TaskHandler @Inject()(taskRepository: TaskRepository) {
   private val logger: Logger = LoggerFactory.getLogger("TaskHandler")
   private implicit val t: Timer[IO] = IO.timer(scala.concurrent.ExecutionContext.global)
 
-  def fetchAndClaimTask(taskName: String, pollingInterval: FiniteDuration): IO[Unit] = {
-    print(s"task name: $taskName; polling interval: $pollingInterval\n\n")
+  def fetchAndClaimTask(taskName: String, pollingInterval: FiniteDuration): IO[Unit] =
     taskRepository.claimTask(taskName, pollingInterval).flatMap {
       case true =>
         IO(logger.info(s"Successfully found and claimed task [$taskName]."))
       case false =>
         val claimError = NoTaskToClaim(taskName)
-        logger.info(claimError.message)
         IO.raiseError(claimError)
     }
-  }
 
   def runSyncTask(
       taskName: String,
@@ -61,7 +58,7 @@ class TaskHandler @Inject()(taskRepository: TaskRepository) {
     } yield ()
 
     syncRun.handleErrorWith { err =>
-      IO(logger.error(s"Encountered task error for [$taskName]", err))
+      IO(logger.info(s"Encountered task error for [$taskName]", err))
     }
   }
 
--- a/modules/portal/app/tasks/TaskHandler.scala
+++ b/modules/portal/app/tasks/TaskHandler.scala
@@ -54,7 +54,7 @@ class TaskHandler @Inject()(taskRepository: TaskRepository) {
       _ <- IO(logger.info(s"Fetched and claimed task [$taskName]."))
       _ <- UserSyncTask
         .syncUsers(userAccountAccessor, authenticator)
-        .handleError(e => IO(logger.error("Encountered error syncing task", e)))
+        .handleErrorWith(e => IO(logger.error("Encountered error syncing task", e)))
       _ <- taskRepository.releaseTask(taskName)
       _ <- IO(logger.info(s"Released task [$taskName]."))
     } yield ()
--- a/modules/portal/app/tasks/TaskHandler.scala
+++ b/modules/portal/app/tasks/TaskHandler.scala
@@ -1,0 +1,91 @@
+/*
+ * Copyright 2018 Comcast Cable Communications Management, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package tasks
+
+import cats.effect.{IO, Timer}
+import cats.implicits._
+import controllers.{Authenticator, Settings, UserAccountAccessor}
+import fs2.Stream
+import javax.inject.{Inject, Singleton}
+import org.slf4j.{Logger, LoggerFactory}
+import vinyldns.core.task._
+
+import scala.concurrent.duration.FiniteDuration
+
+final case class NoTaskToClaim(taskName: String) extends Throwable {
+  def message: String = s"No available task [$taskName] to claim."
+}
+
+@Singleton
+class TaskHandler @Inject()(taskRepository: TaskRepository) {
+  private val logger: Logger = LoggerFactory.getLogger("TaskHandler")
+  private implicit val t: Timer[IO] = IO.timer(scala.concurrent.ExecutionContext.global)
+
+  def fetchAndClaimTask(taskName: String, pollingInterval: FiniteDuration): IO[Unit] =
+    taskRepository.claimTask(taskName, pollingInterval).flatMap {
+      case true =>
+        IO(logger.info(s"Successfully found and claimed task [$taskName]."))
+      case false =>
+        val claimError = NoTaskToClaim(taskName)
+        logger.info(claimError.message)
+        IO.raiseError(claimError)
+    }
+
+  def runSyncTask(
+      taskName: String,
+      pollingInterval: FiniteDuration,
+      userAccountAccessor: UserAccountAccessor,
+      authenticator: Authenticator): IO[Unit] =
+    for {
+      _ <- fetchAndClaimTask(taskName, pollingInterval)
+      _ <- IO(logger.info(s"Fetched and claimed task [$taskName]."))
+      _ <- UserSyncTask.syncUsers(userAccountAccessor, authenticator).handleError(_ => ())
+      _ <- taskRepository.releaseTask(taskName)
+      _ <- IO(logger.info(s"Released task [$taskName]."))
+    } yield ()
+
+  def startTaskStream(
+      settings: Settings,
+      userAccountAccessor: UserAccountAccessor,
+      authenticator: Authenticator): Stream[IO, Unit] = {
+    import UserSyncTask._
+    import settings._
+
+    def runTaskStream(): Stream[IO, Unit] =
+      Stream
+        .awakeEvery[IO](ldapSyncPollingInterval)
+        .evalMap[IO, Unit](
+          _ =>
+            runSyncTask(
+              SYNC_USER_TASK_NAME,
+              ldapSyncPollingInterval,
+              userAccountAccessor,
+              authenticator))
+        .handleErrorWith { error =>
+          logger.error("Encountered error sending sync job", error)
+          runTaskStream()
+        }
+
+    runTaskStream()
+  }
+
+  def run(
+      settings: Settings,
+      userAccountAccessor: UserAccountAccessor,
+      authenticator: Authenticator): IO[Unit] =
+    startTaskStream(settings, userAccountAccessor, authenticator).compile.drain
+}
--- a/modules/portal/app/tasks/TaskHandler.scala
+++ b/modules/portal/app/tasks/TaskHandler.scala
@@ -52,7 +52,9 @@ class TaskHandler @Inject()(taskRepository: TaskRepository) {
     val syncRun = for {
       _ <- fetchAndClaimTask(taskName, pollingInterval)
       _ <- IO(logger.info(s"Fetched and claimed task [$taskName]."))
-      _ <- UserSyncTask.syncUsers(userAccountAccessor, authenticator).handleError(_ => ())
+      _ <- UserSyncTask
+        .syncUsers(userAccountAccessor, authenticator)
+        .handleError(e => IO(logger.error("Encountered error syncing task", e)))
       _ <- taskRepository.releaseTask(taskName)
       _ <- IO(logger.info(s"Released task [$taskName]."))
     } yield ()
--- a/modules/portal/app/tasks/TaskHandler.scala
+++ b/modules/portal/app/tasks/TaskHandler.scala
@@ -49,14 +49,19 @@ class TaskHandler @Inject()(taskRepository: TaskRepository) {
       taskName: String,
       pollingInterval: FiniteDuration,
       userAccountAccessor: UserAccountAccessor,
-      authenticator: Authenticator): IO[Unit] =
-    for {
+      authenticator: Authenticator): IO[Unit] = {
+    val syncRun = for {
       _ <- fetchAndClaimTask(taskName, pollingInterval)
       _ <- IO(logger.info(s"Fetched and claimed task [$taskName]."))
       _ <- UserSyncTask.syncUsers(userAccountAccessor, authenticator).handleError(_ => ())
       _ <- taskRepository.releaseTask(taskName)
       _ <- IO(logger.info(s"Released task [$taskName]."))
     } yield ()
+
+    syncRun.handleErrorWith { err =>
+      IO(logger.error(s"Encountered task error for [$taskName]", err))
+    }
+  }
 
   def startTaskStream(
       settings: Settings,
@@ -65,22 +70,15 @@ class TaskHandler @Inject()(taskRepository: TaskRepository) {
     import UserSyncTask._
     import settings._
 
-    def runTaskStream(): Stream[IO, Unit] =
-      Stream
-        .awakeEvery[IO](ldapSyncPollingInterval)
-        .evalMap[IO, Unit](
-          _ =>
-            runSyncTask(
-              SYNC_USER_TASK_NAME,
-              ldapSyncPollingInterval,
-              userAccountAccessor,
-              authenticator))
-        .handleErrorWith { error =>
-          logger.error("Encountered error sending sync job", error)
-          runTaskStream()
-        }
-
-    runTaskStream()
+    Stream
+      .awakeEvery[IO](ldapSyncPollingInterval)
+      .evalMap[IO, Unit](
+        _ =>
+          runSyncTask(
+            SYNC_USER_TASK_NAME,
+            ldapSyncPollingInterval,
+            userAccountAccessor,
+            authenticator))
   }
 
   def run(
--- a/modules/portal/app/tasks/TaskHandler.scala
+++ b/modules/portal/app/tasks/TaskHandler.scala
@@ -35,7 +35,8 @@ class TaskHandler @Inject()(taskRepository: TaskRepository) {
   private val logger: Logger = LoggerFactory.getLogger("TaskHandler")
   private implicit val t: Timer[IO] = IO.timer(scala.concurrent.ExecutionContext.global)
 
-  def fetchAndClaimTask(taskName: String, pollingInterval: FiniteDuration): IO[Unit] =
+  def fetchAndClaimTask(taskName: String, pollingInterval: FiniteDuration): IO[Unit] = {
+    print(s"task name: $taskName; polling interval: $pollingInterval\n\n")
     taskRepository.claimTask(taskName, pollingInterval).flatMap {
       case true =>
         IO(logger.info(s"Successfully found and claimed task [$taskName]."))
@@ -44,6 +45,7 @@ class TaskHandler @Inject()(taskRepository: TaskRepository) {
         logger.info(claimError.message)
         IO.raiseError(claimError)
     }
+  }
 
   def runSyncTask(
       taskName: String,
--- a/modules/portal/app/tasks/UserSyncTask.scala
+++ b/modules/portal/app/tasks/UserSyncTask.scala
@@ -23,19 +23,19 @@ import vinyldns.core.domain.membership.LockStatus
 
 object UserSyncTask {
   val SYNC_USER_TASK_NAME = "user_sync"
-  private val log: Logger = LoggerFactory.getLogger("Task")
+  private val log: Logger = LoggerFactory.getLogger("UserSyncTask")
 
   def syncUsers(
       userAccountAccessor: UserAccountAccessor,
       authenticator: Authenticator): IO[Unit] = {
-    log.info("Initiating user sync")
+    log.error("Initiating user sync")
     for {
       allUsers <- userAccountAccessor.getAllUsers
       activeUsers = allUsers.filter(u => u.lockStatus != LockStatus.Locked && !u.isTest)
       nonActiveUsers <- authenticator.getUsersNotInLdap(activeUsers)
       lockedUsers <- userAccountAccessor.lockUsers(nonActiveUsers)
       _ <- IO(
-        log.info(
+        log.error(
           s"Users locked: ${lockedUsers.map(_.userName)}. Total locked: ${lockedUsers.size}"))
     } yield ()
   }
--- a/modules/portal/app/tasks/UserSyncTask.scala
+++ b/modules/portal/app/tasks/UserSyncTask.scala
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2018 Comcast Cable Communications Management, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package tasks
+
+import cats.effect.IO
+import controllers.{Authenticator, UserAccountAccessor}
+import org.slf4j.{Logger, LoggerFactory}
+import vinyldns.core.domain.membership.LockStatus
+
+object UserSyncTask {
+  val SYNC_USER_TASK_NAME = "user_sync"
+  private val log: Logger = LoggerFactory.getLogger("Task")
+
+  def syncUsers(
+      userAccountAccessor: UserAccountAccessor,
+      authenticator: Authenticator): IO[Unit] = {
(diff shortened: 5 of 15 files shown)


project/Dependencies.scala
```diff
@@ -101,6 +101,7 @@ object Dependencies {
     "com.typesafe.play"         %% "play-json"                      % "2.6.10",
     "com.amazonaws"             %  "aws-java-sdk-core"              % awsV withSources(),
     "com.amazonaws"             %  "aws-java-sdk-dynamodb"          % awsV withSources(),
+    "com.enragedginger"         %% "akka-quartz-scheduler"          % "1.8.0-akka-2.5.x",
```
mitruly: Chosen over Akka's `Scheduler` as `Scheduler` is more of a "deferrer" (ie. run a task `t` time from now).

Source: https://github.com/enragedginger/akka-quartz-scheduler

modules/portal/conf/reference.conf
```diff
@@ -54,6 +54,22 @@ portal.test_login = true
 portal.test_login = ${?TEST_LOGIN}
 
 
+akka {
+  quartz {
+    defaultTimezone = "UTC"
+    schedules {
+      EveryMinute {
+        description = "Cron job that fires off every minute"
+        expression = "0 * * ? * *"
```
mitruly: Expressions can be verified at: https://www.freeformatter.com/cron-expression-generator-quartz.html

modules/portal/app/tasks/TaskMaster.scala
```diff
@@ -0,0 +1,49 @@
...
+    extends Actor {
+  import tasks.TaskMaster._
+
+  def receive: Actor.Receive = {
+    case SyncUsers =>
+      log.info("Initiating user sync")
+
+      val syncUsers = for {
+        allUsers <- userAccountAccessor.getAllUsers
+        nonLockedUsers = allUsers.filter(_.lockStatus != LockStatus.Locked)
```
mitruly: Only need to check LDAP against currently `Unlocked` users.

modules/portal/app/modules/VinylDNSModule.scala
```diff
@@ -63,6 +67,15 @@ class VinylDNSModule(environment: Environment, configuration: Configuration)
       bind(classOf[UserRepository]).toInstance(repositories.userRepository)
       bind(classOf[UserChangeRepository]).toInstance(repositories.userChangeRepository)
       bind(classOf[HealthService]).toInstance(healthService)
+
+      val actorSystem = ActorSystem()
```
mitruly: The actor will handle asynchronous one-off tasks, specifically the LDAP sync in this case.

nimaeskandary: Does this mean every portal node will be running this job?

mitruly: @nimaeskandary That's a good point - we should have a discussion as to what the best approach is here.

mitruly: Looks like we could potentially use a cluster singleton (source: https://doc.akka.io/docs/akka/snapshot/cluster-singleton.html); I'll update the description.

pauljamescleary: @mitruly best to avoid cluster-singleton as that implies state (uses akka clustering which has been a disaster for us in openstack).

Coordination across nodes is difficult without some kind of membership.

For a stateless solution:

1. We can live with every portal running the job.  Odds are there will be very little cleanup.  Plus, since we follow put semantics it doesn't matter who writes last
2. We can use SQS, but then we have the whole BYOQ but for the portal which is annoying
3. We can use the database to create a very simple lock table.  Once per hour, every node will attempt to acquire the lock.  Whoever gets it, run the job, and then free the lock.  If the lock is more than 2 hours old, then assume the last node crashed and the job is available.  This is kind of how the MySQL Message Queue works, and should be ok

modules/portal/app/tasks/TaskMaster.scala
```diff
@@ -0,0 +1,49 @@
...
+        nonexistentUsers = authenticator.findNonexistentUsers(nonLockedUsers)
+        lockedUsers <- userAccountAccessor.lockUsers(nonexistentUsers)
+        _ <- IO(log.info(s"Users locked: ${lockedUsers.map(_.userName)}"))
+      } yield ()
+
+      syncUsers.unsafeRunSync()
```
pauljamescleary: What happens on failure here?

project/Dependencies.scala
pauljamescleary: If we really want `cron` could also have used fs2 - https://github.com/fthomas/fs2-cron

My FS2 comment is one of we should probably use the same mechanism for this kind of work, and it appears as though FS2 would do the job here.

modules/portal/app/modules/VinylDNSModule.scala
mitruly: The detailed information is greatly appreciated, @pauljamescleary; I think I want to go with the third approach since that sounds the cleanest and works for now. I think realistically if we wanted to implement a more ideal solution where a task actor can run any arbitrary tasks for the portal, the second approach sounds like the way to go.

project/Dependencies.scala
pauljamescleary: also explore if you really need `cron` behavior (i.e. deterministic on the hour).  Could use core `fs2` and just `awakeEvery`

modules/portal/app/tasks/TaskMaster.scala
mitruly: If the task executing on the actor encounters an exception, the default supervisor strategy is to restart the actor:
https://github.com/akka/akka/blob/ef3a19b5d0e970501cde4d78168cce8b940a56ef/akka-actor/src/main/scala/akka/actor/FaultHandling.scala#L153-L175

modules/portal/app/tasks/TaskMaster.scala
```diff
@@ -0,0 +1,49 @@
...
+    extends Actor {
+  import tasks.TaskMaster._
+
+  def receive: Actor.Receive = {
+    case SyncUsers =>
+      log.info("Initiating user sync")
+
+      val syncUsers = for {
+        allUsers <- userAccountAccessor.getAllUsers
+        nonLockedUsers = allUsers.filter(u => u.lockStatus != LockStatus.Locked && !u.isTest)
```
mitruly: Might make sense to exclude super users as well; would like some thoughts.

pauljamescleary: I think you can eliminate the actor all together now that you have FS2 in there.

modules/portal/app/modules/VinylDNSModule.scala
```diff
@@ -46,6 +52,25 @@ class VinylDNSModule(environment: Environment, configuration: Configuration)
...
+  implicit val cs: ContextShift[IO] =
+    IO.contextShift(scala.concurrent.ExecutionContext.Implicits.global)
+  private val logger = LoggerFactory.getLogger("vinyldns.portal.modules.VinylDNSModule")
+
+  def startTaskStream(actor: ActorRef): Stream[IO, Unit] = {
+    def syncLdapTaskStream(): Stream[IO, Unit] =
+      Stream
+        .awakeEvery[IO](settings.ldapSyncPollingInterval)
+        .evalMap[IO, Unit] { _ =>
+          IO(actor ! SyncUsers)
```
pauljamescleary: I would:

1. lose the actor
2. put each step of the processing in here instead of in the Actor

modules/portal/conf/reference.conf
```diff
@@ -116,6 +116,11 @@ LDAP {
     providerUrl = ${?LDAP_PROVIDER_URL}
   }
 
+  sync {
```
pauljamescleary: this should be named `user-sync`

Also, I wonder what the polling-interval should be here.  We certainly do not want to allow users to specify something like 1.millis, that would be pretty terrible.

Should we force the `user-sync.interval` to be hard-coded in terms of hours?  Like ...

```
user-sync {
  enabled = true | false
  hours-interval = number
}
```

modules/portal/app/tasks/TaskMaster.scala
```diff
@@ -0,0 +1,107 @@
...
+
+  def claimTask(name: String)(implicit s: DBSession): Unit =
+    CLAIM_TASK.bind(name).update().apply()
+
+  def unclaimTask(name: String)(implicit s: DBSession): Unit =
+    UNCLAIM_TASK.bind(name).update().apply()
+
+  def fetchAndClaimTask(name: String): IO[Either[Throwable, Unit]] =
+    IO {
+      if (DB.readOnly(implicit s => fetchUnclaimedTask(SYNC_TASK_NAME))) {
```
pauljamescleary: I think we want to do all steps in the same transaction here...

```scala
Db.localTx { implicit s => 
  if (fetchUnclaimedTask) { 
    claimTask
  }
}
```

modules/portal/app/tasks/TaskMaster.scala
```diff
@@ -0,0 +1,107 @@
...
+      .apply()
+      .nonEmpty
+
+  def claimTask(name: String)(implicit s: DBSession): Unit =
+    CLAIM_TASK.bind(name).update().apply()
+
+  def unclaimTask(name: String)(implicit s: DBSession): Unit =
+    UNCLAIM_TASK.bind(name).update().apply()
+
+  def fetchAndClaimTask(name: String): IO[Either[Throwable, Unit]] =
```
pauljamescleary: I don't think we would want to fail the way you have it here.

Just because we cannot claim the task, doesn't mean that we should fail.  We may not be able to claim the task because someone else has it locked.  That is normal and should be expected.

modules/portal/app/tasks/TaskMaster.scala
```diff
@@ -0,0 +1,107 @@
...
+      | WHERE name = ?
+    """.stripMargin
+
+  private val FETCH_UNCLAIMED =
+    sql"""
+      |SELECT *
+      |  FROM task
+      | WHERE (in_flight = 0
+      |    OR updated IS NULL
+      |    OR updated < DATE_SUB(NOW(), INTERVAL 48 HOUR))
```
pauljamescleary: So, the calculation says if the updated timestamp hasn't been updated within 48 hours?  Is that a good "expiration" timeout.  

Perhaps some multiple of the polling interval instead?  

Even if we have multiple people processing the task at the same time, it isn't necessarily terrible as all data carries idempotency / put semantics.  If we lock the user twice, doesn't really matter.

modules/portal/app/controllers/LdapAuthenticator.scala
```diff
@@ -212,11 +213,20 @@ class LdapAuthenticator(
         case _ => ().asRight
       }
     }.asHealthCheck
+
+  def findNonexistentUsers(users: List[User]): List[User] =
```
rebstar6: why is this called `findNonexistentUsers`? its just looking up a list of users

rebstar6: oh weird, its flipped.

maybe `getUsersNotInLdap`? i dunno, hard to word this one

modules/portal/app/controllers/LdapAuthenticator.scala
```diff
@@ -212,11 +213,20 @@ class LdapAuthenticator(
...
+    users.flatMap { user =>
+      lookup(user.userName) match {
+        case Success(_) => None
+        case _ => Some(user)
```
rebstar6: is every failure a user not found? im wary of LDAP connectivity issues causing us to lock all of our users at once

modules/portal/app/tasks/Task.scala
```diff
@@ -0,0 +1,54 @@
...
+  val taskName = "user_sync"
+  private val log: Logger = LoggerFactory.getLogger(classOf[SyncUserTask])
+
+  def performTask: IO[Either[Throwable, Unit]] = {
+    log.info("Initiating user sync")
+
+    val syncUsers = for {
+      allUsers <- EitherT.right(userAccountAccessor.getAllUsers)
+      nonLockedUsers = allUsers.filter(u => u.lockStatus != LockStatus.Locked && !u.isTest)
+      nonexistentUsers = authenticator.findNonexistentUsers(nonLockedUsers)
```
rebstar6: log somewhere in this flow "LDAP sync locked X users" or something. Maybe also log the usernames that were locked.

modules/portal/app/controllers/LdapAuthenticator.scala
mitruly: Good point, @rebstar6; may be very worthwhile to explore how the different bad responses from LDAP (eg. timeout, etc) are converted back in the lookup methods.

modules/dynamodb/src/main/scala/vinyldns/dynamodb/repository/DynamoDBUserRepository.scala
```diff
@@ -270,6 +270,14 @@ class DynamoDBUserRepository private[repository] (
     }
   }
 
+  def getAllUsers: IO[List[User]] =
```
pauljamescleary: I am ok with postponing this to a separate PR, but please add a card for discussion on what we want to do with this.  If we are no longer going to support the dynamodb user repo, we should make a card to remove it.  If we are going to continue to support it, we should have a card to implement these functions.

I think we should move everything to MySQL except the audit tables, and kill with fire the corresponding dynamodb tables

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlUserRepository.scala
```diff
@@ -165,6 +176,30 @@ class MySqlUserRepository(cryptoAlgebra: CryptoAlgebra)
...
+          Seq(
+            'id -> u.id,
+            'userName -> u.userName,
+            'accessKey -> u.accessKey,
+            'data -> toPB(u.withEncryptedSecretKey(cryptoAlgebra)).toByteArray
+          )
+        }
+
+        updates.grouped(1000).foreach { group =>
+          DB.localTx { implicit s =>
```
pauljamescleary: The transaction should not be on each group.  Otherwise, you could fail out of the `save` method, and have partial results saved to the database.  `save` should be atomic.

Put the `DB.localTx` outside of the `updates.grouped`

modules/portal/app/controllers/LdapAuthenticator.scala
pauljamescleary: This is pretty important.

1. If the user is found (positive result), then can skip
2. If the suer is not found, then can skip
3. If we get any other error, then we should fail the entire operation (to be retried in the future)

pauljamescleary:
```scala
  def lookup(username: String): Try[LdapUserDetails] =
    findUserDetails(searchBase, username, authenticator.lookup(_, username, serviceAccount))
```
 if we do not care about the user info, just do

```scala
def ldapUserExists(user: User): IO[Boolean] = ???
```

This way if the

modules/portal/app/modules/VinylDNSModule.scala
```diff
@@ -63,6 +70,20 @@ class VinylDNSModule(environment: Environment, configuration: Configuration)
...
+      if (settings.ldapSyncEnabled) {
+        val syncUserTask = SyncUserTask(
+          new UserAccountAccessor(repositories.userRepository, repositories.userChangeRepository),
+          auth,
+          settings.ldapSyncPollingInterval)
+        TaskHandler.run(syncUserTask).start.unsafeRunAsync {
+          case Right(_) =>
+            logger.error("VINYLDNS PORTAL TASK STREAM STARTED SUCCESSFULLY!!")
+          case Left(startupFailure) =>
+            logger.error(s"VINYLDNS PORTAL TASK STREAM UNABLE TO START $startupFailure")
```
pauljamescleary: for `logger.error`, put the `startupFailure as a separate parameter (I think that is right)...

```
logger.error(s"VINYLDNS PORTAL TASK STREAM UNABLE TO START", startupFailure)
```

modules/portal/app/tasks/Task.scala
```diff
@@ -0,0 +1,54 @@
...
+
+import scala.concurrent.duration.FiniteDuration
+
+trait Task {
+  val taskName: String
+  val pollingInterval: FiniteDuration
+  def performTask: IO[Either[Throwable, Unit]]
+}
+
+case class SyncUserTask(
```
pauljamescleary: Uncomfortable with case classes having behavior like this.

Also, wondering why we need a `Task` to be a `trait`, or why we need a case class

Could we...

```scala
object Task {
  def syncUsers(pollingInterval: FiniteDuration): () => IO[Either[Throwable, Unit]] = () => {
    .. do sync stuff here
  }
}

File: modules/portal/app/tasks/Task.scala [hunk 19]
```diff
@@ -0,0 +1,54 @@
...
+    extends Task {
+  val taskName = "user_sync"
+  private val log: Logger = LoggerFactory.getLogger(classOf[SyncUserTask])
+
+  def performTask: IO[Either[Throwable, Unit]] = {
+    log.info("Initiating user sync")
+
+    val syncUsers = for {
+      allUsers <- EitherT.right(userAccountAccessor.getAllUsers)
+      nonLockedUsers = allUsers.filter(u => u.lockStatus != LockStatus.Locked && !u.isTest)
```
pauljamescleary: hmm, so we are loading and _then_ filtering.  Would it make more sense to filter in the query.... `userAccountAccessor.getActiveUsers` or something like that?

modules/portal/app/tasks/TaskHandler.scala
```diff
@@ -0,0 +1,118 @@
...
+      pollingInterval: FiniteDuration): IO[Either[Throwable, Unit]] =
+    IO {
+      db.localTx { implicit s =>
+        if (fetchUnclaimedTask(name, pollingInterval)) {
+          claimTask(name)
+          logger.info(s"Successfully found and claimed task [$name].")
+          Right(())
+        } else {
+          logger.info(s"No task [$name] to claim.")
+          Left(new Exception(s"No tasks available matching name [$name]."))
```
pauljamescleary: I wouldn't use Exception here, as we know what the failure is.  Perhaps a domain specific error like "NoTaskToClaim".

`IO` will embody any unexpected exceptions (i.e. talking to the database).  `IO[Either[NoTaskToClaim, Unit]]` would or something would be more expressive

modules/portal/app/tasks/TaskHandler.scala
```diff
@@ -0,0 +1,118 @@
...
+import cats.effect.{IO, Timer}
+import fs2.Stream
+import org.slf4j.{Logger, LoggerFactory}
+import scalikejdbc._
+
+import scala.concurrent.duration.FiniteDuration
+
+object TaskHandler extends TaskHandler
+
+trait TaskHandler {
```
pauljamescleary: I don't think we need a separate `trait` and `object` do we?  Can we have just the `object` here?

modules/portal/app/tasks/Task.scala
```diff
@@ -0,0 +1,54 @@
...
+
+case class SyncUserTask(
+    userAccountAccessor: UserAccountAccessor,
+    authenticator: Authenticator,
+    pollingInterval: FiniteDuration)
+    extends Task {
+  val taskName = "user_sync"
+  private val log: Logger = LoggerFactory.getLogger(classOf[SyncUserTask])
+
+  def performTask: IO[Either[Throwable, Unit]] = {
```
pauljamescleary: Curious as to why you chose `IO[Either` as opposed to `IO[Unit]` as the result type.

modules/portal/app/tasks/TaskHandler.scala
mitruly: `trait` is used here so we can overload for testing; otherwise we'd need to convert the test to an integration test with MySQL.

modules/portal/app/tasks/Task.scala
mitruly: This is a good call; initially was trying to reduce the impact of this PR on our MySQL user repo implementation.

modules/portal/app/tasks/Task.scala
mitruly: I broke this out from `TaskHandler`, which uses `IO[Either` for fail-fast; you're right in that I can convert this to `IO[Unit]` (which I'll do).

modules/portal/app/tasks/Task.scala
mitruly: Oh, nvm; we don't have `LockStatus` or `isTest` data exposed outside of the blob :(

modules/portal/app/controllers/UserAccountAccessor.scala
```diff
@@ -53,4 +53,19 @@ class UserAccountAccessor @Inject()(users: UserRepository, changes: UserChangeRe
...
     users.getUserByAccessKey(key)
+
+  def getAllUsers: IO[List[User]] =
+    users.getAllUsers
+
+  def lockUsers(usersToLock: List[User]): IO[List[User]] =
+    for {
+      lockedUsers <- users.save(usersToLock.map(_.copy(lockStatus = LockStatus.Locked)))
+      _ <- IO {
+        usersToLock.foreach { u =>
```
pauljamescleary: `usersToLock.foreach` is side effecting.  Would be preferable to map into a `List[IO]` and then do a `sequence` on that to get an `IO[List]`

modules/portal/app/tasks/Task.scala
pauljamescleary: 😱

nimaeskandary:
If not already should prob be logging a list of a all users that were locked in a sync so it can easily be fed into a reversion script

modules/portal/app/tasks/TaskHandler.scala
```diff
@@ -0,0 +1,137 @@
...
+      |UPDATE task
+      |   SET in_flight = 0, updated = NOW()
+      | WHERE name = ?
+    """.stripMargin
+
+  private val FETCH_UNCLAIMED =
+    sql"""
+      |SELECT *
+      |  FROM task
+      | WHERE (in_flight = 0
```
nimaeskandary: Can you leave comments in the code on this logic? Wondering why `OR updated IS NULL` is necessary

modules/portal/app/modules/VinylDNSModule.scala
```diff
@@ -63,6 +70,22 @@ class VinylDNSModule(environment: Environment, configuration: Configuration)
...
+            settings,
+            new UserAccountAccessor(repositories.userRepository, repositories.userChangeRepository),
+            auth)
+          .start
+          .unsafeRunAsync {
+            case Right(_) =>
+              logger.error("VINYLDNS PORTAL TASK STREAM STARTED SUCCESSFULLY!!")
+            case Left(startupFailure) =>
+              logger.error(s"VINYLDNS PORTAL TASK STREAM UNABLE TO START", startupFailure)
+              startupFailure.printStackTrace()
```
pauljamescleary: If `ldapSynEnabled` and there is an error, we should fail and not start the application, right?  What happens in the `startupFailure` case?

modules/portal/app/tasks/Task.scala
```diff
@@ -0,0 +1,42 @@
...
+  val SYNC_USER_TASK_NAME = "user_sync"
+  private val log: Logger = LoggerFactory.getLogger("Task")
+
+  def syncUsers(
+      userAccountAccessor: UserAccountAccessor,
+      authenticator: Authenticator): IO[Unit] = {
+    log.info("Initiating user sync")
+    for {
+      allUsers <- userAccountAccessor.getAllUsers
+      nonLockedUsers = allUsers.filter(u => u.lockStatus != LockStatus.Locked && !u.isTest)
```
pauljamescleary: `nonLockedUsers` is not a good name now that we are also considering `isTest`, perhaps `activeUsers` or something like that?

modules/portal/app/tasks/TaskHandler.scala
pauljamescleary: This really needs to be broken out.  You are combining the logic of running tasks with a SQL implementation.  Create a `TaskRepository` trait, and provide that to the stream processing steps.

Then, you can simply unit test the `TaskRunner` by passing a mocked `TaskRepository`.

modules/portal/app/tasks/TaskHandler.scala
```diff
@@ -0,0 +1,137 @@
...
+    import Task._
+
+    def runTaskStream(): Stream[IO, Either[Throwable, Unit]] =
+      Stream
+        .awakeEvery[IO](ldapSyncPollingInterval)
+        .evalMap[IO, Either[Throwable, Unit]] { _ =>
+          val doTask = for {
+            _ <- EitherT(fetchAndClaimTask(SYNC_USER_TASK_NAME, ldapSyncPollingInterval))
+            _ <- EitherT.right(IO(logger.info("Fetched and claimed task.")))
+            _ <- EitherT(authenticator.healthCheck()) // Short circuit if LDAP health check fails
```
pauljamescleary: We do not need `authenticator.heatlhCheck()` as `syncUsers` can fail for many reasons

modules/portal/app/tasks/TaskHandler.scala
```diff
@@ -0,0 +1,137 @@
...
+            _ <- EitherT.right(syncUsers(userAccountAccessor, authenticator))
```
pauljamescleary: Why `EitherT.right` here?  As the IO can fail, you should always lift instead.

Given all of your stuff, I don't see why you need `EitherT` at all, as an `IO` is also a monad that can short circuit.

To better test this, as it is the core logic, separate this for comprehension into a separate function so you can easily unit test.  Then you can do a simple `evalMap(functionX)` here instead.

modules/portal/app/tasks/TaskHandler.scala
```diff
@@ -0,0 +1,137 @@
...
+    sql"""
+      |SELECT *
+      |  FROM task
+      | WHERE (in_flight = 0
+      |    OR updated IS NULL
+      |    OR updated < DATE_SUB(NOW(), INTERVAL {expiration_hours} HOUR))
+      |   AND name = {task_name}
+    """.stripMargin
+
+  // $COVERAGE-OFF$
```
pauljamescleary: COVERAGE-OFF?  I think we can leave coverage on and add unit and integration tests.

modules/portal/app/tasks/TaskHandler.scala
mitruly: I'll add a comment - `OR updated IS NULL` is used because the seeded data (in the `V3.XX__Tasks.sql` file) does not set `updated` and `updated` is only set after the task is claimed/unclaimed.

modules/portal/app/tasks/TaskHandler.scala
mitruly: Whoops, should have lifted `syncUsers` rather than `EitherT.right` 👍

The original thought process was that if the task fetch and claim failed, we wouldn't want to continue with syncing (since the `IO` would succeed, but return us a bad response). Could achieve the same behavior with `IO` and `IO.raiseError`, though not sure which is preferable, @pauljamescleary.

modules/portal/app/tasks/TaskHandler.scala
mitruly: Yeah, I wanted to initially avoid that if possible @pauljamescleary since it's only used by the portal, but I finally buckled down and did it so integration tests should be good.

modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlTaskRepositoryIntegrationSpec.scala
```diff
@@ -0,0 +1,136 @@
...
+import scalikejdbc.DB
+import vinyldns.mysql.TestMySqlInstance
+
+import scala.concurrent.duration._
+import scalikejdbc._
+
+class MySqlTaskRepositoryIntegrationSpec extends WordSpec with BeforeAndAfterAll with BeforeAndAfterEach with Matchers {
+  private val repo = TestMySqlInstance.taskRepository.asInstanceOf[MySqlTaskRepository]
+  private val taskName = "task_name"
+  private val insert_statement =
```
pauljamescleary: in scala, do not use lower_snake_case for variable names.  For constants, it is ok to use UPPER_SNAKE_CASE

modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlUserRepositoryIntegrationSpec.scala
```diff
@@ -66,6 +66,13 @@ class MySqlUserRepositoryIntegrationSpec
     super.afterAll()
   }
 
+  "MySqlUserRepository.getAllUsers" should {
+    "return all users" in {
+      repo.getAllUsers.unsafeRunSync() should contain theSameElementsAs
+        users ++ List(case_insensitive_user1, case_insensitive_user2, case_insensitive_user3)
```
pauljamescleary: oh noes, looks like there is precedence with lower_snake_case elsewhere?  Can we clean these up?  We should have a style check for this in our build.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlTaskRepository.scala
```diff
@@ -0,0 +1,80 @@
...
+    sql"""
+         |UPDATE task
+         |   SET in_flight = 1, updated = {current_time}
+         | WHERE name = {name}
+    """.stripMargin
+
+  private val UNCLAIM_TASK =
+    sql"""
+         |UPDATE task
+         |   SET in_flight = 0, updated = {current_time}
```
pauljamescleary: use `{currentTime}` i.e. camel case for parameter names in SQL statements to distinguish from actual database column names.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlTaskRepository.scala
```diff
@@ -0,0 +1,80 @@
...
+       |SELECT *
+       |  FROM task
+       | WHERE (in_flight = 0
+       |    OR updated IS NULL
+       |    OR updated < {updated_time_comparison})
+       |   AND name = {task_name}
+    """.stripMargin
+
+  def unclaimedTaskExists(name: String, pollingInterval: FiniteDuration): IO[Boolean] = {
+    val pollingExpirationHours = pollingInterval.toHours * 2
```
pauljamescleary: `pollingExpirationHours` should be inside the `IO`

modules/portal/app/controllers/LdapAuthenticator.scala
```diff
@@ -212,11 +213,27 @@ class LdapAuthenticator(
         case _ => ().asRight
       }
     }.asHealthCheck
+
+  def getUsersNotInLdap(users: List[User]): IO[List[User]] = IO {
+    users.flatMap { user =>
+      val lookupResult = lookup(user.userName)
+      lookupResult match {
+        case Success(_) => None
+        // If lookup fails, check against service account to confirm that LDAP service is still running
```
pauljamescleary: If LDAP has failed, we should exit the lookup in an exit status, the lock should be released, and we should try again later.  Hopefully we have a health check of some sort, or we can raise an alert, when the lookup process fails.

I think this specific check here is not needed based on my comment above.

modules/portal/app/controllers/LdapAuthenticator.scala
```diff
@@ -212,11 +213,27 @@ class LdapAuthenticator(
         case _ => ().asRight
       }
     }.asHealthCheck
+
+  def getUsersNotInLdap(users: List[User]): IO[List[User]] = IO {
```
pauljamescleary: So, `lookup(username)` results in a `Try[LdapUserDetails]`.  This is a little ugly as the `Try` could be because LDAP connectivity is down, or it could fail because of a `UserDoesNotExistException`.

What we need to do is specifically hunt down all `lookup` calls that fail for ONLY `UserDoesNotExistException`.

If there is any other error, we should pre-emptively exit the loop and release the task.  We need code someplace that will somehow raise an error if we get anything other than a `UserDoesNotExistException`

Finally, we should be able to achieve parallelism here

modules/portal/app/controllers/LdapAuthenticator.scala
```diff
@@ -179,6 +179,10 @@ final case class LdapServiceException(errorMessage: String)
     extends LdapException(s"Encountered error communicating with LDAP service: $errorMessage")
 final case class InvalidCredentials(username: String)
     extends LdapException(s"Provided credentials were invalid for user [$username].")
+final case class NoLdapSearchDomainsConfigured()
```
mitruly: New error defined for when no LDAP search domains are defined via LDAP config.

modules/portal/app/controllers/LdapAuthenticator.scala
```diff
@@ -242,20 +255,16 @@ class LdapAuthenticator(
...
-        case _ =>
-          lookup(serviceAccount.name) match {
-            case Right(_) => Some(user)
-            case _ => None
-          }
-      }
+  // List[User] => List[Either[LdapException, LdapUserDetails]] => List[User]
+  def getUsersNotInLdap(users: List[User]): IO[List[User]] =
+    IO {
+      users.par.filter { u =>
```
mitruly: `.par` here generates a parallel collection, so assuming that the app will handle this in parallel under the hood.

modules/portal/app/controllers/LdapAuthenticator.scala
```diff
@@ -242,20 +255,16 @@ class LdapAuthenticator(
...
+        lookup(u.userName) match {
+          case Left(_: UserDoesNotExistException) => true // Only grab users that do not exist
```
mitruly: With the improvements to LDAP searching, we can simply collect all of the `UserDoesNotExistException` and ignore everything else. If there is an empty list of users, there simply will be no syncing performed and the task will be released.

modules/portal/app/tasks/TaskHandler.scala
```diff
@@ -55,7 +55,7 @@ class TaskHandler @Inject()(taskRepository: TaskRepository) {
     for {
       _ <- fetchAndClaimTask(taskName, pollingInterval)
       _ <- IO(logger.info(s"Fetched and claimed task [$taskName]."))
-      _ <- Task.syncUsers(userAccountAccessor, authenticator)
+      _ <- Task.syncUsers(userAccountAccessor, authenticator).handleError(_ => ())
```
mitruly: If there's an issue with `.syncUsers`, don't short circuit because we want to release the task. We will still short circuit if `fetchAndClaimTask` fails.

modules/dynamodb/src/main/scala/vinyldns/dynamodb/repository/DynamoDBUserRepository.scala
```diff
@@ -270,6 +270,14 @@ class DynamoDBUserRepository private[repository] (
...
+    monitor("repo.User.getAllUsers") {
+      IO.raiseError(
+        UnsupportedDynamoDBRepoFunction(
+          "getAllUsers is not supported by VinylDNS DynamoDB UserRepository")
```
rebstar6: because of this, remove `user` from implementedRepositories in DynamoDBDataStoreProvider

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlUserRepository.scala
```diff
@@ -118,6 +117,18 @@ class MySqlUserRepository(cryptoAlgebra: CryptoAlgebra)
       }
     }
 
+  def getAllUsers: IO[List[User]] =
+    monitor("repo.User.getAllUsers") {
+      IO {
+        DB.readOnly { implicit s =>
+          SQL(BASE_GET_USERS)
```
rebstar6: dont we want this to only get unlocked users? we have no need to check the locked ones id think

rebstar6: ugh locked is in the protobuf, i see it now. thats a PITA, i wish we just didnt protobuf users...

modules/portal/app/controllers/LdapAuthenticator.scala
rebstar6: interesting. I wonder how the IO here fits in here too given the lookup is the IO piece here

pauljamescleary: Shouldn't this be  be `users.map(toIOThing).parSequence`?

modules/portal/app/tasks/Task.scala
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+package tasks
+
+import cats.effect.IO
+import controllers.{Authenticator, UserAccountAccessor}
+import org.slf4j.{Logger, LoggerFactory}
+import vinyldns.core.domain.membership.LockStatus
+
+object Task {
```
rebstar6: this feels the wrong name for this. UserSyncTask maybe?

rebstar6: maybe its ok, it just feels so unintuitive to me

modules/portal/app/controllers/LdapAuthenticator.scala
rebstar6: how about:

```
      users.map { u => 
        IO(lookup(u.userName)).map {
          case Left(_: UserDoesNotExistException) => Some(u) // Only grab users that do not exist
          case _ => None
        }
      }.parSequence.map(_.flatten)
```


i went down a rabbit hole trying to make the called function an IO. which we can do, but this is less impacting

modules/portal/app/controllers/LdapAuthenticator.scala
```diff
@@ -178,6 +179,10 @@ final case class LdapServiceException(errorMessage: String)
     extends LdapException(s"Encountered error communicating with LDAP service: $errorMessage")
 final case class InvalidCredentials(username: String)
     extends LdapException(s"Provided credentials were invalid for user [$username].")
+final case class NoLdapSearchDomainsConfigured()
+    extends LdapException(
+      "No LDAP search domains were configured so " +
+        "user lookup is impossible.")
```
rebstar6: not critical, but this probably doesnt need multi lining

modules/portal/app/tasks/TaskHandler.scala
```diff
@@ -0,0 +1,88 @@
...
+    unclaimedTaskExists(taskName, pollingInterval).flatMap {
+      case true =>
+        claimTask(taskName).as(logger.info(s"Successfully found and claimed task [$taskName]."))
+      case false =>
+        val claimError = NoTaskToClaim(taskName)
+        logger.info(claimError.message)
+        IO.raiseError(claimError)
+    }
+
+  def doTask(
```
rebstar6: this goes along with my other comment on naming. TaskHandler.doTask gives no indication of what it's doing. but it's hard coded to be a sync task. i get that its meant to be generic, but the naming is confusing given the hard coding. i would call it runSyncTask (or something)

modules/portal/app/modules/VinylDNSModule.scala
rebstar6: yeah this startup is a little wacky. this is more consistent with what we do in the api for the command handler:

```
  def configure(): Unit = {
    val startApp = for {
      cryptoConf <- settings.cryptoConfig
      crypto <- CryptoAlgebra.load(cryptoConf)
      repoConfigs <- settings.dataStoreConfigs
      loaderResponse <- DataStoreLoader
        .loadAll[PortalDataAccessor](repoConfigs, crypto, PortalDataAccessorProvider)
      auth = authenticator()
      healthService = new HealthService(auth.healthCheck() :: loaderResponse.healthChecks)
      repositories = loaderResponse.accessor
      _ <- if (settings.ldapSyncEnabled) {
        new TaskHandler(repositories.taskRepository)
          .run(
            settings,
            new UserAccountAccessor(repositories.userRepository, repositories.userChangeRepository),
            auth)
          .start
      } else IO.pure(()).start
    } yield {
      bind(classOf[CryptoAlgebra]).toInstance(crypto)
      bind(classOf[Authenticator]).toInstance(auth)
      bind(classOf[UserRepository]).toInstance(repositories.userRepository)
      bind(classOf[UserChangeRepository]).toInstance(repositories.userChangeRepository)
      bind(classOf[TaskRepository]).toInstance(repositories.taskRepository)
      bind(classOf[HealthService]).toInstance(healthService)
    }

    startApp.unsafeRunSync()
  }
```

the caveat to ^ is im not totally sure what it's doing in the non-enabled case. Starting up an IO.pure(()) should hopefully not matter?

modules/portal/app/tasks/TaskHandler.scala
rebstar6: id at least log the error

modules/portal/app/tasks/Task.scala
```diff
@@ -0,0 +1,42 @@
...
+  val SYNC_USER_TASK_NAME = "user_sync"
+  private val log: Logger = LoggerFactory.getLogger("Task")
+
+  def syncUsers(
```
rebstar6: this PR is super long, making it tough to review. If it's not a crazy task, it may be worth splitting here. You could have `object Task` and the syncUsers function without hooking it into the system at all (no task handler, vinyldnsmodule, etc).

Seems we may be far enough down the line to avoid that, but keep it in mind for next time, 900 lines is about 500 too many for a normal PR

modules/portal/app/tasks/TaskHandler.scala
```diff
@@ -0,0 +1,91 @@
...
+      Stream
+        .awakeEvery[IO](ldapSyncPollingInterval)
+        .evalMap[IO, Unit](
+          _ =>
+            runSyncTask(
+              SYNC_USER_TASK_NAME,
+              ldapSyncPollingInterval,
+              userAccountAccessor,
+              authenticator))
+        .handleErrorWith { error =>
```
rebstar6: not 100% on if the recursion here makes sense. instead, i would just do the `evalMap` here and change runSyncTask to catch the errors:

```  def runSyncTask(
      taskName: String,
      pollingInterval: FiniteDuration,
      authenticator: Authenticator): IO[Unit] = {
    val syncRun = for {
      _ <- fetchAndClaimTask(taskName, pollingInterval)
      _ <- IO(logger.info(s"Fetched and claimed task [$taskName]."))
      _ <- UserSyncTask.syncUsers(userAccountAccessor, authenticator).handleError(_ => ())
      _ <- taskRepository.releaseTask(taskName)
      _ <- IO(logger.info(s"Released task [$taskName]."))
    } yield ()

    syncRun.handleErrorWith { err =>
      IO {
        logger.error(s"Encountered task error for ${taskName}", err)
      }
    }
  }
```

modules/portal/app/controllers/UserAccountAccessor.scala
```diff
@@ -53,4 +56,18 @@ class UserAccountAccessor @Inject()(users: UserRepository, changes: UserChangeRe
...
   def getUserByKey(key: String): IO[Option[User]] =
     users.getUserByAccessKey(key)
+
+  def getAllUsers: IO[List[User]] =
+    users.getAllUsers
+
+  def lockUsers(usersToLock: List[User]): IO[List[User]] =
+    for {
+      lockedUsers <- users.save(usersToLock.map(_.copy(lockStatus = LockStatus.Locked)))
+      _ <- usersToLock
```
rebstar6: cant you use `lockedUsers` here instead of` usersToLock` to avoid the double status copy?

rebstar6: nvm i see it --

```      
lockedUsers <- users.save(updated)
_ <- (usersToLock zip lockedUsers).map {
          case (oldUser, newUser) =>
            changes.save(UserChange
              .UpdateUser(newUser, "system", DateTime.now, oldUser))
      }.parSequence
```

modules/portal/app/modules/VinylDNSModule.scala
```diff
@@ -56,12 +64,21 @@ class VinylDNSModule(environment: Environment, configuration: Configuration)
...
       healthService = new HealthService(auth.healthCheck() :: loaderResponse.healthChecks)
+      repositories = loaderResponse.accessor
+      _ <- if (settings.ldapSyncEnabled) {
+        new TaskHandler(repositories.taskRepository)
+          .run(
+            settings,
+            new UserAccountAccessor(repositories.userRepository, repositories.userChangeRepository),
+            auth)
+          .start
+      } else IO.pure(()).start
```
rebstar6: talked to MP about this, seems good other than that the `IO.pure(()).start` can just be `IO.unit`

modules/portal/app/tasks/TaskHandler.scala
```diff
@@ -0,0 +1,91 @@
...
+  def message: String = s"No available task [$taskName] to claim."
+}
+
+@Singleton
+class TaskHandler @Inject()(taskRepository: TaskRepository) {
+  private val logger: Logger = LoggerFactory.getLogger("TaskHandler")
+  private implicit val t: Timer[IO] = IO.timer(scala.concurrent.ExecutionContext.global)
+
+  def fetchAndClaimTask(taskName: String, pollingInterval: FiniteDuration): IO[Unit] = {
+    print(s"task name: $taskName; polling interval: $pollingInterval\n\n")
```
rebstar6: log dont print :P

mitruly: Heh, think this was a debug statement that I meant to remove

modules/portal/app/tasks/TaskHandler.scala
```diff
@@ -0,0 +1,91 @@
...
+    taskRepository.claimTask(taskName, pollingInterval).flatMap {
+      case true =>
+        IO(logger.info(s"Successfully found and claimed task [$taskName]."))
+      case false =>
+        val claimError = NoTaskToClaim(taskName)
+        logger.info(claimError.message)
```
rebstar6: dont need this log, its printed in the `syncRun.handleErrorWith` now

modules/portal/app/tasks/TaskHandler.scala
```diff
@@ -0,0 +1,91 @@
...
+
+  def runSyncTask(
+      taskName: String,
+      pollingInterval: FiniteDuration,
+      userAccountAccessor: UserAccountAccessor,
+      authenticator: Authenticator): IO[Unit] = {
+    val syncRun = for {
+      _ <- fetchAndClaimTask(taskName, pollingInterval)
+      _ <- IO(logger.info(s"Fetched and claimed task [$taskName]."))
+      _ <- UserSyncTask.syncUsers(userAccountAccessor, authenticator).handleError(_ => ())
```
rebstar6: log this error as logger.error. this one we would want to know

modules/portal/app/tasks/TaskHandler.scala
```diff
@@ -0,0 +1,91 @@
...
+      _ <- taskRepository.releaseTask(taskName)
+      _ <- IO(logger.info(s"Released task [$taskName]."))
+    } yield ()
+
+    syncRun.handleErrorWith { err =>
+      IO(logger.error(s"Encountered task error for [$taskName]", err))
```
rebstar6: this can downgrade to logger.info, it mostly means the sync timeout did expire

modules/portal/app/tasks/UserSyncTask.scala
```diff
@@ -0,0 +1,42 @@
...
+package tasks
+
+import cats.effect.IO
+import controllers.{Authenticator, UserAccountAccessor}
+import org.slf4j.{Logger, LoggerFactory}
+import vinyldns.core.domain.membership.LockStatus
+
+object UserSyncTask {
+  val SYNC_USER_TASK_NAME = "user_sync"
+  private val log: Logger = LoggerFactory.getLogger("Task")
```
rebstar6: name this `UserSyncTask`

modules/portal/app/tasks/TaskHandler.scala
rebstar6: this one still stands
UserSyncTask.syncUsers(userAccountAccessor, authenticator).handleErrorWith(err => IO(logger.error(e))) or something since this is just being swallowed

mitruly: Whoops; added the log.

modules/portal/app/tasks/TaskHandler.scala
```diff
@@ -0,0 +1,90 @@
...
+final case class NoTaskToClaim(taskName: String) extends Throwable {
+  def message: String = s"No available task [$taskName] to claim."
+}
+
+@Singleton
+class TaskHandler @Inject()(taskRepository: TaskRepository) {
+  private val logger: Logger = LoggerFactory.getLogger("TaskHandler")
+  private implicit val t: Timer[IO] = IO.timer(scala.concurrent.ExecutionContext.global)
+
+  def fetchAndClaimTask(taskName: String, pollingInterval: FiniteDuration): IO[Unit] =
```
pauljamescleary: weird that this is `fetchAndClaimTask`, `claimTask` would be better.

This is also strange that it raises an error if no task was claimed, which is an expected result.  `IO.raiseError` should be only when there is something we cannot do anything about.  Here, we can say "ah, nothing to do, another node in the cluster must be running the task".  `EitherT[IO, NoTaskToClaim, Unit]` would be better (or `IO[Either[NoTaskToClaim, Unit]]`

britneywright:
closing, picked up and merged in #718

PR #525, opened 2019-03-15
handle an error during oidc login

nimaeskandary:
There was recently an issue where oidc login failed halfway due to 

```
application - LoginId [some-id] complete with error: ErrorResponse(500,Couldn't retrieve remote JWK set: Read timed out)application - LoginId 
```

Afterwards, it appears our `setOidcSession` function in our `VinylDNS` controller was stuck in a retry loop trying to login with the now stale token. This pr redirects to an error page if the oidc login fails in `setOidcSession`


nimaeskandary:
Screen Shot 2019-03-15 at 11 09 12 AM

modules/portal/app/controllers/VinylDNS.scala
```diff
@@ -153,7 +153,7 @@ class VinylDNS @Inject()(
           Redirect("/index").withSession(ID_TOKEN -> token.toString)
         case Left(err) =>
           Logger.error(s"LoginId [$loginId] complete with error: $err")
-          Status(err.code)(err.message)
+          Redirect(s"/login-error/$loginId").withNewSession
```
britneywright: Do we need the loginId here?

nimaeskandary: Put that in there for one final log on the /login-error/ route

britneywright:
do we need a log messages right before and after the redirect to the login-error route? Can we just have one or the other? Also weird that one says login complete then the next says login failed.

modules/portal/app/views/loginError.scala.html
```diff
@@ -0,0 +1,53 @@
+@(alertMessage: Option[String] = None)(implicit requestHeader: RequestHeader, customLinks: models.CustomLinks, meta: models.Meta)
```
mitruly: Are we missing an `@import helper.CSRF` at the top?

Also seems like the only difference between this file and `noAccess.scala.html` is the `h4` text; can we make this a generic file and pass in the `h4` text as a parameter?

modules/api/src/main/scala/vinyldns/api/engine/ZoneSyncHandler.scala
```diff
@@ -115,10 +115,12 @@ object ZoneSyncHandler extends DnsConversions with Monitored {
                 zone.copy(status = ZoneStatus.Active, latestSync = Some(DateTime.now)),
                 status = ZoneChangeStatus.Synced))
           } else {
-            val dottedRecords = changes.filter { chg =>
-              chg.recordSet.name != zone.name && chg.recordSet.name.contains(".") && chg.recordSet.typ != RecordType.SRV
-            }.map(_.recordSet.name)
-             .mkString(", ")
+            val dottedRecords = changes
```
nimaeskandary: second pr so far thats trying to reformat this file after compile.. figure someone has to get it in at this point but could remove

modules/portal/app/controllers/VinylDNS.scala
```diff
@@ -152,8 +154,13 @@ class VinylDNS @Inject()(
             s"LoginId [$loginId] complete: --LOGIN-- user [${user.userName}] logged in with id ${user.id}")
           Redirect("/index").withSession(ID_TOKEN -> token.toString)
         case Left(err) =>
-          Logger.error(s"LoginId [$loginId] complete with error: $err")
-          Status(err.code)(err.message)
+          Logger.error(s"LoginId [$loginId] failed with error: $err")
+          InternalServerError(
```
rebstar6: is there any session clearing that needs to happen here?

if not, im good with this PR

nimaeskandary: yeah definitely needs that, removed by accident with last commit

PR #527, opened 2019-03-18
have portal wait for mysql

nimaeskandary:
Now that the user repo default is mysql the portal cannot try to start up before mysql or it wail fail to connect and stop the service


build.sbt
```diff
@@ -471,7 +481,6 @@ lazy val sonatypePublishStage = Seq[ReleaseStep](
 lazy val finalReleaseStage = Seq[ReleaseStep] (
   releaseStepCommand("project root"), // use version.sbt file from root
   commitReleaseVersion,
-  tagRelease,
```
nimaeskandary: this tag was really annoying when doing snapshots, might as well just save tags for when do github releases in my opinion

build.sbt
```diff
@@ -157,6 +157,16 @@ lazy val portalDockerSettings = Seq(
 
   // this is the default version, can be overridden
   bashScriptExtraDefines += s"""addJava "-Dvinyldns.base-version=${(version in ThisBuild).value}"""",
+
+  // wait for mysql
+  bashScriptExtraDefines += "(cd /opt/docker/ && ./wait-for-dependencies.sh && cd -)",
```
nimaeskandary: see https://www.playframework.com/documentation/2.6.x/Deploying, the contents of the `dist` directory in the portal module gets plopped in the root of the docker image

build.sbt
```diff
@@ -157,6 +157,16 @@ lazy val portalDockerSettings = Seq(
...
+  dockerCommands ++= Seq(
```
nimaeskandary: same as api

build.sbt
mitruly: More of a quick discussion point: I know this is mimicking what the API does, but is it possible to add a command alias for `wait-for-dependencies` (similar to `preparePortal`)? Feels like it would achieve the same thing without doing the directory switch and switch back in a sub-shell.

modules/portal/dist/wait-for-dependencies.sh
```diff
@@ -0,0 +1,32 @@
...
+
+# the mysql address, default to a local docker setup
+MYSQL_ADDRESS=${MYSQL_ADDRESS:-vinyldns-mysql}
+MYSQL_PORT=${MYSQL_PORT:-3306}
+echo "Waiting for MYSQL to be ready on $MYSQL_ADDRESS:$MYSQL_PORT"
+DATA=""
+RETRY=30
+while [ $RETRY -gt 0 ]
+do
+    DATA=$(nc -vzw1 $MYSQL_ADDRESS $MYSQL_PORT)
```
mitruly: Nitpick, but can we encapsulate the bash variables in double quotes? (eg. `"$MYSQL_ADDRESS $MYSQL_PORT"`)

build.sbt
nimaeskandary: Good point but we wouldn't have sbt installed in the image

PR #529, opened 2019-03-18
update docs on test users

nimaeskandary:
info on test zones and users

### DIFF
--- a/modules/docs/src/main/tut/api/create-zone.md
+++ b/modules/docs/src/main/tut/api/create-zone.md
@@ -70,7 +70,6 @@ id            | string        |  the id of the change.  This is not the id of th
   "created": "2016-12-28T18:00:32Z",
   "changeType": "Create",
   "userId": "vinyl",
-  "id": "dd449c27-bed5-4cd5-95e6-4c54fb20d930",
-  "ia
+  "id": "dd449c27-bed5-4cd5-95e6-4c54fb20d930"
 }
 ```
--- a/modules/docs/src/main/tut/api/create-zone.md
+++ b/modules/docs/src/main/tut/api/create-zone.md
@@ -70,6 +70,7 @@ id            | string        |  the id of the change.  This is not the id of th
   "created": "2016-12-28T18:00:32Z",
   "changeType": "Create",
   "userId": "vinyl",
-  "id": "dd449c27-bed5-4cd5-95e6-4c54fb20d930"
+  "id": "dd449c27-bed5-4cd5-95e6-4c54fb20d930",
+  "ia
 }
 ```
--- a/modules/docs/src/main/tut/api/membership-model.md
+++ b/modules/docs/src/main/tut/api/membership-model.md
@@ -69,7 +69,7 @@ the group, deleting users from the group, toggling other users' admin statuses (
 field         | type        | description |
  ------------ | :---------- | :---------- |
 userName      | string      | This should be the AD username of the user |
-firstName     | string      | First name if the user |
+firstName     | string      | First name of the user |
 lastName      | string      | Last name of the user |
 email         | string      | Email address of the user |
 created       | date-time   | The timestamp GMT when the user was created |
@@ -95,4 +95,4 @@ To get your access and secret keys, log into the VinylDNS portal and then with t
 
 The `isTest` attribute restricts a user's access to Zones. The [Zone Model](../api/zone-model.md) also has its
 own `isTest` attribute that can be set on Zone create. If a user has `isTest` set to `true`,
-then they can **only** manage Zones that also have `isTest` set to `true`
+then they can **only** manage Zones that also have `isTest` set to `true`.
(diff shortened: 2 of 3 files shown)


britneywright:
this PR should update all docs related to `isTest`, so docs need to be updated related to zones as well.

modules/docs/src/main/tut/api/membership-model.md
```diff
@@ -68,12 +68,13 @@ the group, deleting users from the group, toggling other users' admin statuses (
 
 field         | type        | description |
  ------------ | :---------- | :---------- |
-userName      | string      | This should be the NT id of the user |
+userName      | string      | This should be the AD username of the user |
 firstName     | string      | First name fo the user |
```
mitruly: I know it's not tied, but could fix this typo as well (`fo` → `of`)?

modules/docs/src/main/tut/api/create-zone.md
```diff
@@ -70,6 +70,7 @@ id            | string        |  the id of the change.  This is not the id of th
   "created": "2016-12-28T18:00:32Z",
   "changeType": "Create",
   "userId": "vinyl",
-  "id": "dd449c27-bed5-4cd5-95e6-4c54fb20d930"
+  "id": "dd449c27-bed5-4cd5-95e6-4c54fb20d930",
+  "ia
```
britneywright: typos

modules/docs/src/main/tut/api/zone-model.md
```diff
@@ -35,6 +35,7 @@ shared        | boolean     | An indicator that the zone is shared with anyone.
 acl           | ZoneACL     | The access control rules governing the zone.  See the [Zone ACL Rule Attributes](#zone-acl-rule-attr) for more information
 id            | string      | The unique identifier for this zone
 latestSync    | date-time   | The last date and time the zone was synced
+isTest        | boolean     | Defaults to **false**. See [Test Users](../api/membership-model.md#test-users)
```
britneywright: zone should have it's own note. that any user can modify a test zone and maybe that it should be set to true if used in testing.

modules/docs/src/main/tut/api/create-zone.md
mitruly: If we're updating create zone migration as well, should mention that we don't currently allow zone updates to change the `isTest` flag with the workaround to this being abandoning the zone and reconnecting with the desired `isTest` value.

nimaeskandary: I mention zone create in the info section and not zone update which may or may not be enough, but since this is a really minimally used feature I feel like too much on the matter could get outdated and forgot easily, @britneywright do you have thoughts on this?

britneywright: don't think we need to go into that

modules/docs/src/main/tut/api/membership-model.md
```diff
@@ -68,12 +68,13 @@ the group, deleting users from the group, toggling other users' admin statuses (
 
 field         | type        | description |
  ------------ | :---------- | :---------- |
-userName      | string      | This should be the NT id of the user |
-firstName     | string      | First name fo the user |
+userName      | string      | This should be the AD username of the user |
+firstName     | string      | First name if the user |
```
mitruly: `if` → `of`

modules/docs/src/main/tut/api/membership-model.md
```diff
@@ -85,6 +86,13 @@ To get your access and secret keys, log into the VinylDNS portal and then with t
...
+  "id": "1764183c-5e75-4ae6-8833-503cd5f4dcb3",
+  "isTest": false
 }
 ```
+
+### TEST USERS <a id="test-users"></a>
+
+The `isTest` attribute restricts a user's access to Zones. The [Zone Model](../api/zone-model.md) also has its
+own `isTest` attribute that can be set on Zone create. If a user has `isTest` set to `true`,
+then they can **only** manage Zones that also have `isTest` set to `true`
```
mitruly: Missing terminating period.

modules/docs/src/main/tut/api/zone-model.md
```diff
@@ -249,3 +250,9 @@ Will give Read permissions to PTR Record Sets 1000:1000:1000:1000:0000:0000:0000
 ### SHARED ZONES <a id="shared-zones"></a>
 
 Shared zones allow for a more open management of records in VinylDNS. Zone administrators can assign ownership of records to groups. Any user in VinylDNS can claim existing unowned records in shared zones, as well as create records in those zones. Once a record is owned, only users in the record owner group, the zone administrators and those with relevant ACL rules can modify or delete the record. The [batch change API endpoint](../api/create-batchchange) and [batch change area of the portal](../portal/batch-changes) are where users can create new records in shared zones, modify records they own, or claim unowned records. If a zone's shared state changes to false the record ownership access is no longer applicable.
+
+### isTest <a id="is-test"></a>
+
+Zones have a boolean attribute `isTest` that is used so [Test Users](../api/membership-model.md#test-users) can
+update the Zone and its DNS records. `isTest` only impacts test users and access for normal users, groups, acl rules, etc
+is unchanged. This attribute can be set on Zone create
```
britneywright: This attribute can only be set on Zone create.

PR #532, opened 2019-03-18
Add ownerGroupId to mysql recordset table

varsharun:
Fixes #426 .

Changes in this pull request:
- Added a new alter query under migrations section
- Updated MySqlRecordSetRepository to include the optional new field `owner-group-id` 
- Integration tests

### DIFF
--- a/modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlRecordSetRepositoryIntegrationSpec.scala
+++ b/modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlRecordSetRepositoryIntegrationSpec.scala
@@ -272,6 +272,40 @@ class MySqlRecordSetRepositoryIntegrationSpec
       repo.getRecordSet(okZone.id, inserts(1).recordSet.id).unsafeRunSync().map(_.name) shouldBe
         Some(inserts(1).recordSet.name)
     }
+    "works when inserting owner-group-id" in {
+
+      val testRecord = makeTestAddChange(ds, okZone).recordSet
+      val addChange = makeTestAddChange(testRecord.copy(ownerGroupId = Some("someOwner")))
+
+      val dbCalls = for {
+        _ <- repo.apply(ChangeSet(addChange))
+        get <- repo.getRecordSet(testRecord.zoneId, testRecord.id)
+      } yield get
+
+      val get = dbCalls.unsafeRunSync()
+      get shouldBe Some(addChange.recordSet)
+    }
+
+    "works when updating owner-group-id" in {
+
+      val existing = makeTestAddChange(ds, okZone).recordSet
+      val addChange = makeTestAddChange(existing)
+
+      val updateChange = makePendingTestUpdateChange(addChange.recordSet,
+        addChange.recordSet.copy(name = "updated-name", ownerGroupId = Some("someOwner")))
+
+      val dbCalls = for {
+        _ <- repo.apply(ChangeSet(addChange))
+        get <- repo.getRecordSet(addChange.zoneId, addChange.id)
+        _ <- repo.apply(ChangeSet(updateChange))
+        finalGet <- repo.getRecordSet(addChange.zoneId, addChange.id)
+      } yield (get, finalGet)
+
+      val (get, finalGet) = dbCalls.unsafeRunSync()
+      get shouldBe Some(addChange.recordSet)
+      finalGet shouldBe Some(updateChange.recordSet)
+
+    }
   }
 
   "list record sets" should {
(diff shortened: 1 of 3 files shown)


modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
```diff
@@ -92,6 +92,7 @@ class MySqlRecordSetRepository extends RecordSetRepository with Monitored {
             fromRecordType(oldRs.typ),
             toPB(oldRs).toByteArray,
             toFQDN(change.zone.name, oldRs.name),
+            oldRs.ownerGroupId,
```
pauljamescleary: can you add a test for when ownerGroupId is `None`

I presume that `None` maps to `NULL` but best to make sure.

rebstar6: when we get a recordset, we are pulling from the blob anyway. until we are querying the index i dont know that we can test that

rebstar6: may be worth just querying the local mysql to validate this though by hand

pauljamescleary: I meant inserting / UPDATE a null value doesn't blow up the insert and update statement itself.

Here, the updated SQL statement is inserting the owner group id as a column.  Want to be sure that inserting None works without a failure

rebstar6: fyi, ran locally and this is what that value looks like:

```
mysql> select name, owner_group_id from vinyldns.recordset;
+----------------+--------------------------------------+
| name           | owner_group_id                       |
+----------------+--------------------------------------+
| foo            | NULL                                 |
| jenkins        | NULL                                 |
| ok.            | NULL                                 |
| already-exists | NULL                                 |
| asdasd         | 06424926-769f-4eee-8b80-adfd54108f32 |
| ok.            | NULL                                 |
| ok.            | NULL                                 |
| test           | NULL                                 |
+----------------+--------------------------------------+
8 rows in set (0.00 sec)

```

i think were good :)

rebstar6: also of note:
```
mysql> SHOW INDEX FROM vinyldns.recordset
    -> ;
+-----------+------------+----------------------+--------------+----------------+-----------+-------------+----------+--------+------+------------+---------+---------------+
| Table     | Non_unique | Key_name             | Seq_in_index | Column_name    | Collation | Cardinality | Sub_part | Packed | Null | Index_type | Comment | Index_comment |
+-----------+------------+----------------------+--------------+----------------+-----------+-------------+----------+--------+------+------------+---------+---------------+
| recordset |          0 | PRIMARY              |            1 | id             | A         |          16 |     NULL | NULL   |      | BTREE      |         |               |
| recordset |          1 | zone_id_name_index   |            1 | zone_id        | A         |           2 |     NULL | NULL   |      | BTREE      |         |               |
| recordset |          1 | zone_id_name_index   |            2 | name           | A         |          11 |     NULL | NULL   |      | BTREE      |         |               |
| recordset |          1 | zone_id_name_index   |            3 | type           | A         |          16 |     NULL | NULL   |      | BTREE      |         |               |
| recordset |          1 | fqdn_index           |            1 | fqdn           | A         |          11 |     NULL | NULL   |      | BTREE      |         |               |
| recordset |          1 | owner_group_id_index |            1 | owner_group_id | A         |           2 |     NULL | NULL   | YES  | BTREE      |         |               |
+-----------+------------+----------------------+--------------+----------------+-----------+-------------+----------+--------+------+------------+---------+---------------+
6 rows in set (0.00 sec)

```

modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlRecordSetRepositoryIntegrationSpec.scala
```diff
@@ -272,6 +272,50 @@ class MySqlRecordSetRepositoryIntegrationSpec
       repo.getRecordSet(okZone.id, inserts(1).recordSet.id).unsafeRunSync().map(_.name) shouldBe
         Some(inserts(1).recordSet.name)
     }
+    "works when inserting owner-group-id" in {
```
mitruly: Nitpick: can we change "owner-group-id" to "ownerGroupId" or "owner group ID"?

PR #533, opened 2019-03-18
 [Issue-463] Synchronize record set naming restrictions

### LINKED ISSUE #463: Synchronize record set naming restrictions
**Motivation**
Since batch change is a more recent feature, it contains the most up-to-date validations for host name. This includes the length and valid characters within a label:
```
  def validateHostName(name: String): ValidatedNel[DomainValidationError, String] = {
    /*
      Label rules are as follows (from RFC 952; detailed in RFC 1034):
        - Starts with a letter, OR digit (as of RFC 1123)
        - Interior contains letter, digit or hyphen
        - Ends with a letter or digit
      All possible labels permutations:
        - A single letter/digit: [0-9a-zA-Z]{1}
        - A combination of 1-63 letters/digits: [0-9a-zA-Z]{1,63}
        - A single letter/digit followed by up to 61 letters, digits, hyphens or slashes
        and ending with a letter/digit:[0-9a-zA-Z]{1}[0-9a-zA-Z\-]{0,61}[0-9a-zA-Z]{1}
      A valid domain name is a series of one or more <label>s,
      joined by dots/slashes and terminating on a zero-length <label> (ie. dot)
     */
```
Currently, record set naming conventions only perform validations for dotted host and length.

**Scope of change**
Update record set naming validations to be consistent with our existing `DomainValidations` that are used in batch change.

joshulyne:
Fixes #463 .

Changes in this pull request:
- Adding host name validation from `DomainValidations` to `RecordSetValidations` for only certain records (PTR, CNAME)

Comments:
- In `DomainValidatons`, the `validateHostName` method seems to be applied to only these records:
A, AAAA, cname, ptr, mx (exchange), and TXT. Since RecordSetValidations don't seem to have validations for A, AAAA, mx, and TXT records, the validation was only applied to PTR and CNAME. 
- Unsure on the CNAME validation, as it contains the `isNotDotted` check as well
- Pretty crude implementation of this validation, let me know if you guys would like it a different way!

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetValidations.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetValidations.scala
@@ -249,31 +249,19 @@ object RecordSetValidations {
       .leftMap(errors => InvalidRequest(errors.toList.map(_.message).mkString(", ")))
   }
 
-  def validateLabel(name: String): Either[Throwable, Unit] = {
+  def isValidLabelRegex(name: String): Either[Throwable, Unit] = {
+    val label = if (name.contains(".")) name.substring(0, name.indexOf(".")) else name
+
     val validLabelRegex: Regex =
       """^([0-9a-zA-Z]{1}[0-9a-zA-Z\-\/]{0,61}[0-9a-zA-Z]{1}|[0-9a-zA-Z]{1})$""".r
 
-    validLabelRegex.findFirstIn(name) match {
+    validLabelRegex.findFirstIn(label) match {
       case Some(_) => ().asRight
       case None =>
         InvalidRequest(
           s"""Invalid record name: "$name", valid record name labels must be between 1-63 characters OR """ +
             "begin and end in a letter/digit, with hyphens and slashes being the other allowed characters.").asLeft
     }
-  }
-
-  def isValidLabelRegex(name: String): Either[Throwable, Unit] = {
-    val labels = name.split(".")
-    val result = for {
-      label <- labels
-      res = validateLabel(label)
-    } yield res
-
-    val leftResult = InvalidRequest(
-      s"""Invalid record name: "$name", valid record name labels must be between 1-63 characters OR """ +
-        "begin and end in a letter/digit, with hyphens and slashes being the other allowed characters.").asLeft
-
-    if (result.contains(leftResult)) leftResult else ().asRight
   }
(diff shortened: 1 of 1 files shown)


modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetValidations.scala
```diff
@@ -238,6 +251,20 @@ object RecordSetValidations {
       .leftMap(errors => InvalidRequest(errors.toList.map(_.message).mkString(", ")))
   }
 
+  def isValidHostName(name: String): Either[Throwable, Unit] = {
```
mitruly: `DomainValidations` (used in the batch logic) uses a regex to determine whether a provided hostname is a valid FQDN, but `RecordSetValidations` is different in that it is looking at an individual record.

There are basically two validations that are missing from record set validations that are covered by `validFQDNRegex`:
- A label (ie. FQDN segment; the record name in this case) must be between 1-63 characters
- The label has to begin and end in a letter/digit, with hyphens and slashes being the other allowed characters

It may make sense to break out a valid regex for `validLabelRegex` and have this check for the record name.

modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetValidations.scala
```diff
@@ -145,6 +147,9 @@ object RecordSetValidations {
         "CNAME RecordSet cannot have name '@' because it points to zone origin")
       _ <- noRecordWithName
       _ <- isNotDotted(newRecordSet, zone)
+//      _ <- isValidHostName(newRecordSet.name)
```
mitruly: The existing validations should be kept as we don't use dots to terminate record names unless they are apex records; `CNAME` records also cannot be apex.

modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetValidations.scala
joshulyne: oohhh that makes more sense! thanks for clarifying!!

mitruly:
Discussed this offline with @joshulyne and we'll be closing this since we merged in dotted `TXT` record support via record set endpoint (https://github.com/vinyldns/vinyldns/pull/587) and batch endpoint (https://github.com/vinyldns/vinyldns/pull/633).

PR #535, opened 2019-03-19
add more info on integration testing

### LINKED ISSUE #332: Update Integration Tests section in Developer Guide
The  Integration Tests section  of the Developer Guide is outdated. https://github.com/vinyldns/vinyldns/blob/master/DEVELOPER_GUIDE.md#integration-tests

It says there's only integration tests in the API module, but we now have integration tests in the dynamodb and mysql modules as well. Add that information and make it clear that `dockerComposeUp` to start up the background services is required for each module.

britneywright:
Fixes #332

Changes in this pull request:
- Note that we also have mysql and dynamodb module integration tests
- Note that there may be a delay needed between starting the docker containers and running the integration tests in the mysql module

### DIFF
--- a/DEVELOPER_GUIDE.md
+++ b/DEVELOPER_GUIDE.md
@@ -170,11 +170,13 @@ You should now be able to see the zone in the portal at localhost:9001 when logg
 Integration tests are used to test integration with _real_ dependent services.  We use Docker to spin up those
 backend services for integration test development.
 
-1. Integration tests are currently only in the `api` module.  Go to the module in sbt `project api`
+1. Integration tests are currently only in the `api`, `dynamo`, and `mysql` modules.  Go to the target module in sbt, example: `project api`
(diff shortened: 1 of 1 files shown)


DEVELOPER_GUIDE.md
```diff
@@ -170,11 +170,13 @@ You should now be able to see the zone in the portal at localhost:9001 when logg
 Integration tests are used to test integration with _real_ dependent services.  We use Docker to spin up those
 backend services for integration test development.
 
-1. Integration tests are currently only in the `api` module.  Go to the module in sbt `project api`
+1. Integration tests are currently only in the `api`, `dynamo`, and `mysql` modules.  Go to the target module in sbt, example: `project api`
```
mitruly: We also have integration tests in the `sqs` module as well.

britneywright: updated

Issue #311, opened 2018-10-26
Add delete function to GroupRepository trait


nimaeskandary:
**Motivation**
Due to legacy logic groups are not deleted from the DynamoDBGroupRepository, but instead have their status set to `Deleted`. This could be changed, and the groups can now be deleted.

**Scope of change**
* GroupRepository trait should have a `def delete(group: Group): IO[Group]`
* Delete group implementation in DynamoDBGroupRepository
* consider what indexes are needed by DynamoDBGroupRepository if status is not longer used, potentially make a new card for this
* Refactor how group deletions are done in the route

pauljamescleary:
This is done, closing

PR #537, opened 2019-03-19
Delete group - Check if ownerGroupId is in any recordSet

### LINKED ISSUE #436: Delete group logic
Groups can be tied to record ownership, zone admin, and acl rules. Currently, the only validation done before deleting a group is that it is not set to be a zone admin group for any zones. It should be further discussed how we want to handle cases in which the group is set as a record owner, and when the group is used in an ACL rule

varsharun:
Fixes #436 and #311 .

Changes in this pull request:
- not the ownerGroupId for any records - Done - NO TESTS YET
- not in any ACL rules -

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
@@ -85,24 +85,12 @@ class MembershipService(
       _ <- membershipRepo.removeMembers(existingGroup.id, removedMembers).toResult[Set[String]]
     } yield newGroup
 
-  /**
-    * On delete, check that the group is
-    *
-    * not a zone admin (already in place)
-    * not the ownerGroupId for any records (done - no tests yet)
-    * not in any ACL rules
-    * if all those pass, it can be deleted. otherwise return error
-    *
-    * @param groupId id
-    * @param authPrincipal rules
-    * @return Result[Group]
-    */
   def deleteGroup(groupId: String, authPrincipal: AuthPrincipal): Result[Group] =
     for {
       existingGroup <- getExistingGroup(groupId)
       _ <- canEditGroup(existingGroup, authPrincipal).toResult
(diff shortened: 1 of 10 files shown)


modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
```diff
@@ -86,7 +89,8 @@ class MembershipService(
     for {
       existingGroup <- getExistingGroup(groupId)
       _ <- canEditGroup(existingGroup, authPrincipal).toResult
-      _ <- groupCanBeDeleted(existingGroup)
+      _ <- isAZoneAdmin(existingGroup)
```
nimaeskandary: Think we generally name based off what we're saying is good, and it fails otherwise, so suggest these are named `isNotAZoneAdmin` and `isNotRecordOwner`

modules/dynamodb/src/main/scala/vinyldns/dynamodb/repository/DynamoDBRecordSetRepository.scala
```diff
@@ -243,4 +243,13 @@ class DynamoDBRecordSetRepository private[repository] (
         )
       )
     }
+
+  def isNotAOwnerGroupId(ownerGroupId: String): IO[Boolean] =
```
nimaeskandary: this would be a more useful error message if we returned the group id that's causing the failure and bubble it up to the validation

modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
```diff
@@ -233,7 +237,7 @@ class MembershipService(
       }
       .toResult
 
-  def groupCanBeDeleted(group: Group): Result[Unit] =
+  def isAZoneAdmin(group: Group): Result[Unit] =
```
pauljamescleary: in general `isZoneAdmin` instead of `isAZoneAdmin`, avoid using _a_ in names of things

PR #538, opened 2019-03-20
Add record-set under mysql configs for tests

varsharun:
Changes in this pull request:
- Update all of the config files to include `record-set` as part of `mysql` and removed from under `dynamodb` config


modules/docs/src/main/tut/operator/config-api.md
```diff
@@ -267,14 +269,6 @@ vinyldns {
     
     repositories {
       # all repositories with config sections here will be enabled in dynamodb
-      record-set {
-        # Name of the table where recordsets are saved
```
rebstar6: move these helper comments down to the `record-change` section since we dont copy them in each

modules/docs/src/main/tut/operator/config-portal.md
```diff
@@ -100,6 +100,8 @@ vinyldns {
       user {
       # no additional settings for repositories enabled in mysql
       }
+      record-set {
```
rebstar6: the portal doesnt use the recordset repo - remove from here

modules/docs/src/main/tut/operator/config-portal.md
```diff
@@ -289,6 +291,8 @@ mysql {
   repositories {
     user {
     }
+    record-set {
```
rebstar6: remove from here too

nimaeskandary:
Would avoid merging this until we migrate any existing tables right?

modules/docs/src/main/tut/operator/setup-dynamodb.md
```diff
@@ -41,40 +40,6 @@ to complete the setup for the API.
 
 You also need to configure DynamoDB for the portal [Portal Database Configuration](config-portal#database-configuration)
 
-### RecordSet Table
```
rebstar6: interesting

i think leave this section in, just remove at the top where we say that we use the dynamo RSR. Add a note to this section that is something like "Note: the DynamoDB RecordSet repository is only partially implemented. for use you would need to provide implementations of those methods" (or something)

modules/portal/conf/application-test.conf
```diff
@@ -23,6 +23,7 @@ mysql {
 
   repositories {
     user {}
+    record-set{}
```
rebstar6: remove - portal doesnt use this repo

modules/portal/conf/application.conf
```diff
@@ -13,6 +13,7 @@ mysql {
     user {
       # no additional settings for now
     }
+    record-set{ }
```
rebstar6: remove

britneywright:
not sure if we want to wait on the doc updates, although I don't think the timing really matters. I'm good either way 👍🏽

PR #539, opened 2019-03-20
batch change use getRecordSetsByFQDNs

britneywright:
right now recordsets are searched by zone id and recordset name, the getRecordSetsByFQDNs function will allow for batch gets using the input name that was directly input into the batch request

### DIFF
--- a/modules/api/functional_test/live_tests/batch/create_batch_change_test.py
+++ b/modules/api/functional_test/live_tests/batch/create_batch_change_test.py
@@ -552,8 +552,8 @@ def test_create_batch_change_failed(shared_zone_test_context):
         # these records already exist in the backend, but are not synced in zone
         dns_add(shared_zone_test_context.ok_zone, "backend-foo", 200, "A", "1.2.3.4")
         dns_add(shared_zone_test_context.ok_zone, "backend-already-exists", 200, "A", "1.2.3.4")
-        dns_add(shared_zone_test_context.ip6_reverse_zone, "4.3.2.1.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0", 200, "PTR", "test.com.")
-        dns_add(shared_zone_test_context.classless_zone_delegation_zone, "193", 200, "PTR", "test.com")
+        dns_add(shared_zone_test_context.ip6_reverse_zone, "4.3.2.1.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0", 200, "PTR", "another-test.com")
+        dns_add(shared_zone_test_context.classless_zone_delegation_zone, "193", 200, "PTR", "another-test.com")
         result = client.create_batch_change(batch_change_input, status=202)
         completed_batch = client.wait_until_batch_change_completed(result)
 
--- a/modules/api/functional_test/live_tests/batch/create_batch_change_test.py
+++ b/modules/api/functional_test/live_tests/batch/create_batch_change_test.py
@@ -542,16 +542,18 @@ def test_create_batch_change_failed(shared_zone_test_context):
         "comments": "this is optional",
         "changes": [
             get_change_A_AAAA_json("backend-foo.ok.", address="4.5.6.7"),
-            get_change_A_AAAA_json("backend-already-exists.ok.", address="4.5.6.7")
+            get_change_A_AAAA_json("backend-already-exists.ok.", address="4.5.6.7"),
+            get_change_PTR_json("fd69:27cc:fe91::1234"),
+            get_change_PTR_json("192.0.2.196")
         ]
     }
 
     try:
-        # both of these records already exist in the backend, but are not synced in zone
+        # these records already exist in the backend, but are not synced in zone
         dns_add(shared_zone_test_context.ok_zone, "backend-foo", 200, "A", "1.2.3.4")
         dns_add(shared_zone_test_context.ok_zone, "backend-already-exists", 200, "A", "1.2.3.4")
-        dns_add(shared_zone_test_context.ip6_reverse_zone, "fd69:27cc:fe91::1234", 200, "PTR", "test.com.")
-        dns_add(shared_zone_test_context.classless_zone_delegation_zone, "194", 200, "PTR", "another-test.com.")
+        dns_add(shared_zone_test_context.ip6_reverse_zone, "4.3.2.1.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0", 200, "PTR", "test.com.")
+        dns_add(shared_zone_test_context.classless_zone_delegation_zone, "196", 200, "PTR", "test.com")
         result = client.create_batch_change(batch_change_input, status=202)
         completed_batch = client.wait_until_batch_change_completed(result)
 
@@ -560,9 +562,8 @@ def test_create_batch_change_failed(shared_zone_test_context):
     finally:
         dns_delete(shared_zone_test_context.ok_zone, "backend-foo", "A")
         dns_delete(shared_zone_test_context.ok_zone, "backend-already-exists", "A")
-        dns_delete(shared_zone_test_context.ok_zone, "fd69:27cc:fe91::1234", "PTR")
-        dns_delete(shared_zone_test_context.ok_zone, "194", "PTR")
-
+        dns_delete(shared_zone_test_context.classless_zone_delegation_zone, "196", "PTR")
+        dns_delete(shared_zone_test_context.ip6_reverse_zone, "4.3.2.1.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0", "PTR")
 
 def test_empty_batch_fails(shared_zone_test_context):
     """
--- a/modules/api/functional_test/live_tests/batch/create_batch_change_test.py
+++ b/modules/api/functional_test/live_tests/batch/create_batch_change_test.py
@@ -550,6 +550,8 @@ def test_create_batch_change_failed(shared_zone_test_context):
         # both of these records already exist in the backend, but are not synced in zone
         dns_add(shared_zone_test_context.ok_zone, "backend-foo", 200, "A", "1.2.3.4")
         dns_add(shared_zone_test_context.ok_zone, "backend-already-exists", 200, "A", "1.2.3.4")
+        dns_add(shared_zone_test_context.ip6_reverse_zone, "fd69:27cc:fe91::1234", 200, "PTR", "test.com.")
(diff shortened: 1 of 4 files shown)


nimaeskandary:
good if travs is

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -154,16 +154,10 @@ class BatchChangeService(
 
   def getExistingRecordSets(
       changes: ValidatedBatch[ChangeForValidation]): IO[ExistingRecordSets] = {
-    // TODO - this implementation may be problematic with many changes. Need to perf test/later change DB
-    val uniqueGets = changes.getValid.map(change => (change.zone.id, change.recordName)).toSet
+    val uniqueGets = changes.getValid.map(change => change.inputChange.inputName).toSet
+    val allSeq = recordSetRepository.getRecordSetsByFQDNs(uniqueGets)
```
rebstar6: call this allRecordsets or something now, the old names were because of weird typing

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -31,7 +31,7 @@ import vinyldns.core.domain.auth.AuthPrincipal
 import vinyldns.core.domain.batch._
 import vinyldns.core.domain.membership.{Group, GroupRepository}
 import vinyldns.core.domain.record.RecordType._
-import vinyldns.core.domain.record.{RecordSet, RecordSetRepository}
+import vinyldns.core.domain.record.{RecordSetRepository}
```
mitruly: Can remove the surrounding braces if we're only importing a single entity: `import vinyldns.core.domain.record.RecordSetRepository`

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -154,16 +154,10 @@ class BatchChangeService(
...
+    val allRecordSets = recordSetRepository.getRecordSetsByFQDNs(uniqueGets)
 
-    val allIO = uniqueGets.map {
-      case (zoneId, rsName) => recordSetRepository.getRecordSetsByName(zoneId, rsName)
-    }
-
-    val allSeq: IO[List[List[RecordSet]]] = allIO.toList.sequence
-
-    allSeq.map(lst => ExistingRecordSets(lst.flatten))
+    allRecordSets.map(lst => ExistingRecordSets(lst))
```
mitruly: Mapping elements of a collection into a constructor doesn't require explicit unpacking of variables: `allRecordSets.map(ExistingRecordSets)`

modules/api/src/test/scala/vinyldns/api/domain/batch/BatchChangeServiceSpec.scala
```diff
@@ -139,6 +141,13 @@ class BatchChangeServiceSpec
           case (_, _) => List()
         }
       }
+
+    override def getRecordSetsByFQDNs(names: Set[String]): IO[List[RecordSet]] =
+      IO.pure(
+        dbRecordSets
```
mitruly: This is more of a FYI than actual change request (as it doesn't work that cleanly in this scenario) - pattern matching can unpack tuples more comprehensibly:
```
      IO.pure {
        dbRecordSets
          .filter {
            case (_, fqdn) =>
              names.contains(fqdn)
          }
          .map {
            case (rs, _) => rs
          }
          .toList
      }
```

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -154,16 +154,10 @@ class BatchChangeService(
 
   def getExistingRecordSets(
       changes: ValidatedBatch[ChangeForValidation]): IO[ExistingRecordSets] = {
-    // TODO - this implementation may be problematic with many changes. Need to perf test/later change DB
-    val uniqueGets = changes.getValid.map(change => (change.zone.id, change.recordName)).toSet
+    val uniqueGets = changes.getValid.map(change => change.inputChange.inputName).toSet
```
rebstar6: you know this already but commenting to update my status here -
`ChangeForValidation` has the recordName and zone. you are going to have to construct the fqdn from that (rather than using `change.inputChange.inputName`) because PTRs need to search for fqdns like `1.2.3.4.in-addr.arpa.`, not `4.3.2.1`.

I would also want a test where we put a conflicting PTR record (non-batch), and then try to add it in a batch. By this logic it would still go through, but when the update happens, the batch should see the error.

setting myself now X on the PR, but if it gets fixed, the test is added, and it's reviewed while i'm gone, you can override my X

modules/api/src/test/scala/vinyldns/api/domain/batch/BatchChangeServiceSpec.scala
pauljamescleary: 👍 

@britneywright generally avoid `_2` and `_1` for tuples and use pattern matching.  The code is easier to follow

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -154,16 +154,15 @@ class BatchChangeService(
...
-    val uniqueGets = changes.getValid.map(change => (change.zone.id, change.recordName)).toSet
-
-    val allIO = uniqueGets.map {
-      case (zoneId, rsName) => recordSetRepository.getRecordSetsByName(zoneId, rsName)
-    }
-
-    val allSeq: IO[List[List[RecordSet]]] = allIO.toList.sequence
+    val uniqueGets = changes.getValid.map { change =>
+      change.inputChange.typ match {
+        case PTR => s"${change.recordName}.${change.zone.name}"
```
nimaeskandary: Hm, believe this would fail on apex ptr records. Not sure if they're a thing, but did some searching and couldnt find anything saying they're invalid

nimaeskandary: You could use the toFQDN in the mysql recordset repo to work against those

nimaeskandary: Confirmed that an apex ptr in a classless delegation zone will not go through, so guess this can  be disregarded

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
pauljamescleary: @britneywright has this comment been resolved?

britneywright: yup!

britneywright: added to an existing test here https://github.com/vinyldns/vinyldns/pull/539/files#diff-090b2d00e8c30c6696c34e10ed759ac5

modules/api/functional_test/live_tests/batch/create_batch_change_test.py
```diff
@@ -550,6 +550,8 @@ def test_create_batch_change_failed(shared_zone_test_context):
         # both of these records already exist in the backend, but are not synced in zone
         dns_add(shared_zone_test_context.ok_zone, "backend-foo", 200, "A", "1.2.3.4")
         dns_add(shared_zone_test_context.ok_zone, "backend-already-exists", 200, "A", "1.2.3.4")
+        dns_add(shared_zone_test_context.ip6_reverse_zone, "fd69:27cc:fe91::1234", 200, "PTR", "test.com.")
```
mitruly: `batch_change_input` is missing the `get_change_PTR_json` records; current test `completed_batch` prints:
```
{
   "userName": "ok", 
   "status": "Failed", 
   "userId": "ok", 
   "comments": "this is optional", 
   "createdTimestamp": "2019-04-05T19:21:35Z", 
   "changes": [
      {
         "status": "Failed", 
         "recordChangeId": "30988973-3e20-4816-a171-568c1c143ce3", 
         "recordName": "backend-foo", 
         "changeType": "Add", 
         "inputName": "backend-foo.ok.", 
         "zoneId": "7aa3638e-f488-4982-a5e2-6f760717de4b", 
         "record": {
            "address": "4.5.6.7"
         }, 
         "systemMessage": "Failed validating update to DNS for change 30988973-3e20-4816-a171-568c1c143ce3:backend-foo: Incompatible record already exists in DNS.", 
         "ttl": 200, 
         "type": "A", 
         "id": "c9ec37e4-9312-441e-9be1-e788975f28f9", 
         "zoneName": "ok."
      }, 
      {
         "status": "Failed", 
         "recordChangeId": "907a0d94-0cdb-4185-a388-0782685fc028", 
         "recordName": "backend-already-exists", 
         "changeType": "Add", 
         "inputName": "backend-already-exists.ok.", 
         "zoneId": "7aa3638e-f488-4982-a5e2-6f760717de4b", 
         "record": {
            "address": "4.5.6.7"
         }, 
         "systemMessage": "Failed validating update to DNS for change 907a0d94-0cdb-4185-a388-0782685fc028:backend-already-exists: Incompatible record already exists in DNS.", 
         "ttl": 200, 
         "type": "A", 
         "id": "7bce0779-9b9a-4f55-8260-9c56a1694ab5", 
         "zoneName": "ok."
      }
   ], 
   "id": "6f6e4af7-2b72-4d5b-8106-eaf838285d66"
}
```

modules/api/functional_test/live_tests/batch/create_batch_change_test.py
```diff
@@ -550,6 +550,8 @@ def test_create_batch_change_failed(shared_zone_test_context):
...
+        dns_add(shared_zone_test_context.classless_zone_delegation_zone, "194", 200, "PTR", "another-test.com.")
```
mitruly: Need to use a different record than `194` here; `194` is already in `2.0.192.in-addr.arpa` hosts file.

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
britneywright: the  conflicting record for the ipv6 reverse zone seems to go through completely in this branch and in master, the record for the classless zone goes passes initial validations then fails on the backend conflict.

PR #542, opened 2019-03-21
Delete group - Check if ownerGroupId is in any recordSet

varsharun:
Fixes #.

Changes in this pull request:
- check if ownerGroupId is not in any recordset before deleting group

### DIFF
--- a/modules/api/functional_test/live_tests/batch/get_batch_change_test.py
+++ b/modules/api/functional_test/live_tests/batch/get_batch_change_test.py
@@ -97,10 +97,18 @@ def test_get_batch_change_with_deleted_record_owner_group_success(shared_zone_te
         record_set_list = [(change['zoneId'], change['recordSetId']) for change in completed_batch['changes']]
         record_to_delete = set(record_set_list)
 
-        # delete group
+        #delete records and owner group
+        temp = record_to_delete.copy()
+        for result_rs in temp:
+            delete_result = client.delete_recordset(result_rs[0], result_rs[1], status=202)
+            client.wait_until_recordset_change_status(delete_result, 'Complete')
+            record_to_delete.remove(result_rs)
+        temp.clear()
+
         client.delete_group(group_to_delete['id'], status=200)
         del completed_batch['ownerGroupName']
 
+        #the batch should not be updated with deleted group data
         result = client.get_batch_change(batch_change['id'], status=200)
         assert_that(result, is_(completed_batch))
         assert_that(result['ownerGroupId'], is_(group_to_delete['id']))
--- a/modules/api/functional_test/live_tests/batch/list_batch_change_summaries_test.py
+++ b/modules/api/functional_test/live_tests/batch/list_batch_change_summaries_test.py
@@ -226,6 +226,14 @@ def test_list_batch_change_summaries_with_deleted_record_owner_group_passes(shar
         record_set_list = [(change['zoneId'], change['recordSetId']) for change in completed_batch['changes']]
         record_to_delete = set(record_set_list)
 
+        #delete records and owner group
+        temp = record_to_delete.copy()
+        for result_rs in temp:
+            delete_result = client.delete_recordset(result_rs[0], result_rs[1], status=202)
+            client.wait_until_recordset_change_status(delete_result, 'Complete')
+            record_to_delete.remove(result_rs)
+        temp.clear()
+
         # delete group
         client.delete_group(group_to_delete['id'], status=200)
 
--- a/modules/api/functional_test/live_tests/zone_history_context.py
+++ b/modules/api/functional_test/live_tests/zone_history_context.py
@@ -159,7 +159,6 @@ class ZoneHistoryContext(object):
         result =  self.client.delete_recordset(self.zone['id'], rs['id'], status=202)
         return result, result['recordSet']
 
-
     def confirm_member_in_group(self, client, group):
         retries = 2
         success = group in client.list_all_my_groups(status=200)
--- a/modules/api/functional_test/live_tests/zones/sync_zone_test.py
+++ b/modules/api/functional_test/live_tests/zones/sync_zone_test.py
@@ -95,6 +95,7 @@ def test_sync_zone_success(shared_zone_test_context):
     client = shared_zone_test_context.ok_vinyldns_client
     zone_name = 'sync-test'
     updated_rs_id = None
+    check_rs = None
 
     zone = {
         'name': zone_name,
@@ -216,6 +217,11 @@ def test_sync_zone_success(shared_zone_test_context):
                 client.wait_until_recordset_change_status(change, 'Complete')
 
     finally:
+        # reset the ownerGroupId for foo record
+        if check_rs:
+            check_rs['ownerGroupId'] = None
+            update_response = client.update_recordset(check_rs, status=202)
+            client.wait_until_recordset_change_status(update_response, 'Complete')
         if 'id' in zone:
             dns_update(zone, 'foo', 38400, 'A', '2.2.2.2')
             dns_delete(zone, 'newrs', 'A')
--- a/modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
@@ -85,24 +85,12 @@ class MembershipService(
       _ <- membershipRepo.removeMembers(existingGroup.id, removedMembers).toResult[Set[String]]
     } yield newGroup
 
-  /**
-    * On delete, check that the group is
-    *
-    * not a zone admin (already in place)
-    * not the ownerGroupId for any records (done - no tests yet)
-    * not in any ACL rules
-    * if all those pass, it can be deleted. otherwise return error
-    *
-    * @param groupId id
-    * @param authPrincipal rules
-    * @return Result[Group]
-    */
   def deleteGroup(groupId: String, authPrincipal: AuthPrincipal): Result[Group] =
     for {
       existingGroup <- getExistingGroup(groupId)
       _ <- canEditGroup(existingGroup, authPrincipal).toResult
-      _ <- isAZoneAdmin(existingGroup) //checks if the group id is not a Zone admin
-      _ <- isAOwnerGroupId(existingGroup) //checks if group is the owner of a record
+      _ <- isAZoneAdmin(existingGroup)
+      _ <- isOwnerGroup(existingGroup)
       _ <- groupChangeRepo
         .save(GroupChange.forDelete(existingGroup, authPrincipal))
         .toResult[GroupChange]
@@ -259,13 +247,13 @@ class MembershipService(
       }
       .toResult
 
-  def isAOwnerGroupId(group: Group): Result[Unit] =
+  def isOwnerGroup(group: Group): Result[Unit] =
     recordSetRepo
-      .getRecordSetByOwnerGroupId(group.id)
+      .isNotAOwnerGroupId(group.id)
       .map { rs =>
         ensuring(
           InvalidGroupRequestError(s"${group.name} is the owner for a record. Cannot delete.")) {
-          rs.isEmpty
+          rs
         }
       }
       .toResult
--- a/modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
@@ -89,8 +89,8 @@ class MembershipService(
     for {
       existingGroup <- getExistingGroup(groupId)
       _ <- canEditGroup(existingGroup, authPrincipal).toResult
-      _ <- isZoneAdmin(existingGroup)
-      _ <- isRecordOwnerGroup(existingGroup)
+      _ <- isNotZoneAdmin(existingGroup)
+      _ <- isNotRecordOwnerGroup(existingGroup)
       _ <- groupChangeRepo
         .save(GroupChange.forDelete(existingGroup, authPrincipal))
         .toResult[GroupChange]
@@ -237,7 +237,7 @@ class MembershipService(
       }
       .toResult
 
-  def isZoneAdmin(group: Group): Result[Unit] =
+  def isNotZoneAdmin(group: Group): Result[Unit] =
     zoneRepo
       .getZonesByAdminGroupId(group.id)
       .map { zones =>
@@ -247,13 +247,13 @@ class MembershipService(
       }
       .toResult
 
-  def isRecordOwnerGroup(group: Group): Result[Unit] =
+  def isNotRecordOwnerGroup(group: Group): Result[Unit] =
     recordSetRepo
-      .isRecordOwnerGroup(group.id)
+      .getRecordSetOwnerGroup(group.id)
       .map { rs =>
         ensuring(
-          InvalidGroupRequestError(s"${group.name} is the owner for a record. Cannot delete.")) {
-          !rs
+          InvalidGroupRequestError(s"${group.name} is the owner for a record set $rs. Cannot delete.")){
+          rs.isEmpty
         }
       }
       .toResult
--- a/modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
@@ -23,6 +23,7 @@ import vinyldns.core.domain.auth.AuthPrincipal
 import vinyldns.core.domain.membership.LockStatus.LockStatus
 import vinyldns.core.domain.zone.ZoneRepository
 import vinyldns.core.domain.membership._
+import vinyldns.core.domain.record.RecordSetRepository
 
 object MembershipService {
   def apply(dataAccessor: ApiDataAccessor): MembershipService =
@@ -31,7 +32,8 @@ object MembershipService {
       dataAccessor.userRepository,
       dataAccessor.membershipRepository,
       dataAccessor.zoneRepository,
-      dataAccessor.groupChangeRepository
+      dataAccessor.groupChangeRepository,
+      dataAccessor.recordSetRepository
     )
 }
 
@@ -40,7 +42,8 @@ class MembershipService(
     userRepo: UserRepository,
     membershipRepo: MembershipRepository,
     zoneRepo: ZoneRepository,
-    groupChangeRepo: GroupChangeRepository)
+    groupChangeRepo: GroupChangeRepository,
+    recordSetRepo: RecordSetRepository)
     extends MembershipServiceAlgebra {
 
   import MembershipValidations._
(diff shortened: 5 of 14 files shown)


modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
```diff
@@ -243,6 +247,17 @@ class MembershipService(
       }
       .toResult
 
+  def isOwnerGroup(group: Group): Result[Unit] =
```
britneywright: we might want to call this `isRecordOwnerGroup` to be more explicit.

modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
```diff
@@ -233,7 +237,7 @@ class MembershipService(
       }
       .toResult
 
-  def groupCanBeDeleted(group: Group): Result[Unit] =
+  def isZoneAdmin(group: Group): Result[Unit] =
```
britneywright: @nimaeskandary made the comment in the other PR, this should be `isNotZoneAdmin` or `isNotZoneAdminGroup`

modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
```diff
@@ -243,6 +247,17 @@ class MembershipService(
       }
       .toResult
 
+  def isOwnerGroup(group: Group): Result[Unit] =
+    recordSetRepo
+      .isNotRecordOwnerGroup(group.id)
+      .map { rs =>
+        ensuring(
+          InvalidGroupRequestError(s"${group.name} is the owner for a record. Cannot delete.")) {
```
mitruly: It might be a good discussion point as to what information is useful to return if the owner group is associated with at least one record. With larger matches, it wouldn't make sense to have record-level information (eg. `zoneId`, `name`, `id`, etc.).

Wondering if it might make sense to do a `SELECT COUNT(*)` and then we can return the number of records the owner group is associated with. Or we might consider just returning a collection of records with matching owner group if we decide to eventually have record set filtering account for `ownerGroupId`.

mitruly: Could also refactor the `.map`:
```
      .map {
        ensuring(InvalidGroupRequestError(s"${group.name} is the owner for a record. Cannot delete."))(_)
      }
```

modules/dynamodb/src/main/scala/vinyldns/dynamodb/repository/DynamoDBRecordSetRepository.scala
```diff
@@ -243,4 +243,13 @@ class DynamoDBRecordSetRepository private[repository] (
         )
       )
     }
+
+  def isNotRecordOwnerGroup(ownerGroupId: String): IO[Boolean] =
+    monitor("repo.RecordSet.isNotAOwnerGroupId") {
+      IO.raiseError(
+        UnsupportedDynamoDBRepoFunction(
+          s"isNotOwnerGroupId is not supported by VinylDNS DynamoDB RecordSetRepository id=${ownerGroupId}"
```
mitruly: We don't need to include the `id=${ownerGroupId}` here since this method is unsupported and we're just raising an error.

modules/dynamodb/src/test/scala/vinyldns/dynamodb/repository/DynamoDBRecordSetRepositorySpec.scala
```diff
@@ -395,4 +395,13 @@ class DynamoDBRecordSetRepositorySpec
         .unsafeRunSync()
     }
   }
+
+  "DynamoDBRecordSetRepository.isNotOwnerGroupId" should {
+    "return an error if used" in {
+      val store = new TestDynamoRecordSetRepo
+      a[UnsupportedDynamoDBRepoFunction] should be thrownBy store
```
mitruly: `a[UnsupportedDynamoDBRepoFunction]` → `an[UnsupportedDynamoDBRepoFunction]`

modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlRecordSetRepositoryIntegrationSpec.scala
```diff
@@ -478,4 +478,25 @@ class MySqlRecordSetRepositoryIntegrationSpec
       result2 should contain theSameElementsAs List(change1.recordSet, change2.recordSet)
     }
   }
+
+  "isNotOwnerGroupId" should {
+    "return true when id is ownerGroupId for a record set" in {
```
mitruly: `return false when ownerGroupId matches at least one record set` or something similar makes more sense here.

modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlRecordSetRepositoryIntegrationSpec.scala
```diff
@@ -478,4 +478,25 @@ class MySqlRecordSetRepositoryIntegrationSpec
...
+        get <- repo.getRecordSet(testRecord.zoneId, testRecord.id)
+      } yield get
+
+      dbCalls.unsafeRunSync()
+
+      val result = repo.isNotRecordOwnerGroup("someOwner").unsafeRunSync()
+      result shouldBe false
+    }
+
+    "return false when no record set has the id as ownerGroupId" in {
```
mitruly: `return true when no record set matches ownerGroupId` is clearer.

modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
```diff
@@ -86,7 +89,8 @@ class MembershipService(
     for {
       existingGroup <- getExistingGroup(groupId)
       _ <- canEditGroup(existingGroup, authPrincipal).toResult
-      _ <- groupCanBeDeleted(existingGroup)
+      _ <- isZoneAdmin(existingGroup)
```
britneywright: these should be negatives `isNotZoneAdmin` and `isNotRecordOwnerGroup` because it's valid if the group is not a zone admin or a record owner and will fail if it's a zone admin or a record owner

varsharun: I did not change isZoneAdmin only renamed it. So the isRecordOwnerGroup is similar to isZoneAdmin. im checking for positive results but negating for our purpose. So if the id `isRecordOwnerGroup` is true then error.

modules/api/src/test/scala/vinyldns/api/repository/EmptyRepositories.scala
```diff
@@ -49,6 +49,8 @@ trait EmptyRecordSetRepo extends RecordSetRepository {
   def getRecordSetCount(zoneId: String): IO[Int] = IO.pure(0)
 
   def getRecordSetsByFQDNs(names: Set[String]): IO[List[RecordSet]] = IO.pure(List())
+
+  def isRecordOwnerGroup(ownerGroupId: String): IO[Boolean] = IO.pure(true)
```
nimaeskandary: Think the error message you can produce by this may be a blocker, without the name of the record the user wouldn't be able to move forward and it would be difficult for admins to help them debug

varsharun: should I return a list of records for which this group id is an admin? How long will this list be? @nimaeskandary

nimaeskandary: Maybe just the first one and they can keep working with the error messages until they resolve everything

nimaeskandary: Or all of them may work too if you to a toString on a list or seq

varsharun: Oo I like that idea. Will make that change and update the PR. Was there anything else from your side?

modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
nimaeskandary: I think a good example of the the convention we use is the RecordSetService file. Names like `isZoneAdmin` and `isRecordOwnerGroup` may be confusing from a readability perspective because coming from another file, the developer would assume you are checking that the the group **is** a zone admin and passing if so. I think changing to britneys suggestion would help keep things consistent with that convention and how the devs read through the for comprehensions

modules/api/functional_test/live_tests/batch/get_batch_change_test.py
```diff
@@ -97,14 +95,16 @@ def test_get_batch_change_with_deleted_record_owner_group_success(shared_zone_te
         record_set_list = [(change['zoneId'], change['recordSetId']) for change in completed_batch['changes']]
         record_to_delete = set(record_set_list)
 
-        # delete group
-        client.delete_group(group_to_delete['id'], status=200)
+        # delete group should not be possible
+        rdDelete = client.delete_group(group_to_delete['id'], status=400)
+        assert_that(rdDelete, contains_string("Cannot delete"))
         del completed_batch['ownerGroupName']
```
nimaeskandary: should not be deleting this field from the expected anymore

modules/api/functional_test/live_tests/batch/get_batch_change_test.py
```diff
@@ -97,10 +97,17 @@ def test_get_batch_change_with_deleted_record_owner_group_success(shared_zone_te
         record_set_list = [(change['zoneId'], change['recordSetId']) for change in completed_batch['changes']]
         record_to_delete = set(record_set_list)
```
rebstar6: theres only 1 record, no need to be throwing in a set and looping

rebstar6: oh i see, nvm. also may make sense to just delete the record now. you have to do it eventually anyway, might as well save the extra calls.

id do something like
```
record_to_delete = set(record_set_list)
for result_rs in set(record_set_list):
            delete_result = client.delete_recordset(result_rs[0], result_rs[1], status=202)
            record_to_delete.remove(result_rs)
```

after the loop, record_to_delete should be empty if everything succeeds

modules/api/functional_test/live_tests/batch/list_batch_change_summaries_test.py
```diff
@@ -226,6 +226,13 @@ def test_list_batch_change_summaries_with_deleted_record_owner_group_passes(shar
         record_set_list = [(change['zoneId'], change['recordSetId']) for change in completed_batch['changes']]
         record_to_delete = set(record_set_list)
 
+        #remove ownerGroupId from recordset
+        for result_rs in record_to_delete:
```
rebstar6: same comment here as above

modules/api/functional_test/live_tests/zones/sync_zone_test.py
```diff
@@ -196,6 +196,11 @@ def test_sync_zone_success(shared_zone_test_context):
         check_rs = client.get_recordset(zone['id'], updated_rs_id)['recordSet']
         assert_that(check_rs['ownerGroupId'], is_(shared_zone_test_context.ok_group['id']))
 
+        #reset the ownerGroupId for foo record
```
rebstar6: move this into the finally, its not needed here i dont think

modules/api/functional_test/live_tests/zone_history_context.py
```diff
@@ -159,6 +159,9 @@ def delete_recordset(self, rs):
         result =  self.client.delete_recordset(self.zone['id'], rs['id'], status=202)
         return result, result['recordSet']
 
+    def get_recordset(self, rs):
```
rebstar6: is this needed? i dont see zone history test updates

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
```diff
@@ -275,6 +283,20 @@ class MySqlRecordSetRepository extends RecordSetRepository with Monitored {
         }
       }
     }
+
+  def getRecordSetIdOwnerGroup(ownerGroupId: String): IO[String] =
```
rebstar6: return an `IO[Option[String]]` instead of a string. your empty case here should be None, not ""

rebstar6: getFirstOwnedRecordByGroup maybe? naming on this one is tough, open to alternatives

modules/api/functional_test/live_tests/batch/get_batch_change_test.py
varsharun: alright will make this change. and remove the `finally` section.

modules/api/functional_test/live_tests/zones/sync_zone_test.py
varsharun: I tried to move it to finally but it gives me a connection timeout exception.

rebstar6: make sure its before the DNS updates. make it the 1st thing in finally

modules/api/functional_test/live_tests/batch/get_batch_change_test.py
rebstar6: keep finally actually, record_to_delete should be empty, but if its not finally will clean it up

modules/api/functional_test/live_tests/batch/get_batch_change_test.py
```diff
@@ -97,10 +97,16 @@ def test_get_batch_change_with_deleted_record_owner_group_success(shared_zone_te
...
 
-        # delete group
+        #remove ownerGroupId from recordset
+        for result_rs in record_to_delete.copy():
```
rebstar6: the copy wont do what we want (i think). the original record_to_delete is unchanged, which means `finally` will be deleting on a still full list.

id say bag this idea, lets just do `for result_rs in record_to_delete: ...` once here and then remove it from finally

running locally, the test is to run it twice without restarting the api. if this runs 2x successfully, it is cleaning itself up sufficiently. if not, something bad remains in the database.

modules/api/functional_test/live_tests/batch/list_batch_change_summaries_test.py
```diff
@@ -226,6 +226,12 @@ def test_list_batch_change_summaries_with_deleted_record_owner_group_passes(shar
         record_set_list = [(change['zoneId'], change['recordSetId']) for change in completed_batch['changes']]
         record_to_delete = set(record_set_list)
 
+        #remove ownerGroupId from recordset
+        for result_rs in record_to_delete.copy():
```
rebstar6: same as above comment

modules/api/functional_test/live_tests/zones/sync_zone_test.py
```diff
@@ -196,6 +197,9 @@ def test_sync_zone_success(shared_zone_test_context):
         check_rs = client.get_recordset(zone['id'], updated_rs_id)['recordSet']
         assert_that(check_rs['ownerGroupId'], is_(shared_zone_test_context.ok_group['id']))
 
+        check_rs['ownerGroupId'] = None
```
rebstar6: remove this line

modules/api/functional_test/live_tests/zones/sync_zone_test.py
```diff
@@ -216,6 +220,11 @@ def test_sync_zone_success(shared_zone_test_context):
                 client.wait_until_recordset_change_status(change, 'Complete')
 
     finally:
+        # reset the ownerGroupId for foo record
+        if check_rs:
```
rebstar6: good, this should work

modules/api/src/test/scala/vinyldns/api/repository/EmptyRepositories.scala
```diff
@@ -49,6 +49,8 @@ trait EmptyRecordSetRepo extends RecordSetRepository {
   def getRecordSetCount(zoneId: String): IO[Int] = IO.pure(0)
 
   def getRecordSetsByFQDNs(names: Set[String]): IO[List[RecordSet]] = IO.pure(List())
+
+  def getFirstOwnedRecordByGroup(ownerGroupId: String): IO[String] = IO.pure("")
```
rebstar6: IO.pure(None) here

modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlRecordSetRepositoryIntegrationSpec.scala
```diff
@@ -478,4 +478,25 @@ class MySqlRecordSetRepositoryIntegrationSpec
...
+
+      dbCalls.unsafeRunSync()
+
+      val result = repo.getFirstOwnedRecordByGroup("someOwner").unsafeRunSync()
+      result shouldBe testRecord.id
+    }
+
+    "return empty string when no record set has the id as ownerGroupId" in {
+      val result = repo.getFirstOwnedRecordByGroup("notFound").unsafeRunSync()
+      result shouldBe ""
```
rebstar6: these tests need updating for the types now. result here should be None, above should be Some(testRecord.id)

modules/api/functional_test/live_tests/batch/get_batch_change_test.py
varsharun: How about 
```temp = record_to_delete
        #remove ownerGroupId from recordset
        for result_rs in temp:
            delete_result = client.delete_recordset(result_rs[0], result_rs[1], status=202)
            record_to_delete.remove(result_rs)
        temp.clear()```

Comment (rebstar6) [reply modules/api/functional_test/live_tests/batch/get_batch_change_test.py]: maybe haha. print temp after, and see what you get, i dont know off the top of my head

Comment (varsharun) [reply modules/api/functional_test/live_tests/batch/get_batch_change_test.py]:
```temp = record_to_delete.copy()
        #remove ownerGroupId from recordset
        for result_rs in temp:
            delete_result = client.delete_recordset(result_rs[0], result_rs[1], status=202)
            record_to_delete.remove(result_rs)
        temp.clear()
```
This worked.

modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlRecordSetRepositoryIntegrationSpec.scala
```diff
@@ -478,4 +478,25 @@ class MySqlRecordSetRepositoryIntegrationSpec
       result2 should contain theSameElementsAs List(change1.recordSet, change2.recordSet)
     }
   }
+
+  "getRecordSetIdOwnerGroup" should {
+    "return recordset id when groupid is ownerGroupId for the record set" in {
```
britneywright: capitalize the `I` in "groupId"

modules/api/src/test/scala/vinyldns/api/domain/membership/MembershipServiceSpec.scala
```diff
@@ -465,6 +477,17 @@ class MembershipServiceSpec
...
       }
+
+      "return an error if the group is an owner for a record set" in {
+        doReturn(IO.pure(Some(okGroup))).when(mockGroupRepo).getGroup(anyString)
+        doReturn(IO.pure(Some("somerecordsetid")))
+          .when(mockRecordSetRepo)
+          .getFirstOwnedRecordByGroup(anyString())
+        val error = leftResultOf(underTest.deleteGroup("ok", okAuth).value)
+
+        error shouldBe a[InvalidGroupRequestError]
```
mitruly: Should change `a[InvalidGroupRequestError]` → `an[InvalidGroupRequestError]`

modules/api/functional_test/live_tests/batch/get_batch_change_test.py
```diff
@@ -97,10 +97,18 @@ def test_get_batch_change_with_deleted_record_owner_group_success(shared_zone_te
         record_set_list = [(change['zoneId'], change['recordSetId']) for change in completed_batch['changes']]
         record_to_delete = set(record_set_list)
 
-        # delete group
+        temp = record_to_delete.copy()
+        #remove ownerGroupId from recordset
```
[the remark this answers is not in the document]
varsharun: oops!! leftover comment will fix it

PR #545, opened 2019-03-22
make logs on change more friendly

rebstar6:
dont love how the old way is queried, this should be more friendly

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/engine/RecordSetChangeHandler.scala
+++ b/modules/api/src/main/scala/vinyldns/api/engine/RecordSetChangeHandler.scala
@@ -171,33 +171,24 @@ object RecordSetChangeHandler {
 
     state match {
       case Pending(change) =>
-        logger.info(
-          s"PENDING for change ${change.id}:${change.recordSet.name} change type ${change.changeType}")
-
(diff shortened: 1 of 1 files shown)


modules/api/src/main/scala/vinyldns/api/engine/RecordSetChangeHandler.scala
```diff
@@ -171,35 +171,38 @@ object RecordSetChangeHandler {
...
       case done: Completed =>
-        logger.info(
-          s"COMPLETED for change [${done.change.id}:${done.change.recordSet.name}], " +
-            s"zone name = [${done.change.zone.name}], isTestZone = [${done.change.zone.isTest}]," +
-            s"change type = [${done.change.changeType}], result [${done.change.status}]")
+        logger.info(s"CHANGE COMPLETED; ${getChangeLog(done.change)}")
         IO.pure(done)
     }
+
+    def getChangeLog(change: RecordSetChange): String = {
```
mitruly: Wonder if it might make more sense to just have this as the `toString` override for `RecordSetChange`, but not sure what currently uses it.

rebstar6: nah there are other times we want the full change info

PR #546, opened 2019-03-22
Delete Group - Checking if id is present in Zone ACL

varsharun:
Fixes #.

Changes in this pull request:
- new method in ZoneRepository to check if the id is in any of the zone ACLs

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
@@ -85,18 +85,6 @@ class MembershipService(
       _ <- membershipRepo.removeMembers(existingGroup.id, removedMembers).toResult[Set[String]]
     } yield newGroup
 
-  /**
-    * On delete, check that the group is
-    *
-    * not a zone admin (already in place)
-    * not the ownerGroupId for any records (done - no tests yet)
-    * not in any ACL rules
-    * if all those pass, it can be deleted. otherwise return error
-    *
-    * @param groupId id
-    * @param authPrincipal rules
-    * @return Result[Group]
-    */
   def deleteGroup(groupId: String, authPrincipal: AuthPrincipal): Result[Group] =
     for {
       existingGroup <- getExistingGroup(groupId)
--- a/modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
@@ -91,6 +91,7 @@ class MembershipService(
       _ <- canEditGroup(existingGroup, authPrincipal).toResult
       _ <- isNotZoneAdmin(existingGroup)
       _ <- isNotRecordOwnerGroup(existingGroup)
+      _ <- isZoneAclGroupId(existingGroup)
       _ <- groupChangeRepo
         .save(GroupChange.forDelete(existingGroup, authPrincipal))
         .toResult[GroupChange]
@@ -241,9 +242,8 @@ class MembershipService(
     zoneRepo
       .getZonesByAdminGroupId(group.id)
       .map { zones =>
-        ensuring(InvalidGroupRequestError(s"${group.name} is the admin of a zone. Cannot delete.")) {
-          zones.isEmpty
-        }
+        ensuring(InvalidGroupRequestError(s"${group.name} is the admin of a zone. Cannot delete."))(
+          zones.isEmpty)
       }
       .toResult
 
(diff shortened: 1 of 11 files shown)


modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneRepository.scala
```diff
@@ -227,6 +235,20 @@ class MySqlZoneRepository extends ZoneRepository with ProtobufConversions with M
       }
     }
 
+  def getZoneAclGroupId(groupId: String): IO[String] =
```
rebstar6: same here as the other PR. in general we shouldnt use the empty string to mean lack of existence of something. thats what Option is for :)

make this return IO[Option[String]]

modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
```diff
@@ -91,6 +91,7 @@ class MembershipService(
       _ <- canEditGroup(existingGroup, authPrincipal).toResult
       _ <- isNotZoneAdmin(existingGroup)
       _ <- isNotRecordOwnerGroup(existingGroup)
+      _ <- isNotZoneAclGroupId(existingGroup)
```
britneywright: I would call this something like `isNotInZoneAclRule`

modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
```diff
@@ -259,6 +259,18 @@ class MembershipService(
       }
       .toResult
 
+  def isNotInZoneAclRule(group: Group): Result[Unit] =
+    zoneRepo
+      .getFirstOwnedZoneAclGroupId(group.id)
+      .map { zId =>
+        ensuring(
+          InvalidGroupRequestError(
+            s"${group.name} has an ACL rule for a zone including $zId. Cannot delete.")) {
```
mitruly: Can we change this to `(zones.isEmpty)` or the other `ensuring` to `{zones.isEmpty}` so they can be consistent?

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneRepository.scala
```diff
@@ -115,6 +115,14 @@ class MySqlZoneRepository extends ZoneRepository with ProtobufConversions with M
       |  FROM zone
        """.stripMargin
 
+  private final val GET_ZONE_ACCESS_BY_ADMIN_GROUP_ID =
+    sql"""
+         |SELECT DISTINCT zone_id
```
mitruly: This is more of a comment: I think this is fine because my understanding is that `zone_access` isn't really used (anymore). Otherwise if group/user data gets really large, could potentially run into a scenario where `accessor_id` is really a `user_id` and we encounter false positives.

modules/api/src/test/scala/vinyldns/api/domain/membership/MembershipServiceSpec.scala
```diff
@@ -488,6 +491,16 @@ class MembershipServiceSpec
...
       }
 
+      "return an error if the group has an ACL rule on a zone" in {
+        doReturn(IO.pure(Some(okGroup))).when(mockGroupRepo).getGroup(anyString)
+        doReturn(IO.pure(Some("someId")))
+          .when(mockZoneRepo)
+          .getFirstOwnedZoneAclGroupId(anyString())
+        val error = leftResultOf(underTest.deleteGroup("ok", okAuth).value)
+
+        error shouldBe a[InvalidGroupRequestError]
```
mitruly: `a[InvalidGroupRequestError]` → `an[InvalidGroupRequestError]`

modules/api/src/test/scala/vinyldns/api/domain/membership/MembershipServiceSpec.scala
```diff
@@ -833,6 +846,22 @@ class MembershipServiceSpec
       }
     }
 
+    "isNotZoneAclGroupId" should {
+      "return true when a groupId is not in any zone ACL" in {
```
mitruly: Can we change the test sub-title to "return successfully when a groupID is not in any zone ACL"? Technically it returns `Right(Unit)` and `true` might be confusing.

modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
varsharun: `zones` in other context is the actual List[Zone] but here its just an id.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneRepository.scala
pauljamescleary: @mitruly zone access is used for permissions / access control checks.  It is wedged down in where we build the zone search

modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
mitruly: Whoops, I meant `(zId.isEmpty)`

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneRepository.scala
mitruly: Thanks for the clarification, @pauljamescleary.

PR #548, opened 2019-03-25
Delete Group - Delete the Group from DynamoDB

### LINKED ISSUE #189: Delete DynamoDB Groups
**Motivation**
Currently, we are not deleting groups from DynamoDB, we are just marking them as "Deleted".  The issue is that we are repeatedly running black box tests creating and deleting groups.  Overtime, these unused "deleted" groups will impact query performance and table cost.

**Scope of change**
Delete groups when they are deleted.

varsharun:
Fixes #189 #436 .

Changes in this pull request:
- Modify the `delete()` method to actually delete the group from Dynamo

### DIFF
--- a/modules/api/functional_test/live_tests/batch/get_batch_change_test.py
+++ b/modules/api/functional_test/live_tests/batch/get_batch_change_test.py
@@ -64,7 +64,7 @@ def test_get_batch_change_with_record_owner_group_success(shared_zone_test_conte
             client.wait_until_recordset_change_status(delete_result, 'Complete')
 
 
-def test_get_batch_change_with_deleted_record_owner_group_failed(shared_zone_test_context):
+def test_get_batch_change_with_deleted_record_owner_group_success(shared_zone_test_context):
     """
     Test that if the owner group no longer exists that getting a batch change will still succeed,
     with the ownerGroupName attribute set to None
@@ -80,7 +80,9 @@ def test_get_batch_change_with_deleted_record_owner_group_failed(shared_zone_tes
 
     record_to_delete = []
     try:
+
         group_to_delete = client.create_group(temp_group, status=200)
+
         batch_change_input = {
             "comments": "this is optional",
             "changes": [
@@ -95,7 +97,6 @@ def test_get_batch_change_with_deleted_record_owner_group_failed(shared_zone_tes
         record_set_list = [(change['zoneId'], change['recordSetId']) for change in completed_batch['changes']]
         record_to_delete = set(record_set_list)
 
-<<<<<<< HEAD
         #delete records and owner group
         temp = record_to_delete.copy()
         for result_rs in temp:
@@ -105,18 +106,13 @@ def test_get_batch_change_with_deleted_record_owner_group_failed(shared_zone_tes
         temp.clear()
 
         client.delete_group(group_to_delete['id'], status=200)
-=======
-        # delete group should not be possible
-        rdDelete = client.delete_group(group_to_delete['id'], status=400)
-        assert_that(rdDelete, contains_string("Cannot delete"))
->>>>>>> Functional test modification
         del completed_batch['ownerGroupName']
 
         #the batch should not be updated with deleted group data
         result = client.get_batch_change(batch_change['id'], status=200)
-        assert_that(result, is_not(completed_batch))
+        assert_that(result, is_(completed_batch))
         assert_that(result['ownerGroupId'], is_(group_to_delete['id']))
-        assert_that(result, is_(has_key('ownerGroupName')))
+        assert_that(result, is_not(has_key('ownerGroupName')))
 
     finally:
         for result_rs in record_to_delete:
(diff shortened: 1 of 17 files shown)


modules/dynamodb/src/main/scala/vinyldns/dynamodb/repository/DynamoDBGroupRepository.scala
```diff
@@ -99,7 +99,11 @@ class DynamoDBGroupRepository private[repository] (
     }
 
   def delete(group: Group): IO[Group] =
-    save(group.copy(status = GroupStatus.Deleted))
+    monitor("repo.Group.delete") {
+      log.info(s"Deleting group ${group.id} ${group.name}.")
+      val request = new DeleteItemRequest().withTableName(groupTableName)
+      dynamoDBHelper.deleteItem(request).map(_ => group)
```
nimaeskandary: Don't think this is creating the correct delete request, the delete request also needs to the items key, in this case something like
```
val key = new java.util.HashMap[String, AttributeValue]()
item.put(GROUP_ID, new AttributeValue(group.id))
```
Can that be added and have the integration spec verify the group is no longer in the db after the delete is complete?

varsharun: yeah Im not using the `group` anywhere 👎 will change this thanks @nimaeskandary :)

modules/dynamodb/src/it/scala/vinyldns/dynamodb/repository/DynamoDBGroupRepositoryIntegrationSpec.scala
```diff
@@ -165,5 +165,29 @@ class DynamoDBGroupRepositoryIntegrationSpec extends DynamoDBIntegrationSpec {
 
       test.unsafeRunSync().get.description shouldBe None
     }
+
+    "delete a group" in {
```
rebstar6: expand this test a little. wanna make sure that it's only deleting the one group you specified, not all the groups in the table.

put 3 groups, delete 1 or 2 of them, and make sure just the one you deleted was removed

modules/dynamodb/src/it/scala/vinyldns/dynamodb/repository/DynamoDBGroupRepositoryIntegrationSpec.scala
```diff
@@ -165,5 +165,29 @@ class DynamoDBGroupRepositoryIntegrationSpec extends DynamoDBIntegrationSpec {
...
+
+      val getAfterDeleted =
+        for {
+          get <- repo.getGroup("test-deleted-group-get-groups")
+          getAll <- repo.getAllGroups()
+        } yield (get, getAll)
+
+      val (get, getAll) = getAfterDeleted.unsafeRunSync()
+      get shouldBe None
+      !getAll.exists(_.id == "test-deleted-group-get-groups")
```
rebstar6: to my comment above:
you are close here. if there is stuff seeded in the table, you can just look for one specific item here to confirm that that item is still in the table (rather than having to add new things)

modules/api/functional_test/live_tests/batch/get_batch_change_test.py
```diff
@@ -64,7 +64,7 @@ def test_get_batch_change_with_record_owner_group_success(shared_zone_test_conte
             client.wait_until_recordset_change_status(delete_result, 'Complete')
 
 
-def test_get_batch_change_with_deleted_record_owner_group_success(shared_zone_test_context):
+def test_get_batch_change_with_deleted_record_owner_group_failed(shared_zone_test_context):
```
mitruly: Can we use present tense here to stay consistent with test naming? Either `test_get_batch_change_with_deleted_record_owner_group_failure` or `test_get_batch_change_with_deleted_record_owner_group_fails` would be good.

modules/api/functional_test/live_tests/batch/get_batch_change_test.py
```diff
@@ -80,9 +80,7 @@ def test_get_batch_change_with_deleted_record_owner_group_success(shared_zone_te
 
     record_to_delete = []
     try:
-
         group_to_delete = client.create_group(temp_group, status=200)
-
```
mitruly: We need `group_to_delete = None` above this `try` statement so that we can delete the group at the end in the `finally` clause (so that we can run this test multiple times in succession without hitting a `409`):
```
        if group_to_delete:
            client.delete_group(group_to_delete['id'])
```

modules/api/src/test/scala/vinyldns/api/repository/EmptyRepositories.scala
```diff
@@ -49,6 +49,8 @@ trait EmptyRecordSetRepo extends RecordSetRepository {
   def getRecordSetCount(zoneId: String): IO[Int] = IO.pure(0)
 
   def getRecordSetsByFQDNs(names: Set[String]): IO[List[RecordSet]] = IO.pure(List())
+
+  def getRecordSetOwnerGroup(ownerGroupId: String): IO[String] = IO.pure("")
```
mitruly: These should be `IO[Option[String]]` (similar to other PR comments).

PR #550, opened 2019-03-26
set recordset as unimplemented in dynamo

rebstar6:


rebstar6:
will followup with config-api.md change, i wanna get this in for rebases

Issue #477, opened 2019-02-06
Large record data can look strange in the UI


pauljamescleary:
![image](https://user-images.githubusercontent.com/17361753/52350515-c5477900-29f6-11e9-8f59-9a93f40b9408.png)

To replicate:
- In the portal, make a new record in a zone with long record data. Probably best to make a TXT record type with record data that is 256 characters (the maximum allowed).

To fix:
This may be resolved by restricting the width of the Record Data column and wrapping the text. Open to other suggestions to handle displaying the long data. The table is found in `modules/portal/views/zones/zoneTabs/manageRecords.scala.html`

pauljamescleary:
Oof, unsure how to wrap that but we should wrap if possible.

We may want to consider elipses --to--> popup for more info (expand / collapse), something along those lines

britneywright:
@pauljamescleary i implemented wrapping in the PR #637

PR #555, opened 2019-03-29
multi-record support for txt in portal and batch

### LINKED ISSUE #547: Multi-record support for TXT records
**Describe the bug**
Presently, VinylDNS does not support multiple records for TXT records **in the UI**.  It is supported everywhere else in the system

**VinylDNS Version**
Latest

**Expected behavior**
Support a multi-record record set for TXT record types. 

Should be supported in the UI and the API

Should support large TXT rdata, as it can be 65000 characters

britneywright:
Fixes #547.

Changes in this pull request:
- allow input of multi-record TXT record sets in the portal
- display multi-record TXT record sets in the portal
- change the "more..." and "less.." links in multi-record record sets from list level items to span and rename as "Show more..." and "Show fewer..."

Screen Shot 2019-03-29 at 2 25 01 PM

Screen Shot 2019-03-29 at 2 23 54 PM

Screen Shot 2019-03-29 at 2 24 06 PM

API:

- allow multiple TXT records with the same record name in batch changes


pauljamescleary:
@britneywright thanks for the screen shots.  Is the multi-record TEXT just separated by newline in the same text area?

I think we may need to review the design.  Each TXT record in a record set has the potential to be very large (65,000 characters).

britneywright:
> @britneywright thanks for the screen shots. Is the multi-record TEXT just separated by newline in the same text area?
> 
> I think we may need to review the design. Each TXT record in a record set has the potential to be very large (65,000 characters).

Correct, they're separated by new lines. I can test long record data and see how they work.

britneywright:
Realized we weren't allowing TXT records with the same input name in batch changes on the API side, updated that as well.

britneywright:
> Confirmed that this works with #568, but concerned that we're not wrapping the columns which stretches the layout.

That's a separate issue https://github.com/vinyldns/vinyldns/issues/477

PR #561, opened 2019-04-01
docs add singleBatchChangeIds attribute to recordset change

### LINKED ISSUE #551: docs: add "singleBatchChangeIds" attribute to RecordSetChanges
noticed "singleBatchChangesIds" was missing in the example response in https://www.vinyldns.io/api/create-recordset.html. We may also want to note that the response is in fact a recordSetChange and not a recordSet.

Also need to update:

- https://www.vinyldns.io/api/update-recordset.html
- https://www.vinyldns.io/api/get-recordset-change.html
- https://www.vinyldns.io/api/list-recordset-changes.html

britneywright:
Fixes #551.

Changes in this pull request:
- update response attributes for:
 - create recordset
 - get recordset change
 - list recordset change
 - update recordset
- update response examples for:
 - create recordset
 - get recordset change
 - update recordset

note: list recordset change does not currently include singleBatchChangeIds in the response.


modules/docs/src/main/tut/api/create-recordset.md
```diff
@@ -61,6 +61,8 @@ userId        | string        | the user id that initiated the change |
 changeType    | string        | type of change requested (Create, Update, Delete); in this case Create |
 created       | string        | the time (GMT) the change was initiated |
 id            | string        | the id of the change.  This is not the id of the recordset |
+status        | string        | the status of the change (Pending, Complete, Failed) |
+singleBatchChangeIds |  array of IDs  | if the recordset change was part of a batch change, the IDs of the single changes that comprise the recordset change
```
mitruly: We use `BatchChangeStatus` and `SingleChangeStatus` as types in `get-batchchange.md` and other docs, so we should probably have this type be `RecordSetChangeStatus`.

Having the type of `singleBatchChangeIds` as `array of SingleBatchChange Id objects` or `array of strings` might make more sense to be consistent with the other pages (eg. in `get-group.md`)

modules/docs/src/main/tut/api/list-recordset-changes.md
```diff
@@ -97,7 +98,8 @@ maxItems      | int           | The maxItems parameter that was sent in on the H
           "type": "CNAME",
           "id": "f62235df-5372-443c-9ba4-bdd3fca452f4"
         },
-        "id": "68fd6dbe-0da8-4280-bcf3-37f54528dc41"
+        "id": "68fd6dbe-0da8-4280-bcf3-37f54528dc41",
+        "singleBatchChangeIds": []
```
mitruly: Wonder if we should populate this with some sample UUIDs since this is one of the few endpoints where this would potentially be non-empty (as normal create record set endpoint would not create it since the record set needs to be created through batch change to populate a single batch change ID).

britneywright: I meant to take these out for now. we don't currently include singleBatchChangeIds in the list recordset changes response.

PR #562, opened 2019-04-01
[WIP] portal: sort columns in records table

### LINKED ISSUE #541: Add Sort feature in Record Table columns
I'm unable to sort columns in a zone's Records table. All entries get sorted by the their name, by default.

Add a sort feature to each column in the table

N/A

**Additional context**
N/A

britneywright:
Fixes #541.

Changes in this pull request:
- adds sorting to Name, Type and TTL columns fo the records table in the portal

order by name:
Screen Shot 2019-04-01 at 12 51 24 PM

order by name, reverse: 
Screen Shot 2019-04-01 at 12 51 33 PM

### DIFF
--- a/modules/portal/app/views/zones/zoneTabs/manageRecords.scala.html
+++ b/modules/portal/app/views/zones/zoneTabs/manageRecords.scala.html
@@ -98,9 +98,21 @@
       <table id="recordDataTable" class="table table-hover table-striped">
             <thead>
                 <tr>
-                    <th>Name</th>
-                    <th>Type</th>
-                    <th>TTL</th>
+                    <th>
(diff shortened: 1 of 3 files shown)


pauljamescleary:
@britneywright this is portal only, right?  We would need to add sorting to the API as well

mitruly:
To @pauljamescleary's point on needing the API to handle sorting, when reducing the record paging parameter (eg. `var recordsPaging = pagingService.getNewPagingParams(2);`) in `controller.records.js`, the sorting only affects the records that get returned by the API for that particular page.

britneywright:
ah yeah forgot about the api pagination 🤦🏽‍♀️

britneywright:
Closing for now, need to work on the API side

PR #568, opened 2019-04-03
correctly handle txt record data

### LINKED ISSUE #563: TXT records exceeding 255 characters do not fail properly
**Describe the bug**
When creating a TXT record with too many characters, the record fails in processing (an UPDATE is issued) as opposed to failing up front on record submission

**VinylDNS Version**
8.0.0

**To Reproduce**
1. Go to a zone
2. Click on Create RecordSet
3. Choose TXT record
4. Enter any name for the record name
5. Enter the following long value for the record data...
6. Click the Save button
7. Record is submitted, and fails in verification

```
This is an example of a really long string that I wrote let us see how many characters can actually flow through to the backend dns server before the entire thing fails and blows up maybe in processing maybe up front but that really doesnt matter as the most important point is that we have to exceed 300 characters in the entire string so that we can try and force some kind of issue to happen in the backend itself and 65000 characters is indeed a lot of characters I dont think I will get that far but let us see
```

**Expected behavior**
VinylDNS should _split_ the very long string into 255 character chunks.  The entire string has to be less than 65000 characters (including extra quotes and spaces).

The very long string above should be processed properly, and return as a set of 255 character strings instead.

The dnsjava library can accept a `String[]` array during creation.  We should use that constructor for `TXTRecord` instead.

**Screenshots**

**Stack trace or error log output**

```
2019-04-01 20:05:43,800 [denis] INFO  | [scala-execution-context-global-22] (v.a.d.d.DnsConnection:163) | AKKA_SOURCE= | DnsConnection.send - Sending DNS Message ;; ->>HEADER<<- opcode: UPDATE, status: NOERROR, id: 60228
;; flags: ; qd: 1 an: 0 au: 0 ad: 0 
;; ZONE:
;;	denis.comcast.net., type = SOA, class = IN

;; PREREQUISITES:

;; UPDATE RECORDS:

;; ADDITIONAL RECORDS:

;; Message size: 0 bytes
...received response Right(NoError(;; ->>HEADER<<- opcode: UPDATE, status: NOERROR, id: 49753
;; flags: qr ; qd: 1 an: 0 au: 0 ad: 0 
;; ZONE:
;;	denis.comcast.net., type = SOA, class = IN

;; PREREQUISITES:

;; UPDATE RECORDS:

;; ADDITIONAL RECORDS:

;; Message size: 119 bytes)) 
```

**Additional context**

There is a bug in how we convert from Vinyl Record Sets to dnsjava record sets in `DnsConversions`...

```scala
  def toDnsRRset(recordSet: RecordSet, zoneName: String): Either[Throwable, DNS.RRset] =
    Either.catchNonFatal {
      val dnsRecordSet = new DNS.RRset()
      toDnsRecords(recordSet, zoneName).map(_.foreach(dnsRecordSet.addRR))

      dnsRecordSet
    }
```

In the above code, `toDnsRecords` actually fails `Left` properly.  However, we do `map` which is not fired, and then `_.foreach` in there.  The result is that we convert a **bad** `TXT` record to an **empty** record set that we attempt to add.

The DNS backend gives us a `NOERROR` because there is nothing to do.  We fail in `Verify` step because nothing has actually happened.

We should fix this bug here, as it is silently swallowing issues **bad**.  But we should fail in `DnsJsonProtocol` when we parse the `TXT` record so it never gets to processing at all.

britneywright:
Fixes #563.

Changes in this pull request:
- group TXT record data as a list of 255 character length strings
- also updated the `toDnsRRset` so it doesn't swallow errors

### DIFF
--- a/modules/api/functional_test/live_tests/recordsets/create_recordset_test.py
+++ b/modules/api/functional_test/live_tests/recordsets/create_recordset_test.py
@@ -1517,6 +1517,7 @@ def test_create_long_txt_record_succeeds(shared_zone_test_context):
     try:
         rs_create = client.create_recordset(long_txt_rs, status=202)
         rs = client.wait_until_recordset_change_status(rs_create, 'Complete')['recordSet']
+        assert_that(rs['records'][0]['text'], is_(record_data))
     finally:
         try:
             delete_result = client.delete_recordset(rs['zoneId'], rs['id'], status=202)
--- a/modules/api/functional_test/live_tests/recordsets/create_recordset_test.py
+++ b/modules/api/functional_test/live_tests/recordsets/create_recordset_test.py
@@ -1507,6 +1507,23 @@ def test_create_record_with_existing_cname_wildcard_succeed(shared_zone_test_con
                 pass
 
 
+def test_create_long_txt_record_succeeds(shared_zone_test_context):
+    client = shared_zone_test_context.ok_vinyldns_client
+
+    zone = shared_zone_test_context.system_test_zone
+    record_data = 'a' * 64763
+    long_txt_rs = get_recordset_json(zone, 'long-txt-record', 'TXT', [{'text': record_data}])
+
+    try:
+        rs_create = client.create_recordset(long_txt_rs, status=202)
+        rs = client.wait_until_recordset_change_status(rs_create, 'Complete')['recordSet']
+    finally:
+        try:
+            delete_result = client.delete_recordset(rs['zoneId'], rs['id'], status=202)
+            client.wait_until_recordset_change_status(delete_result, 'Complete')
+        except:
+            pass
+
 def test_dotted_host_create_fails(shared_zone_test_context):
     """
     Tests that a dotted host recordset create fails
--- a/modules/api/src/main/scala/vinyldns/api/domain/dns/DnsConversions.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/dns/DnsConversions.scala
@@ -274,7 +274,7 @@ trait DnsConversions {
 
   def fromTXTRecord(r: DNS.TXTRecord, zoneName: DNS.Name, zoneId: String): RecordSet =
     fromDnsRecord(r, zoneName, zoneId) { data =>
-      List(TXTData(data.getStrings.asScala.mkString(",")))
+      List(TXTData(data.getStrings.asScala.mkString("")))
     }
 
   def toDnsRecords(recordSet: RecordSet, zoneName: String): Either[Throwable, List[DNS.Record]] =
@@ -346,17 +346,18 @@ trait DnsConversions {
           new DNS.SPFRecord(recordName, DNS.DClass.IN, ttl, text)
 
         case TXTData(text) =>
-          new DNS.TXTRecord(recordName, DNS.DClass.IN, ttl, text)
+          val texts = text.grouped(255).toList
+          new DNS.TXTRecord(recordName, DNS.DClass.IN, ttl, texts.asJava)
       }
     }
 
-  def toDnsRRset(recordSet: RecordSet, zoneName: String): Either[Throwable, DNS.RRset] =
-    Either.catchNonFatal {
-      val dnsRecordSet = new DNS.RRset()
-      toDnsRecords(recordSet, zoneName).map(_.foreach(dnsRecordSet.addRR))
-
+  def toDnsRRset(recordSet: RecordSet, zoneName: String): Either[Throwable, DNS.RRset] = {
+    val dnsRecordSet = new DNS.RRset()
+    toDnsRecords(recordSet, zoneName).map { thing =>
+      thing.foreach(dnsRecordSet.addRR)
       dnsRecordSet
     }
+  }
 
   def toDnsRecordType(typ: RecordType): Integer = typ match {
     case RecordType.A => DNS.Type.A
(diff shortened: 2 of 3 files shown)


modules/api/src/test/scala/vinyldns/api/domain/dns/DnsConversionsSpec.scala
```diff
@@ -336,6 +345,11 @@ class DnsConversionsSpec
       val result = rightValue(toDnsRRset(testTXT, testZoneName))
       verifyMatch(result, testTXT)
     }
+
+    "fail to convert a bad SPF record set" in {
```
britneywright: SPF might need to be updated like TXT record data, but I'm putting this test in for now to confirm that we aren't swallowing errors that might occur during the `toDnsRRset` function. we can create a separate issue and PR to fix SPF and then this test can be removed or modified.

modules/api/src/main/scala/vinyldns/api/domain/dns/DnsConversions.scala
```diff
@@ -274,7 +274,7 @@ trait DnsConversions {
 
   def fromTXTRecord(r: DNS.TXTRecord, zoneName: DNS.Name, zoneId: String): RecordSet =
     fromDnsRecord(r, zoneName, zoneId) { data =>
-      List(TXTData(data.getStrings.asScala.mkString(",")))
+      List(TXTData(data.getStrings.asScala.mkString("")))
```
mitruly: No need to pass in empty string since that is default behavior; can just be:
`List(TXTData(data.getStrings.asScala.mkString))`

britneywright: fixed

modules/api/src/main/scala/vinyldns/api/domain/dns/DnsConversions.scala
```diff
@@ -346,17 +346,18 @@ trait DnsConversions {
...
 
-  def toDnsRRset(recordSet: RecordSet, zoneName: String): Either[Throwable, DNS.RRset] =
-    Either.catchNonFatal {
-      val dnsRecordSet = new DNS.RRset()
-      toDnsRecords(recordSet, zoneName).map(_.foreach(dnsRecordSet.addRR))
-
+  def toDnsRRset(recordSet: RecordSet, zoneName: String): Either[Throwable, DNS.RRset] = {
+    val dnsRecordSet = new DNS.RRset()
+    toDnsRecords(recordSet, zoneName).map { thing =>
+      thing.foreach(dnsRecordSet.addRR)
```
pauljamescleary: the variable name `thing` is a longer version of `x`.  When naming variables, if we really do not care about the variable then `x` is preferable.  In this case, we are getting a `List[DNS.Record]` as the thing, so a name like `records` would be more appropriate.

modules/api/src/test/scala/vinyldns/api/domain/dns/DnsConversionsSpec.scala
pauljamescleary: SPF is not implemented correctly, I agree that is a separate PR

modules/api/src/test/scala/vinyldns/api/domain/dns/DnsConversionsSpec.scala
```diff
@@ -124,6 +124,15 @@ class DnsConversionsSpec
     DateTime.now,
     None,
     List(SPFData("spf")))
+  private val testLongSPF = RecordSet(
```
pauljamescleary: Add test for `fromTXTRecord` that it does the `mkString` appropriately combining the elements

modules/api/src/test/scala/vinyldns/api/domain/dns/DnsConversionsSpec.scala
rebstar6: oh yeah, SPF isnt even supposed to be a type, its a sub of TXT...

rebstar6: for later

modules/api/functional_test/live_tests/recordsets/create_recordset_test.py
```diff
@@ -1507,6 +1507,23 @@ def test_create_record_with_existing_cname_wildcard_succeed(shared_zone_test_con
...
+def test_create_long_txt_record_succeeds(shared_zone_test_context):
+    client = shared_zone_test_context.ok_vinyldns_client
+
+    zone = shared_zone_test_context.system_test_zone
+    record_data = 'a' * 64763
+    long_txt_rs = get_recordset_json(zone, 'long-txt-record', 'TXT', [{'text': record_data}])
+
+    try:
+        rs_create = client.create_recordset(long_txt_rs, status=202)
+        rs = client.wait_until_recordset_change_status(rs_create, 'Complete')['recordSet']
```
pauljamescleary: add a `get` once it is created and compare / assert on the record data.

modules/api/functional_test/live_tests/recordsets/create_recordset_test.py
```diff
@@ -1507,6 +1507,24 @@ def test_create_record_with_existing_cname_wildcard_succeed(shared_zone_test_con
...
+        assert_that(rs['records'][0]['text'], is_(record_data))
+    finally:
```
mitruly: To @pauljamescleary's point, should add something like
```
        rs_result = client.get_recordset(rs['zoneId'], rs['id'], status=200)['recordSet']
        assert_that(rs_result['records'][0]['text'], is_(record_data))
```
to the end of the `try`.

mitruly: Nevermind; appears that this was addressed.

PR #569, opened 2019-04-04
upgrade to scala 2.12.18

nimaeskandary:
Pretty much ready to submit a pr with the scala js portal rewrite I've been working on. That branch is feature flagged and adds a `/v2` route to the existing portal when the flag is turned on

scala js has a known bug with 2.12.17 so want to get this first commit in first to make sure travis is good with 2.12.18

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: scala
 
 scala:
   - 2.12.7
+  - 2.12.8
 
 python:
   - 2.7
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -2,6 +2,7 @@ language: scala
 
 scala:
   - 2.12.7
```
mitruly: Should we just bump this up instead of adding one? Travis CI is adding another job (https://travis-ci.org/vinyldns/vinyldns/builds/515889680?utm_source=github_status&utm_medium=notification) to this, heh

PR #574, opened 2019-04-09
Update Gentelella template

britneywright:
Changes in this pull request:
- update gentelella to use latest from Github
- inject oidc-finish script in html, this must've missed in a previous PR. caught when running `preparePortal`

### DIFF
--- a/modules/portal/package.json
+++ b/modules/portal/package.json
@@ -13,7 +13,7 @@
     "angular-ui-bootstrap": "2.5.6",
     "bootstrap": "4.1.2",
     "font-awesome": "4.7.0",
-    "gentelella": "1.4.0",
+    "gentelella": "github:colorlibhq/gentelella",
     "grunt": "^1.0.3",
     "grunt-angular-templates": "^1.1.0",
     "grunt-contrib-clean": "^1.0.0",
--- a/modules/portal/package.json
+++ b/modules/portal/package.json
@@ -13,7 +13,7 @@
     "angular-ui-bootstrap": "2.5.6",
     "bootstrap": "4.1.2",
     "font-awesome": "4.7.0",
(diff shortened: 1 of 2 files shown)


modules/portal/package.json
```diff
@@ -13,7 +13,7 @@
     "angular-ui-bootstrap": "2.5.6",
     "bootstrap": "4.1.2",
     "font-awesome": "4.7.0",
-    "gentelella": "1.4.0",
+    "gentelella": "github:colorlibhq/gentelella",
```
mitruly: Depending on the timeline of Gentelella releasing a tag, I'd be a lot more comfortable pinpointing an actual release tag once they publish a (minor) version with these changes so that:
- Users who are pulling our portal will consistently have the same code changes
- We don't forget to fix a version when a tagged release actually gets published to npm

britneywright: I've asked if they plan to do a new release, we'll see what they say. https://github.com/ColorlibHQ/gentelella/pull/790#issuecomment-481370850

rebstar6: you can probably pin to a specific commit hash - have a google on that, would be better than master

modules/portal/package.json
```diff
@@ -13,7 +13,7 @@
     "angular-ui-bootstrap": "2.5.6",
     "bootstrap": "4.1.2",
     "font-awesome": "4.7.0",
-    "gentelella": "1.4.0",
+    "gentelella": "git@github.com:colorlibhq/gentelella.git#eea0150c65d51cb1f1a03f18bcc3bfa2a8db9312",
```
mitruly: `modules/portal/prepare-portal.sh` hangs at this stage for me, so I don't think this is working as expected:
`loadDep:grunt-injector: sill install loadAllDepsIntoIdealTree`

PR #575, opened 2019-04-09
increase batch change limit default to 1000

britneywright:
follow-up to #509.

now that we've migrated recordsets to mysql and get existing recordsets more efficiently we can increase the batch change limit.

Changes in this pull request:
- set default batch-change-limit to 1000
- remove tests that check for specific limit (every instance of VinylDNS doesn't have to have the same batch change limit!)

### DIFF
--- a/docker/api/docker.conf
+++ b/docker/api/docker.conf
@@ -189,7 +189,7 @@ vinyldns {
     }
   ]
 
-  batch-change-limit = 20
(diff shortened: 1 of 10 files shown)


britneywright:
> If we're changing docs as well, `create-batchchange.md` has a reference for the `413` error; otherwise I'm good.

updated

PR #576, opened 2019-04-09
[WIP]Relax domain name validation requirements for batch

varsharun:
Fixes #.

Changes in this pull request:
- TB updated soon
-


modules/api/src/main/scala/vinyldns/api/domain/DomainValidations.scala
```diff
@@ -25,7 +25,7 @@ import scala.util.matching.Regex
  */
 object DomainValidations {
   val validFQDNRegex: Regex =
-    """^(?:([0-9a-zA-Z]{1,63}|[0-9a-zA-Z]{1}[0-9a-zA-Z\-\/]{0,61}[0-9a-zA-Z]{1})\.)*$""".r
+    """^(?:([0-9a-zA-Z_]{1,63}|[0-9a-zA-Z]{1}[0-9a-zA-Z\-\/_]{0,61}[0-9a-zA-Z]{1})\.)*$""".r
```
mitruly: Just a note: I think we should also allow underscores for the other list of possible characters; at least for the starting character in case the zone starts with an underscore (eg. `_some.domain.`):
```
"""^(?:([0-9a-zA-Z_]{1,63}|[0-9a-zA-Z_]{1}[0-9a-zA-Z\-\/_]{0,61}[0-9a-zA-Z_]{1})\.)*$""".r
```

Not sure if it's valid, but I was able to create a record named `___` against a local instance of VinylDNS.

varsharun:
Closing this PR. There are future work required regarding Batch changes

PR #579, opened 2019-04-11
Remove distinct from zone sql queries

varsharun:
Fixes #.

Changes in this pull request:
- Removed `DISTINCT` from GET_ZONES_BY_ADMIN_GROUP_ID
- Removed `DISTINCT` from BASE_ZONE_SEARCH_SQL
- Removed `DISTINCT` from GET_ZONE_ACCESS_BY_ADMIN_GROUP_ID

### DIFF
--- a/modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneRepository.scala
+++ b/modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneRepository.scala
@@ -98,14 +98,14 @@ class MySqlZoneRepository extends ZoneRepository with ProtobufConversions with M
 
   private final val GET_ZONES_BY_ADMIN_GROUP_ID =
     sql"""
-       |SELECT DISTINCT id, data
+       |SELECT data
        |  FROM zone z
        | WHERE z.admin_group_id = (?)
         """.stripMargin
 
   private final val BASE_ZONE_SEARCH_SQL =
     """
-      |SELECT DISTINCT z.id, z.data, z.name
+      |SELECT z.data
       |  FROM zone z
        """.stripMargin
 
@@ -117,7 +117,7 @@ class MySqlZoneRepository extends ZoneRepository with ProtobufConversions with M
 
   private final val GET_ZONE_ACCESS_BY_ADMIN_GROUP_ID =
     sql"""
-         |SELECT DISTINCT zone_id
+         |SELECT zone_id
          |  FROM zone_access z
          | WHERE z.accessor_id = (?)
          | LIMIT 1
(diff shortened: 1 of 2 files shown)


modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneRepository.scala
```diff
@@ -98,14 +98,14 @@ class MySqlZoneRepository extends ZoneRepository with ProtobufConversions with M
...
-       |SELECT DISTINCT id, data
+       |SELECT data
        |  FROM zone z
        | WHERE z.admin_group_id = (?)
         """.stripMargin
 
   private final val BASE_ZONE_SEARCH_SQL =
     """
-      |SELECT DISTINCT z.id, z.data, z.name
+      |SELECT z.id, z.data, z.name
```
pauljamescleary: Perhaps we can lose `z.id` and `z.name` here?

varsharun: @pauljamescleary In the issue 117 I read the following 
The new way adds the zone name for paging and removes the offset LIMIT...

```
SELECT z.id, z.data, z.name
  FROM zone z
        JOIN zone_access za ON z.id = za.zone_id
  AND za.accessor_id IN (?,?,?,?)
     AND z.name > ?  ORDER BY z.name ASC LIMIT ?
```

This is the SQL that generates the above, hence left the id and name. Should I still remove it?

pauljamescleary: You should be able to remove it from the `SELECT` clause (leave it in the rest of the query).  The reason is that we are returning zone id and name from the database call and not using them (taking up network bandwidth and time)

varsharun: Oh alright thanks 👍

rebstar6: arent we gonna need name for paging at some point anyway though?

rebstar6: guess you could bring it back then 🤷‍♀️

pauljamescleary: name is in the protobuf

pauljamescleary: RIght, we already have it in the protobuf, so it will be there for serialization

PR #582, opened 2019-04-15
Chunk up zone sync logging and drop records prior to deserialization into RecordSet model

### LINKED ISSUE #581: Group logging into fixed sizes
When attempting a zone sync (either directly or through zone connect), unsupported records are flagged as `Unknown` and dropped; the issue is that as we don't limit the number of records when logging, dropping 100K+ records is problematic for our log ingest.

We should also confirm that `Unknown` records are dropped.

*Edit*: We should move up the dropping of unknown records to before records get deserialized.

mitruly:
Fixes #581.

Changes in this pull request:
- Group logging into smaller chunks
- Move record drop logic from `ZoneSyncHandler` to `DnsZoneViewLoader` (prior to deserialization of `DNS.Record` type to VinylDNS' `RecordSet` model)

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneViewLoader.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneViewLoader.scala
@@ -17,12 +17,13 @@
 package vinyldns.api.domain.zone
 
 import cats.effect._
+import org.slf4j.{Logger, LoggerFactory}
 import org.xbill.DNS
 import org.xbill.DNS.{TSIG, ZoneTransferIn}
 import vinyldns.api.VinylDNSConfig
 import vinyldns.api.crypto.Crypto
 import vinyldns.api.domain.dns.DnsConversions
-import vinyldns.core.domain.record.RecordSetRepository
+import vinyldns.core.domain.record.{RecordSetRepository, RecordType}
 import vinyldns.core.domain.zone.Zone
 import vinyldns.core.route.Monitored
 
@@ -33,6 +34,7 @@ trait ZoneViewLoader {
 }
 
 object DnsZoneViewLoader extends DnsConversions {
+  private implicit val logger: Logger = LoggerFactory.getLogger("vinyldns.engine.ZoneSyncHandler")
 
   def dnsZoneTransfer(zone: Zone): ZoneTransferIn = {
     val conn =
@@ -59,6 +61,7 @@ case class DnsZoneViewLoader(zone: Zone, zoneTransfer: Zone => ZoneTransferIn)
     extends ZoneViewLoader
     with DnsConversions
     with Monitored {
+  import DnsZoneViewLoader._
 
   def load: () => IO[ZoneView] =
     () =>
@@ -70,8 +73,29 @@ case class DnsZoneViewLoader(zone: Zone, zoneTransfer: Zone => ZoneTransferIn)
           val rawDnsRecords: List[DNS.Record] =
             xfr.getAXFR.asScala.map(_.asInstanceOf[DNS.Record]).toList.distinct
 
+          // not accepting unknown record types
+          val (supportedRecords, droppedRecords) = rawDnsRecords.partition { record =>
+            fromDnsRecordType(record.getType) != RecordType.UNKNOWN
+          }
+
           val dnsZoneName = zoneDnsName(zone.name)
-          val recordSets = rawDnsRecords.map(toRecordSet(_, dnsZoneName, zone.id))
+
+          if (droppedRecords.nonEmpty) {
+            droppedRecords
+              .grouped(1000)
+              .foreach { droppedGroup =>
+                val droppedInfo = droppedGroup
+                  .map(record =>
+                    relativize(record.getName, dnsZoneName) + " " + fromDnsRecordType(
+                      record.getType))
+                  .mkString(", ")
+
+                logger.warn(
+                  s"Zone sync for zone [$dnsZoneName] dropped ${droppedGroup.length} recordsets: $droppedInfo")
+              }
+          }
+
+          val recordSets = supportedRecords.map(toRecordSet(_, dnsZoneName, zone.id))
 
           ZoneView(zone, recordSets)
         }
--- a/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneViewLoader.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneViewLoader.scala
@@ -17,7 +17,6 @@
 package vinyldns.api.domain.zone
 
 import cats.effect._
-import org.slf4j.{Logger, LoggerFactory}
 import org.xbill.DNS
 import org.xbill.DNS.{TSIG, ZoneTransferIn}
 import vinyldns.api.VinylDNSConfig
@@ -34,8 +33,6 @@ trait ZoneViewLoader {
 }
 
 object DnsZoneViewLoader extends DnsConversions {
-  private val logger: Logger = LoggerFactory.getLogger("vinyldns.api.domain.zone.DnsZoneViewLoader")
-
   def dnsZoneTransfer(zone: Zone): ZoneTransferIn = {
     val conn =
       ZoneConnectionValidator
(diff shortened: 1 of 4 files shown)


modules/api/src/main/scala/vinyldns/api/engine/ZoneSyncHandler.scala
```diff
@@ -123,9 +127,12 @@ object ZoneSyncHandler extends DnsConversions with Monitored {
               .map(_.recordSet.name)
               .mkString(", ")
```
nimaeskandary: dotted records is a string of record names here right?

mitruly: @nimaeskandary Yep!

nimaeskandary: how is that grouped logic working then, shouldn't it group before the strings are made?

mitruly: @nimaeskandary Yep - you're right; good catch

modules/api/src/main/scala/vinyldns/api/engine/ZoneSyncHandler.scala
```diff
@@ -121,11 +125,14 @@ object ZoneSyncHandler extends DnsConversions with Monitored {
                 chg.recordSet.typ != RecordType.SRV
```
nimaeskandary: btw, we are about to allow dotted hosts on txt, should we just add that here now?

modules/api/src/main/scala/vinyldns/api/engine/ZoneSyncHandler.scala
```diff
@@ -121,11 +125,14 @@ object ZoneSyncHandler extends DnsConversions with Monitored {
...
+              .grouped(1000)
+              .map(_.mkString(", "))
             if (dottedRecords.nonEmpty) {
-              logger.info(
-                s"Zone sync for '${zone.name}' (id: ${zone.id}) " +
-                  s"includes the following dotted host records: [$dottedRecords]")
+              dottedRecords
+                .foreach { dottedGroupString =>
+                  logger.info(s"Zone sync for '${zone.name}' (id: ${zone.id}) " +
+                    s"includes the following ${dottedGroupString.length} dotted host records: [$dottedGroupString]")
```
nimaeskandary: This `dottedGroupString.length` would be the length of the names string not how many names, should we just remove the if on line 130 and do the logging in the map on 129 for easier access on how many names per log

modules/api/src/main/scala/vinyldns/api/engine/ZoneSyncHandler.scala
mitruly: I think it might be better to have that all in a separate PR.

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneViewLoader.scala
```diff
@@ -70,8 +73,29 @@ case class DnsZoneViewLoader(zone: Zone, zoneTransfer: Zone => ZoneTransferIn)
...
           val dnsZoneName = zoneDnsName(zone.name)
-          val recordSets = rawDnsRecords.map(toRecordSet(_, dnsZoneName, zone.id))
+
+          if (droppedRecords.nonEmpty) {
+            droppedRecords
+              .grouped(1000)
+              .foreach { droppedGroup =>
+                val droppedInfo = droppedGroup
+                  .map(record =>
+                    relativize(record.getName, dnsZoneName) + " " + fromDnsRecordType(
```
mitruly: Using `DnsConversions` helper functions here similar like our actual deserialization into `RecordSet`.

modules/api/src/main/scala/vinyldns/api/engine/ZoneSyncHandler.scala
```diff
@@ -95,41 +95,30 @@ object ZoneSyncHandler extends DnsConversions with Monitored {
 
         recordSetChanges.flatMap { allChanges =>
           val changesWithUserIds = allChanges.map(_.withUserId(zoneChange.userId))
-          // not accepting unknown record types
-          val (changes, dropped) = changesWithUserIds.partition { rs =>
```
mitruly: Removing this since our check for supported record types is now in `DnsZoneViewLoader`.

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneViewLoader.scala
nimaeskandary: fromDnsRecordType will just be UNKNOWN anyway right?

mitruly: Yep

mitruly: Hm, I think I see what you're saying; I was considering just having `UNKNOWN` here as well but there are potentially specific types that we wouldn't want to sync in either. Can hardcode this as `UNKNOWN` here and save some processing, but would like feedback from others on this first.

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneViewLoader.scala
```diff
@@ -33,6 +34,7 @@ trait ZoneViewLoader {
 }
 
 object DnsZoneViewLoader extends DnsConversions {
+  private implicit val logger: Logger = LoggerFactory.getLogger("vinyldns.api.domain.zone.DnsZoneViewLoader")
```
rebstar6: implicit?

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneViewLoader.scala
```diff
@@ -70,8 +73,29 @@ case class DnsZoneViewLoader(zone: Zone, zoneTransfer: Zone => ZoneTransferIn)
...
 
+          // not accepting unknown record types
+          val (supportedRecords, droppedRecords) = rawDnsRecords.partition { record =>
+            fromDnsRecordType(record.getType) != RecordType.UNKNOWN
+          }
+
           val dnsZoneName = zoneDnsName(zone.name)
-          val recordSets = rawDnsRecords.map(toRecordSet(_, dnsZoneName, zone.id))
+
+          if (droppedRecords.nonEmpty) {
```
rebstar6: i dont even know if we need this. maybe kill the logs on dropped as a whole

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneViewLoader.scala
mitruly: Whoops

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneViewLoader.scala
mitruly: Do we want to maybe just log the drop count?

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneViewLoader.scala
```diff
@@ -70,8 +73,19 @@ case class DnsZoneViewLoader(zone: Zone, zoneTransfer: Zone => ZoneTransferIn)
...
+            logger.warn(
+              s"Zone sync for zone [$dnsZoneName] dropped ${droppedRecords.length} unsupported record sets")
```
mitruly: Opted to just log total of dropped (unsupported) records.

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneViewLoader.scala
```diff
@@ -70,8 +73,19 @@ case class DnsZoneViewLoader(zone: Zone, zoneTransfer: Zone => ZoneTransferIn)
           val rawDnsRecords: List[DNS.Record] =
             xfr.getAXFR.asScala.map(_.asInstanceOf[DNS.Record]).toList.distinct
 
+          // not accepting unknown record types
+          val (supportedRecords, droppedRecords) = rawDnsRecords.partition { record =>
```
rebstar6: just filter and drop, no logs. the partition itself may be problematic

PR #583, opened 2019-04-15
Change paging to not use offset in MySqlZoneRepository

varsharun:
Fixes #.

Changes in this pull request:
- Change `startFrom` in `listZone` to `Option[String]`

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneProtocol.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneProtocol.scala
@@ -201,8 +201,8 @@ object RecordSetChangeInfo {
 case class ListZonesResponse(
     zones: List[ZoneSummaryInfo],
     nameFilter: Option[String],
-    startFrom: Option[Int] = None,
-    nextId: Option[Int] = None,
+    startFrom: Option[String] = None,
+    nextId: Option[String] = None,
     maxItems: Int = 100)
 
 // Errors
--- a/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneService.scala
@@ -132,21 +132,20 @@ class ZoneService(
   def listZones(
       authPrincipal: AuthPrincipal,
       nameFilter: Option[String] = None,
-      startFrom: Option[Int] = None,
+      startFrom: Option[String] = None,
       maxItems: Int = 100): Result[ListZonesResponse] = {
     for {
       zones <- zoneRepository.listZones(authPrincipal, nameFilter, startFrom, maxItems)
-      groupIds = zones.map(_.adminGroupId).toSet
+      groupIds = zones.zones.map(_.adminGroupId).toSet
       groups <- groupRepository.getGroups(groupIds)
-      zoneSummaryInfos = zoneAdminGroupMapping(zones, groups)
+      zoneSummaryInfos = zoneAdminGroupMapping(zones.zones, groups)
     } yield
       ListZonesResponse(
-        zones = zoneSummaryInfos,
-        nameFilter = nameFilter,
-        startFrom = startFrom,
-        nextId = if (zones.size < maxItems) None else startFrom.orElse(Some(0)).map(_ + zones.size),
-        maxItems = maxItems
-      )
+        zoneSummaryInfos,
+        zones.zonesFilter,
+        zones.startFrom,
+        zones.nextId,
+        zones.maxItems)
   }.toResult
 
   def zoneAdminGroupMapping(zones: List[Zone], groups: Set[Group]): List[ZoneSummaryInfo] =
--- a/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneService.scala
@@ -135,17 +135,18 @@ class ZoneService(
       startFrom: Option[String] = None,
       maxItems: Int = 100): Result[ListZonesResponse] = {
     for {
-      zones <- zoneRepository.listZones(authPrincipal, nameFilter, startFrom, maxItems)
-      groupIds = zones.zones.map(_.adminGroupId).toSet
+      listZonesResult <- zoneRepository.listZones(authPrincipal, nameFilter, startFrom, maxItems)
+      zones = listZonesResult.zones
+      groupIds = zones.map(_.adminGroupId).toSet
       groups <- groupRepository.getGroups(groupIds)
-      zoneSummaryInfos = zoneAdminGroupMapping(zones.zones, groups)
+      zoneSummaryInfos = zoneAdminGroupMapping(zones, groups)
     } yield
       ListZonesResponse(
         zoneSummaryInfos,
-        zones.zonesFilter,
-        zones.startFrom,
-        zones.nextId,
-        zones.maxItems)
+        listZonesResult.zonesFilter,
+        listZonesResult.startFrom,
+        listZonesResult.nextId,
+        listZonesResult.maxItems)
   }.toResult
 
   def zoneAdminGroupMapping(zones: List[Zone], groups: Set[Group]): List[ZoneSummaryInfo] =
--- a/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneServiceAlgebra.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneServiceAlgebra.scala
@@ -39,7 +39,7 @@ trait ZoneServiceAlgebra {
   def listZones(
       authPrincipal: AuthPrincipal,
       nameFilter: Option[String],
-      startFrom: Option[Int],
+      startFrom: Option[String],
       maxItems: Int): Result[ListZonesResponse]
 
   def listZoneChanges(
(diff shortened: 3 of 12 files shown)


modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneProtocol.scala
```diff
@@ -197,9 +197,9 @@ object RecordSetChangeInfo {
 case class ListZonesResponse(
     zones: List[ZoneSummaryInfo],
     nameFilter: Option[String],
-    startFrom: Option[Int] = None,
-    nextId: Option[Int] = None,
-    maxItems: Int = 100)
+    startFrom: Option[String] = None,
+    nextId: Option[String] = None,
+    maxItems: Option[Int] = None)
```
nimaeskandary: maxItems shouldn't be optional, will default to 100 if not there

modules/core/src/main/scala/vinyldns/core/domain/zone/ListZonesResults.scala
```diff
@@ -0,0 +1,24 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package vinyldns.core.domain.zone
+
+case class ListZonestResults(
```
nimaeskandary: "ListZonesResult"

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneRepository.scala
```diff
@@ -261,16 +254,55 @@ class MySqlZoneRepository extends ZoneRepository with ProtobufConversions with M
       user: User,
       groupIds: Seq[String],
       zoneNameFilter: Option[String],
-      offset: Option[Int],
-      pageSize: Int) = {
+      startFrom: Option[String],
```
nimaeskandary: previously this was returning a sql object and not executing the query, if that is going to change function comments should change as well `* @return A fully bound scalike SQL that can be run`

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneRepository.scala
```diff
@@ -261,16 +254,55 @@ class MySqlZoneRepository extends ZoneRepository with ProtobufConversions with M
...
+        )
+      }
+    }
 
-    val (withAccessorCheck, accessors) = withAccessors(user, groupIds)
-    val (withZoneNameSQL, zoneNameParams) = withZoneNameFilter(withAccessorCheck, zoneNameFilter)
-    val sortedSQL = withZoneNameSQL + " ORDER BY z.name ASC"
-    val (finalSQL, pagingParams) = withPaging(sortedSQL, offset, pageSize)
+  /* Append a WHERE clause for the zone name if it is provided */
+  private def withZoneNameFilter(
```
nimaeskandary: It may be worth refactoring this to use a fold for readability. Something like collect the opts as a list of strings, then do 

```
val rest = opts.foldLeft(""){
    case (acc, opt) if acc.isEmpty => s"WHERE $opt"
    case (acc, opt) => s"AND $opt"
}
```

mitruly: Or `StringBuilder` with `foreach`; check out `getUsers` in `MySqlUserRepository`.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneRepository.scala
```diff
@@ -261,16 +254,53 @@ class MySqlZoneRepository extends ZoneRepository with ProtobufConversions with M
       user: User,
       groupIds: Seq[String],
       zoneNameFilter: Option[String],
-      offset: Option[Int],
-      pageSize: Int) = {
+      startFrom: Option[String],
+      maxItems: Option[Int]): IO[ListZonesResults] =
+    IO {
+      DB.readOnly { implicit s =>
+        val (withAccessorCheck, accessors) = withAccessors(user, groupIds)
```
mitruly: This is more of clean up (and would like more input from @rebstar6 or @pauljamescleary before embarking on it), but I think we can get rid of the helper functions (eg. `withAccessors`, `withZoneNameFilter`, etc). Using `StringBuilder` and `.toList` method for `Option` avoids the helpers with the chained `map`.

For example, this is just replacing for `zoneNameFilter`:
```
      DB.readOnly { implicit s =>
        val (withAccessorCheck, accessors) = withAccessors(user, groupIds)
        val sb = new StringBuilder
        sb.append(withAccessorCheck) // Adding accessor part of query
        zoneNameFilter.foreach(_ => sb.append(" WHERE z.name LIKE ?"))
        val (finalSQL, pagingParams) = withSortedPaging(sb.toString(), startFrom, maxItems)
```

Then the `SQL(...).bind()` can just append the `Option` values:
```
        val results: List[Zone] = SQL(finalSQL)
          .bind(accessors ++ zoneNameFilter.toList ++ pagingParams: _*)
```

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneServiceAlgebra.scala
```diff
@@ -37,8 +37,8 @@ trait ZoneServiceAlgebra {
   def listZones(
       authPrincipal: AuthPrincipal,
       nameFilter: Option[String],
-      startFrom: Option[Int],
-      maxItems: Int): Result[ListZonesResponse]
+      startFrom: Option[String],
+      maxItems: Option[Int] = Some(100)): Result[ListZonesResponse]
```
nimaeskandary: unresolved, this should not be an option to be consistent with the other endpoints

mitruly:
Seems like there are some failed tests now in both functional and unit tests related to paging.

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneService.scala
```diff
@@ -124,21 +124,20 @@ class ZoneService(
   def listZones(
       authPrincipal: AuthPrincipal,
       nameFilter: Option[String] = None,
-      startFrom: Option[Int] = None,
+      startFrom: Option[String] = None,
       maxItems: Int = 100): Result[ListZonesResponse] = {
     for {
       zones <- zoneRepository.listZones(authPrincipal, nameFilter, startFrom, maxItems)
```
rebstar6: nitpick - call this variable `listZonesResult` or something. the `zones.zones` makes the typing hard to follow

modules/api/src/test/scala/vinyldns/api/domain/zone/ZoneServiceSpec.scala
```diff
@@ -437,23 +439,25 @@ class ZoneServiceSpec
...
-        .listZones(abcAuth, None, None, 100)
+        .listZones(abcAuth, None, None)
       doReturn(IO.pure(Set(okGroup))).when(mockGroupRepo).getGroups(any[Set[String]])
 
       val result: ListZonesResponse = rightResultOf(underTest.listZones(abcAuth).value)
       val expectedZones =
         List(abcZoneSummary, xyzZoneSummary).map(_.copy(adminGroupName = "Unknown group name"))
       result.zones shouldBe expectedZones
-      result.maxItems shouldBe 100
+      result.maxItems shouldBe Some(100)
```
rebstar6: maxItems is a value not an option i thought (?)

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneRepository.scala
```diff
@@ -208,22 +208,15 @@ class MySqlZoneRepository extends ZoneRepository with ProtobufConversions with M
...
-            authPrincipal.memberGroupIds,
-            zoneNameFilter,
-            offset,
-            pageSize)
-            .map(extractZone(1))
-            .list()
-            .apply()
-        }
-      }
+      buildZoneSearch(
```
rebstar6: if all this function is doing is calling this other function, we could probably just have all the stuff in that other function in here

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneRepository.scala
```diff
@@ -205,23 +205,52 @@ class MySqlZoneRepository extends ZoneRepository with ProtobufConversions with M
...
-            authPrincipal.memberGroupIds,
-            zoneNameFilter,
-            offset,
-            pageSize)
+          val (withAccessorCheck, accessors) =
+            withAccessors(authPrincipal.signedInUser, authPrincipal.memberGroupIds)
+          val sb = new StringBuilder
+          sb.append(withAccessorCheck)
+          zoneNameFilter.foreach(flt => sb.append(s" WHERE z.name LIKE '%$flt%'"))
+          startFrom.foreach { os =>
```
rebstar6: i dont love the toString.contains

does something like this make more sense?

`          val filters = List(
            zoneNameFilter.map(flt => s"z.name LIKE '%$flt%"),
            startFrom.map(os => s"z.name > '$os'")
          ).flatten

          if (filters.nonEmpty) {
            sb.append(" WHERE ")
            sb.append(filters.mkString(" AND "))
          }
`

PR #584, opened 2019-04-15
max zone size

nimaeskandary:
set a max for the number of records that can be in a zone to prevent possible out of memory errors

### DIFF
--- a/modules/api/src/it/resources/application.conf
+++ b/modules/api/src/it/resources/application.conf
@@ -9,9 +9,7 @@ vinyldns {
   approved-name-servers = [
     "172.17.42.1."
   ]
-
-  max-zone-size = 100 # number of records that can be in a zone
-
+  
   mysql {
     settings {
       # see https://github.com/brettwooldridge/HikariCP
--- a/modules/api/src/it/resources/application.conf
+++ b/modules/api/src/it/resources/application.conf
@@ -9,7 +9,7 @@ vinyldns {
   approved-name-servers = [
     "172.17.42.1."
   ]
-  
+
   mysql {
     settings {
       # see https://github.com/brettwooldridge/HikariCP
@@ -26,6 +26,7 @@ vinyldns {
       }
 
     }
+
     repositories {
       zone {
         # no additional settings for now
--- a/modules/api/src/it/resources/application.conf
+++ b/modules/api/src/it/resources/application.conf
@@ -10,6 +10,8 @@ vinyldns {
     "172.17.42.1."
   ]
 
+  max-zone-size = 100 # number of records that can be in a zone
+
   mysql {
     settings {
       # see https://github.com/brettwooldridge/HikariCP
--- a/modules/api/src/it/resources/application.conf
+++ b/modules/api/src/it/resources/application.conf
@@ -26,7 +26,6 @@ vinyldns {
       }
 
     }
-
     repositories {
       zone {
         # no additional settings for now
--- a/modules/api/src/it/scala/vinyldns/api/domain/zone/ZoneViewLoaderIntegrationSpec.scala
+++ b/modules/api/src/it/scala/vinyldns/api/domain/zone/ZoneViewLoaderIntegrationSpec.scala
@@ -47,8 +47,7 @@ class ZoneViewLoaderIntegrationSpec extends WordSpec with Matchers {
 
     "return a failure if the zone is larger than the max zone size" in {
       assertThrows[ZoneTooLargeError](
-        DnsZoneViewLoader(Zone("vinyldns.", "test@test.com"))
-          .copy(maxZoneSize = 1)
+        DnsZoneViewLoader(Zone("vinyldns.", "test@test.com"), DnsZoneViewLoader.dnsZoneTransfer, 1)
           .load()
           .unsafeRunSync()
       )
--- a/modules/api/src/it/scala/vinyldns/api/domain/zone/ZoneViewLoaderIntegrationSpec.scala
+++ b/modules/api/src/it/scala/vinyldns/api/domain/zone/ZoneViewLoaderIntegrationSpec.scala
@@ -44,5 +44,14 @@ class ZoneViewLoaderIntegrationSpec extends WordSpec with Matchers {
           .load()
           .unsafeRunSync())
     }
+
+    "return a failure if the zone is larger than the max zone size" in {
+      assertThrows[ZoneTooLargeError](
+        DnsZoneViewLoader(Zone("vinyldns.", "test@test.com"))
+          .copy(maxZoneSize = 1)
+          .load()
+          .unsafeRunSync()
+      )
+    }
   }
 }
--- a/modules/api/src/main/resources/application.conf
+++ b/modules/api/src/main/resources/application.conf
@@ -31,8 +31,6 @@ akka.http {
 vinyldns {
   data-stores = ["mysql", "dynamodb"]
 
-  max-zone-size = 100 # number of records that can be in a zone
-
   mysql {
     settings {
       # see https://github.com/brettwooldridge/HikariCP
--- a/modules/api/src/main/resources/application.conf
+++ b/modules/api/src/main/resources/application.conf
@@ -31,6 +31,8 @@ akka.http {
 vinyldns {
   data-stores = ["mysql", "dynamodb"]
 
+  max-zone-size = 100 # number of records that can be in a zone
+
   mysql {
     settings {
       # see https://github.com/brettwooldridge/HikariCP
--- a/modules/api/src/main/resources/reference.conf
+++ b/modules/api/src/main/resources/reference.conf
@@ -8,6 +8,8 @@ vinyldns {
   # if we should start up polling for change requests, set this to false for the inactive cluster
   processing-disabled = false
 
+  max-zone-size = 60000 # number of records that can be in a zone
+
   queue {
     class-name = "vinyldns.sqs.queue.SqsMessageQueueProvider"
 
--- a/modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
+++ b/modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
@@ -113,5 +113,5 @@ object VinylDNSConfig {
       vinyldnsConfig.getStringList(key).asScala.toList
     } else List()
 
-  lazy val maxZoneSize: Int = vinyldnsConfig.getInt("max-zone-size")
+  lazy val maxZoneSize: Int = vinyldnsConfig.as[Option[Int]]("max-zone-size").getOrElse(60000)
 }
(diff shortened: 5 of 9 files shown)


modules/api/src/main/resources/application.conf
```diff
@@ -31,6 +31,8 @@ akka.http {
 vinyldns {
   data-stores = ["mysql", "dynamodb"]
 
+  max-zone-size = 100 # number of records that can be in a zone
```
nimaeskandary: shouldnt be here

nimaeskandary:
Screen Shot 2019-04-15 at 12 42 45 PM

nimaeskandary:
Screen Shot 2019-04-15 at 12 42 31 PM

modules/api/src/it/resources/application.conf
```diff
@@ -9,7 +9,7 @@ vinyldns {
   approved-name-servers = [
     "172.17.42.1."
   ]
-
+
```
mitruly: Very subtle

modules/api/src/main/resources/reference.conf
```diff
@@ -8,6 +8,8 @@ vinyldns {
   # if we should start up polling for change requests, set this to false for the inactive cluster
   processing-disabled = false
 
+  max-zone-size = 60000 # number of records that can be in a zone
```
mitruly: Wonder if `max-zone-sync-size` is clearer.

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneViewLoader.scala
```diff
@@ -70,10 +73,20 @@ case class DnsZoneViewLoader(zone: Zone, zoneTransfer: Zone => ZoneTransferIn)
           val rawDnsRecords: List[DNS.Record] =
             xfr.getAXFR.asScala.map(_.asInstanceOf[DNS.Record]).toList.distinct
 
-          val dnsZoneName = zoneDnsName(zone.name)
-          val recordSets = rawDnsRecords.map(toRecordSet(_, dnsZoneName, zone.id))
+          if (rawDnsRecords.length > maxZoneSize)
```
mitruly: From our initial discussion, we wanted to only include records that we are syncing in (excluding the dropped records); should probably block this until https://github.com/vinyldns/vinyldns/pull/582 is merged in (which brings the dropping to `ZoneViewLoader` from `ZoneSyncHandler`.

modules/api/src/it/resources/application.conf
rebstar6: i dig it

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneViewLoader.scala
```diff
@@ -70,10 +73,20 @@ case class DnsZoneViewLoader(zone: Zone, zoneTransfer: Zone => ZoneTransferIn)
...
+            throw ZoneTooLargeError(
```
rebstar6: wow throw ew

rebstar6: also, we need to check on what the retry logic is here. this should *not* retry in the case that the zone is too large. we would want the sync to fail and a system message to indicate that in the zone change for this sync (read: sync message must be deleted from queue).

mitruly:
Also was thinking that since we use a `diff` in `ZoneSyncHandler.scala` to identify if records currently in VinylDNS actually differ against the DNS backend, it might make more sense to apply the limit in `runSync` after we know how many actual record we're trying to write.

nimaeskandary:
@mitruly Not so sure that would solve the issue, the zone view loader returns a list of records sets: https://github.com/vinyldns/vinyldns/blob/master/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneViewLoader.scala#L74, so if we wait for https://github.com/vinyldns/vinyldns/blob/master/modules/api/src/main/scala/vinyldns/api/engine/ZoneSyncHandler.scala#L89 that may cause the out of mem error depending on the zone

mitruly:
@nimaeskandary Another issue would be the following scenario:
- User syncs in zone for 60k records
- User makes a few changes
- Maybe some DNS records get created out-of-band (via DNS backend directly)
- User hits sync and hits an error where sync exceeds max limit

mitruly:
With https://github.com/vinyldns/vinyldns/blob/ef30ce3fdc906c2b7ff0226d6d1dd548ed1f01d0/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneViewLoader.scala#L76-L79 change, we shouldn't be deserializing all of those changes into memory, right?

nimaeskandary:
At that point that we may be better off saying the zone is too big for vinyl by definition of our sync limit, because if the zone was abandoned they could not reconnect, because then the rules of whether a zone can be in or not would be more clear

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneViewLoader.scala
nimaeskandary: Yeah would have IO.raised but is already in an IO

nimaeskandary: And here we catch throwables and just complete the zone change without syncing so the queue will not retry https://github.com/vinyldns/vinyldns/blob/master/modules/api/src/main/scala/vinyldns/api/engine/ZoneSyncHandler.scala#L152

modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
```diff
@@ -112,4 +112,6 @@ object VinylDNSConfig {
     if (vinyldnsConfig.hasPath(key)) {
       vinyldnsConfig.getStringList(key).asScala.toList
     } else List()
+
+  lazy val maxZoneSize: Int = vinyldnsConfig.getInt("max-zone-size")
```
rebstar6: pull as an option and getOrElse 60000

modules/api/src/it/scala/vinyldns/api/domain/zone/ZoneViewLoaderIntegrationSpec.scala
```diff
@@ -44,5 +44,14 @@ class ZoneViewLoaderIntegrationSpec extends WordSpec with Matchers {
           .load()
           .unsafeRunSync())
     }
+
+    "return a failure if the zone is larger than the max zone size" in {
+      assertThrows[ZoneTooLargeError](
+        DnsZoneViewLoader(Zone("vinyldns.", "test@test.com"))
+          .copy(maxZoneSize = 1)
```
rebstar6: i think this copy thing is just `DnsZoneViewLoader(Zone("vinyldns.", "test@test.com"), DnsZoneViewLoader.dnsZoneTransfer, 1)`

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneViewLoader.scala
rebstar6: yuck. i still really dont like the throw, how crazy are things if we make that return an IO[Either] (or EitherT)?

modules/api/src/main/resources/reference.conf
britneywright: should add this to the docs around configs, with more context on what it's intended for. It's just for checking the size of the zone during sync, not hard limiting the number of records that can exist in a zone.

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneViewLoader.scala
pauljamescleary: just wrap a few steps in `IO`, and then you can have one of the steps do the `IO.raiseError`, can use flatMap or a for comprehension...

```scala
for {
  rawDnsRecords <- IO(xxx)
  _ <- IO(ensureOr?)
  recordSets <- IO(yyy)
} yield ZoneView()
```

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneViewLoader.scala
```diff
@@ -54,31 +56,38 @@ object DnsZoneViewLoader extends DnsConversions {
...
-            rawDnsRecords.filter(record => fromDnsRecordType(record.getType) != RecordType.UNKNOWN)
-
-          val dnsZoneName = zoneDnsName(zone.name)
-
-          val recordSets = supportedRecords.map(toRecordSet(_, dnsZoneName, zone.id))
-
-          ZoneView(zone, recordSets)
-        }
+        for {
+          rawDnsRecords <- IO {
```
rebstar6: IO split here feels arbitrary, and is hard to read. switch to:

```
  def load: () => IO[ZoneView] =
    () =>
      monitor("dns.loadZoneView") {
        for {
          zoneXfr <- IO {
            val xfr = zoneTransfer(zone)
            xfr.run()
            xfr.getAXFR.asScala.map(_.asInstanceOf[DNS.Record]).toList.distinct
          }
          rawDnsRecords = zoneXfr.filter(record => fromDnsRecordType(record.getType) != RecordType.UNKNOWN)
          _ <- if (rawDnsRecords.length > maxZoneSize)
            IO.raiseError(ZoneTooLargeError(zone, rawDnsRecords.length, maxZoneSize))
          else IO.pure(Unit)
          dnsZoneName <- IO(zoneDnsName(zone.name))
          recordSets = rawDnsRecords.map(toRecordSet(_, dnsZoneName, zone.id))
        } yield ZoneView(zone, recordSets)
    }
```

was not sure on last line (rawDnsRecords.map..). i dont think that needs to be wrapped in IO, but if so, wrap it.

for the error to work, need a :
```
object ZoneTooLargeError {
  def apply(zone: Zone, len: Int, maxZoneSize: Int): ZoneTooLargeError = {
    val msg = s"ZoneTooLargeError: Zone '${zone.name}' (id: '${zone.id}') contains $len records" +
      s"which exceeds the max of $maxZoneSize"
    new ZoneTooLargeError(msg)
  }
}
```

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneViewLoader.scala
```diff
@@ -54,31 +56,31 @@ object DnsZoneViewLoader extends DnsConversions {
...
-          ZoneView(zone, recordSets)
-        }
+          }
+          rawDnsRecords = zoneXfr.filter(record =>
+            fromDnsRecordType(record.getType) != RecordType.UNKNOWN)
+          _ <- if (rawDnsRecords.length > maxZoneSize)
+            IO.raiseError(ZoneTooLargeError(zone, rawDnsRecords.length, maxZoneSize))
+          else IO.pure(Unit)
+          dnsZoneName <- IO(zoneDnsName(zone.name))
+          recordSets = rawDnsRecords.map(toRecordSet(_, dnsZoneName, zone.id))
```
rebstar6: sorry, actually, this can throw, so wrap in IO. otherwise 👍

PR #586, opened 2019-04-15
Use getRecordSetsByFQDNs

britneywright:
Changes in this pull request:
- Add a filter in getRecordSetsByFQDNs to only return recordsets in existing zones

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
@@ -31,7 +31,7 @@ import vinyldns.core.domain.auth.AuthPrincipal
 import vinyldns.core.domain.batch._
 import vinyldns.core.domain.membership.{Group, GroupRepository}
 import vinyldns.core.domain.record.RecordType._
-import vinyldns.core.domain.record.{RecordSet, RecordSetRepository}
+import vinyldns.core.domain.record.RecordSetRepository
 import vinyldns.core.domain.zone.ZoneRepository
 
 object BatchChangeService {
@@ -154,14 +154,15 @@ class BatchChangeService(
 
   def getExistingRecordSets(
       changes: ValidatedBatch[ChangeForValidation]): IO[ExistingRecordSets] = {
-    val uniqueGets = changes.getValid.map(change => (change.zone.id, change.recordName)).toSet
-    val allIO = uniqueGets.map {
-      case (zoneId, rsName) =>
-        recordSetRepository.getRecordSetsByName(zoneId, rsName)
-    }
-    val allSeq: IO[List[List[RecordSet]]] = allIO.toList.sequence
+    val uniqueGets = changes.getValid.map{ change =>
+      change.inputChange.typ match {
+        case PTR => s"${change.recordName}.${change.zone.name}"
+        case _ => change.inputChange.inputName
+      }
+    }.toSet
+    val allRecordSets = recordSetRepository.getRecordSetsByFQDNs(uniqueGets)
 
-    allSeq.map(lst => ExistingRecordSets(lst.flatten))
+    allRecordSets.map(ExistingRecordSets)
   }
 
   def getOwnerGroup(ownerGroupId: Option[String]): BatchResult[Option[Group]] = {
--- a/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
@@ -161,12 +161,11 @@ class BatchChangeService(
         case _ => change.inputChange.inputName
       }
     }.toSet
-    val allRecordSets = recordSetRepository.getRecordSetsByFQDNs(uniqueGets)
-    val recordSetsWithExistingZone = allRecordSets.map { lst =>
-      lst.filter(rs => zoneMap.getById(rs.zoneId))
-    }
 
-    recordSetsWithExistingZone.map(ExistingRecordSets)
+    for {
+      allRecordSets <- recordSetRepository.getRecordSetsByFQDNs(uniqueGets)
+      recordSetsWithExistingZone = allRecordSets.filter(rs => zoneMap.getById(rs.zoneId))
+    } yield ExistingRecordSets(recordSetsWithExistingZone)
   }
 
   def getOwnerGroup(ownerGroupId: Option[String]): BatchResult[Option[Group]] = {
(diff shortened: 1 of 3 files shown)


modules/api/src/main/scala/vinyldns/api/domain/batch/BatchTransformations.scala
```diff
@@ -35,6 +35,8 @@ object BatchTransformations {
   final case class ExistingZones(zones: Set[Zone]) {
     val zoneMap: Map[String, Zone] = zones.map(z => (z.name, z)).toMap
 
+    def getById(id: String): Set[Zone] = zones.filter(zn => zn.id.equals(id))
```
mitruly: I'm tempted to say that it's probably more efficient if we add the criteria of ` WHERE zone_id IN (...)` to `BASE_FIND_RECORDSETS_BY_FQDNS` (in `MySqlRecordSetRepository`) and just pass in the `zoneMap.zones.map(_.id)` for the parameters to get the unique zone IDs.

nimaeskandary: Hmm if we already have ExistingZones processed then this may be faster than adding to the sql query @mitruly

mitruly: @nimaeskandary `zoneMap` _is_ `ExistingZones`; it's basically a set of unique zones that we have. I think it's faster to have MySQL do it then returning a list of record sets and then cross-reference that back against the unique zones list to filter out the record sets.

rebstar6: we just need id exists, not the full get. can do `zones.exists(zn => zn.id.equals(id))` and return a bool.

rebstar6: dont hit mysql again, no need to :P

modules/api/src/test/scala/vinyldns/api/domain/batch/BatchChangeServiceSpec.scala
```diff
@@ -371,15 +371,34 @@ class BatchChangeServiceSpec
...
-      val result = await(underTest.getExistingRecordSets(in))
+      val zoneMap = ExistingZones(Set(apexZone, baseZone, ptrZone, delegatedPTRZone, ipv6PTRZone))
+      val result = await(underTest.getExistingRecordSets(in, zoneMap))
 
       val expected =
         List(existingApex, existingNonApex, existingPtr, existingPtrDelegated, existingPtrV6)
       result.recordSets should contain theSameElementsAs expected
     }
+
+    "combine gets for each valid record with existing zone" in {
```
rebstar6: make sure we test where the DB returns a conflicting record, but the ID is bad (so it's not really a conflict)

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchTransformations.scala
britneywright: updated

modules/api/src/test/scala/vinyldns/api/domain/batch/BatchChangeServiceSpec.scala
britneywright: I added a recordset with the same fqdn as existingApex but with a zone ID that isn't connected to any of the existing zones.

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -153,15 +153,20 @@ class BatchChangeService(
...
-    val uniqueGets = changes.getValid.map(change => (change.zone.id, change.recordName)).toSet
-    val allIO = uniqueGets.map {
-      case (zoneId, rsName) =>
-        recordSetRepository.getRecordSetsByName(zoneId, rsName)
+      changes: ValidatedBatch[ChangeForValidation],
+      zoneMap: ExistingZones): IO[ExistingRecordSets] = {
+    val uniqueGets = changes.getValid.map { change =>
+      change.inputChange.typ match {
+        case PTR => s"${change.recordName}.${change.zone.name}"
+        case _ => change.inputChange.inputName
```
rebstar6: is it possible that trailing dots will be an issue for us here? do we consistently have the trailing dot in the fqdns stored in the db, and these items ^^?

britneywright: I don't think there's any issues. We have the `ensureTrailingDot` function, so there's always going to be a trailing dot on the non-PTR changes and the zones end in trailing dots.

britneywright:
I removed the trailing dots from the input names the BatchChangeServiceSpec, to confirm they still eventually match to FQDNS with trailing dots.

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -153,15 +153,20 @@ class BatchChangeService(
...
+      }
+    }.toSet
+    val allRecordSets = recordSetRepository.getRecordSetsByFQDNs(uniqueGets)
```
mitruly: These `map`s are a bit difficult to follow; the following feels cleaner imo:
```
    for {
      allRecordSets <- recordSetRepository.getRecordSetsByFQDNs(uniqueGets)
      recordSetsWithExistingZone = allRecordSets.filter(rs => zoneMap.getById(rs.zoneId))
    } yield ExistingRecordSets(recordSetsWithExistingZone)
```

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
mitruly: I also checked creating a record set through both record set and batch endpoint (not using terminating dot) and looked in MySQL:
```
mysql> select rs.name, rs.fqdn from recordset as rs inner join zone as zn on rs.zone_id=zn.id where zn.name='ok.';
+-----------------------+---------------------------+
| name                  | fqdn                      |
+-----------------------+---------------------------+
| already-exists        | already-exists.ok.        |
| batch-not-dotted      | batch-not-dotted.ok.      |
| foo                   | foo.ok.                   |
| jenkins               | jenkins.ok.               |
| ok.                   | ok.                       |
| ok.                   | ok.                       |
| ok.                   | ok.                       |
| record-set-not-dotted | record-set-not-dotted.ok. |
| test                  | test.ok.                  |
+-----------------------+---------------------------+
9 rows in set (0.01 sec)
```

rebstar6: 👍  figured we should check :P

PR #589, opened 2019-04-17
Enable shared zone toggle for zone admins that are also support users

mitruly:
Currently, only super users can toggle the shared flag for zone creates/updates. We need more flexibility and don't want to open it up to all zone admins, so a middle ground is a zone admin that is also a support user.

Changes in this pull request:
- Enable functionality
- Create unit tests

### DIFF
--- a/modules/api/src/test/scala/vinyldns/api/domain/zone/ZoneServiceSpec.scala
+++ b/modules/api/src/test/scala/vinyldns/api/domain/zone/ZoneServiceSpec.scala
@@ -164,7 +164,25 @@ class ZoneServiceSpec
       resultZone.shared shouldBe true
     }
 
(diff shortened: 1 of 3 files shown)


modules/api/src/test/scala/vinyldns/api/domain/zone/ZoneServiceSpec.scala
```diff
@@ -164,7 +164,22 @@ class ZoneServiceSpec
       resultZone.shared shouldBe true
     }
 
-    "return a NotAuthorizedError if zone is shared and user is not a super user" in {
+    "succeed if zone is shared and user is both a zone admin and support user" in {
```
mitruly: Didn't need to add tests for zone admin prohibited since there are already existing tests that use the zone admin group.

PR #588, opened 2019-04-16
Expose get zone by name route in the API

varsharun:
Fixes #.

Changes in this pull request:
- Added a route to `ZoneRouting` `GET /zones/name/<segment>`
- Added a new method to `ZoneService`to `getZonesByName`
- Unit and Integrations tests
- Functional tests

PR #593, opened 2019-04-19
portal: add getZoneByName API route

britneywright:
Supports #588

Changes in this pull request:
- add an API route in the portal for getZoneByName (`/api/zones/name/:name`)

### DIFF
--- a/modules/portal/test/controllers/TestApplicationData.scala
+++ b/modules/portal/test/controllers/TestApplicationData.scala
@@ -180,9 +180,10 @@ trait TestApplicationData { this: Mockito =>
   )
 
   val hobbitZoneId = "uuid-abcdef-12345"
+  val hobbitZoneName = "hobbits"
   val hobbitZone: JsValue = Json.parse(s"""{
       | "id":             "${hobbitZoneId}",
-      | "name":           "hobbits",
+      | "name":           "${hobbitZoneName}",
       | "email":          "hobbitAdmin@shire.me",
       | "status":         "Active",
       | "account":        "system",
--- a/modules/portal/test/controllers/TestApplicationData.scala
+++ b/modules/portal/test/controllers/TestApplicationData.scala
@@ -133,7 +133,7 @@ trait TestApplicationData { this: Mockito =>
 
   val hobbitGroupId = "uuid-12345-abcdef"
   val hobbitGroup: JsValue = Json.parse(s"""{
-       | "id":          "${hobbitGroupId}",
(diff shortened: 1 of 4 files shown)


modules/portal/test/controllers/VinylDNSSpec.scala
```diff
@@ -1742,6 +1742,122 @@ class VinylDNSSpec extends Specification with Mockito with TestApplicationData w
...
+
+            status(result) must beEqualTo(OK)
+            hasCacheHeaders(result)
+            contentAsJson(result) must beEqualTo(hobbitZone)
+          }
+        }
+      }
+      "return a not found (404) if the zone does not exist" in new WithApplication(app) {
+        Server.withRouter(ServerConfig(port = Some(simulatedBackendPort), mode = Mode.Test)) {
+          case backendGET(p"/zones/name/not-hobbits") =>
```
pauljamescleary: My one comment here is that I am not sure that the backend API is "good REST".

From a convention standpoint, think of `/zones` as "the collection of all zones".  Could be an in memory List of zones if you will.

What is `/zones/name`?  From a usage standpoint doesn't seem to make a lot of sense.

Not sure what the exact guidance is on the backend API, just pointing out that we should revisit the URL path to it

modules/portal/test/controllers/TestApplicationData.scala
```diff
@@ -180,9 +180,10 @@ trait TestApplicationData { this: Mockito =>
   )
 
   val hobbitZoneId = "uuid-abcdef-12345"
+  val hobbitZoneName = "hobbits"
   val hobbitZone: JsValue = Json.parse(s"""{
       | "id":             "${hobbitZoneId}",
-      | "name":           "hobbits",
+      | "name":           "${hobbitZoneName}",
```
mitruly: If we're using a variable directly, curly braces aren't needed.

modules/portal/test/controllers/VinylDNSSpec.scala
```diff
@@ -1742,6 +1742,122 @@ class VinylDNSSpec extends Specification with Mockito with TestApplicationData w
       }
     }
 
+    ".getZoneByName" should {
+      "return ok (200) if the zone is found" in new WithApplication(app) {
+        Server.withRouter(ServerConfig(port = Some(simulatedBackendPort), mode = Mode.Test)) {
+          case backendGET(p"/zones/name/${hobbitZoneName}") =>
```
mitruly: Curly braces not needed here.

Issue #521, opened 2019-03-06
Support whitelisting access for dotted hosts


pauljamescleary:
**Is your feature request related to a problem? Please describe.**
Legacy DNS installations may have a predominance of dotted hosts that are not cleaned up.  Currently, VinylDNS does not allow the maintenance of dotted host records.

**Describe the solution you'd like**
1. Allow create and update of dotted hosts in VinylDNS
2. Whitelist the domains that are allowed to contain a dotted host via config.  Do not allow a new or updated dotted host that is not in the whitelist
3. Allow only super users and support users to manage whitelisted dotted hosts by default
4. Allow ACL rule access to manage whitelisted dotted host records
5. Add verbiage in a tooltip for dotted hosts when displayed to indicate the issue with dotted hosts

britneywright:
#702 and #703 take care of support for updating and deleting existing dotted hosts.

britneywright:
we also already have a warning sign and tooltip in the portal for dotted host records.

sebschrader:
Are you still considering adding support for dotted hosts? We have tons of dotted hosts in our zones and make heavy use of them.

This limitation is also making use of wildcard records `*` very cumbersome.

PR #594, opened 2019-04-22
Adding NAPTR support

tpretz:
Addition of support for the NAPTR RR Type

Clone of SRV support behaviour, adjusting as needed.

Room to improve validation, beyond my immediate abilities as a non-scala developer.

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/backend/CommandHandler.scala
+++ b/modules/api/src/main/scala/vinyldns/api/backend/CommandHandler.scala
@@ -51,7 +51,8 @@ object CommandHandler {
       count: MessageCount,
       pollingInterval: FiniteDuration,
       pauseSignal: SignallingRef[IO, Boolean],
-      connections: ConfiguredDnsConnections)(implicit timer: Timer[IO]): Stream[IO, Unit] = {
+      connections: ConfiguredDnsConnections,
+      maxOpen: Int = 4)(implicit timer: Timer[IO]): Stream[IO, Unit] = {
 
     // Polls queue for message batches, connected to the signal which is toggled in the status endpoint
     val messageSource = startPolling(mq, count, pollingInterval).pauseWhen(pauseSignal)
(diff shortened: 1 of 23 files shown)


nimaeskandary:
Thanks @tpretz! Did a quick glance so far and this is pretty sweet! Leaving a reminder for someone to make an issue to add NAPTR Batch Change Support when this is merged

pauljamescleary:
Thanks @tpretz for the PR.  This is really good stuff :) .

pauljamescleary:
A few comments:

1. Should add a functional test (add a python test under `modules/api/functional_test` in the `create_recordset_test.py`.  I asked @nimaeskandary to help out with that
2. We need to add cards to our backlog to update the different client libraries for support.  Those should all be pretty small

@tpretz do you use the client libraries at all, or only the web portal?

tpretz:
My initial usecase, that drove this addition, is just via the portal.
(essentially self-service for a mobile team to drive their telco infra zones which use NAPTR heavily)

So, i don't have any other pull requests pending right now.
(although, the GUI modals are a bit cramped with the amount of fields NAPTRs require, didn't want to adjust that in this PR tho as its a bit out of scope for this RR addition)

My only other deviation is to disable the dotted hosts check (i know its generally good practice to avoid these, but with some of these infrastructure zones its not really practical to delegate at every ".")
I can see a proper solution is already raised for this in #521 
Being able to opt a zone out of these checks is going to be essential to on-board telco zones, or legacy zones prior to a cleanup. 

Anyway, thanks for making this project opensource, it's been very useful and i hope to expand my deployment to other usecases soon.

pauljamescleary:
@tpretz Would you have time to meet up so we can discuss your use case / deployment?  Would love to get feedback to help us drive our backlog

rebstar6:
Also, feel free to add yourself to https://github.com/vinyldns/vinyldns/blob/master/AUTHORS.md in this PR if you'd like!

nimaeskandary:
@tpretz made a pr against yours with a couple func tests https://github.com/tpretz/vinyldns/pull/1

nimaeskandary:
Screen Shot 2019-04-24 at 10 28 32 AM

made a couple records in the portal as well

PR #595, opened 2019-04-23
Expose backendId for zones

varsharun:
Fixes #.

Changes in this pull request:
- Add `backendId` to zone creation as an optional
- Add `backendId` to manage zone page

### DIFF
--- a/modules/portal/app/views/zones/zoneTabs/manageZone.scala.html
+++ b/modules/portal/app/views/zones/zoneTabs/manageZone.scala.html
@@ -68,6 +68,15 @@
                                 </div>
                             </div>
 
+                            <div class="form-group">
+                                <label class="col-md-3 control-label">DNS Backend ID (optional)</label>
+                                <div class="col-md-9">
+                                    <div>
+                                        <input id="zone-update-backendID" type="text" ng-model="updateZoneInfo.backendId" class="form-control" ng-disabled="!isZoneAdmin"/>
+                                    </div>
+                                </div>
+                            </div>
+
                         </div>
                         <div class="col-md-6">
 
@@ -178,6 +187,7 @@
                                         ng-click="clearUpdateTransferConnection()" type="button">Clear Connection</button>
                             </div>
                         </div>
+
                     </div>
                 </div>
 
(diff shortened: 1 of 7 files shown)


modules/portal/app/views/zones/zoneTabs/manageZone.scala.html
```diff
@@ -68,6 +68,15 @@ <h3 class="panel-title">Zone Info</h3>
                                 </div>
                             </div>
 
+                            <div class="form-group">
```
mitruly: I think we need to handle empty strings in the modals; when trying to connect to a new zone and typing something in the backend ID field and manually clearing it out or doing something similar in the manage zone screen, I encounter the following message:
```
HTTP 400 (Bad Request): Invalid backendId: [Some()]; please check system configuration
```

We should treat empty Strings like `None`, maybe by deleting from the JSON request object?.

modules/portal/public/templates/zoneconnection-modal.html
```diff
@@ -32,6 +32,26 @@
...
+            </div>
+
+            <div class="dns-connection-form">
+                <h4>DNS Zone Backend ID (Optional)</h4>
+                <hr>
+            </div>
+            <modal-element label="Backend Id">
+                <input id="zone-backendId" type="text" ng-model="currentZone.backendId" name="backendIdVal" class="form-control"/>
+                <span class="help-block">
+                    The ID for a pre-configured DNS connection configuration. Please contact your DNS admin team
```
pauljamescleary: Hmmm, it would be better if this were a drop down list, right?  A simple endpoint in the API could expose the backend ids via a GET since it is just in Config.

mitruly: @pauljamescleary It might make sense to have a dropdown list, but we should maintain that most users won't be using this (since they would use the default keys). There is also the possibility of multiple backend IDs working with the same zone, so it might be better from a need-to-know-basis perspective to keep this freeform.

pauljamescleary: Is there anything stopping the user from entering in _both_ a BackendID _and_ the DNS server information?  They should be exclusive.  

Also, keep in mind this is a product @mitruly , so we cannot anticipate necessarily how other people may use this.

From a product perspective, it is better to have a drop down list rather than free form.

nimaeskandary:
closing in favor of https://github.com/vinyldns/vinyldns/pull/607

PR #598, opened 2019-04-24
Release 0.9.0

nimaeskandary:
commit 6683c8f was release to docker and maven for api, portal, and core


nimaeskandary:
that auto commit message is odd, it made it 0.9.0 in the diff before releasing not 0.8.1-SNAPSHOT

PR #599, opened 2019-04-25
Portal batch change import

britneywright:
Add an "import CSV" feature to the portal batch change form, so users don't have to manually enter all of their individual changes in the form.

The CSV is only for changes, not for the batch change comment or owner group ID fields.

For initial review: test out functionality. Does it work as expected/desired? Try to break it or find flaws in the flow or data parsing. Use the `modules/docs/src/main/resources/microsite/static/batch-csv-sample.csv` to start, play around with that as well.


britneywright:
Screenshots and flow of the current state of the import feature in the batch change form:

- In the New Batch Change form there is an "Import CSV" next to the "Add a Change" button. Below the buttons is a link to Download the CSV, this will take users to the ["Create a Batch Change"](https://www.vinyldns.io/portal/batch-changes#create-a-batch-change) documentation that explains the CSV and has a downloadable sample. Those doc updates are in this PR.
batch change form

- If the user uploads a valid CSV (validation is that the CSV is of the file type "text/csv") an Upload button with the CSV file name appears. The "Import CSV" button changes to "Change File" to indicate the user can choose a different file before uploading.
batch change form csv upload button

- After the user clicks the Upload button the contents of the csv are parsed and added as single record changes in the batch change. Some validations/formatting notes:
  - all fields are trimmed to avoid possible white space issues
  - CSV data is now case-insensitive. The casing will be changed as needed (e.g. the API only accepts uppercase record type values)
  - The changeType field accepts any string with "delete" so a user could enter the accepted "DeleteRecordSet" or simply "delete" and it will be set to "DeleteRecordSet"
  - TTL is converted to an integer
  - record data keys are created based on the record type

- The "Import CSV" form is reset so the "Upload" button is no longer visible and the "Import CSV" button is visible rather than the "Change File" CSV.

batch change form imported data

- If the user chooses a file that is not a CSV they will see an error message, "[File name] is not a CSV file." The "Import CSV" button changes to the "Change File" button
batch change improper file type

nimaeskandary:
commit history looks a bit off, looks like if it needs a rebase or some cherry picking if that doesn't work

britneywright:
> commit history looks a bit off, looks like if it needs a rebase or some cherry picking if that doesn't work

fixed

modules/portal/app/views/batchChanges/batchChangeNew.scala.html
```diff
@@ -191,8 +191,20 @@ <h4>Changes</h4>
...
+                                <ng-form name="csvForm" isolate-form>
+                                    <label class="batch-change-csv-label btn btn-default" for="batchChangeCsv" id="batchChangeCsvImportLabel">
+                                        <span ng-if="csvForm.batchChangeCsv.$pristine || csvForm.batchChangeCsv.$error.filled"><span class="glyphicon glyphicon-import"></span> Import CSV</span>
+                                        <span ng-if="csvForm.batchChangeCsv.$dirty && csvForm.batchChangeCsv.$$success.filled"><span class="glyphicon glyphicon-folder-open"></span> Change File</span>
+                                    </label>
+                                    <input type="file" id="batchChangeCsv" ng-model="csvInput" name="batchChangeCsv" filled csv>
+                                    <button class="btn btn-success" ng-click="uploadCSV()" type="button" ng-if="csvForm.batchChangeCsv.$dirty && csvForm.batchChangeCsv.$valid" id="batchChangeCsvUpload"><span class="glyphicon glyphicon-upload"></span> Upload {{csvForm.batchChangeCsv.$viewValue}}</button>
```
mitruly: Nitpick - there is slightly unexpected behavior when canceling an import with an existing file selected:
- Select a CSV file using `Import CSV` button
- Click the `Import CSV` button again and then hit cancel; initial file will be cleared

My expectation here when not selecting a new file would be to keep the existing file since the file would not get imported unless clicking the `Upload CSV` button.

rebstar6:
tried it out, overall approach is good.

some small stuff:
- dont import empty lines; excel added one for me, that should just be passed over
- something weird happens if you import a 2nd time on the same page (after you upload...dunno why someone would do this, so low priority)

i say clean it up and then get final reviews on it

modules/portal/app/views/batchChanges/batchChangeNew.scala.html
britneywright: That's not my behavior, that's the way the input file type/browser works, if you get the prompt to select a file and hit cancel the indication is you've chosen nothing.

britneywright:
> * something weird happens if you import a 2nd time on the same page (after you upload...dunno why someone would do this, so low priority)

what's the something weird that happens?

modules/docs/src/main/tut/portal/batch-changes.md
```diff
@@ -31,7 +31,10 @@ This does not apply to zone administrators or users with specific ACL access rul
 ### Create a Batch Change
 1. Go to the Batch Changes section of the site.
 1. Select the *New Batch Change* button.
-1. Add a description and record changes. Select the *Add a Change* button to add additional rows for data entry as needed.
+1. Add a description.
+1. Add record changes in one of two ways:
+ - Select the *Add a Change* button to add additional rows for data entry as needed.
+ - Select the *Import CSV* button to choose a CSV file of the record changes, then select *Upload* to confirm your submission. Download a sample CSV [here](../static/batch-csv-sample.csv). **Note** The header row is required.
```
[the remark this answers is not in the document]
britneywright: > Is the download URL correct? It's directing me to https://www.vinyldns.io/portal/batch-changes#create-a-batch-change (when running on localhost).

yes in my comment earlier, that's where the information will be when the PR is merged in.

> We might also want to be explicit about the order of the columns (Change Type, Record Type, Input Name, TTL, Record Data) since this is a hard-coded order and we're not associating the data rows with the ordering of the first row.

Do you think we need to be more explicit in the documentation or elsewhere?

modules/portal/app/views/batchChanges/batchChangeNew.scala.html
mitruly: This is when one already chooses the first file and walk through the process to choose a second file but instead cancel; it doesn't seem normal that canceling the file change would wipe out the initial file unless something is preemptively clearing the selected file.

That said, this issue is also low priority so I'm fine with it we can't change its behavior.

britneywright: here's a jsfiddle of the file type input https://jsfiddle.net/5kxvcpm9/

modules/portal/public/lib/batch-change/batch-change.directive.js
```diff
@@ -38,6 +38,7 @@
                                 '([0-9a-fA-F]{1,4}:){1,4}:((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])',
                                 ')$'
     ].join(''));
+    var CSV_REGEX = /\.csv$/;
```
nimaeskandary: we prob dont to check for .csv right can we just try/catch on reading it?

britneywright: oops meant to delete that, not using it

modules/portal/public/lib/batch-change/batch-change-new.controller.js
```diff
@@ -44,13 +44,16 @@
...
             $scope.confirmSubmit = function(form) {
+                console.log(form.$error)
                 if(form.$invalid){
                     form.$setSubmitted();
                     $scope.formStatus = "pendingSubmit";
                 }
             };
 
             $scope.createBatchChange = function() {
+                console.log("creating a batch change")
```
mitruly: Can we combine these two log statements?

modules/portal/app/views/batchChanges/batchChangeNew.scala.html
```diff
@@ -191,8 +191,20 @@ <h4>Changes</h4>
...
+                                    </label>
+                                    <input type="file" id="batchChangeCsv" ng-model="csvInput" name="batchChangeCsv" class="batchChangeCsv" filled csv>
+                                    <button class="btn btn-success" ng-click="uploadCSV(csvForm.batchChangeCsv.$viewValue)" type="button" ng-if="csvForm.batchChangeCsv.$dirty && csvForm.batchChangeCsv.$valid" id="batchChangeCsvUpload"><span class="glyphicon glyphicon-upload"></span> Upload {{csvForm.batchChangeCsv.$viewValue.name}}</button>
+                                    <p ng-show="csvForm.batchChangeCsv.$error.csv && csvForm.batchChangeCsv.$$success.filled" class="batch-change-error-help">{{csvForm.batchChangeCsv.$viewValue.name}} is not a CSV file.</p>
+                                    <p><a href="https://www.vinyldns.io/portal/batch-changes#create-a-batch-change">Download sample CSV</a></p>
+                                </ng-form>
+                            </div>
+                            <p ng-if="newBatch.changes.length >= batchChangeLimit">Limit reached. Cannot add more than {{batchChangeLimit}} records per batch change.</p>
```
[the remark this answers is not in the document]
britneywright: yes it's possible to import more, but it will be invalidated when a user submits it. I wouldn't truncate anything, the user should decide how they want to modify their changes to proceed.

modules/portal/public/lib/batch-change/batch-change-new.controller.js
britneywright: removed them

modules/portal/public/lib/batch-change/batch-change-new.controller.js
```diff
@@ -108,5 +109,61 @@
...
+                if (file.type == "text/csv"){
+                    $scope.newBatch.changes = [];
+                    var reader = new FileReader();
+                    reader.onload = function(e) {
+                        var rows = e.target.result.split("\n");
+                        for(var i = 1; i < rows.length; i++) {
+                        if (rows[i].length < 10) { continue; }
+                            parseRow(rows[i])
+                        }
+                        $timeout();
```
mitruly: `$timeout()` here feels a bit hacky here; `$scope.$apply()` works for me.

modules/portal/public/lib/batch-change/batch-change-new.controller.js
```diff
@@ -108,5 +109,61 @@
...
+                    resetForm();
+                } else {
+                    $scope.alerts.push({type: 'danger', content: 'Import failed. Not a valid CSV file.'});
+                };
+
+                function parseRow(row) {
+                    var change = {};
+                    var headers = ["changeType", "type", "inputName", "ttl", "record"];
+                    var rowContent = row.split(",");
+                    for(var j = 0; j < rowContent.length; j++) {
```
mitruly: Should add a space after `for`.

modules/portal/public/lib/batch-change/batch-change-new.controller.js
```diff
@@ -108,5 +109,61 @@
...
+                        if (headers[j] == "changeType") {
+                            if (rowContent[j].trim().match(/add/i)) {
```
mitruly: We don't need to `trim()` before the `.match` since we're just checking and not saving this value; this would just be extra processing.

modules/docs/src/main/resources/microsite/static/batch-csv-sample.csv
```diff
@@ -0,0 +1,12 @@
...
+Add,A,test1.example.,200,1.2.3.4
+Add,AAAA+PTR,test2.example.,200,fd69:27cc:fe91::60
+Add,AAAA,test3.example.com.,200,fd69:27cc:fe91::60
+Add,CNAME,test4.example.com.,200,test0.example.com.
+Add,PTR,192.0.2.193,200,test.example.com.
+DeleteRecordSet,A+PTR,test5.example.com.,,1.1.1.1
+DeleteRecordSet,A,test6.exmaple.com.,,
+DeleteRecordSet,AAAA+PTR,test7.example.com.,,fd69:27cc:fe91::60
+DeleteRecordSet,CNAME,test8.example.com.,,
+DeleteRecordSet,PTR,193.0.2.192,,
```
mitruly: Since we're filtering blank lines, can we add a new line to the sample file?

modules/portal/public/lib/batch-change/batch-change-new.controller.js
```diff
@@ -108,5 +109,61 @@
...
             }
+
+            $scope.uploadCSV = function(file) {
+                if (file.type == "text/csv"){
+                    $scope.newBatch.changes = [];
+                    var reader = new FileReader();
+                    reader.onload = function(e) {
+                        var rows = e.target.result.split("\n");
+                        for(var i = 1; i < rows.length; i++) {
+                        if (rows[i].length < 10) { continue; }
```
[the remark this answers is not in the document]
britneywright: empty lines aren't completely empty, they still have the commas. probably need to do some regex here.

modules/portal/public/lib/batch-change/batch-change.directive.js
```diff
@@ -71,4 +71,49 @@
...
+            require: 'ngModel',
+            link: function(scope, elm, attrs, ctrl) {
+                var fileType;
+                ctrl.$validators.csv = function(modelValue, viewValue) {
+                    elm.on('change', function(e){
+                        if (e.target.files.length > 0) {
+                            fileType = e.target.files[0].type;
+                            ctrl.$setViewValue(e.target.files[0]);
+                        } else {
+                            ctrl.$setViewValue();
```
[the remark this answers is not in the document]
britneywright: yes it's local

PR #604, opened 2019-04-29
case ignored regexes for high value domain and name servers

nimaeskandary:
Ignored case for high value domain and name server regexes

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
+++ b/modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
@@ -25,6 +25,7 @@ import vinyldns.api.crypto.Crypto
 import com.comcast.ip4s._
 import net.ceedubs.ficus.Ficus._
 import net.ceedubs.ficus.readers.EnumerationReader._
+import vinyldns.api.domain.zone.ZoneRecordValidations
 import vinyldns.core.domain.record.RecordType
 
 import scala.collection.JavaConverters._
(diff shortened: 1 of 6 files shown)


modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
```diff
@@ -58,10 +59,11 @@ object VinylDNSConfig {
   lazy val cryptoConfig: Config = vinyldnsConfig.getConfig("crypto")
   lazy val system: ActorSystem = ActorSystem("VinylDNS", VinylDNSConfig.config)
   lazy val approvedNameServers: List[Regex] =
```
rebstar6: need to check the thing comparing to more so than the regex provided. if we hvd `.*high-value.*` and someone tries to make the record `HIGH-VALUE.test.com`, that should fail the hvd check.

rebstar6: `isNotHighValueFqdn` at a minimum would be impacted

nimaeskandary: ah okay I see what you're saying

nimaeskandary: @rebstar6 I added more validations, let me know if that covers what you were thinking

rebstar6: ooooh i see, i was reading this wrong the 1st time

PR #607, opened 2019-04-29
Zone BackendID in portal

nimaeskandary:
taking over https://github.com/vinyldns/vinyldns/pull/595

### DIFF
--- a/modules/api/functional_test/live_tests/zones/get_zone_test.py
+++ b/modules/api/functional_test/live_tests/zones/get_zone_test.py
@@ -130,10 +130,10 @@ def test_get_zone_by_name_returns_404_when_not_found(shared_zone_test_context):
     client.get_zone_by_name("zone_name_does_not_exist", status=404)
 
 
-def test_get_zone_backendids(shared_zone_test_context):
+def test_get_zone_backend_ids(shared_zone_test_context):
     """
     Test that you can get possible backend ids for zones
     """
     client = shared_zone_test_context.ok_vinyldns_client
-    response = client.get_backendids(status = 200)
+    response = client.get_backend_ids(status = 200)
     assert_that(response, is_([u'func-test-backend']))
--- a/modules/api/functional_test/live_tests/zones/get_zone_test.py
+++ b/modules/api/functional_test/live_tests/zones/get_zone_test.py
@@ -128,3 +128,12 @@ def test_get_zone_by_name_returns_404_when_not_found(shared_zone_test_context):
     client = shared_zone_test_context.ok_vinyldns_client
 
     client.get_zone_by_name("zone_name_does_not_exist", status=404)
+
+
+def test_get_zone_backendids(shared_zone_test_context):
+    """
+    Test that you can get possible backend ids for zones
+    """
+    client = shared_zone_test_context.ok_vinyldns_client
+    response = client.get_backendids(status = 200)
+    assert_that(response, is_([u'func-test-backend']))
--- a/modules/api/functional_test/live_tests/zones/get_zone_test.py
+++ b/modules/api/functional_test/live_tests/zones/get_zone_test.py
@@ -19,6 +19,7 @@ def test_get_zone_by_id(shared_zone_test_context):
     assert_that(retrieved['id'], is_(shared_zone_test_context.system_test_zone['id']))
     assert_that(retrieved['adminGroupName'], is_(shared_zone_test_context.ok_group['name']))
 
+
 def test_get_shared_zone_by_id(shared_zone_test_context):
     """
     Test get an existing shared zone by id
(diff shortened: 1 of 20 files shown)


modules/api/functional_test/vinyldns_python.py
```diff
@@ -398,6 +398,16 @@ def get_zone_by_name(self, zone_name, **kwargs):
 
         return data
 
+    def get_backendids(self, **kwargs):
```
pauljamescleary: oof, nitpick coming...

`def get_backend_ids`

And make the route `/backends/ids`

modules/portal/public/templates/zoneconnection-modal.html
```diff
@@ -32,6 +32,32 @@
...
+            </div>
+            <modal-element label="Backend Id">
+                <select id="zone-backendId" name="backendIdVal" class="form-control" ng-model="currentZone.backendId">
+                    <option value=""></option>
+                    <option ng-repeat="id in backendIds" value="{{id}}">
+                        {{id}}
+                    </option>
+                </select>
+                <span class="help-block">
+                    The Id for a pre-configured DNS connection configuration. Please contact your DNS admin team
```
mitruly: `Id` → `ID`

modules/portal/public/templates/zoneconnection-modal.html
```diff
@@ -32,6 +32,32 @@
...
+                    If none of the following options are filled, default connection settings are used.
+                </span>
+            </div>
+
+            <div class="dns-connection-form">
+                <h4>DNS Zone Backend ID (Optional)</h4>
+                <hr>
+            </div>
+            <modal-element label="Backend Id">
+                <select id="zone-backendId" name="backendIdVal" class="form-control" ng-model="currentZone.backendId">
```
mitruly: When choosing a backend ID and then clearing it before submitting the modal, I encounter the following error:
`HTTP 400 (Bad Request): Invalid backendId: [Some()]; please check system configuration`

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneService.scala
```diff
@@ -211,6 +212,9 @@ class ZoneService(
     } yield zoneChange
   }
 
+  def getBackendIds(): Result[List[String]] =
+    IO(VinylDNSConfig.configuredDnsConnections.dnsBackends.map(_.id)).toResult
```
mitruly: `IO` isn't needed here since `toResult` will convert it properly.

modules/api/functional_test/live_tests/zones/get_zone_test.py
```diff
@@ -128,3 +128,12 @@ def test_get_zone_by_name_returns_404_when_not_found(shared_zone_test_context):
     client = shared_zone_test_context.ok_vinyldns_client
 
     client.get_zone_by_name("zone_name_does_not_exist", status=404)
+
+
```
britneywright: extra line here

modules/portal/app/views/zones/zones.scala.html
```diff
@@ -80,6 +80,7 @@ <h3><span class="fa fa-table"></span> Zones {{ getZonePageTitle() }}</h3>
                                     <th>Name</th>
                                     <th>Email</th>
                                     <th>Admin Group</th>
+                                    <th>Backend ID</th>
```
britneywright: Do we actually need the backend IDs in this table?

modules/portal/app/views/zones/zoneTabs/manageZone.scala.html
```diff
@@ -68,6 +68,19 @@ <h3 class="panel-title">Zone Info</h3>
                                 </div>
                             </div>
 
+                            <div class="form-group" ng-if="isZoneAdmin">
```
britneywright: think this should always be visible, but disabled if not a zone admin. Like what we do with Zone Email

modules/portal/app/views/zones/zones.scala.html
nimaeskandary: I'm indifferent, was there when I took the pr over and just kept it

modules/portal/app/views/zones/zoneTabs/manageZone.scala.html
nimaeskandary: changed

modules/portal/app/views/zones/zones.scala.html
nimaeskandary: took it out

modules/api/functional_test/live_tests/zones/get_zone_test.py
nimaeskandary: That file was inconsistent so added the new lines between all the functions which is how we do it in other test files as file, in general python functions have two lines separating them unless they are class methods

PR #609, opened 2019-04-29
limit super user to normal access on record changes

rebstar6:
Fixes #.

Changes in this pull request:
-
-

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/AccessValidations.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/AccessValidations.scala
@@ -40,7 +40,7 @@ object AccessValidations extends AccessValidationAlgebra {
       NotAuthorizedError(
         s"""User '${auth.signedInUser.userName}' cannot create or modify zone '$zoneName' because
            |they are not in the Zone Admin Group '$zoneAdminGroupId'""".stripMargin
-          .replace("\n", " ")))(auth.canEditAll || auth.isGroupMember(zoneAdminGroupId))
+          .replace("\n", " ")))(auth.isSuper || auth.isGroupMember(zoneAdminGroupId))
 
   def canAddRecordSet(
       auth: AuthPrincipal,
@@ -94,7 +94,7 @@ object AccessValidations extends AccessValidationAlgebra {
       auth: AuthPrincipal,
       recordSets: List[RecordSetInfo],
       zone: Zone): List[RecordSetListInfo] =
-    if (auth.canEditAll || auth.isGroupMember(zone.adminGroupId))
+    if (auth.isSuper || auth.isGroupMember(zone.adminGroupId))
       recordSets.map(RecordSetListInfo(_, AccessLevel.Delete))
     else {
       val rulesForUser = zone.acl.rules.filter(ruleAppliesToUser(auth, _))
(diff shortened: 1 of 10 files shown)


modules/api/src/main/scala/vinyldns/api/domain/AccessValidations.scala
```diff
@@ -183,12 +183,12 @@ object AccessValidations extends AccessValidationAlgebra {
...
       recordOwnerGroupId: Option[String] = None): AccessLevel = auth match {
     case testUser if testUser.isTestUser && !zone.isTest => AccessLevel.NoAccess
-    case admin if admin.canEditAll || admin.isGroupMember(zone.adminGroupId) =>
+    case admin if admin.isGroupMember(zone.adminGroupId) =>
       AccessLevel.Delete
     case recordOwner
         if zone.shared && sharedRecordAccess(recordOwner, recordType, recordOwnerGroupId) =>
       AccessLevel.Delete
-    case supportUser if supportUser.canReadAll =>
+    case support if support.canReadAll =>
```
mitruly: Should we also add `isSuper` here to give at least `Read` access? My understanding is that super users are still between support users and admins, with the added ability to choose owner groups that they do not belong to.

nimaeskandary: can read all checks for the super and support flag

nimaeskandary: could maybe rename to supportOrSuper

mitruly: Good call, think renaming would make it clearer, but I'm good with this.

mitruly:
Seems like there are a couple of failing tests:
```
- should update relative NS record without trailing dot *** FAILED *** (64 milliseconds)
[info]   java.util.NoSuchElementException: None.get
[info]   at scala.None$.get(Option.scala:366)
[info]   at scala.None$.get(Option.scala:364)
[info]   at vinyldns.api.ResultHelpers.rightValue(ResultHelpers.scala:68)
[info]   at vinyldns.api.ResultHelpers.rightValue$(ResultHelpers.scala:68)
[info]   at vinyldns.api.domain.record.RecordSetServiceIntegrationSpec.rightValue(RecordSetServiceIntegrationSpec.scala:43)
[info]   at vinyldns.api.domain.record.RecordSetServiceIntegrationSpec.$anonfun$new$16(RecordSetServiceIntegrationSpec.scala:348)
[info]   at org.scalatest.concurrent.Futures.whenReady(Futures.scala:619)
[info]   at org.scalatest.concurrent.Futures.whenReady$(Futures.scala:617)
[info]   at vinyldns.api.domain.record.RecordSetServiceIntegrationSpec.whenReady(RecordSetServiceIntegrationSpec.scala:43)
[info]   at vinyldns.api.domain.record.RecordSetServiceIntegrationSpec.$anonfun$new$15(RecordSetServiceIntegrationSpec.scala:347)
[info]   ...
[info] - should delete successfully for super user in shared zone *** FAILED *** (19 milliseconds)
[info]   'Left(vinyldns.api.domain.zone.NotAuthorizedError: User super does not have access to delete shared-record.shared-zone-test.)' was not an Right, but should have been. (RecordSetServiceIntegrationSpec.scala:551)
```

PR #610, opened 2019-04-30
is test on zone create

nimaeskandary:
make zone isTest match user isTest

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneChangeGenerator.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneChangeGenerator.scala
@@ -33,8 +33,7 @@ object ZoneChangeGenerator {
       status: ZoneChangeStatus = Pending): ZoneChange =
     ZoneChange(
       zone
-        .copy(id = UUID.randomUUID().toString, created = DateTime.now, status = ZoneStatus.Syncing,
-          isTest = authPrincipal.signedInUser.isTest),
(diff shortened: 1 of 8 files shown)


modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneChangeGenerator.scala
```diff
@@ -33,7 +33,8 @@ object ZoneChangeGenerator {
       status: ZoneChangeStatus = Pending): ZoneChange =
     ZoneChange(
```
rebstar6: this isnt the right place. remove isTest from createZoneInput if we arent using it, and then on the line `zoneToCreate = Zone(createZoneInput)`, pass in the auth (or test flag directly)

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneService.scala
```diff
@@ -70,7 +70,7 @@ class ZoneService(
       _ <- zoneDoesNotExist(createZoneInput.name)
       _ <- adminGroupExists(createZoneInput.adminGroupId)
       _ <- canChangeZone(auth, createZoneInput.name, createZoneInput.adminGroupId).toResult
-      zoneToCreate = Zone(createZoneInput)
+      zoneToCreate = Zone(createZoneInput, auth.signedInUser.isTest)
```
rebstar6: `auth.isTestUser`

Issue #143, opened 2018-08-27
Allow users to search for zones they do not manage


pauljamescleary:
**Is your feature request related to a problem? Please describe.**
A user only can see / search for zones that they have access to.  If a zone is registered in VinylDNS, and the user does not have access, they have no way to know it is there.

**Describe the solution you'd like**
1. Allow users to _search_ for a zone they do not have access to and have it display.  
1. User should not be able to _click_ into the zone, but just see the "summary" detail on the search result list.
1. User should see the group email in the summary detail so they can contact the owner to potentially gain access to the zone.

britneywright:
this was resolved with #689

PR #613, opened 2019-05-03
WIP: Query all zones

pauljamescleary:
Allows a user to search across all zones in the system.

* Added a `searchAll` flag to the list zones route, service, and repository methods
* Added a new `getAccess` method to the `ZoneRepository` to determine which zones a user actually has access to
* Added a `hasAccess` flag to the `ZoneSummaryInfo` type that is returned that indicates if a user has access to the zone
* Updated the `ZoneService` to check for access before building the `ZoneSummaryInfo`
* Updated the portal to consider whether a user has access when displaying the zones in the zones page
* Updated the portal to add a check box to allow for "search all"

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,5 @@ release.version
 .ensime_cache
 package-lock.json
 *trustStore.jks
+.bloop
+.metals
--- a/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneProtocol.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneProtocol.scala
@@ -81,10 +81,11 @@ case class ZoneSummaryInfo(
     adminGroupId: String,
     adminGroupName: String,
     latestSync: Option[DateTime],
-    backendId: Option[String])
+    backendId: Option[String],
+    hasAccess: Boolean)
 
 object ZoneSummaryInfo {
-  def apply(zone: Zone, groupName: String): ZoneSummaryInfo =
+  def apply(zone: Zone, groupName: String, hasAccess: Boolean = true): ZoneSummaryInfo =
     ZoneSummaryInfo(
       name = zone.name,
       email = zone.email,
@@ -100,7 +101,8 @@ object ZoneSummaryInfo {
       adminGroupId = zone.adminGroupId,
       adminGroupName = groupName,
       latestSync = zone.latestSync,
-      zone.backendId
+      zone.backendId,
+      hasAccess
     )
 }
 
--- a/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneProtocol.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneProtocol.scala
@@ -66,6 +66,11 @@ object ZoneInfo {
     )
 }
 
+object ZoneAccessLevel extends Enumeration {
+  type ZoneAccessLevel = Value
+  val ReadOnly, ACLAccess, ZoneAdmin, SuperUser = Value
+}
+
 case class ZoneSummaryInfo(
     name: String,
     email: String,
@@ -81,11 +86,10 @@ case class ZoneSummaryInfo(
     adminGroupId: String,
     adminGroupName: String,
     latestSync: Option[DateTime],
-    backendId: Option[String],
-    hasAccess: Boolean)
+    backendId: Option[String])
 
 object ZoneSummaryInfo {
-  def apply(zone: Zone, groupName: String, hasAccess: Boolean = true): ZoneSummaryInfo =
+  def apply(zone: Zone, groupName: String): ZoneSummaryInfo =
     ZoneSummaryInfo(
       name = zone.name,
       email = zone.email,
@@ -101,8 +105,7 @@ object ZoneSummaryInfo {
       adminGroupId = zone.adminGroupId,
       adminGroupName = groupName,
       latestSync = zone.latestSync,
-      zone.backendId,
-      hasAccess
+      zone.backendId
     )
 }
 
@@ -205,7 +208,8 @@ case class ListZonesResponse(
     nameFilter: Option[String],
     startFrom: Option[String] = None,
     nextId: Option[String] = None,
-    maxItems: Int = 100)
+    maxItems: Int = 100,
+    searchAll: Boolean = false)
 
 // Errors
 case class InvalidRequest(msg: String) extends Throwable(msg)
--- a/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneProtocol.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneProtocol.scala
@@ -66,11 +66,6 @@ object ZoneInfo {
     )
 }
 
-object ZoneAccessLevel extends Enumeration {
-  type ZoneAccessLevel = Value
-  val ReadOnly, ACLAccess, ZoneAdmin, SuperUser = Value
-}
-
 case class ZoneSummaryInfo(
     name: String,
     email: String,
--- a/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneService.scala
@@ -134,9 +134,15 @@ class ZoneService(
       authPrincipal: AuthPrincipal,
       nameFilter: Option[String] = None,
       startFrom: Option[String] = None,
-      maxItems: Int = 100): Result[ListZonesResponse] = {
+      maxItems: Int = 100,
+      searchAll: Boolean = false): Result[ListZonesResponse] = {
     for {
-      listZonesResult <- zoneRepository.listZones(authPrincipal, nameFilter, startFrom, maxItems)
+      listZonesResult <- zoneRepository.listZones(
+        authPrincipal,
+        nameFilter,
+        startFrom,
+        maxItems,
+        searchAll)
       zones = listZonesResult.zones
       groupIds = zones.map(_.adminGroupId).toSet
       groups <- groupRepository.getGroups(groupIds)
--- a/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneService.scala
@@ -16,6 +16,7 @@
 
 package vinyldns.api.domain.zone
 
+import cats.effect.{ContextShift, IO}
 import cats.implicits._
 import vinyldns.api.{Interfaces, VinylDNSConfig}
 import vinyldns.api.domain.AccessValidationAlgebra
@@ -59,6 +60,9 @@ class ZoneService(
   import accessValidation._
   import zoneValidations._
   import Interfaces._
+
+  private implicit val cs: ContextShift[IO] =
+    IO.contextShift(scala.concurrent.ExecutionContext.global)
 
   def connectToZone(
       createZoneInput: CreateZoneInput,
@@ -136,6 +140,16 @@ class ZoneService(
       startFrom: Option[String] = None,
       maxItems: Int = 100,
       searchAll: Boolean = false): Result[ListZonesResponse] = {
+
+    def getGroupsAndAccess(
+        groupIds: Set[String],
+        zoneIds: Seq[String]): IO[(Set[Group], Seq[String])] =
+      if (zoneIds.isEmpty) {
+        groupRepository.getGroups(groupIds).map(groups => (groups, Nil))
+      } else {
+        (groupRepository.getGroups(groupIds), zoneRepository.getAccess(zoneIds, authPrincipal)).parTupled
+      }
+
     for {
       listZonesResult <- zoneRepository.listZones(
         authPrincipal,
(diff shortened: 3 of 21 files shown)


britneywright:
this is #143 right?

britneywright:
I think the checkbox should be below or to the right of the text field

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneProtocol.scala
```diff
@@ -100,7 +101,8 @@ object ZoneSummaryInfo {
       adminGroupId = zone.adminGroupId,
       adminGroupName = groupName,
       latestSync = zone.latestSync,
-      zone.backendId
+      zone.backendId,
+      hasAccess
```
britneywright: is it worth adding or changing this to levels of access. essentially no access, read-only access (support users and acl users), write access (zone admins and super users"?

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneRepository.scala
```diff
@@ -283,20 +283,51 @@ class MySqlZoneRepository extends ZoneRepository with ProtobufConversions with M
       }
     }
 
-  private def withAccessors(user: User, groupIds: Seq[String]): (String, Seq[Any]) =
+  def getAccess(zones: Seq[String], authPrincipal: AuthPrincipal): IO[Seq[String]] =
```
nimaeskandary: Think this should return a `Set`, we have a weird case where you can be in an admin group and have an acl rule on you for the same zone (not that it makes sense, you just could), so `accessor_id IN($accessorQuestionMarks)` will have two matches

nimaeskandary: Meh, guess is doesn't matter with how you're using it

nimaeskandary: if the user is super or support we can skip this query no?

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneService.scala
```diff
@@ -134,13 +138,30 @@ class ZoneService(
...
+      maxItems: Int = 100,
+      searchAll: Boolean = false): Result[ListZonesResponse] = {
+
+    def getGroupsAndAccess(
+        groupIds: Set[String],
+        zoneIds: Seq[String]): IO[(Set[Group], Seq[String])] =
+      if (zoneIds.isEmpty) {
+        groupRepository.getGroups(groupIds).map(groups => (groups, Nil))
+      } else {
+        (groupRepository.getGroups(groupIds), zoneRepository.getAccess(zoneIds, authPrincipal)).parTupled
```
rebstar6: is the parTuple adding a lot? why not just have these as separate lines in the for, following this is a tricky

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneProtocol.scala
rebstar6: +1 on that. it also seems to me that it makes more sense to split this piece off anyway. we can have a "search all" or "search yours" without pulling in the new `hasAccess` bool (or ideally accessLevel) and do that as a followup

also of note, arguably all shared zones should have "true" for hasAccess

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneService.scala
```diff
@@ -134,13 +138,30 @@ class ZoneService(
...
+      }
```
rebstar6: also, im not convinced we need the `zoneRepository.getAccess` function. would have to mess with the mysql, but id think a left join (vs normal join) on the initial call could retrieve all the info we need from that

that said, it does embed quite a bit of logic in the datastore directly

modules/portal/app/views/zones/zones.scala.html
```diff
@@ -42,14 +42,23 @@ <h3><span class="fa fa-table"></span> Zones {{ getZonePageTitle() }}</h3>
...
-                                        </button>
-                                    </span>
-                                    <input id="zone-search-text" ng-model="query" type="text" class="form-control"  placeholder="Zone Name"/>
+                            <form class="form-inline" ng-submit="refreshZones()">
+                                <div class="form-group" style="padding-left: 10px;">
+                                    <label class="check">
+                                        <input type="checkbox"
+                                               ng-model="searchAll" name="searchAll"
+                                               class="icheckbox_minimal-grey"/>
+                                        Search All Zones?</label>
```
rebstar6: i feel like people arent going to know what this means...why not always search all?

maybe more of a view toggle between "My Zones" and "All Zones", or something

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneService.scala
pauljamescleary: We want to do things in parallel if possible as these are independent.  I can tinker with it and see if I can make it a little more readable

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneService.scala
pauljamescleary: I really should have made a note of why I did it this way.  I did explore the left join thing at the time, but it wasn't working out.

I can revisit

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneRepository.scala
rebstar6: super definitely.

support yes if this remains a bool, no if theres more details in access leveling

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneProtocol.scala
pauljamescleary: I did the hasAccess to complete the UI, and to fix the bug at the same time.  @britneywright I think has a fix merged for the bug, so they could be separate PRs.  I can see doing that if the check access is tough.  Without the access indicator, this PR becomes ridiculously small (which means it gets merged quickly amirite?)

.gitignore
```diff
@@ -25,3 +25,5 @@ release.version
 .ensime_cache
 package-lock.json
 *trustStore.jks
+.bloop
```
britneywright: what is bloop???

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneService.scala
```diff
@@ -134,13 +138,30 @@ class ZoneService(
       authPrincipal: AuthPrincipal,
       nameFilter: Option[String] = None,
       startFrom: Option[String] = None,
-      maxItems: Int = 100): Result[ListZonesResponse] = {
+      maxItems: Int = 100,
+      searchAll: Boolean = false): Result[ListZonesResponse] = {
```
britneywright: can you update the attributes in the docs for this. and when this gets merged in we should make follow up cards in all of the tools to make sure it's supported.

modules/portal/app/views/zones/zones.scala.html
```diff
@@ -42,14 +42,23 @@ <h3><span class="fa fa-table"></span> Zones {{ getZonePageTitle() }}</h3>
 
                         <!-- SEARCH BOX -->
```
britneywright: just making note we'll want to update the portal screenshot in the docs for this.

.gitignore
pauljamescleary: bloop comes with metals, metals is a new scala thing for working in IDEs, specifically I use it in VS Code

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneService.scala
pauljamescleary: @rebstar6 the reason for the separate method and separate call is that `hasAccess` or `accessLevel` would be an attribute of the relationship between the user and the zone.

In the existing code, we return a `Zone` object from all of the `getters` in the repository.

We cannot add an access indicator to `Zone`.

However, the `ZoneSummaryInfo` makes sense because that is effectively the calling users' `View` of the zone data.

modules/portal/app/views/zones/zones.scala.html
britneywright: if we go "My Zones" / "All Zones" might be worth looking at allowing super and support users to make that distinction as well. We currently return all zones if a user is super or support, they have no way of seeing just the zones they have access to without their added access level.

pauljamescleary: We definitely will need "Advanced Search" at some point, although probably not for zones.  I went with the placement mimicking the "Group Manager" check box.

Can always have it checked and says "Only My Zones" or something to that effect.

modules/portal/public/lib/controllers/controller.zones.spec.js
```diff
@@ -74,12 +74,13 @@ describe('Controller: ZonesController', function () {
         var expectedMaxItems = 100;
         var expectedStartFrom = undefined;
         var expectedQuery = this.scope.query;
+        var expectedSearchAll = false;
```
mitruly: Optional: Could just have this defined once at the top of the file before the `beforeEach` since we're using the same value for each test.

modules/api/src/test/scala/vinyldns/api/domain/zone/ZoneServiceSpec.scala
```diff
@@ -518,9 +518,11 @@ class ZoneServiceSpec
     }
 
     "return Unknown group name if zone admin group cannot be found" in {
-      doReturn(IO.pure(ListZonesResults(List(abcZone, xyzZone))))
+      val resultZones = List(abcZone, xyzZone)
```
mitruly: Optional: Could have `resultZones` just defined once at the top since the same value is being assigned.

pauljamescleary: I actually deleted this as it was old from before we decided to move access stuff out

modules/portal/app/views/zones/zones.scala.html
```diff
@@ -42,14 +42,23 @@ <h3><span class="fa fa-table"></span> Zones {{ getZonePageTitle() }}</h3>
...
+                                <div class="form-group" style="padding-left: 10px;">
+                                    <div class="input-group">
+                                        <span class="input-group-btn">
+                                            <button id="zone-search-button" type="submit" class="btn btn-primary btn-left-round">
+                                              <span class="fa fa-search"></span>
+                                            </button>
+                                        </span>
+                                        <input id="zone-search-text" ng-model="query" type="text" class="form-control"  placeholder="Zone Name"/>
+                                    </div>
                                 </div>
```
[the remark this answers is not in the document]
pauljamescleary: @mitruly I think the view will be "read only" though, which should be ok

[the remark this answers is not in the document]
mitruly: @pauljamescleary As long as this is the path that we are going with for access, I am good with this. Just noting that I am currently running into an error when trying to access a zone that I am not explicitly given access to.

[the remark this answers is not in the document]
britneywright: The view isn't read-only, the zone page loads but none of the data does because it's a forbidden request. There shouldn't be view buttons on zones a user can't view.  We would need to change the restrictions in `getZone` in the API.

modules/portal/public/lib/controllers/controller.zones.spec.js
pauljamescleary: I used the `expectedSearchAll` to be consistent with the other checks.

modules/portal/app/views/zones/zones.scala.html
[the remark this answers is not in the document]
pauljamescleary: @mitruly actually this is bad, as we get a whole lot of 403s because the user cannot access the zone.

So, we should hide the VIEW button if the user does not have access right now.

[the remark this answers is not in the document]
rebstar6: but we took out the access part of this right? check on shared zones, people should have view on shared. assuming they do, in practice this will be not as bad as it seems in testing

or really just open up getZone while were here and let the people view all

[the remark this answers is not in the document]
britneywright: we don't distinguish between shared and private zones for validating if users can see a zone.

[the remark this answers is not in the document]
mitruly: @britneywright Don't think this matters right now until we open up read-only access to all users since users cannot currently view shared zones that they don't have explicit access.

pauljamescleary:
Turns out that we really need the hasAccess check if this is to move forward :/

pauljamescleary:
Going to close this right now, as our plans is to:

1. Allow most of the things read-only access globally
2. Add an access level to all of the things (default being ReadOnly, with other access levels depending on the type of entity)
3. Add a `searchAll` option to all `list` endpoints for all entities to return the appropriate access level for each.

Seems like a series of smaller PRs to come in.

britneywright:
@pauljamescleary I created some new issues related to this #645 and #643 there's also similar ones for the group endpoints

PR #615, opened 2019-05-07
manage DS records in the portal

### LINKED ISSUE #605: Cannot manage DS records in the portal
**Describe the bug**
DS records appear in the portal if they exist, but one cannot create, update, or delete them

**VinylDNS Version**
0.9.0

**Expected behavior**
1. I can create, update, or delete a DS record set

britneywright:
Fixes #605.

Changes in this pull request:
- add the DS record type to the record modal in the portal, for recordset creates and updates
- handle DS record type in the Records controller and service
- update records table to properly display DS record data


modules/docs/src/main/tut/api/recordset-model.md
```diff
@@ -68,6 +68,11 @@ AAAA         | address     | string      |
 <br>         |             |             |
 CNAME        | cname       | string      |
 <br>         |             |             |
+DS           | keytag      | integer     |
+DS           | algorithm   | integer     |
+DS           | digest type | integer     |
```
mitruly: This should be `digesttype` since that's what we deserialize from the request JSON (in `DnsJsonProtocol`).

britneywright: fixed

Issue #45, opened 2018-08-01
Roadmap doc


britneywright:
This already exists, but could use some sprucing. More details where useful.

PR #616, opened 2019-05-07
only show abandon zone button to zone admins

### LINKED ISSUE #606: Granting access via ACL Rules shows Abandon Zone button
**Describe the bug**
When a user or group is given ACL access to a zone, the user sees the "Abandon Zone" button in the portal.

**VinylDNS Version**
0.9.0

**To Reproduce**
1. Connect to a zone in VinylDNS
2. Add an ACL rule that grants access to a user who is NOT a zone admin OR super OR support user
3. Log in as this other user in the ACL rule
4. The user sees the zone in their zone list.  The user also sees the Abandon Zone button

**Expected behavior**
The `Abandon Zone` button should only be visible to zone admins, super users, or support users.

britneywright:
Fixes #606.

Changes in this pull request:
- only show the Abandon zone button to zone admins (members of the group that owns the zone and super users)
- note: support users don't have the ability to change zones so shouldn't see the button
- renamed `isGroupMember` function to `canAccessGroup` that applies to zone admins, support users and super users. This is used to determine if the group name should be linked to in the Zones table.
- added `isZoneAdmin` function that applies to zone admins and super users. This is used to to determine if the Abandon zone button is visible.

### DIFF
--- a/modules/portal/app/views/zones/zoneTabs/manageZone.scala.html
+++ b/modules/portal/app/views/zones/zoneTabs/manageZone.scala.html
@@ -198,17 +198,25 @@
 
             </div>
             <div class="form-group panel-footer" ng-if="isZoneAdmin">
-                <div class="col-md-12">
-                    <span ng-if="currentManageZoneState == manageZoneState.UPDATE">
-                        <button id="zone-update-button" type="button"
-                                ng-disabled="!objectsDiffer(updateZoneInfo, zoneInfo) && updateZoneInfo.hiddenKey == '' &&
-                                             updateZoneInfo.hiddenTransferKey == ''"
-                                class="btn btn-primary pull-right"
-                                ng-click="clickUpdateZone()">Update Zone</button>
-                        <button type="button" class="btn btn-primary" ng-click="refreshZone()"
-                                ng-disabled="!objectsDiffer(updateZoneInfo, zoneInfo) && updateZoneInfo.hiddenKey == '' &&
-                                             updateZoneInfo.hiddenTransferKey == ''">Reset</button>
-                    </span>
+                <div class="col-md-3">
+                    <button ng-if="currentManageZoneState == manageZoneState.UPDATE" type="button" class="btn btn-primary" ng-click="refreshZone()"
+                            ng-disabled="!objectsDiffer(updateZoneInfo, zoneInfo) && updateZoneInfo.hiddenKey == '' &&
+                                         updateZoneInfo.hiddenTransferKey == ''">
+                        Reset
+                    </button>
+                </div>
+                <div class="vinyldns-flex-right">
+                    <button ng-if="currentManageZoneState == manageZoneState.UPDATE" id="zone-update-button" type="button"
+                            ng-disabled="!objectsDiffer(updateZoneInfo, zoneInfo) && updateZoneInfo.hiddenKey == '' &&
+                                         updateZoneInfo.hiddenTransferKey == ''"
+                            class="btn btn-primary"
+                            ng-click="clickUpdateZone()">
+                        Update Zone
+                    </button>
+                    <button id="zone-delete-{{zone.name}}" class="btn btn-danger btn-rounded mb-control"
+                            ng-click="confirmDeleteZone();">
+                        Abandon Zone
+                    </button>
 
                     <span class="update-zone-msg" ng-if="currentManageZoneState == manageZoneState.CONFIRM_UPDATE">
                         <span>Are you sure you want to update this zone?&emsp;</span>
@@ -224,7 +232,7 @@
     <!-- END ZONE INFO FORM -->
 
     <!-- START ACL TABLE -->
-<div class="panel panel-default" ng-if="isZoneAdmin">
+<div class="panel panel-default">
 
     <div class="panel-heading">
         <h3 class="panel-title">
@@ -237,7 +245,7 @@
             <button id="refresh-acl-rules-button" class="btn btn-default" ng-click="refreshZone()">
                 <span class="fa fa-refresh"></span> Refresh
             </button>
-            <button id="create-acl-rule-button" class="btn btn-default" ng-click="clickCreateAclRule()">
+            <button ng-if="isZoneAdmin" id="create-acl-rule-button" class="btn btn-default" ng-click="clickCreateAclRule()">
                 <span class="fa fa-plus"></span> Create ACL Rule
             </button>
         </div>
@@ -281,7 +289,7 @@
                     </td>
 
                     <td>
-                        <span>
+                        <span ng-if="isZoneAdmin">
                             <button class="btn btn-info btn-sm" id="update-acl-rule-button-{{$index}}"
                                 ng-click="clickUpdateAclRule($index)">
                                 Update
@@ -460,3 +468,34 @@
     </modal>
 </form>
     <!-- END ACL MODAL -->
+
+<div class="modal fade in" id="delete_zone_connection_modal">
+    <div class="modal-dialog">
+        <div class="modal-content">
+            <div class="modal-header">
+                <button type="button" class="close" data-dismiss="modal" aria-label="Close"><span aria-hidden="true">×</span>
+                </button>
+                <div class="modal-title">Abandon Zone?</div>
+            </div>
+            <div class="modal-body">
+                <p>Are you sure you want to abandon the <strong>{{zoneInfo.name}}</strong> zone? This will disconnect the zone from Vinyl,
+                    but recordsets will still exist unless deleted from
+                    <a ng-href="/zones/{{zoneInfo.id}}">
+                        <strong>Manage Records</strong></a></p>
+            </div>
+            <div class="modal-footer">
+                <form name="deleteZoneConnectionForm" ng-submit="deleteZoneConnection()" role="form" class="form-horizontal" novalidate>
+                    <button type="submit" id="zone-delete-yes-button"
+                            class="btn btn-danger pull-right"
+                            ng-click="submitDeleteZone(zoneInfo.id)">
+                        Abandon
+                    </button>
+                    <button type="button" class="btn btn-primary"
+                            data-dismiss="modal">
+                        Close
+                    </button>
+                </form>
+            </div>
+        </div>
+    </div>
+</div>
--- a/modules/portal/app/views/zones/zoneTabs/manageZone.scala.html
+++ b/modules/portal/app/views/zones/zoneTabs/manageZone.scala.html
@@ -232,7 +232,7 @@
     <!-- END ZONE INFO FORM -->
 
     <!-- START ACL TABLE -->
-<div class="panel panel-default">
+<div class="panel panel-default" ng-if="canReadZone">
 
     <div class="panel-heading">
         <h3 class="panel-title">
--- a/modules/portal/public/lib/controllers/controller.records.js
+++ b/modules/portal/public/lib/controllers/controller.records.js
@@ -576,7 +576,6 @@ angular.module('controller.records', [])
     }
 
     function profileFailure(results) {
-        $scope.profile = $scope.profile || {};
         handleError(results, 'profileService::getAuthenticatedUserData-catch');
     }
 
--- a/modules/portal/public/lib/controllers/controller.records.js
+++ b/modules/portal/public/lib/controllers/controller.records.js
@@ -41,6 +41,7 @@ angular.module('controller.records', [])
     $scope.recordsetChanges = {};
     $scope.currentRecord = {};
     $scope.zoneInfo = {};
+    $scope.profile = {};
 
     var loadZonesPromise;
     var loadRecordsPromise;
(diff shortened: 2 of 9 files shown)


modules/portal/public/lib/controllers/controller.zones.js
```diff
@@ -61,11 +61,23 @@ angular.module('controller.zones', [])
...
-        var groupMember = $scope.myGroups.find(function(group) {
+    $scope.canAccessGroup = function(groupId) {
+        var groupAccess = $scope.myGroups.find(function(group) {
             return groupId === group.id;
         });
-        return groupMember !== undefined
+        return groupAccess !== undefined
+    };
+
+    $scope.isZoneAdmin = function(groupId) {
```
rebstar6: this is inconsistent with `function determineAdmin() // isZoneAdmin` in controller.records.js. Those two are doing the same thing, we should make sure they are the same (or in some shared place if possible)

mitruly: Based off of the `Expected Behavior` outlined in https://github.com/vinyldns/vinyldns/issues/606, we should also be checking if the user is a support user.

britneywright: isZoneAdmin is wrong in controller.records.js, well really that we didn't update the logic after adding the support user role. I'll update it.

modules/portal/public/lib/controllers/controller.zones.js
```diff
@@ -61,11 +61,23 @@ angular.module('controller.zones', [])
         }
     });
 
-    $scope.isGroupMember = function(groupId) {
-        var groupMember = $scope.myGroups.find(function(group) {
+    $scope.canAccessGroup = function(groupId) {
```
mitruly: Not sure if this function's name should be changed because functionality-wise, we're still only checking for membership to a given group.

modules/portal/public/lib/controllers/controller.zones.js
britneywright: support users don't have the ability to abandon zones, that is an oversight in the issue.

modules/portal/public/lib/controllers/controller.records.js
```diff
@@ -339,11 +347,17 @@ angular.module('controller.records', [])
...
     function determineAdmin(){
+        var groupMember;
         var groupIds = $scope.myGroups.map(function(grp) {return grp['id']});
-        $scope.isZoneAdmin = groupIds.indexOf($scope.zoneInfo.adminGroupId) > -1;
+        var theGroupIndex = groupIds.indexOf($scope.zoneInfo.adminGroupId);
+        if (theGroupIndex > -1) {
+            var stuff = $scope.myGroups[theGroupIndex].members.map(function(member) {return member['id']});
+            groupMember = stuff.indexOf($scope.profile.id) > -1;
+        }
+        $scope.isZoneAdmin = groupMember || $scope.profile.isSuper
```
britneywright: had to rewrite this to not use find because our current test framework doesn't support it. That's also why the similar functions in the zone specs aren't tested. Needed to revive this PR https://github.com/vinyldns/vinyldns/pull/457/files. I don't think it's a blocker for this though.

mitruly: This implementation seems to be more complicated than it needs to be. `$scope.myGroups` is populated by `getMembership()` via `$scope.refreshZone` on load - do we really need to double-check whether the profile ID matches when we are getting groups based on user?

We should be able to have this simply be:
```
    function determineAdmin(){
        var groupIds = $scope.myGroups.map(function(grp) {return grp['id']});
        $scope.isZoneADmin = groupIds.indexOf($scope.zoneInfo.adminGroupId) > -1 || $scope.profile.isSuper;
    }
```

modules/portal/public/lib/controllers/controller.records.js
```diff
@@ -63,6 +63,14 @@ angular.module('controller.records', [])
 
     $scope.isZoneAdmin = false;
 
+    profileService.getAuthenticatedUserData().then(function (results) {
```
mitruly: We should be able to remove this since we are getting everything that we need from groups service.

Note: Having functions executed that aren't functions themselves in plain script isn't ideal; if we need this function, should wrap it as a function and have it called when the page refreshes.

modules/portal/public/lib/controllers/controller.zones.js
mitruly: This comment still stands; the functionality of this is still explicitly checking group membership.

modules/portal/public/lib/controllers/controller.zones.js
mitruly: Same comment as in `controller.records.js` - is there any situation where profile group info would be different from the groups retrieved from group service?

modules/portal/public/lib/controllers/controller.records.js
britneywright: for super users and support users `myGroups` is actually all of the groups in Vinyl. So I have to look at the members of the group to distinguish between who is actually a zone admin vs who is just a support user, because support users don't have the ability to change zones. As it stands now support users see the abandon zone and sync zone buttons, but if they click those buttons they will get errors because the API does not allow it. And we don't have intentions of allowing that in the API at this time. So this is aimed at correctly reflecting user abilities based on their roles.

modules/portal/public/lib/controllers/controller.records.js
mitruly: Since we're keeping this, we should follow the existing format:
```
    $scope.profile = {}

    function retrieveProfile() {
      function success(response) {
        $log.log('profileService::getAuthenticatedUserData-success');
        $scope.profile = response.data
      }
      return profileService
        .getAuthenticatedUserData()
        .then(success)
        .catch(function (error){
          handleError(error, 'profileService::getAuthenticatedUserData-failure');
        });
    }
```

Keep in mind that `retrieveProfile` will have to be loaded before `refreshZone` is invoked.

modules/portal/public/lib/controllers/controller.zones.js
mitruly: Feels kind of hacky to return an assignment in a value. Keep in mind that each `find` is a function, which you have to return up to the calling function; the following works:
```
    $scope.isZoneAdmin = function(groupId) {
        var groupMember = $scope.myGroups.find(function(group) {
            if (group.id === groupId) {
                return group.members.find(function(member){
                    return member.id === $scope.profile.id;
                });
            };
        });
        return groupMember !== undefined || $scope.profile.isSuper
    };
```

modules/portal/public/lib/controllers/controller.records.js
```diff
@@ -339,11 +340,17 @@ angular.module('controller.records', [])
     }
 
     function determineAdmin(){
+        var groupMember;
         var groupIds = $scope.myGroups.map(function(grp) {return grp['id']});
-        $scope.isZoneAdmin = groupIds.indexOf($scope.zoneInfo.adminGroupId) > -1;
+        var theGroupIndex = groupIds.indexOf($scope.zoneInfo.adminGroupId);
+        if (theGroupIndex > -1) {
+            var stuff = $scope.myGroups[theGroupIndex].members.map(function(member) {return member['id']});
```
rebstar6: lol var stuff

modules/portal/public/lib/controllers/controller.zones.js
```diff
@@ -61,11 +61,22 @@ angular.module('controller.zones', [])
...
-        return groupMember !== undefined
+        return groupAccess !== undefined
+    };
+
+    $scope.isZoneAdmin = function(groupId) {
+        var groupMember = $scope.myGroups.find(function(group) {
+            if (groupId === group.id) {
+                return group.members.find(function(member){
+                    return $scope.profile.id === member.id;
+                });
```
rebstar6: this double find is oof. i see why were doing it, but i do wonder if it will substantially slow things down when you have 100 zones on the page...

is there a good way to test that?

britneywright: what if we remove abandon zone from the Zones table all together and put it in the Manage Zone tab?

britneywright:
### Updates
I removed the Abandon zone button from the Zones tab

Benefits:
- Removes the need to check the zone members for every zone admin group to determine whether the Abandon zone button should be visible. 
- Frees up space in the table

zones-list-zone-admin

---

Moved the Abandon Zone button into the Manage Zone tab in the Zone Detail page

Benefits:
  - Makes sense to put it here since this is the only places where changes to the zone itself take place

manage-zone-zone-admin

---

Allow support users and ACL users to view the ACL rules (without the create and update buttons)

Benefits:
  - This was restricted to zone admins, support users and super users previously, with no restrictions on the create and update buttons. Now allows ACL users to see their rules without the ability to edit.

manage zone - acl user

---

### Expectations by user type:
It should be noted all of these expectations are supported by the API, the portal has lagged behind in correctly reflecting what users can and can't do.
- Super users:
  - Zones table: **can** see links to zone admin group
  - Manage Records tab: **can** see Sync Zone button
  - Manage Zone tab: **can** see Abandon Zone button and related modal
  - ACL rules table: **can** see ACL rules table, Create Rule button and Update buttons on existing rules

- Zone admin:
  - Zones table: **can** see links to zone admin group
  - Manage Records tab: **can** see Sync Zone button
  - Manage Zone tab: **can** see Abandon Zone button and related modal
  - ACL rules table: **can** see ACL rules table, Create Rule button and Update buttons on existing rules

- Support admin:
  - Zones table: **can** see links to zone admin group
  - Manage Records tab: **cannot** see Sync Zone button
  - Manage Zone tab: **cannot** see Abandon Zone button and related modal
  - ACL rules table: **can** see ACL rules table, **cannot** see Create Rule button and Update buttons on existing rules

- User with ACL-only zone access:
  - Zones table: **cannot** see links to zone admin group
  - Manage Records tab: **cannot** see Sync Zone button
  - Manage Zone tab: **cannot** see Abandon Zone button and related modal
  - ACL rules table: **can** see ACL rules table, **cannot** see Create Rule button and Update buttons on existing rules

modules/portal/public/lib/controllers/controller.records.js
```diff
@@ -359,15 +361,17 @@ angular.module('controller.records', [])
...
+        $scope.isZoneAdmin = groupMember || $scope.profile.isSuper
     }
 
-    $scope.isGroupMember = function(groupId) {
-        var groupMember = $scope.myGroups.find(function(group) {
-            return groupId === group.id;
-        });
-        return groupMember !== undefined
+    $scope.canAccessGroup = function(groupId) {
+        $scope.myGroupIds.indexOf(groupId) > -1
```
mitruly: Functions should `return` a value if they are evaluate an expression; otherwise invoking this function could result in unintended behavior.

modules/portal/app/views/zones/zoneTabs/manageZone.scala.html
```diff
@@ -198,17 +198,25 @@ <h4>Transfer Connection (Optional)</h4>
 
             </div>
             <div class="form-group panel-footer" ng-if="isZoneAdmin">
-                <div class="col-md-12">
```
rebstar6: why are we changing stuff in manage zone? this card was just for the zone list

rebstar6: nevermind, got it

modules/portal/app/views/zones/zoneTabs/manageZone.scala.html
```diff
@@ -281,7 +289,7 @@ <h3 class="panel-title">
                     </td>
 
                     <td>
-                        <span>
+                        <span ng-if="isZoneAdmin">
```
rebstar6: we used to be hiding the whole ACL panel for non-zone admins (or supers). That should still be in place

modules/portal/public/lib/controllers/controller.records.js
```diff
@@ -359,15 +361,17 @@ angular.module('controller.records', [])
     }
 
     function determineAdmin(){
```
rebstar6: i think something like this makes more sense:

```
    function determineAdmin(){
        $scope.isZoneAdmin = $scope.profile.isSuper || isInAdminGroup()
    }

    function isInAdminGroup() {
        var groupMember = false;
        var theGroupIndex = $scope.myGroupIds.indexOf($scope.zoneInfo.adminGroupId);
        var grp = $scope.myGroups.find(function(grp) {return grp['id'] == $scope.zoneInfo.adminGroupId})
        if (grp !== undefined) {
          u = grp.members.find(function(user) {return user['id'] == $scope.profile.id})
          groupMember = u !== undefined
        }
        return groupMember
    }
```

2 reasons: 1, the find feels like what we want more than doing the index then get. if we need the thing, index across 2 lists feels iffy. and 2, im hoping (though cant confirm) that JS is smart enough to not even call the 2nd function if isSuper is true.

britneywright: just a note about find and why we can't use it if we're testing currently. Gotta revive that PR to upgrade our javascript testing tools: https://github.com/vinyldns/vinyldns/pull/45. will make a new issue for it.

mitruly: Not sure if it's worthwhile to get too involved with the portal/tests if the current portal is going to get deprecated with https://github.com/vinyldns/vinyldns/pull/621.

mitruly: @rebstar6 You are correct about the `isSuper` check order:
```
> var doot = true || 1+1
undefined
> doot
true
> var doot = false || 1+1
undefined
> doot
2
```

britneywright:
updated my explanations of changes https://github.com/vinyldns/vinyldns/pull/616#issuecomment-491160689

PR #621, opened 2019-05-07
Scala js client

nimaeskandary:
This will be a fun pr, anticipating it will be in review for a long time 🙂

Things I still need to port over from the JavaScript code base: 

* [x] CSV import
* [x] disable/hide group view buttons based on access
* [x] disable/hide grouplist buttons based on access
* [x] disable/hide manage record buttons based on access
* [x] disable/hide manage zone buttons based on access
* [x] disable/hide manage access buttons based on access
* [x] disable/hide zonelist buttons based on access
* [x] move zone delete to zone edit screen
* [x] make travis happy

The goals of this pr were to port over the same functionality and design of the existing client but to do so in Scala instead of the mix of play HTML and JavaScript that the existing client has. Hoping that having the HTML and all the logic supporting it in the same place will make the client easier to maintain in the future. I also was very pleased with how many errors the compiler can catch through the course of this since Scala has the added type safety JavaScript does not. 

The client heavily depends on a [scala-js react library](https://github.com/japgolly/scalajs-react). This library is a facade to React js, it adds a type safe way to access all react functions and classes. React is also the only npm dependency the client depends on other than gentelella.

### How it works

A new module is added to `build.sbt` called `client`. This is a [scala js](https://www.scala-js.org/) module as opposed to a JVM module, so at compile time it will use the scala js compiler and turn it into a single JavaScript file that contains all the HTML and logic. When the portal module compiles it will also compile the client module and add the JavaScript code to public assets.

The portal serves this file to a users web browser. From that point they have the single page app loaded and will no longer fetch pages from the portal node. React maintains a virtual dom depending on the route and renders different components and pages accordingly. 

### React components

A component can be pretty much anything, e.g. a form, a page, a div. When to make a new component as opposed to keeping it in the parent depends on how much logic that component performs, and if you want to share it with other components. 

React components aid in performance by only re-rendering when they need to, e.g. if the virtual dom has several components rendered and you only are interacting with one form, it will only re-generate the HTML for that section. This also is what lets the app be a single page app, the sidebar and topnav never have to re-render.  

Each component in the client is modeled by a Scala class. A component class has two useful case classes **Props**, which can contain attributes passed to the component from its parent component, and **State**, whose attributes can be modified.

That sounds side effecty, but luckily the State case class is modified in a functional way, by adding the transformation to a Callback queue maintained by React. When React executes the State transformation, it will re render the component with the new State. This is useful because you can have business logic on how to render different HTML based on the State, and is the general flow for all component interactions: user does something that triggers an event -> add a state altercation to the callback queue -> re render that component with the new state. 

### Tests

Tests can be ran via `sbt ";project client;test"`

No coverage for now, tracking this issue https://github.com/scoverage/sbt-scoverage/issues/284 and this one https://github.com/scoverage/scalac-scoverage-plugin/issues/196. It seems scoverage on a scala js project works for some and not others, so will see if that is sorted out. 

### Running locally

Need to add `v2client.enabled = true` to your portal module app conf or local conf. After that you can access it from the play project on the route `/v2`

You can also access a docker-up-vinyldns version at http://18.188.190.179:9001/v2/


nimaeskandary:
Also seems like travis ran out of memory compiling the client module 🙂 will need to look into that

.gitignore
```diff
@@ -25,3 +25,5 @@ release.version
 .ensime_cache
 package-lock.json
 *trustStore.jks
+node_modules
```
nimaeskandary: This is for scoverage though we are not using it. If the scoverage bug is ever fixed, we need to npm install js-dom into root for the client tests to work. Don't have to do that now though

build.sbt
```diff
@@ -379,10 +379,36 @@ lazy val portal = (project in file("modules/portal")).enablePlugins(PlayScala, A
...
+  .enablePlugins(AutomateHeaderPlugin, ScalaJSBundlerPlugin)
+  .settings(sharedSettings)
+  .settings(testSettings)
+  .settings(scalaStyleCompile ++ scalaStyleTest)
+  .settings(
+    name := "client",
+    libraryDependencies ++= clientDependencies.value ++ clientTestDependencies.value.map(_ % "test"),
+    npmDependencies in Compile ++= clientNpmDependencies,
+    webpackBundlingMode := BundlingMode.LibraryOnly(),
+    unmanagedSourceDirectories in Compile ++= (unmanagedSourceDirectories in (core, Compile)).value
```
nimaeskandary: You can utilize things from the core module as long as the scala js compiler likes it. Many scala libraries now support going through the scala js compiler. Joda Date Time though is an example of a library that does't, so we cant use any case classes that us the Joda Date Time type

modules/portal/public/scalajs/datalist-polyfill.js
```diff
@@ -0,0 +1,6 @@
+/*
```
nimaeskandary: The `datalist` html5 element is how the group select fields are done, as it lets you search for a group by name or id. All major browsers support it now, but this polyfill is in case someone has an older browser, e.g. IE9

modules/portal/public/scalajs/loadScalaJsLibrary.js
```diff
@@ -0,0 +1,5 @@
+var exports = window;
```
nimaeskandary: https://scalacenter.github.io/scalajs-bundler/reference.html

modules/portal/app/views/v2client.scala.html
```diff
@@ -0,0 +1,44 @@
...
+		<link rel="stylesheet" type="text/css" id="theme" href="/public/gentelella/vendors/bootstrap/dist/css/bootstrap.min.css"/>
+		<link rel="stylesheet" type="text/css" href="/public/gentelella/vendors/font-awesome/css/font-awesome.css">
+		<link rel="stylesheet" type="text/css" id="theme" href="/public/gentelella/build/css/custom.css"/>
+		<link rel="stylesheet" type="text/css" id="theme" href="/public/css/vinyldns.css"/>
+		<link rel="stylesheet" type="text/css" id="theme" href="/public/css/theme-overrides.css"/>
+	</head>
+	<body onload="ReactApp.main('root')">
+		<div id="root"></div>
+
+		<!-- choose full opt if it exists, fast opt otherwise -->
```
nimaeskandary: If you're compiling the portal via the sbt `run` command then a "fast optimized" version of the client is compiled, if you are actually building an executable then the "full optimized" version is compiled. This logic loads the full opt first if it exists and falls back to the fast opt for local development

nimaeskandary:
This is feature complete now, need to figure out this Travis out of memory issue though

.jvmopts
```diff
@@ -1,5 +1,5 @@
 -Xms512M
--Xmx4096M
+-Xmx2048M
```
nimaeskandary: travis doesn't actually have 4096M to give up for the jvm heap, we just never hit that limit before

mitruly:
### Initial functionality review
Normal VinylDNS UI interaction:
- Need more spacing in "view" sections (eg. zone view, group view, etc)
- Logout should prompt a confirmation box/modal
- Clicking on `Zones` from the sidebar doesn't put `/zones` in URL (though `/v2/zones` _does_ redirect properly)
- Alert/confirmation when creating group is different
- Modal validations happen immediately upon loading rather than on submission (ie. when opening a modal, all required fields are displaying as missing)
- Confirmation prompt for submitting create group/recordset modals and batch create is another key stroke
- Certain web confirmation pop-ups feel like they should be like the confirmation modals in v3 portal
- Connect to zone modal checkbox for each connection type feels complex; can this just be a section where you click and it shows a dropdown section for that field (eg. backend ID, connection info, etc)
- Delete ACL prompt should be ACL description, not user associated
- Batch change deserialization errors should probably be extracted before returning to the user (example message that pops up in alert: `creating batch request [400] [Bad Request]
{"errors":["A must be a valid IPv4 Address"]}`)

Miscellaneous UI:
- Showing numbered paging buttons is confusing when there is no next page
- Having update/delete buttons that we can't click can be confusing in the sense where users may think that it's possible to change those with certain permissions (eg. zone apex SOA and NS records)
- Modal for ACL rules and selecting multiple types should support shift + click functionality like our current functionality rather than appending to an array of types
- Error alerts remain on screen until closed
- Certain dialogue text boxes don't wrap the text properly (eg. sync zone)

nimaeskandary:
Thanks for the notes @mitruly! Wanted to start some dialogue on a few of your review bullets:

* Modal validations happen immediately upon loading - someone comments today in gitter about how not having required fields highlighted was confusing. I'm more inclined to validate right away as opposed to expect people to make a mistake and validate after, so want to open that to discussion
* The alert box auto removal for success and not for errors was intentional. A success goes away after 5 seconds, while errors stay so users have time to fully digest it and ask for support based off the error message
* Confirmation boxes are an interesting point. Browsers have a built in prompt modal that returns a boolean. Building out code for modals that do the same thing but look a little prettier seemed silly to me for confirmations 
* Would love to sit down and chat about the best way to design the connect to zone modal. Since we have so many options now its a lot to pack in

modules/client/src/main/scala/vinyldns/client/models/README.txt
```diff
@@ -0,0 +1,16 @@
+A note about these models:
+
+Though they are similar to the models in vinyldns.core, they exist to utilize the uPickle library for json reading
+and writing. Attribute names of classes have to match how the API reads and writes json exactly.
+
+To utilize uPickle, attribute types have to be something uPickle knows how to read and write. These can be Scala
+built-ins, or other custom models in this directory.
+
+Would have loved to somehow incorporate models in vinyldns.core, but certain types in there are not compatible
+with Scala.js, such as Joda Data Time. You can still depend on core if you want, but just make sure you do not try
```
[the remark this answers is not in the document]
nimaeskandary: True, I should make a larger readme probably at module root and make serialization a section

modules/portal/app/views/v2client.scala.html
```diff
@@ -0,0 +1,44 @@
...
+			val src = toAsset(name)
+			val loader = toAsset("scalajs/loadScalaJsLibrary.js")
+			<div>
+				<script src={s"$src"} type="text/javascript"></script>
+				<script src={s"$loader"} type="text/javascript"></script>
+				{
+					Seq(s"client-opt.js", s"client-fastopt.js").find(exists).map { name =>
+					val src = toAsset(name)
+					<script src={s"$src"} type="text/javascript"></script>
+					}.get
```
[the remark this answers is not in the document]
nimaeskandary: Will think about the best way to handle that, it failing would be an odd case because if that fails it means the scala js module failed to compile

modules/client/src/test/scala/vinyldns/client/SharedTestData.scala
```diff
@@ -0,0 +1,189 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package vinyldns.client
+
+/*
```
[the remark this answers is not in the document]
nimaeskandary: oops, don't know how that got in there

pauljamescleary:
Closing this temporarily as we do need to revisit the Portal

PR #622, opened 2019-05-08
delete recordsets from repo on zone delete

nimaeskandary:
When VinylDNS accepts a zone delete it removes the zone from the zone database but not the zone's recordsets from the recordset database. This orphans recordsets in the repo that turn into junk data. Even if someone reconnected to the zone the zone will have a new ID and duplicate all its recordsets in the recordset repo.

* Added logic to ZoneChangeHandler such that on a successful zone delete it will call a function to delete all recordsets with that zone id
* The ZoneChangeHandler will process the successful delete zone change even if the recordset delete fails
* Recordsets are NOT deleted from the DNS backend, just the recordsets table

### DIFF
--- a/modules/api/src/it/scala/vinyldns/api/domain/zone/ZoneServiceIntegrationSpec.scala
+++ b/modules/api/src/it/scala/vinyldns/api/domain/zone/ZoneServiceIntegrationSpec.scala
@@ -18,18 +18,20 @@ package vinyldns.api.domain.zone
 
 import cats.effect._
 import org.joda.time.DateTime
-import org.scalatest.concurrent.PatienceConfiguration
+import org.scalatest._
+import org.scalatest.concurrent.{PatienceConfiguration, ScalaFutures}
 import org.scalatest.mockito.MockitoSugar
 import org.scalatest.time.{Seconds, Span}
-import vinyldns.api.{DynamoDBApiIntegrationSpec, MySqlApiIntegrationSpec, ResultHelpers}
+import scalikejdbc.DB
+import vinyldns.api.{MySqlApiIntegrationSpec, ResultHelpers}
 import vinyldns.api.domain.AccessValidations
 import vinyldns.api.domain.record.RecordSetChangeGenerator
 import vinyldns.api.engine.TestMessageQueue
 import vinyldns.core.domain.auth.AuthPrincipal
-import vinyldns.core.domain.membership.{Group, GroupRepository, User, UserRepository}
+import vinyldns.core.domain.membership.{GroupRepository, UserRepository}
 import vinyldns.core.domain.record._
-import vinyldns.dynamodb.repository.{DynamoDBRecordSetRepository, DynamoDBRepositorySettings}
 import vinyldns.core.TestZoneData.testConnection
+import vinyldns.core.TestMembershipData.{okAuth, okGroup, okUser}
 import vinyldns.core.domain.zone._
 
 import scala.concurrent.Await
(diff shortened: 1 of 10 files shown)


modules/api/src/it/scala/vinyldns/api/domain/zone/ZoneServiceIntegrationSpec.scala
```diff
@@ -75,32 +74,42 @@ class ZoneServiceIntegrationSpec
...
     typ = RecordType.A,
     ttl = 38400,
     status = RecordSetStatus.Active,
     created = DateTime.now,
     records = List(AData("10.1.1.1")))
 
-  private val changeSetSOA = ChangeSet(RecordSetChangeGenerator.forAdd(testRecordSOA, zone))
-  private val changeSetNS = ChangeSet(RecordSetChangeGenerator.forAdd(testRecordNS, zone))
-  private val changeSetA = ChangeSet(RecordSetChangeGenerator.forAdd(testRecordA, zone))
+  private val changeSetSOA = ChangeSet(RecordSetChangeGenerator.forAdd(testRecordSOA, liveZone))
```
britneywright: do you still need to call this liveZone instead of zone?

nimaeskandary: went back in and got rid of live zone completely, no point if we have okZone in shared test data

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
```diff
@@ -296,6 +302,23 @@ class MySqlRecordSetRepository extends RecordSetRepository with Monitored {
         }
       }
     }
+
+  def deleteRecordSetsInZone(zoneId: String, zoneName: String): IO[Int] =
```
mitruly: I think it's better to have `deleteRecordSetsInZone` return an `IO[Unit]` since: there is nothing that we're realistically using back in the API from this result, and 2. we're logging already logging the error/dropped record set count in the `MySqlRecordSetRepository`.

Think we should be able to do something like:
```
  def deleteRecordSetsInZone(zoneId: String, zoneName: String): IO[Unit] =
    monitor("repo.RecordSet.deleteRecordSetsInZone") {
      IO {
        val numDeleted = DB.localTx { implicit s =>
          DELETE_RECORDSETS_IN_ZONE
            .bind(zoneId)
            .update()
            .apply()
        }
        logger.info(s"Deleted $numDeleted records from zone $zoneName (zone id: $zoneId)")
      }.handleErrorWith { error =>
        IO(logger.error(s"Failed deleting records from zone $zoneName (zone id: $zoneId)", error))
      }
    }
```

modules/api/src/main/scala/vinyldns/api/engine/ZoneChangeHandler.scala
```diff
@@ -36,6 +36,24 @@ object ZoneChangeHandler {
               status = ZoneChangeStatus.Failed,
               systemMessage = Some(duplicateZoneError.message))
           )
+        case Right(_) if zoneChange.changeType == ZoneChangeType.Delete =>
```
mitruly: Assuming that we change `deleteRecordSetsInZone` to an `IO[Unit]`, we can simplify this to:
```
        case Right(_) if zoneChange.changeType == ZoneChangeType.Delete =>
          recordSetRepository
            .deleteRecordSetsInZone(zoneChange.zone.id, zoneChange.zone.name)
            .flatMap(_ =>
              zoneChangeRepository.save(zoneChange.copy(status = ZoneChangeStatus.Synced)))
```

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
nimaeskandary: Hm good thought

PR #625, opened 2019-05-10
Include `*.csv` files in the built microsite

calvellido:
Changes in this pull request:

 * Include `.csv` files as part of the ones to be included when building the site.

This is needed due to an [sbt-microsites setting](https://github.com/47deg/sbt-microsites/blob/0a8387363c2c87179c433ca406fd85d428ece6dc/src/main/scala/microsites/MicrositesPlugin.scala#L113) where only some extensions are included as part the built microsite.


britneywright:
Thanks @calvellido!

PR #628, opened 2019-05-10
group duplicate zones in listZones query

### LINKED ISSUE #614: Duplicate zones in list response if admin group + acl
**Describe the bug**

If a user is in the zone admin group and they have an acl rule on them then the list zones response will add the zone twice

```
{
  "zones": [
    {
      "name": "ok.",
      "email": "test@test.com",
      "status": "Active",
      "created": "2019-05-04T18:06:39Z",
      "updated": "2019-05-04T19:13:07Z",
      "id": "4933e02e-b10a-4e66-b8a4-7345138ecef7",
      "account": "system",
      "shared": false,
      "acl": {
        "rules": [
          {
            "accessLevel": "Read",
            "groupId": "92274cbe-87c2-4e7a-b2e7-2cb01df7acd1",
            "recordTypes": [
              
            ]
          },
          {
            "accessLevel": "Read",
            "userId": "testuser",
            "recordTypes": [
              
            ]
          }
        ]
      },
      "adminGroupId": "92274cbe-87c2-4e7a-b2e7-2cb01df7acd1",
      "adminGroupName": "test2",
      "latestSync": "2019-05-04T18:06:39Z"
    },
    {
      "name": "ok.",
      "email": "test@test.com",
      "status": "Active",
      "created": "2019-05-04T18:06:39Z",
      "updated": "2019-05-04T19:13:07Z",
      "id": "4933e02e-b10a-4e66-b8a4-7345138ecef7",
      "account": "system",
      "shared": false,
      "acl": {
        "rules": [
          {
            "accessLevel": "Read",
            "groupId": "92274cbe-87c2-4e7a-b2e7-2cb01df7acd1",
            "recordTypes": [
              
            ]
          },
          {
            "accessLevel": "Read",
            "userId": "testuser",
            "recordTypes": [
              
            ]
          }
        ]
      },
      "adminGroupId": "92274cbe-87c2-4e7a-b2e7-2cb01df7acd1",
      "adminGroupName": "test2",
      "latestSync": "2019-05-04T18:06:39Z"
    },
    {
      "name": "vinyldns.",
      "email": "test@test.com",
      "status": "Active",
      "created": "2019-05-03T18:02:35Z",
      "id": "1588bfab-7978-4826-9ff6-9ef5e01d978f",
      "account": "system",
      "shared": false,
      "acl": {
        "rules": [
          
        ]
      },
      "adminGroupId": "92274cbe-87c2-4e7a-b2e7-2cb01df7acd1",
      "adminGroupName": "test2",
      "latestSync": "2019-05-03T18:02:36Z"
    }
  ],
  "maxItems": 100
}
```
**VinylDNS Version**
0.9.2-SNAPSHOT

britneywright:
Fixes #614

Changes in this pull request:
- Group zones by ID instead of returning all zones to prevent duplicates
- Update spec to use `contain theSameElementsAs` instead of `contain allElementsOf` so duplicate zones in the response would fail

### DIFF
--- a/modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneRepository.scala
+++ b/modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneRepository.scala
@@ -237,8 +237,7 @@ class MySqlZoneRepository extends ZoneRepository with ProtobufConversions with M
             sb.append(filters.mkString(" AND "))
           }
 
-          sb.append(s" GROUP BY z.id ")
-          sb.append(s" ORDER BY z.name ASC ")
+          sb.append(s" GROUP BY z.name ")
           sb.append(s" LIMIT $maxItems")
 
           val query = sb.toString
--- a/modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneRepository.scala
+++ b/modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneRepository.scala
@@ -237,6 +237,7 @@ class MySqlZoneRepository extends ZoneRepository with ProtobufConversions with M
             sb.append(filters.mkString(" AND "))
           }
 
+          sb.append(s" GROUP BY z.id ")
           sb.append(s" ORDER BY z.name ASC ")
           sb.append(s" LIMIT $maxItems")
 
--- a/modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlZoneRepositoryIntegrationSpec.scala
+++ b/modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlZoneRepositoryIntegrationSpec.scala
@@ -443,8 +443,54 @@ class MySqlZoneRepositoryIntegrationSpec
             acl = ZoneACL()
           )
 
-        // we are going to have 5 zones that havea different admin group id
+        // we are going to have 5 zones that have a different admin group id
         if (num % 2 == 0) z.copy(adminGroupId = differentAdminGroupId) else z
+      }
+
+      val sorted = testZones.sortBy(_.name)
+      val filtered = sorted.filter(_.adminGroupId == testZoneAdminGroupId)
+      val expectedFirstPage = filtered.take(2)
+      val expectedSecondPage = filtered.slice(2, 4)
+      val expectedThirdPage = filtered.slice(4, 6)
+
+      // make sure our auth is a member of the testZoneAdminGroup
+      val auth = AuthPrincipal(dummyUser, Seq(testZoneAdminGroupId))
+
+      saveZones(testZones).unsafeRunSync()
+
+      repo.listZones(auth, None, None, 2)
+        .unsafeRunSync().zones should contain theSameElementsInOrderAs expectedFirstPage
+
+      repo.listZones(auth, None, Some(filtered(1).name), 2)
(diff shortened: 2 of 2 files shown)


nimaeskandary:
`distinct` was removed from the query in https://github.com/vinyldns/vinyldns/pull/579/files since it negatively impacted performance when there are many zones in the db. Wonder if since we only get a max of 100 at a time it may be better to remove duplicates after `RecordSetRepository.listZones` is called

nimaeskandary:
Recommend syncing up with @pauljamescleary and running the same performance test he did for https://github.com/vinyldns/vinyldns/pull/579 to make sure it wouldn't cause the same issue

nimaeskandary:
Worried a `group by` will just have the same execution path that the `distinct` did given there is added overhead from the `limit`

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneRepository.scala
```diff
@@ -237,6 +237,7 @@ class MySqlZoneRepository extends ZoneRepository with ProtobufConversions with M
             sb.append(filters.mkString(" AND "))
           }
 
+          sb.append(s" GROUP BY z.id ")
           sb.append(s" ORDER BY z.name ASC ")
```
mitruly: Should be able to just do `GROUP BY z.name` here and remove `ORDER BY z.name ASC`; both `id` and `name` are unique keys in the table and `GROUP BY` will inherently sort in alpha order.

mitruly:
Discussing with @nimaeskandary, think it also makes sense to add an index to our `zone` table for `name` to speed up the ordering.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneRepository.scala
nimaeskandary: Possible solution. All in all we should probably be doing performance testing on both given we have had issues with this query recently

rebstar6: sorry if im late on this game, is this better than `SELECT DISTINCT z.data` in the initial query?

rebstar6: oh, oops, scroll up. got it

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneRepository.scala
```diff
@@ -237,7 +237,7 @@ class MySqlZoneRepository extends ZoneRepository with ProtobufConversions with M
             sb.append(filters.mkString(" AND "))
           }
 
-          sb.append(s" ORDER BY z.name ASC ")
```
rebstar6: add a test in `MySqlZoneRepositoryIntegrationSpec` - you can duplicate the problem there. before this fix, with an ACL added, it will return dups. now it should be one.

wanna also ensure paging still works properly :)

britneywright: made a new paging test for a user with zone admin and ACL access.

britneywright:
no noticeable time difference between `ORDER BY` AND `GROUP BY` when retrieving all of the zones.

PR #632, opened 2019-05-15
Add Notifier to Vinyl core

dscleaver:
Add the Notifier trait and support for loading Notifiers from config

Changes in this pull request:
- Add Notifier trait
- Add Notification wrapper class
- Add NotifierConfig class
- Add NotifierProvider trait
- Add NotifierLoader object
- Add AllNotifiers wrapper for loaded Notifier instances

### DIFF
--- a/modules/core/src/main/scala/vinyldns/core/notifier/AllNotifiers.scala
+++ b/modules/core/src/main/scala/vinyldns/core/notifier/AllNotifiers.scala
@@ -18,13 +18,11 @@ package vinyldns.core.notifier
 
 import cats.effect.{ContextShift, IO}
 import cats.implicits._
-import scala.concurrent.ExecutionContext
 import org.slf4j.LoggerFactory
 
-final case class AllNotifiers(notifiers: List[Notifier]) extends Notifier {
+final case class AllNotifiers(notifiers: List[Notifier])(implicit val cs: ContextShift[IO]) {
 
   private val logger = LoggerFactory.getLogger("AllNotifiers")
-  private implicit val cs: ContextShift[IO] = IO.contextShift(ExecutionContext.global)
 
   def notify(notification: Notification[_]): IO[Unit] =
     for {
--- a/modules/core/src/main/scala/vinyldns/core/notifier/AllNotifiers.scala
+++ b/modules/core/src/main/scala/vinyldns/core/notifier/AllNotifiers.scala
@@ -1,0 +1,38 @@
+/*
+ * Copyright 2018 Comcast Cable Communications Management, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package vinyldns.core.notifier
+
+import cats.effect.{ContextShift, IO}
+import cats.implicits._
+import scala.concurrent.ExecutionContext
+import org.slf4j.LoggerFactory
+
+final case class AllNotifiers(notifiers: List[Notifier]) extends Notifier {
+
+  private val logger = LoggerFactory.getLogger("AllNotifiers")
+  private implicit val cs: ContextShift[IO] = IO.contextShift(ExecutionContext.global)
+
+  def notify(notification: Notification[_]): IO[Unit] =
+    for {
+      _ <- notifiers.parTraverse(notify(_, notification))
+    } yield ()
+
+  def notify(notifier: Notifier, notification: Notification[_]): IO[Unit] =
+    notifier.notify(notification).handleErrorWith { e =>
+      IO { logger.error(s"Notifier ${notifier.getClass.getSimpleName} failed.", e) }
+    }
+}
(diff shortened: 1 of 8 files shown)


modules/core/src/main/scala/vinyldns/core/notifier/NotifierLoader.scala
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+package vinyldns.core.notifier
+import vinyldns.core.domain.membership.UserRepository
+import cats.effect.IO
+import cats.implicits._
+
+object NotifierLoader {
+
+  def loadAll(configs: List[NotifierConfig], userRepository: UserRepository): IO[Notifier] =
```
rebstar6: even in the case of a single (or no) notifiers, we are saying we will wrap in an AllNotifiers. I think that's fine, though i would then probably just have this return IO[AllNotifiers] and AllNotifiers doesnt need to extend the Notifier trait

also of note, if we didnt have this precedent, id say just have the load be object:apply for AllNotifiers. i could be convinced both to leave it here or put it there.

dscleaver: hmmm.. I think I was comfortable hiding the specific implementation details. I can see your point though that contractually, `AllNotifiers` would then capture the "this should never fail" requirement

modules/core/src/main/scala/vinyldns/core/notifier/AllNotifiers.scala
```diff
@@ -0,0 +1,38 @@
...
+
+import cats.effect.{ContextShift, IO}
+import cats.implicits._
+import scala.concurrent.ExecutionContext
+import org.slf4j.LoggerFactory
+
+final case class AllNotifiers(notifiers: List[Notifier]) extends Notifier {
+
+  private val logger = LoggerFactory.getLogger("AllNotifiers")
+  private implicit val cs: ContextShift[IO] = IO.contextShift(ExecutionContext.global)
```
pauljamescleary: probably makes sense to have the context shift passed in implicitly, either on the notify or more likely the case class itself.

modules/core/src/test/scala/vinyldns/core/notifier/NotifierLoaderSpec.scala
```diff
@@ -0,0 +1,131 @@
...
+import cats.effect.IO
+import org.mockito.Mockito._
+
+import scala.collection.JavaConverters._
+
+object Mocks extends MockitoSugar {
+  val mockUserRepository: UserRepository = mock[UserRepository]
+}
+
+object MockNotifierProvider extends MockitoSugar {
```
pauljamescleary: why are these separate objects?

modules/core/src/test/scala/vinyldns/core/notifier/NotifierLoaderSpec.scala
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    "return a notifier for valid config of one notifier" in {
+      val notifier = NotifierLoader.loadAll(List(goodConfig), mockUserRepository).unsafeRunSync()
+
+      notifier shouldNot be(null)
+
+      val notification = Notification(3)
+
+      reset(mockNotifier)
```
pauljamescleary: If this is to reset mocks between tests, use the `BeforeEach` from ScalaTest and do a `reset` in the `def beforeEach`

modules/core/src/main/scala/vinyldns/core/notifier/AllNotifiers.scala
dscleaver: I was thinking this originally. I think it makes sense to pass it through load... grab it if you need it.

modules/core/src/test/scala/vinyldns/core/notifier/NotifierLoaderSpec.scala
```diff
@@ -0,0 +1,131 @@
...
+  import Mocks._
+  import MockNotifierProvider._
+
+  "loadAll" should {
+
+    "return some notifier with no configs" in {
+
+      val notifier = NotifierLoader.loadAll(List.empty, mockUserRepository).unsafeRunSync()
+
+      notifier shouldNot be(null)
```
mitruly: `notifier shouldBe a[Notifier]` seems better than non-`null` check

modules/core/src/test/scala/vinyldns/core/notifier/AllNotifiersSpec.scala
```diff
@@ -0,0 +1,70 @@
...
+
+  "notifier" should {
+    "notify all contained notifiers" in {
+
+      val notifier = AllNotifiers(mockNotifiers)
+
+      val notification = Notification("anything")
+
+      mockNotifiers.foreach { mock =>
+        reset(mock)
```
nimaeskandary: An alternative to sharing a single mock object and calling reset is having a fixture that provides a new mock for each test

```
trait Fixture {
    val mockNotifiers = List.fill(3)(mock[Notifier])
}
"notifier" should {
    "notify all contained notifiers" in new Fixture {...}
    "suppress errors from notifiers" in new Fixture{...}
}
```

modules/core/src/test/scala/vinyldns/core/notifier/NotifierLoaderSpec.scala
nimaeskandary: Made a comment above about fixtures as an alternative to beforeEach

modules/core/src/test/scala/vinyldns/core/notifier/NotifierLoaderSpec.scala
dscleaver: I wasn't thinking... changed it.

modules/core/src/test/scala/vinyldns/core/notifier/NotifierLoaderSpec.scala
```diff
@@ -0,0 +1,132 @@
...
+  override def beforeEach: Unit =
+    reset(mockNotifier)
+
+  "loadAll" should {
+
+    "return some notifier with no configs" in {
+
+      val notifier = NotifierLoader.loadAll(List.empty, mockUserRepository).unsafeRunSync()
+
+      notifier shouldBe a[AllNotifiers]
```
mitruly: Can use `an[AllNotifiers]` here.

PR #633, opened 2019-05-15
dotted host zone discovery in batch

rebstar6:
Needs tests

the idea here is to open up zone discovery to be able to find zones for legal dotted hosts.

For now, just throwing all TXT records to this, but could be expanded to others if we decide thats the way to go.

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
@@ -220,10 +220,8 @@ class BatchChangeService(
       change: ChangeInput,
       zoneMap: ExistingZones): SingleValidation[ChangeForValidation] = {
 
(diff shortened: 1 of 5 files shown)


modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -100,15 +100,36 @@ class BatchChangeService(
       }
       .toBatchResult
 
-  def getZonesForRequest(changes: ValidatedBatch[ChangeInput]): IO[ExistingZones] = {
+  // zone name possibilities for all non-PTR changes
+  def getPossibleNonPtrZoneNames(nonPtr: List[ChangeInput]): Set[String] = {
+    def getZonesForDottedTypes(dotted: ChangeInput): Set[String] =
+      dotted.inputName
```
rebstar6: will move this to a helper. its finding all the zone options, so for test.example.com, this should get List(test.example.com, example.com, com)

PR #636, opened 2019-05-18
Wildcard search

pauljamescleary:
Support wildcard searches for zones and record sets

* Updated MySqlRecordSetRepository to replace `*` with `%` for the `listRecordSets`
* Updated  MySqlZoneRepository to replace `*` with `%` for `listZones`
* Added integration tests to verify searches
* Updated api documentation to indicate wildcards are supported
* Fixed the `Aws4Authenticator`.  The main issue is how we are calculating the signature server side.  We are `urlDecoding` the query params _first_, and then calculate signature.  However, the way that it works on the client (verified in python and java) is that the query params are **url encoded** before signature calculation.  We may have other issues with query strings in the future.  In otherwords, the client says `recordNameFilter=%2Alist%2A` is signed, whereas on the server we are calculating signature based on `recordNameFilter=*list*`

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/route/Aws4Authenticator.scala
+++ b/modules/api/src/main/scala/vinyldns/api/route/Aws4Authenticator.scala
@@ -239,6 +239,7 @@ class Aws4Authenticator {
       .
       // and doesn't encode '~' at all
       replaceAllLiterally("%7E", "~")
+      .replaceAllLiterally("*", "%2A") // aws encodes the asterisk specially
 
   private def hexString(bs: Array[Byte]) =
     bs.foldLeft("")((out, b) => f"$out%s${b & 0x0ff}%02x")
--- a/modules/docs/src/main/tut/api/list-recordsets.md
+++ b/modules/docs/src/main/tut/api/list-recordsets.md
@@ -16,7 +16,7 @@ Retrieves a list of RecordSets from the zone
 
 name          | type          | required?   | description |
  ------------ | ------------- | ----------- | :---------- |
-recordNameFilter    | string        | no          | One or more characters contained in the name of the record set to search for.  For example `vinyl`.  This is a contains search only, no wildcards or regular expressions are supported |
+recordNameFilter    | string        | no          | Characters that are part of the record name to search for.  The wildcard character `*` is supported, for example `www*` |
 startFrom     | *any*         | no          | In order to advance through pages of results, the startFrom is set to the `nextId` that is returned on the previous response.  It is up to the client to maintain previous pages if the client wishes to advance forward and backward.   If not specified, will return the first page of results |
 maxItems      | integer       | no          | The number of items to return in the page.  Valid values are 1 to 100. Defaults to 100 if not provided. |
 
--- a/modules/docs/src/main/tut/api/list-recordsets.md
+++ b/modules/docs/src/main/tut/api/list-recordsets.md
@@ -16,7 +16,7 @@ Retrieves a list of RecordSets from the zone
 
 name          | type          | required?   | description |
  ------------ | ------------- | ----------- | :---------- |
-recordNameFilter    | string        | no          | Characters that are part of the record name to search for.  The wildcard character `*` is supported, for example `www*` |
+recordNameFilter    | string        | no          | Characters that are part of the record name to search for.  The wildcard character `*` is supported, for example `www*`.  Omit the wildcard when searching for an exact record name. |
 startFrom     | *any*         | no          | In order to advance through pages of results, the startFrom is set to the `nextId` that is returned on the previous response.  It is up to the client to maintain previous pages if the client wishes to advance forward and backward.   If not specified, will return the first page of results |
 maxItems      | integer       | no          | The number of items to return in the page.  Valid values are 1 to 100. Defaults to 100 if not provided. |
 
--- a/modules/docs/src/main/tut/api/list-zones.md
+++ b/modules/docs/src/main/tut/api/list-zones.md
@@ -16,7 +16,7 @@ Retrieves the list of zones a user has access to.  The zone name is only sorted 
 
 name          | type          | required?   | description |
  ------------ | ------------- | ----------- | :---------- |
-nameFilter    | string        | no          | One or more characters contained in the name of the zone to search for.  For example `www-`.  This is a contains search only, no wildcards or regular expressions are supported |
+nameFilter    | string        | no          | Characters that are part of the zone name to search for.  The wildcard character `*` is supported, for example `www*` |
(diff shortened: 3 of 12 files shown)


modules/api/src/main/scala/vinyldns/api/route/Aws4Authenticator.scala
```diff
@@ -239,6 +239,7 @@ class Aws4Authenticator {
       .
       // and doesn't encode '~' at all
       replaceAllLiterally("%7E", "~")
+      .replaceAllLiterally("*", "%2A") // aws encodes the asterisk specially
```
pauljamescleary: I think we may have other issues with our custom Aws4 stuff with url encoded characters in the query strings.  This is a short term workaround for just the asterisk character

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
```diff
@@ -191,7 +191,7 @@ class MySqlRecordSetRepository extends RecordSetRepository with Monitored {
 
           val params = (Some('zoneId -> zoneId) ++
             startFrom.map(n => 'startFrom -> n) ++
-            recordNameFilter.map(f => 'nameFilter -> s"%$f%") ++
+            recordNameFilter.map(f => 'nameFilter -> f.replace('*', '%')) ++
```
rebstar6: hmmm on doing this in the RSR, `*` actually has meaning for wildcards...

pauljamescleary: This is correct.  We could add special escaping for wildcards (like "*").  99.9999999% of record names are not wildcard, so I recommend we do that in a follow up PR where we can think escapes for searches.

modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlRecordSetRepositoryIntegrationSpec.scala
```diff
@@ -359,9 +356,58 @@ class MySqlRecordSetRepositoryIntegrationSpec
...
       val startFrom = Some(newRecordSets(1).name)
-      val found = repo.listRecordSets(okZone.id, startFrom, Some(3), Some("z")).unsafeRunSync()
+      val found = repo.listRecordSets(okZone.id, startFrom, Some(3), Some("*z*")).unsafeRunSync()
+      found.recordSets.map(_.name) should contain theSameElementsInOrderAs expectedNames
+    }
+    "return record sets using starts with wildcard" in {
+      val recordNames = List("aaa", "aab", "ccc")
+      val expectedNames = recordNames.filter(_.startsWith("aa"))
+
+      val newRecordSets =
```
mitruly: This is more of a question: for single actions, would it be better to not use a `for`-comprehension and do something like `val newRecordSets = recordNames.map(name => aaaa.copy(zoneId = okZone.id, name = name, id = UUID.randomUUID().toString))`?

pauljamescleary: Not sure I follow.

In general, always use for comprehensions unless you have only one line in the for comprehnesion, then use straight map or flatmap

Also in general, keep with the convention already in the file that you are working with.  If you decide to deviate from what is already in the file, then refactor EVERYTHING to be consistent with the update you made.

pauljamescleary: Finally, in unit and integration tests _use caution_ when trying to be too **DRY**.   Often times, it is better to see all the test setup inside the test itself rather than having to hunt it down.  This can be a style thing and is a hot button for debate, but tests can be a pain in the butt if you strive too much for reuse.

modules/docs/src/main/tut/api/list-zones.md
```diff
@@ -16,7 +16,7 @@ Retrieves the list of zones a user has access to.  The zone name is only sorted
 
 name          | type          | required?   | description |
  ------------ | ------------- | ----------- | :---------- |
-nameFilter    | string        | no          | One or more characters contained in the name of the zone to search for.  For example `www-`.  This is a contains search only, no wildcards or regular expressions are supported |
+nameFilter    | string        | no          | Characters that are part of the zone name to search for.  The wildcard character `*` is supported, for example `www*` |
```
britneywright: based on this PR the nameFilter requires an exact match of the zone name unless the wildcard characters are used, we could be more explicit in explaining that.

modules/docs/src/main/tut/api/list-recordsets.md
```diff
@@ -16,7 +16,7 @@ Retrieves a list of RecordSets from the zone
 
 name          | type          | required?   | description |
  ------------ | ------------- | ----------- | :---------- |
-recordNameFilter    | string        | no          | One or more characters contained in the name of the record set to search for.  For example `vinyl`.  This is a contains search only, no wildcards or regular expressions are supported |
+recordNameFilter    | string        | no          | Characters that are part of the record name to search for.  The wildcard character `*` is supported, for example `www*` |
```
britneywright: same as my comment in list-zones

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
rebstar6: sounds good

britneywright:
should we add a tooltip in the portal to mention how the wildcard search works?

pauljamescleary:
@britneywright I don't think we need to put a tooltip in the portal for this.  I think the documentation should be sufficient here.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneRepository.scala
```diff
@@ -228,7 +228,7 @@ class MySqlZoneRepository extends ZoneRepository with ProtobufConversions with M
           sb.append(withAccessorCheck)
 
           val filters = List(
-            zoneNameFilter.map(flt => s"z.name LIKE '%$flt%'"),
+            zoneNameFilter.map(flt => s"z.name LIKE '${flt.replace('*', '%')}'"),
```
nimaeskandary: should this be replaceAll?

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
nimaeskandary: should this be replaceAll?

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneRepository.scala
nimaeskandary: Ah nvm

PR #637, opened 2019-05-20
wrap text in the Records table

### LINKED ISSUE #477: Large record data can look strange in the UI

britneywright:
fixes #477 

Issue:
Really long record data, record names, etc. in the Records table does not wrap and makes the table extend off the page.

Screenshots of current table:
records table

records table in mobile view

Changes in this pull request:
- in the Records table in the portal added a class that sets the `word-break` to `break-all` so text can break between any characters (not just spaces and hyphens) to wrap onto multiple lines

Screenshots of portal with wrapping added:
updated records table

updated mobile view

### DIFF
--- a/modules/portal/app/views/batchChanges/batchChangeDetail.scala.html
+++ b/modules/portal/app/views/batchChanges/batchChangeDetail.scala.html
@@ -95,28 +95,28 @@
                                 <!--record data name based on record type-->
                                     <td ng-if="change.type=='A'||change.type=='AAAA'">
                                         <ul>
-                                            <li>{{change.record.address}}</li>
+                                            <li class="wrap-long-text">{{change.record.address}}</li>
                                         </ul>
                                     </td>
                                     <td ng-if="change.type=='CNAME'">
                                         <ul>
-                                            <li>{{change.record.cname}}</li>
+                                            <li class="wrap-long-text">{{change.record.cname}}</li>
                                         </ul>
                                     </td>
                                     <td ng-if="change.type=='PTR'">
                                         <ul>
-                                            <li>{{change.record.ptrdname}}</li>
+                                            <li class="wrap-long-text">{{change.record.ptrdname}}</li>
                                         </ul>
                                     </td>
                                     <td ng-if="change.type=='TXT'">
                                         <ul>
-                                            <li>{{change.record.text}}</li>
+                                            <li class="wrap-long-text">{{change.record.text}}</li>
                                         </ul>
                                     </td>
                                     <td ng-if="change.type=='MX'">
                                         <ul>
-                                            <li>Preference: {{change.record.preference}}</li>
-                                            <li>Exchange: {{change.record.exchange}}</li>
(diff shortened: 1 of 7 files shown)


mitruly:
If we decide to wrap the record name, the record change name will also have to be wrapped:
unwrapped_dotted_txt_record_change_name

mitruly:
@britneywright The record set name wraps, but the record change name still doesn't wrap (in both the `Manage Records` and `Change History` tabs) - are these going to be handled in a separate PR?

britneywright:
@mitruly I was only planning to address the Records table in this PR, but it's probably not too much additional effort to address all the tables in portal.

mitruly:
@britneywright Right, I was referring to the record set change table in addition to the record set table since some of that information (eg. name) is shared data. I'm fine with having non-record table wrapping addressed in a separate PR (if there are wrapping issues), but it makes sense to address the record set and record change history tables at the same time.

modules/portal/app/views/batchChanges/batchChangeDetail.scala.html
```diff
@@ -95,28 +95,28 @@ <h3 class="panel-title">Changes</h3>
...
                                     <td ng-if="change.type=='PTR'">
                                         <ul>
-                                            <li>{{change.record.ptrdname}}</li>
+                                            <li class="wrap-long-text">{{change.record.ptrdname}}</li>
                                         </ul>
                                     </td>
                                     <td ng-if="change.type=='TXT'">
                                         <ul>
-                                            <li>{{change.record.text}}</li>
+                                            <li class="wrap-long-text">{{change.record.text}}</li>
```
mitruly: If we're wrapping batch change detail table, can we also wrap the record and input name in this screen?

modules/portal/app/views/zones/zoneTabs/manageRecords.scala.html
```diff
@@ -29,14 +29,14 @@ <h3 class="panel-title">
...
-                    <td>{{change.recordSet.name}}</td>
+                    <td class="wrap-long-text">{{change.recordSet.name}}</td>
                     <td>{{change.recordSet.type}}</td>
                     <td>{{change.changeType}}</td>
                     <td>{{change.userName}}</td>
                     <td>
                         <span class="label label-{{ getRecordChangeStatusLabel(change.status) }}">{{ change.status }}</span>
                     </td>
-                    <td class="col-md-3">
+                    <td class="col-md-3 wrap-long-text">
```
mitruly: Not sure if this is an issue, but the change table above `Recent Record Changes` section is stretched wider than before. Looks a bit strange with the extra whitespace if all record names are short.

PR #639, opened 2019-05-21
Call Notifiers on BatchChange completion

dscleaver:
Fixes #.

Changes in this pull request:
- Load Notifiers from config
- Change updateSingleChanges to return the full BatchChange.
- Call Notifiers on successful BatchChange

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
+++ b/modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
@@ -55,10 +55,12 @@ object VinylDNSConfig {
 
   lazy val notifierConfigs: IO[List[NotifierConfig]] =
     vinyldnsConfig
-      .getConfigList("notifiers")
+      .getStringList("notifiers")
       .asScala
       .toList
-      .map(loadConfigF[IO, NotifierConfig](_))
+      .map { configKey =>
+        loadConfigF[IO, NotifierConfig](vinyldnsConfig, configKey)
+      }
       .parSequence
 
   lazy val restConfig: Config = vinyldnsConfig.getConfig("rest")
--- a/modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
+++ b/modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
@@ -33,6 +33,7 @@ import scala.util.matching.Regex
 import vinyldns.core.domain.zone.{ConfiguredDnsConnections, DnsBackend, ZoneConnection}
 import vinyldns.core.queue.MessageQueueConfig
 import vinyldns.core.repository.DataStoreConfig
+import vinyldns.core.notifier.NotifierConfig
 
 object VinylDNSConfig {
 
@@ -50,6 +51,14 @@ object VinylDNSConfig {
       .map { configKey =>
         loadConfigF[IO, DataStoreConfig](vinyldnsConfig, configKey)
       }
+      .parSequence
+
+  lazy val notifierConfigs: IO[List[NotifierConfig]] =
+    vinyldnsConfig
+      .getConfigList("notifiers")
+      .asScala
+      .toList
+      .map(loadConfigF[IO, NotifierConfig](_))
       .parSequence
 
   lazy val restConfig: Config = vinyldnsConfig.getConfig("rest")
--- a/modules/api/src/main/scala/vinyldns/api/engine/RecordSetChangeHandler.scala
+++ b/modules/api/src/main/scala/vinyldns/api/engine/RecordSetChangeHandler.scala
@@ -27,6 +27,9 @@ import vinyldns.core.domain.record._
 
 import scala.concurrent.ExecutionContext
 import scala.concurrent.duration._
+import vinyldns.core.notifier.AllNotifiers
+import vinyldns.core.notifier.Notification
+import vinyldns.core.domain.batch.BatchChangeStatus
 
 object RecordSetChangeHandler {
 
@@ -38,13 +41,15 @@ object RecordSetChangeHandler {
   def apply(
       recordSetRepository: RecordSetRepository,
       recordChangeRepository: RecordChangeRepository,
-      batchChangeRepository: BatchChangeRepository)(
+      batchChangeRepository: BatchChangeRepository,
+      notifiers: AllNotifiers)(
       implicit timer: Timer[IO]): (DnsConnection, RecordSetChange) => IO[RecordSetChange] =
     (conn, recordSetChange) => {
       process(
         recordSetRepository,
         recordChangeRepository,
         batchChangeRepository,
+        notifiers,
         conn,
         recordSetChange)
     }
@@ -53,6 +58,7 @@ object RecordSetChangeHandler {
       recordSetRepository: RecordSetRepository,
       recordChangeRepository: RecordChangeRepository,
       batchChangeRepository: BatchChangeRepository,
+      notifiers: AllNotifiers,
       conn: DnsConnection,
       recordSetChange: RecordSetChange)(implicit timer: Timer[IO]): IO[RecordSetChange] =
     for {
@@ -64,7 +70,16 @@ object RecordSetChangeHandler {
       singleBatchChanges <- batchChangeRepository.getSingleChanges(
         recordSetChange.singleBatchChangeIds)
       singleChangeStatusUpdates = updateBatchStatuses(singleBatchChanges, completedState.change)
-      _ <- batchChangeRepository.updateSingleChanges(singleChangeStatusUpdates)
+      batchChangeOption <- batchChangeRepository.updateSingleChanges(singleChangeStatusUpdates)
+      _ <- batchChangeOption.map { bc =>
+        bc.status match {
+          case BatchChangeStatus.Complete =>
+            notifiers.notify(Notification(bc))
+          case BatchChangeStatus.Failed =>
+            notifiers.notify(Notification(bc))
+          case _ => IO.unit
+        }
+      }.sequence
     } yield completedState.change
 
   def updateBatchStatuses(
--- a/modules/api/src/main/scala/vinyldns/api/engine/RecordSetChangeHandler.scala
+++ b/modules/api/src/main/scala/vinyldns/api/engine/RecordSetChangeHandler.scala
@@ -73,11 +73,10 @@ object RecordSetChangeHandler {
       batchChangeOption <- batchChangeRepository.updateSingleChanges(singleChangeStatusUpdates)
       _ <- batchChangeOption.map { bc =>
         bc.status match {
-          case BatchChangeStatus.Complete =>
+          case BatchChangeStatus.Pending =>
+            IO.unit
+          case _ =>
             notifiers.notify(Notification(bc))
-          case BatchChangeStatus.Failed =>
-            notifiers.notify(Notification(bc))
-          case _ => IO.unit
         }
       }.sequence
     } yield completedState.change
--- a/modules/api/src/test/resources/application.conf
+++ b/modules/api/src/test/resources/application.conf
@@ -88,6 +88,15 @@ vinyldns {
     }
   }
 
+  notifiers = [
+    {
+      class-name = "someclass"
+      settings = {
+        value = "test"
+      }
+    }
+  ]
+
   backends = [
     {
       id = "test"
--- a/modules/api/src/test/resources/application.conf
+++ b/modules/api/src/test/resources/application.conf
@@ -88,14 +88,14 @@ vinyldns {
     }
   }
 
-  notifiers = [
-    {
-      class-name = "someclass"
-      settings = {
-        value = "test"
-      }
+  notifiers = [ "test-notifier"]
+
+  test-notifier {
+    class-name = "someclass"
+    settings {
+      value = "test"
     }
-  ]
+  }
 
   backends = [
     {
--- a/modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlBatchChangeRepository.scala
+++ b/modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlBatchChangeRepository.scala
@@ -62,6 +62,15 @@ class MySqlBatchChangeRepository
          | WHERE bc.id = ?
         """.stripMargin
 
+  private final val GET_BATCH_CHANGE_METADATA_FROM_SINGLE_CHANGE =
+    sql"""
+         |SELECT bc.id, bc.user_id, bc.user_name, bc.created_time, bc.comments, bc.owner_group_id
+         |  FROM single_change sc
+         |  JOIN batch_change bc
+         |    ON sc.batch_change_id = bc.id
+         | WHERE sc.id = ?
+        """.stripMargin
+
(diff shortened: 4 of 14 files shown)


modules/api/src/main/scala/vinyldns/api/engine/RecordSetChangeHandler.scala
```diff
@@ -64,7 +70,16 @@ object RecordSetChangeHandler {
       singleBatchChanges <- batchChangeRepository.getSingleChanges(
         recordSetChange.singleBatchChangeIds)
       singleChangeStatusUpdates = updateBatchStatuses(singleBatchChanges, completedState.change)
-      _ <- batchChangeRepository.updateSingleChanges(singleChangeStatusUpdates)
+      batchChangeOption <- batchChangeRepository.updateSingleChanges(singleChangeStatusUpdates)
+      _ <- batchChangeOption.map { bc =>
```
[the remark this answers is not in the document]
dscleaver: ah great... that was a misunderstanding on my part.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlBatchChangeRepository.scala
```diff
@@ -118,34 +127,59 @@ class MySqlBatchChangeRepository
...
+                    'recordSetChangeId -> singleChange.recordChangeId,
+                    'recordSetId -> singleChange.recordSetId,
+                    'zoneId -> singleChange.zoneId,
+                    'id -> singleChange.id
+                  )
+                case Left(e) => throw e
+              }
             }
+            UPDATE_SINGLE_CHANGE.batchByName(batchParams: _*).apply()
+            GET_BATCH_CHANGE_METADATA_FROM_SINGLE_CHANGE
```
mitruly: Do we need this to specifically return an `Option[BatchChange]`? The result of `storeQueuingFailures` is not used anywhere and we could get rid of everything after the initial query `apply`.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlBatchChangeRepository.scala
```diff
@@ -118,34 +127,59 @@ class MySqlBatchChangeRepository
...
+              .bind(singleChanges.head.id)
+              .map { result =>
+                BatchChange(
```
mitruly: Since types are cheap, might be better to have a `BatchChangeMetadata` class instead of artificially setting `changes` as `Nil` if we keep this.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlBatchChangeRepository.scala
```diff
@@ -118,34 +127,59 @@ class MySqlBatchChangeRepository
...
+                    'inputName -> singleChange.inputName,
+                    'changeType -> changeType.toString,
+                    'data -> data.toByteArray,
+                    'status -> singleChange.status.toString,
+                    'recordSetChangeId -> singleChange.recordChangeId,
+                    'recordSetId -> singleChange.recordSetId,
+                    'zoneId -> singleChange.zoneId,
+                    'id -> singleChange.id
+                  )
+                case Left(e) => throw e
```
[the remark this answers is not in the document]
dscleaver: Looks like this code was working like this previously. Maybe it should be a follow-on?

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlBatchChangeRepository.scala
dscleaver: I copied this set of code from `getBatchChangeMetadata`, which I'm not really loving the sort of extra copy-pasta in this method. It's itching for some refactoring, but a good approach wasn't readily apparent. One reason to share and get more thought behind it.

modules/api/src/main/scala/vinyldns/api/engine/RecordSetChangeHandler.scala
```diff
@@ -64,7 +70,16 @@ object RecordSetChangeHandler {
...
+        bc.status match {
+          case BatchChangeStatus.Complete =>
```
pauljamescleary: we have a race condition here, make a separate card as it feels like an edge case and it requires some real thinking about how we process batch changes.

If we have 1000 batch changes processing across nodes, it is possible that we never see a Complete status and never notify.

dscleaver: the BatchChange is fetched in the transaction that processes the complete status of the single changes. This was to prevent the race condition... are you still concerned that there is one here?.

modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
```diff
@@ -52,6 +53,14 @@ object VinylDNSConfig {
       }
       .parSequence
 
+  lazy val notifierConfigs: IO[List[NotifierConfig]] =
+    vinyldnsConfig
+      .getConfigList("notifiers")
+      .asScala
+      .toList
+      .map(loadConfigF[IO, NotifierConfig](_))
```
rebstar6: test this config setup; i think you will need to do `.map(loadConfigF[IO, NotifierConfig](vinyldnsConfig, _))` given that the notifier configs will likely be nested in vinyldns

modules/api/src/main/scala/vinyldns/api/engine/RecordSetChangeHandler.scala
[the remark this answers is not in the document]
rebstar6: ooh kinda not actually...PartialFailure *may* be complete, it may also still have some pending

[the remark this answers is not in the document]
rebstar6: oh wait scratch that, MLY is right, our match is a little confusing

modules/api/src/main/scala/vinyldns/api/engine/RecordSetChangeHandler.scala
```diff
@@ -64,7 +70,16 @@ object RecordSetChangeHandler {
...
+            notifiers.notify(Notification(bc))
+          case BatchChangeStatus.Failed =>
```
rebstar6: match on Pending to IO.unit; everything else goes to the notifier

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlBatchChangeRepository.scala
rebstar6: BatchChangeSummary is probably the item you want

rebstar6: something like 

```
  private final val GET_SINGLE_BATCH_CHANGE_SUMMARY =
    sql"""
         |       SELECT bc.id, bc.user_id, bc.user_name, bc.created_time, bc.comments, bc.owner_group_id,
         |              SUM( case sc.status when 'Failed' then 1 else 0 end ) AS fail_count,
         |              SUM( case sc.status when 'Pending' then 1 else 0 end ) AS pending_count,
         |              SUM( case sc.status when 'Complete' then 1 else 0 end )  AS complete_count
         |         FROM single_change sc
         |         JOIN batch_change bc
         |           ON sc.batch_change_id = bc.id
         |        WHERE sc.id = ?
         |        GROUP BY bc.id
        """.stripMargin
```
(hacked from GET_BATCH_CHANGE_SUMMARY)

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlBatchChangeRepository.scala
```diff
@@ -118,34 +127,59 @@ class MySqlBatchChangeRepository
...
+              }
             }
+            UPDATE_SINGLE_CHANGE.batchByName(batchParams: _*).apply()
+            GET_BATCH_CHANGE_METADATA_FROM_SINGLE_CHANGE
+              .bind(singleChanges.head.id)
```
rebstar6: calling head on the list will yell at you...may be why travis is failing.

can do 
```
    NonEmptyList.fromList(singleChanges) match {
      case None => IO.pure(None)
      case Some(changes) => <all the stuff>
    }
```
since you can `.head` on a NEL

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlBatchChangeRepository.scala
dscleaver: We just discussed this offline, the notifier wants all of the BatchChange detail to craft emails etc.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlBatchChangeRepository.scala
dscleaver: ah right... languages that can't figure out that I've already checked the list to see if it was empty.
I think Travis is yelling at me for a nondeterministic test, but I'm not sure

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlBatchChangeRepository.scala
```diff
@@ -62,6 +62,15 @@ class MySqlBatchChangeRepository
          | WHERE bc.id = ?
         """.stripMargin
 
+  private final val GET_BATCH_CHANGE_METADATA_FROM_SINGLE_CHANGE =
+    sql"""
+         |SELECT bc.id, bc.user_id, bc.user_name, bc.created_time, bc.comments, bc.owner_group_id
```
rebstar6: i havent tested this, but i wonder performance-wise if there is a difference between what you have here and something like

```
SELECT bc.id, bc.user_id, bc.user_name, bc.created_time, bc.comments, bc.owner_group_id
  FROM batch_change bc
  JOIN (SELECT * from single_change where id = ?) sc
    ON sc.batch_change_id = bc.id
```

no idea how smart mysql is, would have to query plan the two, it seems joining on the already reduced set would be the way to go though

dscleaver: I like that better actually.

modules/api/src/test/resources/application.conf
```diff
@@ -88,6 +88,15 @@ vinyldns {
     }
   }
 
+  notifiers = [
+    {
```
rebstar6: not sure i love this nesting. part of the reason we set the enabled datastores as keys to later definitions is that it's hard to override parts of the list. As setup, i think everyone would have to override `notifiers=[]` as a whole. The alternative is doing:

```
notifiers = ["email-notifier"]

email-notifier {
   class-name = "..."
   settings = {...}
}
```
in which case you can just override `email-notifier.settings.xyz`

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlBatchChangeRepository.scala
```diff
@@ -118,34 +126,49 @@ class MySqlBatchChangeRepository
...
-                  'zoneId -> singleChange.zoneId,
-                  'id -> singleChange.id
-                )
-              case Left(e) => throw e
+  def updateSingleChanges(singleChanges: List[SingleChange]): IO[Option[BatchChange]] =
+    if (singleChanges.isEmpty)
+      IO.pure(None)
+    else
+      monitor("repo.BatchChangeJDBC.updateSingleChanges") {
+        logger.info(
```
rebstar6: i think technically our log lines should be in the IO's

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlBatchChangeRepository.scala
```diff
@@ -118,34 +126,49 @@ class MySqlBatchChangeRepository
...
+                  )
+                case Left(e) => throw e
+              }
             }
+            UPDATE_SINGLE_CHANGE.batchByName(batchParams: _*).apply()
+            GET_BATCH_CHANGE_METADATA_FROM_SINGLE_CHANGE
+              .bind(singleChanges.head.id)
+              .map(extractBatchChange(None))
+              .first
+              .apply()
```
rebstar6: pulled to helpers to make this more readable. functionally, this should change nothing:

```
  def updateSingleChanges(singleChanges: List[SingleChange]): IO[Option[BatchChange]] = {

    def convertSingleChangeToParams(singleChange: SingleChange): Seq[(Symbol, AnyRef)] =
      toPB(singleChange) match {
        case Right(data) =>
          val changeType = SingleChangeType.from(singleChange)
          Seq(
            'inputName -> singleChange.inputName,
            'changeType -> changeType.toString,
            'data -> data.toByteArray,
            'status -> singleChange.status.toString,
            'recordSetChangeId -> singleChange.recordChangeId,
            'recordSetId -> singleChange.recordSetId,
            'zoneId -> singleChange.zoneId,
            'id -> singleChange.id
          )
        case Left(e) => throw e
      }

    def getBatchFromSingleChangeId(singleChangeId: String): Option[BatchChange] = {
      GET_BATCH_CHANGE_METADATA_FROM_SINGLE_CHANGE
        .bind(singleChanges.head.id)
        .map(extractBatchChange(None))
        .first
        .apply()
        .map { batchMeta =>
          val changes = GET_SINGLE_CHANGES_BY_BCID
            .bind(batchMeta.id)
            .map(extractSingleChange(1))
            .list()
            .apply()
          batchMeta.copy(changes = changes)
        }
    }
    
    if (singleChanges.isEmpty)
      IO.pure(None)
    else
      monitor("repo.BatchChangeJDBC.updateSingleChanges") {
        logger.info(
          s"Updating single change statuses: ${singleChanges.map(ch => (ch.id, ch.status))}")
        IO {
          DB.localTx { implicit s =>
            for {
              headChange <- singleChanges.headOption
              batchParams = singleChanges.map(convertSingleChangeToParams)
              _ = UPDATE_SINGLE_CHANGE.batchByName(batchParams: _*).apply()
              updatedBatch <- getBatchFromSingleChangeId(headChange.id)
            } yield updatedBatch
          }
        }
      }
  }
```

thoughts?

rebstar6: or for the last (and now im yak shaving):

```
    singleChanges match {
      case Nil => IO.pure(None)
      case headChange :: tl =>
        monitor("repo.BatchChangeJDBC.updateSingleChanges") {
          IO {
            logger.info(
              s"Updating single change statuses: ${singleChanges.map(ch => (ch.id, ch.status))}")
            DB.localTx { implicit s =>
              val batchParams = singleChanges.map(convertSingleChangeToParams)
              UPDATE_SINGLE_CHANGE.batchByName(batchParams: _*).apply()
              getBatchFromSingleChangeId(headChange.id)
            }
          }
        }
    }
```

dscleaver: I went with the first one, but eliminated the superfluous if

PR #640, opened 2019-05-21
mysql recordset fix paging for duplicate names

### LINKED ISSUE #565: mysql recordset paging does not work for records with duplicate names
**Describe the bug**
Does not appear to work when records have same name. The dummy. zone has three records named dummy and paging only returns one of them

**VinylDNS Version**
0.8.1-SNAPSHOT

**To Reproduce**

1. start up api
1. connect to `dummy.` zone
1. `GET zones/id/recordsets?maxItems=1`
    `returns the A record with name and nextId already-exists`
1. `GET zones/id/recordsets?maxItems=1&startFrom=already-exists`
    `return SOA record with name and nextId dummy.`
1. `GET zones/id/recordsets?maxItems=1&startFrom=dummy`
    `return A record with name and nextId foo`
1. keep querying, you will only get 6 out of the 8 records

**Screenshots**

**Additional context**
This is probably because we return the name as the nextId. May want to consider the id or hash of some sort

nimaeskandary:
Fixes #565 .

* recordset paging is done by name and type
* paging will either match on `(name >= prev.name && type > prev.type) || (name > prev.name)`
* this is to address cases in which a page boundary record shares its name with other records that must still be paged
* in the previous logic, `name > prev.name`, duplicate records that were still not paged would be lost 
* in the new logic, a page will match for records with the previous last items name as long as the record does not have the same type

### DIFF
--- a/modules/api/functional_test/live_tests/recordsets/list_recordsets_test.py
+++ b/modules/api/functional_test/live_tests/recordsets/list_recordsets_test.py
@@ -209,36 +209,36 @@ def test_list_recordsets_default_size_is_100(rs_fixture):
 
 def test_list_recordsets_duplicate_names(rs_fixture):
     """
-    Test that paging keys work for records with duplicate names, in this SOA and NS
+    Test that paging keys work for records with duplicate names
     """
     client = rs_fixture.client
     ok_zone = rs_fixture.test_context
 
-    apex_ns = False
-    apex_soa = False
+    created = []
 
-    start_key = None
-    offset = 0
-    while offset < 17:
-        list_results = client.list_recordsets(ok_zone['id'], status=200, start_from=start_key, max_items=1)
-        rs_fixture.check_recordsets_page_accuracy(list_results, size=1, offset=offset, nextId=True,
-                                                  startFrom=start_key, maxItems=1)
-        start_key = list_results['nextId']
-        offset = offset + 1
+    try:
+        record_data_a = [{'address': '1.1.1.1'}]
+        record_data_txt = [{'text': 'some=value'}]
 
-        record = list_results['recordSets'][0]
-        if record['name'] == "ok.":
-            if record['type'] == "SOA":
-                apex_soa = True
-            if record['type'] == "NS":
-                apex_ns = True
+        record_json_a = get_recordset_json(ok_zone, '0', 'A', record_data_a, ttl=100)
+        record_json_txt = get_recordset_json(ok_zone, '0', 'TXT', record_data_txt, ttl=100)
 
-    list_results = client.list_recordsets(ok_zone['id'], status=200, max_items=1, start_from=start_key)
-    rs_fixture.check_recordsets_page_accuracy(list_results, size=0, offset=offset, nextId=False,
-                                              maxItems=1, startFrom=start_key)
+        create_response = client.create_recordset(record_json_a, status=202)
+        created.append(client.wait_until_recordset_change_status(create_response, 'Complete')['recordSet']['id'])
 
-    assert_that(apex_ns, is_(True))
-    assert_that(apex_soa, is_(True))
+        create_response = client.create_recordset(record_json_txt, status=202)
+        created.append(client.wait_until_recordset_change_status(create_response, 'Complete')['recordSet']['id'])
+
+        list_results = client.list_recordsets(ok_zone['id'], status=200, start_from=None, max_items=1)
+        assert_that(list_results['recordSets'][0]['id'], is_(created[0]))
+
+        list_results = client.list_recordsets(ok_zone['id'], status=200, start_from=list_results['nextId'], max_items=1)
+        assert_that(list_results['recordSets'][0]['id'], is_(created[1]))
+
+    finally:
+        for id in created:
+            client.delete_recordset(ok_zone['id'], id, status=202)
+            client.wait_until_recordset_deleted(ok_zone['id'], id)
 
 
 def test_list_recordsets_with_record_name_filter_all(rs_fixture):
--- a/modules/api/functional_test/live_tests/recordsets/list_recordsets_test.py
+++ b/modules/api/functional_test/live_tests/recordsets/list_recordsets_test.py
@@ -207,6 +207,40 @@ def test_list_recordsets_default_size_is_100(rs_fixture):
     rs_fixture.check_recordsets_page_accuracy(list_results, size=17, offset=0, maxItems=100)
 
 
+def test_list_recordsets_duplicate_names(rs_fixture):
+    """
+    Test that paging keys work for records with duplicate names, in this SOA and NS
+    """
+    client = rs_fixture.client
+    ok_zone = rs_fixture.test_context
+
+    apex_ns = False
+    apex_soa = False
+
+    start_key = None
+    offset = 0
+    while offset < 17:
+        list_results = client.list_recordsets(ok_zone['id'], status=200, start_from=start_key, max_items=1)
+        rs_fixture.check_recordsets_page_accuracy(list_results, size=1, offset=offset, nextId=True,
+                                                  startFrom=start_key, maxItems=1)
+        start_key = list_results['nextId']
+        offset = offset + 1
+
+        record = list_results['recordSets'][0]
+        if record['name'] == "ok.":
+            if record['type'] == "SOA":
+                apex_soa = True
+            if record['type'] == "NS":
+                apex_ns = True
+
+    list_results = client.list_recordsets(ok_zone['id'], status=200, max_items=1, start_from=start_key)
+    rs_fixture.check_recordsets_page_accuracy(list_results, size=0, offset=offset, nextId=False,
+                                              maxItems=1, startFrom=start_key)
+
+    assert_that(apex_ns, is_(True))
+    assert_that(apex_soa, is_(True))
+
+
 def test_list_recordsets_with_record_name_filter_all(rs_fixture):
     """
     Test listing all recordsets whose name contains a substring, all recordsets have substring 'list' in name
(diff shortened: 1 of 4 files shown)


modules/api/functional_test/live_tests/recordsets/list_recordsets_test.py
```diff
@@ -207,6 +207,40 @@ def test_list_recordsets_default_size_is_100(rs_fixture):
     rs_fixture.check_recordsets_page_accuracy(list_results, size=17, offset=0, maxItems=100)
 
 
+def test_list_recordsets_duplicate_names(rs_fixture):
```
britneywright: why not create two records an, A and a TXT, for example with the name "a" then you don't have to loop through all of the zone records, you could just check against the first two. and the assertions would be that the A record is the first record from the first list_recordsets call and the TXT record is first record from the second list_recordsets call.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
```diff
@@ -357,4 +366,31 @@ object MySqlRecordSetRepository extends ProtobufConversions {
...
     else absoluteRecordSetName + absoluteZoneName
   }
+
+  case class PagingKey(recordName: String, recordType: Int)
+
+  object PagingKey {
+    val delimiterRegex = "<recordName\\.\\.\\.\\.recordType>"
+    val delimiter = "<recordName....recordType>"
+
+    def fromStartFrom(startFrom: Option[String]): Option[PagingKey] =
```
mitruly: Since this is a companion object with only one constructor for `PagingKey`, we could just name this `apply`:
```
    def apply(startFrom: Option[String]): Option[PagingKey] =
      startFrom.flatMap { sf =>
        val tokens = sf.split(delimiterRegex)
        if (tokens.length != 2) None
        else {
          val recordName = tokens.head
          Try(tokens(1).toInt) match {
            case Success(recordType) => Some(PagingKey(recordName, recordType))
            case Failure(_) => None
          }
        }
      }
```

The reference to `val pagingKey = PagingKey.fromStartFrom(startFrom)` can then become `val pagingKey = PagingKey(startFrom)`.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
```diff
@@ -357,4 +366,31 @@ object MySqlRecordSetRepository extends ProtobufConversions {
     if (absoluteRecordSetName.equals(absoluteZoneName)) absoluteZoneName
     else absoluteRecordSetName + absoluteZoneName
   }
+
+  case class PagingKey(recordName: String, recordType: Int)
+
+  object PagingKey {
+    val delimiterRegex = "<recordName\\.\\.\\.\\.recordType>"
```
mitruly: The delimiter may be a bit overzealous; could just use a character that would never be in a record name or record type (integer) (eg. `|`).

modules/mysql/src/test/scala/vinyldns/mysql/repository/MySqlRecordSetRepositorySpec.scala
```diff
@@ -70,4 +71,43 @@ class MySqlRecordSetRepositorySpec extends WordSpec with Matchers {
...
+  "PagingKey.fromStartFrom" should {
+    "return None if startFrom is None" in {
+      PagingKey.fromStartFrom(None) shouldBe None
+    }
+
+    "return None if startFrom is malformed" in {
+      val empty = ""
+      val noDelimiter = "nodelim"
+      val justDelimiter = s"${PagingKey.delimiter}"
+      val noType = s"name${PagingKey.delimiter}"
```
mitruly: We could add a no record name case here.

modules/api/functional_test/live_tests/recordsets/list_recordsets_test.py
```diff
@@ -207,6 +207,40 @@ def test_list_recordsets_default_size_is_100(rs_fixture):
...
+    """
+    Test that paging keys work for records with duplicate names, in this SOA and NS
+    """
+    client = rs_fixture.client
+    ok_zone = rs_fixture.test_context
+
+    apex_ns = False
```
mitruly: We should rename these booleans for clarity (eg. `apex_ns` → `apex_ns_record_found`, etc.).

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
```diff
@@ -205,7 +212,9 @@ class MySqlRecordSetRepository extends RecordSetRepository with Monitored {
           // if size of results is less than the number returned, we don't have a next id
           // if maxItems is None, we don't have a next id
           val nextId =
-            maxItems.filter(_ == results.size).flatMap(_ => results.lastOption.map(_.name))
+            maxItems
```
mitruly: Wondering if we should be trying to grab `maxItems + 1` items and returning the last item if `results.size > maxItems`; with the current setup, there is an edge case where a `nextId` may be returned despite there not really being a next page. We could then actually use `nextId` as a real `startFrom`.

mitruly: Wondering if we should be trying to grab `maxItems + 1` items and returning the last item if `results.size > maxItems`; with the current setup, there is an edge case where a `nextId` may be returned despite there not really being a next page. We could then actually use `nextId` as a real `startFrom`.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
nimaeskandary: yeah was iffy on it too. Would prefer it is something that does not have to be url encoded. Figured empty labels are not legal record names so lots of dots were good, can take out the extra stuff on the left or right

modules/api/functional_test/live_tests/recordsets/list_recordsets_test.py
nimaeskandary: true, may change

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
nimaeskandary: I played with this and it actually introduces a bit of logic around doing the query with maxItems plus one for the look ahead and doing some annoying stuff for the take and nextId check. I would rather that is a different pr that also changes maxItems to Int (which I think it should be) in this function def as opposed to Option[Int] because that would make the logic cleaner

mitruly: Think I'm fine with this, since we are aware of this @nima

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
```diff
@@ -357,4 +366,33 @@ object MySqlRecordSetRepository extends ProtobufConversions {
...
+
+  case class PagingKey(recordName: String, recordType: Int)
+
+  object PagingKey {
+    val delimiterRegex = "<\\.\\.\\.\\.>"
+    val delimiter = "<....>"
+
+    def apply(startFrom: Option[String]): Option[PagingKey] =
+      startFrom.flatMap { sf =>
+        val tokens = sf.split(delimiterRegex)
```
rebstar6: the flatmap/if/match/etc here is a little wonky

```
      for {
        sf <- startFrom
        tokens = sf.split(delimiterRegex)
        _ <- if (tokens.length != 2) None else Some(())
        recordName <- tokens.headOption
        recordType <- Try(tokens(1).toInt).toOption
      } yield Some(PagingKey(recordName, recordType))
```

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
```diff
@@ -183,14 +185,19 @@ class MySqlRecordSetRepository extends RecordSetRepository with Monitored {
...
+          val opts =
+            (pagingKey.as(
+              "AND ((name >= {startFromName} AND type > {startFromType}) OR name > {startFromName})") ++
+              recordNameFilter.as("AND name LIKE {nameFilter}") ++
+              Some("ORDER BY name ASC, type ASC") ++
+              maxItems.as("LIMIT {maxItems}")).toList.mkString(" ")
 
           val params = (Some('zoneId -> zoneId) ++
-            startFrom.map(n => 'startFrom -> n) ++
+            pagingKey.map(pk => 'startFromName -> pk.recordName) ++
```
rebstar6: please check with @mdb to make sure that this wont have any TF impacts.

It should not, we used to take a string, and are now still taking a string (just a different string) but still, id like to be certain

modules/mysql/src/test/scala/vinyldns/mysql/repository/MySqlRecordSetRepositorySpec.scala
```diff
@@ -81,13 +81,11 @@ class MySqlRecordSetRepositorySpec extends WordSpec with Matchers {
       val empty = ""
       val noDelimiter = "nodelim"
       val justDelimiter = s"${PagingKey.delimiter}"
-      val noName = s"${PagingKey.delimiter}1"
```
nimaeskandary: An empty string is still a valid string and will be returned as an element in a split

mdb:
Am I correct in perceiving this yields no change in client behavior? I don't _believe_ this will impact Terraform, though I'm a bit foggy on what's going on here, admittedly.

nimaeskandary:
@mdb correct, clients treat paging exactly how they usually do (funnel from nextId to startFrom), but paging key is not just the record name anymore and has some metadata on type

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
britneywright: did you make the new issue for this?

nimaeskandary: will do

PR #647, opened 2019-05-22
change Zone shared attribute label to Access

britneywright:
We said if we ever came up with a better name for the Zone shared attribute in the portal we would change it from Type. Today is that day! Going with "Access". The possible values remain the same as "Shared" or "Private".


britneywright:
no doc updates needed for this one.

PR #648, opened 2019-05-22
Change homepage from Zones to New Batch Change

britneywright:
with the addition of shared zones the expectation is majority of portal users will use the new batch change form rather than managing records in individual zones. This PR changes the default home page in the portal to the New Batch Change screen. Also changes the order of the navigation.

VinylDNS home page screenshot


pauljamescleary:
I think the batch change summaries would be the default landing page

britneywright:
> I think the batch change summaries would be the default landing page

misunderstanding, updated now.

PR #650, opened 2019-05-23
add attributes for batch change review process

britneywright:
First PR for batch change review feature: https://github.com/vinyldns/rfcs/pull/9

Changes in this pull request:
- add batch change attributes to model and database:
  - approvalStatus: optional field with four possible types: AutoApproved, PendingApproval, ManuallyApproved, ManuallyRejected. Made an index since we'll query on this field.
  - reviewerId: optional field, ID of the user reviewing the batch change.
  - reviewComment: optional field, string for a comment by the reviewer
  - reviewTimeStamp: optional field, dateTime of the review

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
@@ -303,7 +303,8 @@ class BatchChangeService(
         batchChangeInput.comments,
         DateTime.now,
         changes,
-        batchChangeInput.ownerGroupId
+        batchChangeInput.ownerGroupId,
+        BatchChangeApprovalStatus.AutoApproved
       ).asRight
     } else {
       InvalidBatchChangeResponses(batchChangeInput.changes, transformed).asLeft
--- a/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
@@ -303,11 +303,7 @@ class BatchChangeService(
         batchChangeInput.comments,
         DateTime.now,
         changes,
-        batchChangeInput.ownerGroupId,
-        batchChangeInput.approvalStatus,
-        batchChangeInput.reviewerId,
-        batchChangeInput.reviewComment,
-        batchChangeInput.reviewTimestamp
+        batchChangeInput.ownerGroupId
       ).asRight
     } else {
       InvalidBatchChangeResponses(batchChangeInput.changes, transformed).asLeft
--- a/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
@@ -20,6 +20,7 @@ import cats.data._
 import cats.effect._
 import cats.implicits._
 import org.joda.time.DateTime
+import org.slf4j.{Logger, LoggerFactory}
 import vinyldns.api.domain.DomainValidations._
 import vinyldns.api.domain.ReverseZoneHelpers.ptrIsInZone
 import vinyldns.api.domain.batch.BatchChangeInterfaces._
@@ -60,6 +61,7 @@ class BatchChangeService(
 
   import batchChangeValidations._
 
+  val logger: Logger = LoggerFactory.getLogger(classOf[BatchChangeService])
   def applyBatchChange(
       batchChangeInput: BatchChangeInput,
       auth: AuthPrincipal): BatchResult[BatchChange] =
@@ -70,8 +72,9 @@ class BatchChangeService(
       zoneMap <- getZonesForRequest(inputValidatedSingleChanges).toBatchResult
       changesWithZones = zoneDiscovery(inputValidatedSingleChanges, zoneMap)
       recordSets <- getExistingRecordSets(changesWithZones, zoneMap).toBatchResult
+      withTtl = doTtlMapping(changesWithZones, recordSets)
       validatedSingleChanges = validateChangesWithContext(
-        changesWithZones,
+        withTtl,
         recordSets,
         auth,
         batchChangeInput.ownerGroupId)
@@ -177,6 +180,21 @@ class BatchChangeService(
       recordSetsWithExistingZone = allRecordSets.filter(rs => zoneMap.getById(rs.zoneId))
     } yield ExistingRecordSets(recordSetsWithExistingZone)
   }
+
+  def doTtlMapping(
+      changes: ValidatedBatch[ChangeForValidation],
+      existingRecordSets: ExistingRecordSets): ValidatedBatch[ChangeForValidation] =
+    changes.mapValid {
+      case add: AddChangeForValidation =>
+        existingRecordSets
+          .get(add.recordKey)
+          .map { rs =>
+            add.copy(existingRecordTtl = Some(rs.ttl))
+          }
+          .getOrElse(add)
+          .validNel
+      case del: DeleteChangeForValidation => del.validNel
+    }
 
   def getOwnerGroup(ownerGroupId: Option[String]): BatchResult[Option[Group]] = {
     val ownerGroup = for {
--- a/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
@@ -303,7 +303,12 @@ class BatchChangeService(
         batchChangeInput.comments,
         DateTime.now,
         changes,
-        batchChangeInput.ownerGroupId).asRight
+        batchChangeInput.ownerGroupId,
+        batchChangeInput.approvalStatus,
+        batchChangeInput.reviewerId,
+        batchChangeInput.reviewComment,
+        batchChangeInput.reviewTimestamp
+      ).asRight
     } else {
       InvalidBatchChangeResponses(batchChangeInput.changes, transformed).asLeft
     }
--- a/modules/api/src/test/scala/vinyldns/api/repository/InMemoryBatchChangeRepository.scala
+++ b/modules/api/src/test/scala/vinyldns/api/repository/InMemoryBatchChangeRepository.scala
@@ -77,6 +77,10 @@ class InMemoryBatchChangeRepository extends BatchChangeRepository {
           sc.createdTimestamp,
           singleChangesFromRepo,
           sc.ownerGroupId,
+          None,
+          None,
+          None,
+          None,
           sc.id)
       }
     }
@@ -121,6 +125,10 @@ class InMemoryBatchChangeRepository extends BatchChangeRepository {
         sc.createdTimestamp,
         changes,
         sc.ownerGroupId,
+        None,
+        None,
+        None,
+        None,
         sc.id)
     } yield BatchChangeSummary(batchChange)
     val sorted = batchChangeSummaries.sortBy(_.createdTimestamp)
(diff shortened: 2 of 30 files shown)


modules/mysql/src/main/resources/db/migration/V3.14__BatchChangePendingAttributes.sql
```diff
@@ -0,0 +1,9 @@
+CREATE SCHEMA IF NOT EXISTS ${dbName};
+
+USE ${dbName};
+
+ALTER TABLE batch_change ADD approval_status CHAR(20) NULL;
```
nimaeskandary: The approval status is not always going to be 20 characters, we should make this varchar no?

modules/mysql/src/main/resources/db/migration/V3.14__BatchChangePendingAttributes.sql
```diff
@@ -0,0 +1,9 @@
...
+ALTER TABLE batch_change ADD reviewer_id CHAR(36) NULL;
+ALTER TABLE batch_change ADD review_comment VARCHAR(1024) NULL;
+ALTER TABLE batch_change ADD review_timestamp DATETIME NULL;
+CREATE INDEX approval_status_index ON batch_change (approval_status);
```
nimaeskandary: curious if we have a use case scoped out for this index

modules/api/src/test/scala/vinyldns/api/repository/InMemoryBatchChangeRepository.scala
```diff
@@ -76,6 +76,10 @@ class InMemoryBatchChangeRepository extends BatchChangeRepository {
           sc.createdTimestamp,
           singleChangesFromRepo,
           sc.ownerGroupId,
```
rebstar6: im increasingly disliking this in-mem repo as it's getting more and more only semi-correct.

i wonder what the effort level would be to switch to EmptyBatchChange repo like we do with all the others and stop trying to update this.

im gonna look into it

modules/mysql/src/main/resources/db/migration/V3.14__BatchChangePendingAttributes.sql
britneywright: We'll query for all batch changes that are PendingApproval

mitruly: Not sure if the `single_change` table is going to be handled separately in a different PR, but I can imagine down the road:
- Adding a similar status to `approval_status` to `single_change`
- Adding an index to `single_change` to the effect of `(batch_change_id, approval_status, zone_id, input_name)` and other necessary indices

modules/mysql/src/main/resources/db/migration/V3.14__BatchChangePendingAttributes.sql
```diff
@@ -0,0 +1,9 @@
+CREATE SCHEMA IF NOT EXISTS ${dbName};
+
+USE ${dbName};
+
+ALTER TABLE batch_change ADD approval_status VARCHAR(20) NULL;
```
mitruly: I think having a `type TINYINT NULL` similar to how we construct our `recordset` table is better since any naming changes/casing/etc would be handled in the API when we deserialize from its table value (number).

britneywright: I'm open to either. `VARCHAR` is consistent with the `status` and `change_type` fields in the `single_change` table.

modules/mysql/src/main/resources/db/migration/V3.14__BatchChangePendingAttributes.sql
britneywright: i'm not aware of any plans to modify the SingleChange model

modules/mysql/src/main/resources/db/migration/V3.14__BatchChangePendingAttributes.sql
pauljamescleary: I know we haven't been super disciplined on this, but now that we are moving entirely to SQL, memory is at a premium when it comes to indexes.

TINYINT would be much better than VARCHAR as it would take up a lot less memory in the index.  We should enact a policy moving forward that all statuses and enums are implemented using something like TINYINT.

Note: I know that MySQL has enums natively, but I am unclear on the pros/cons of those when it comes to space and such.

britneywright:
I still need to handle the serialization in this PR, more commits coming

modules/mysql/src/main/resources/db/migration/V3.14__BatchChangePendingAttributes.sql
```diff
@@ -0,0 +1,9 @@
+CREATE SCHEMA IF NOT EXISTS ${dbName};
+
+USE ${dbName};
+
+ALTER TABLE batch_change ADD approval_status TINYINT NULL;
+ALTER TABLE batch_change ADD reviewer_id CHAR(36) NULL;
+ALTER TABLE batch_change ADD review_comment VARCHAR(1024) NULL;
```
nimaeskandary: Good with this pr, putting a reminder that we have a validation at the front gate to make sure comments cannot be over 1024 characters

modules/core/src/main/scala/vinyldns/core/domain/batch/BatchChange.scala
```diff
@@ -28,6 +29,10 @@ case class BatchChange(
     createdTimestamp: DateTime,
     changes: List[SingleChange],
     ownerGroupId: Option[String] = None,
+    approvalStatus: Option[BatchChangeApprovalStatus] = None,
```
rebstar6: approvalStatus shouldnt be an option

rebstar6: serialize this out to `autoApproved` if not defined in the DB for the existing stuff. probably no default on the type itself though

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -303,7 +303,9 @@ class BatchChangeService(
         batchChangeInput.comments,
         DateTime.now,
         changes,
-        batchChangeInput.ownerGroupId).asRight
+        batchChangeInput.ownerGroupId,
+        BatchChangeApprovalStatus.AutoApproved
```
britneywright: this may change when the validations are in place down the road to determine what is autoapproved and what isn't.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlBatchChangeRepository.scala
```diff
@@ -330,4 +350,26 @@ class MySqlBatchChangeRepository
...
+
+  def toApprovalStatus(key: Int): BatchChangeApprovalStatus =
+    key match {
+      case 2 => BatchChangeApprovalStatus.PendingApproval
+      case 3 => BatchChangeApprovalStatus.ManuallyApproved
+      case 4 => BatchChangeApprovalStatus.ManuallyRejected
+      case _ => BatchChangeApprovalStatus.AutoApproved
+    }
+
+  def toReviewTimestamp(result: Option[Timestamp]): Option[DateTime] =
```
britneywright: joda DateTime will ignore null values and respond with the current time, which we don't want if there isn't a time in the database for reviewTimestamp.

modules/core/src/main/scala/vinyldns/core/domain/batch/BatchChange.scala
```diff
@@ -70,7 +80,11 @@ case class BatchChangeInfo(
     ownerGroupId: Option[String],
     id: String,
     status: BatchChangeStatus,
-    ownerGroupName: Option[String]
+    ownerGroupName: Option[String],
```
rebstar6: do we want these in BatchChangeInfo for now?

im thinking no, leave that to endpoint cards (which will flag off for now anyway)

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlBatchChangeRepository.scala
rebstar6: this is `result.map(new org.joda.time.DateTime(_))`

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlBatchChangeRepository.scala
```diff
@@ -330,4 +350,26 @@ class MySqlBatchChangeRepository
...
+      case BatchChangeApprovalStatus.ManuallyApproved => 3
+      case BatchChangeApprovalStatus.ManuallyRejected => 4
+    }
+
+  def toApprovalStatus(key: Int): BatchChangeApprovalStatus =
+    key match {
+      case 2 => BatchChangeApprovalStatus.PendingApproval
+      case 3 => BatchChangeApprovalStatus.ManuallyApproved
+      case 4 => BatchChangeApprovalStatus.ManuallyRejected
+      case _ => BatchChangeApprovalStatus.AutoApproved
```
rebstar6: define the 1/2/3/4 cases here, i dont like a default because it wont warn in the future if we add another one and forget it

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlBatchChangeRepository.scala
```diff
@@ -230,7 +238,11 @@ class MySqlBatchChangeRepository
                 pending + failed + complete,
                 BatchChangeStatus.fromSingleStatuses(pending > 0, failed > 0, complete > 0),
                 Option(res.string("owner_group_id")),
-                res.string("id")
+                res.string("id"),
+                toApprovalStatus(res.int("approval_status")),
```
rebstar6: i asusme this `res.int` needs to be wrapped in an option:

Option(res.string("approval_status")).map(toApprovalStatus).getOrElse(AutoApproved)

modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlBatchChangeRepositoryIntegrationSpec.scala
```diff
@@ -189,8 +201,9 @@ class MySqlBatchChangeRepositoryIntegrationSpec
       areSame(f.unsafeRunSync(), Some(bcARecords))
     }
 
-    "save/get a batch change with empty comments, ownerGroup" in {
-      val testBatch = randomBatchChange.copy(comments = None, ownerGroupId = None)
+    "save/get a batch change with empty comments, ownerGroup, reviewerId, reviewComment, reviewTimeStamp" in {
```
rebstar6: are we testing anywhere that we save/get a change with non-empty ^^?

modules/core/src/main/scala/vinyldns/core/domain/batch/BatchChangeSummary.scala
```diff
@@ -30,6 +31,10 @@ case class BatchChangeSummary(
     status: BatchChangeStatus,
     ownerGroupId: Option[String],
     id: String = UUID.randomUUID().toString,
+    approvalStatus: BatchChangeApprovalStatus = BatchChangeApprovalStatus.AutoApproved,
```
rebstar6: same here, batch summary is used in the responses back to the user. keep it unchanged (just drop the extra info) for now until we want that to be displayed to people

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
rebstar6: autoapproved here is good for now 👍

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlBatchChangeRepository.scala
mitruly: Agree with ^^; the following works: `res.intOpt("approval_status").map(toApprovalStatus).getOrElse(BatchChangeApprovalStatus.AutoApproved)`

In a similar vein, can conveniently use scalikejdbc's `.stringOpt` instead of manually wrapping in an option (eg. `result.stringOpt("reviewer_id"),`).

PR #653, opened 2019-05-24
ensure trailing dot in list zones search

### LINKED ISSUE #652: Zone search needs to add a trailing dot
**Describe the bug**
Searching for the zone "foo.com" brings up nothing you need to search "foo.com.

**VinylDNS Version**
0.9.2

**Additional context**
Pay special attention to the wildcard being added. If they add a wildcard to the end of the query, then dont add

britneywright:
Fixes #652.

Changes in this pull request:
- use the `ensureTrailingDot` DomainHelper method to add a trailing dot to the passed in name_filter if it doesn't have one
- all zones end in a trailing dot so this will have no impact on whether `*` are included for wildcard search or not.

### DIFF
--- a/modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlZoneRepositoryIntegrationSpec.scala
+++ b/modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlZoneRepositoryIntegrationSpec.scala
@@ -393,7 +393,7 @@ class MySqlZoneRepositoryIntegrationSpec
     "apply the zone filter as a normal user" in {
 
       val testZones = Seq(
-        testZone("system-tes.t", adminGroupId = "foo"),
+        testZone("system-test.", adminGroupId = "foo"),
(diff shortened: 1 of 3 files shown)


modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlZoneRepositoryIntegrationSpec.scala
```diff
@@ -393,9 +393,9 @@ class MySqlZoneRepositoryIntegrationSpec
     "apply the zone filter as a normal user" in {
 
       val testZones = Seq(
-        testZone("system-test", adminGroupId = "foo"),
-        testZone("system-temp", adminGroupId = "foo"),
-        testZone("system-nomatch", adminGroupId = "bar")
+        testZone("system-tes.t", adminGroupId = "foo"),
```
mitruly: Think the dot should be at the end here.

PR #656, opened 2019-05-28
Fail if zone id is not equal on recordset update

### LINKED ISSUE #353: Remove zoneID from getRecordSet
We require the user to enter the recordSetId and zoneId on getRecordSet, but we dont actually need or use the zoneId to get it from the database. Once we have the recordSet, we can pull the zoneId from there for any access considerations.

We should remove the zoneId from this call.

For backwards compatibility, if someone includes a zoneId in the request, we should just drop it (do not error).

Docs should also be updated accordingly.

### LINKED ISSUE #638: Issue update of record set changing zone id fails
**Describe the bug**
When updating a record set, it is possible to change the `zoneId` on the record set when the `PUT` is issued.  This would effectively results in a "move" of the record set to a new zone.  However, this is bad as it breaks history (record set changes point to the old zone).

**VinylDNS Version**
0.9.0

**To Reproduce**
1. Create two zones
2. Create a record set in zone a
3. Using one of the clients, issue an update to the record set, changing the zone id from zone a to zone b
4. Record set update gets submitted, but ultimately fails trying to save to zone b

**Expected behavior**
We should fail quickly and _NOT_ allow this to happen.  Do a logic check that compares the `zoneId` of the new zone against the `zoneId` of the existing record set.  If they change in the course of an UPDATE record set, then fail.

nimaeskandary:
Fixes #638 

* If the recordSet zoneId and the zoneId in the route do not match return a 422

Can likely close #353

### DIFF
--- a/modules/api/functional_test/live_tests/recordsets/update_recordset_test.py
+++ b/modules/api/functional_test/live_tests/recordsets/update_recordset_test.py
@@ -2368,7 +2368,7 @@ def test_update_ds_bad_ttl(shared_zone_test_context):
             client.delete_recordset(result_rs['zoneId'], result_rs['id'], status=(202,404))
             client.wait_until_recordset_deleted(result_rs['zoneId'], result_rs['id'])
 
-def test_update_zone_id_fails(shared_zone_test_context):
+def test_update_fails_when_payload_and_route_zone_id_does_not_match(shared_zone_test_context):
     """
     Test that a 422 is returned if the zoneId in the body and route do not match
     """
@@ -2388,10 +2388,38 @@ def test_update_zone_id_fails(shared_zone_test_context):
         update['zoneId'] = shared_zone_test_context.dummy_zone['id']
 
         url = urljoin(client.index_url, u'/zones/{0}/recordsets/{1}'.format(zone[u'id'], update[u'id']))
-        response, data = client.make_request(url, u'PUT', client.headers, json.dumps(update), not_found_ok=True,
+        response, error = client.make_request(url, u'PUT', client.headers, json.dumps(update), not_found_ok=True,
                                              status=422)
 
-        assert_that(data, is_("Cannot update RecordSet's zoneId attribute"))
+        assert_that(error, is_("Cannot update RecordSet's zoneId attribute"))
+
+    finally:
+        if created:
+            delete_result = client.delete_recordset(zone['id'], created['id'], status=202)
+            client.wait_until_recordset_change_status(delete_result, 'Complete')
+
+
+def test_update_fails_when_payload_and_actual_zone_id_do_not_match(shared_zone_test_context):
+    """
+    Test that a 422 is returned if the zoneId in the body and the recordSets actual zoneId do not match
+    """
+
+    client = shared_zone_test_context.ok_vinyldns_client
+    zone = shared_zone_test_context.ok_zone
+
+    created = None
+
+    try:
+        record_json = get_recordset_json(zone, 'test_update_zone_id', 'A', [{'address': '1.1.1.1'}])
+        create_response = client.create_recordset(record_json, status=202)
+        created = client.wait_until_recordset_change_status(create_response, 'Complete')['recordSet']
+
+        update = created
+        update['zoneId'] = shared_zone_test_context.dummy_zone['id']
+
+        error = client.update_recordset(update, status=422)
+
+        assert_that(error, is_("Cannot update RecordSet's zoneId attribute"))
 
     finally:
         if created:
--- a/modules/api/functional_test/live_tests/recordsets/update_recordset_test.py
+++ b/modules/api/functional_test/live_tests/recordsets/update_recordset_test.py
@@ -7,6 +7,8 @@ from vinyldns_python import VinylDNSClient
 from test_data import TestData
 from vinyldns_context import VinylDNSTestContext
 import time
+import json
+from requests.compat import urljoin
 
 
 def test_update_a_with_same_name_as_cname(shared_zone_test_context):
@@ -2365,3 +2367,33 @@ def test_update_ds_bad_ttl(shared_zone_test_context):
         if result_rs:
             client.delete_recordset(result_rs['zoneId'], result_rs['id'], status=(202,404))
             client.wait_until_recordset_deleted(result_rs['zoneId'], result_rs['id'])
+
+def test_update_zoneId_fails(shared_zone_test_context):
+    """
+    Test that a 422 is returned if the zoneId in the body and route do not match
+    """
+
+    client = shared_zone_test_context.ok_vinyldns_client
+    zone = shared_zone_test_context.ok_zone
+
+    created = None
+
+    try:
+        record_json = get_recordset_json(zone, 'test_update_zone_id', 'A', [{'address': '1.1.1.1'}])
+        create_response = client.create_recordset(record_json, status=202)
+        created = client.wait_until_recordset_change_status(create_response, 'Complete')['recordSet']
+
+        update = created
+        update['ttl'] = update['ttl'] + 100
+        update['zoneId'] = shared_zone_test_context.dummy_zone['id']
+
+        url = urljoin(client.index_url, u'/zones/{0}/recordsets/{1}'.format(zone[u'id'], update[u'id']))
+        response, data = client.make_request(url, u'PUT', client.headers, json.dumps(update), not_found_ok=True,
+                                             status=422)
+
+        assert_that(data, is_("Cannot update RecordSet's zoneId attribute"))
+
+    finally:
+        if created:
+            delete_result = client.delete_recordset(zone['id'], created['id'], status=202)
+            client.wait_until_recordset_change_status(delete_result, 'Complete')
--- a/modules/api/functional_test/live_tests/recordsets/update_recordset_test.py
+++ b/modules/api/functional_test/live_tests/recordsets/update_recordset_test.py
@@ -2368,7 +2368,7 @@ def test_update_ds_bad_ttl(shared_zone_test_context):
             client.delete_recordset(result_rs['zoneId'], result_rs['id'], status=(202,404))
             client.wait_until_recordset_deleted(result_rs['zoneId'], result_rs['id'])
 
-def test_update_zoneId_fails(shared_zone_test_context):
+def test_update_zone_id_fails(shared_zone_test_context):
     """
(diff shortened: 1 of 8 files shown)


modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetService.scala
```diff
@@ -86,6 +86,7 @@ class RecordSetService(
       change <- RecordSetChangeGenerator.forUpdate(existing, recordSet, zone, Some(auth)).toResult
       // because changes happen to the RS in forUpdate itself, converting 1st and validating on that
       rsForValidations = change.recordSet
+      _ <- hasSameZoneId(existing.zoneId, rsForValidations.zoneId).toResult
```
rebstar6: to your comment, i would still make this change. in the route itself you can also do a 
```
            validate(
              zoneId == rs.zoneId,
              "bad bad!"
            )
```
before calling updateRecordSet and that should cover both cases

modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetValidations.scala
```diff
@@ -250,4 +250,12 @@ object RecordSetValidations {
         if (authPrincipal.isSuper || authPrincipal.isGroupMember(groupId)) ().asRight
         else InvalidRequest(s"""User not in record owner group with id "$groupId"""").asLeft
     }
+
+  def hasSameZoneId(existingZoneId: String, newZoneId: String): Either[Throwable, Unit] =
+    Either.cond(
+      existingZoneId == newZoneId,
+      (),
+      InvalidRequest(
+        s"zoneId in body ('$newZoneId') must equal the existing RecordSet zoneId ('$existingZoneId')")
```
mitruly: Think `match` sounds better than `equal` here.

modules/api/src/main/scala/vinyldns/api/route/RecordSetRouting.scala
```diff
@@ -96,8 +96,15 @@ trait RecordSetRoute extends Directives {
           put {
             monitor("Endpoint.updateRecordSet") {
               entity(as[RecordSet]) { rs =>
-                execute(recordSetService.updateRecordSet(rs, authPrincipal)) { rc =>
-                  complete(StatusCodes.Accepted, rc)
+                handleRejections(invalidContentHandler) {
```
mitruly: I encounter the following error message when trying to start up the API:
```
java.lang.AbstractMethodError: Method vinyldns/api/route/VinylDNSService.vinyldns$api$route$RecordSetRoute$_setter_$vinyldns$api$route$RecordSetRoute$$invalidContentHandler_$eq(Lakka/http/scaladsl/server/RejectionHandler;)V is abstract
	at vinyldns.api.route.VinylDNSService.vinyldns$api$route$RecordSetRoute$_setter_$vinyldns$api$route$RecordSetRoute$$invalidContentHandler_$eq(VinylDNSService.scala)
	at vinyldns.api.route.RecordSetRoute.$init$(RecordSetRouting.scala:158)
	at vinyldns.api.route.VinylDNSService.<init>(VinylDNSService.scala:125)
	at vinyldns.api.Boot$.$anonfun$runApp$25(Boot.scala:137)
	at cats.effect.IO$Map.apply(IO.scala:1407)
	at cats.effect.IO$Map.apply(IO.scala:1403)
	at cats.effect.internals.IORunLoop$.cats$effect$internals$IORunLoop$$loop(IORunLoop.scala:137)
	at cats.effect.internals.IORunLoop$RestartCallback.signal(IORunLoop.scala:336)
	at cats.effect.internals.IORunLoop$RestartCallback.run(IORunLoop.scala:347)
	at cats.effect.internals.Trampoline.cats$effect$internals$Trampoline$$immediateLoop(Trampoline.scala:70)
	at cats.effect.internals.Trampoline.startLoop(Trampoline.scala:36)
	at cats.effect.internals.TrampolineEC$JVMTrampoline.super$startLoop(TrampolineEC.scala:93)
	at cats.effect.internals.TrampolineEC$JVMTrampoline.$anonfun$startLoop$1(TrampolineEC.scala:93)
	at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:23)
	at scala.concurrent.BlockContext$.withBlockContext(BlockContext.scala:85)
	at cats.effect.internals.TrampolineEC$JVMTrampoline.startLoop(TrampolineEC.scala:93)
	at cats.effect.internals.Trampoline.execute(Trampoline.scala:43)
	at cats.effect.internals.TrampolineEC.execute(TrampolineEC.scala:44)
	at cats.effect.internals.IORunLoop$RestartCallback.apply(IORunLoop.scala:355)
	at cats.effect.internals.IORunLoop$RestartCallback.apply(IORunLoop.scala:303)
	at cats.effect.internals.IOParMap$ParMapRunnable.complete(IOParMap.scala:100)
	at cats.effect.internals.IOParMap$ParMapRunnable.$anonfun$callbackA$1(IOParMap.scala:71)
	at cats.effect.internals.IOParMap$ParMapRunnable.$anonfun$callbackA$1$adapted(IOParMap.scala:64)
	at cats.effect.internals.IORunLoop$.cats$effect$internals$IORunLoop$$loop(IORunLoop.scala:134)
	at cats.effect.internals.IORunLoop$RestartCallback.signal(IORunLoop.scala:336)
	at cats.effect.internals.IORunLoop$RestartCallback.apply(IORunLoop.scala:357)
	at cats.effect.internals.IORunLoop$RestartCallback.apply(IORunLoop.scala:303)
	at cats.effect.internals.IOShift$Tick.run(IOShift.scala:36)
	at java.util.concurrent.ForkJoinTask$RunnableExecuteAction.exec(ForkJoinTask.java:1402)
	at java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:289)
	at java.util.concurrent.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1056)
	at java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1692)
	at java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:157)
```

mitruly: Per discussion with @nimaeskandary, `sbt clean` remedied this issue.

modules/api/src/test/scala/vinyldns/api/route/RecordSetRoutingSpec.scala
```diff
@@ -374,6 +375,21 @@ class RecordSetRoutingSpec
       case zoneDeleted.id => Left(ZoneInactiveError(zoneId))
       case notAuthorizedZone.id => Left(NotAuthorizedError(zoneId))
       case syncingZone.id => Left(ZoneUnavailableError(zoneId))
+      case invalidChangeZone.id =>
+        Right(
+          RecordSetChange(
+            zone = invalidChangeZone,
+            recordSet = recordSets
+              .get(rsId)
```
mitruly: Since `recordSets` is a map, we can replace `recordSets.get(rsId).get` with `recordSets(rsId)`.

modules/api/functional_test/live_tests/recordsets/update_recordset_test.py
```diff
@@ -2365,3 +2367,33 @@ def test_update_ds_bad_ttl(shared_zone_test_context):
         if result_rs:
             client.delete_recordset(result_rs['zoneId'], result_rs['id'], status=(202,404))
             client.wait_until_recordset_deleted(result_rs['zoneId'], result_rs['id'])
+
+def test_update_zoneId_fails(shared_zone_test_context):
```
mitruly: Nitpick: since python is snake-case, can we use `zone_id` instead of `zoneId` in the test description (though it's the attribute casing)?

modules/api/src/test/scala/vinyldns/api/route/RecordSetRoutingSpec.scala
nimaeskandary: changed

modules/api/functional_test/live_tests/recordsets/update_recordset_test.py
nimaeskandary: changed

modules/api/functional_test/live_tests/recordsets/update_recordset_test.py
```diff
@@ -2365,3 +2367,33 @@ def test_update_ds_bad_ttl(shared_zone_test_context):
         if result_rs:
             client.delete_recordset(result_rs['zoneId'], result_rs['id'], status=(202,404))
             client.wait_until_recordset_deleted(result_rs['zoneId'], result_rs['id'])
+
+def test_update_zone_id_fails(shared_zone_test_context):
```
rebstar6: why did we take out the other check? what if we have:
---
rname: test-record, zid: 123, rid: 456 (fqdn as test-record.example.com)
and the update is 
rname: test-record, zid: 789, rid: 456 (fqdn as test-record.new-zone.com, zid passed in consistently to both places)
---
this should fail. will it now?

mitruly: Looking at how things are handled in `RecordSetRouting`, @rebstar6 is correct in that the update endpoint (source: https://github.com/vinyldns/vinyldns/blob/c3b437b377cb3de5531754f0f6debb450608f58e/modules/api/src/main/scala/vinyldns/api/route/RecordSetRouting.scala#L96-L103) doesn't actually use the zone ID segment so comparing that they match doesn't appear sufficient.

nimaeskandary: Hm, I think that the function header for `getRecordSet(zoneId: String, recordSetId: String)` is misleading. I think altering that function to return None if the record is not in the zone would solve this cleaner

```
// Note: In MySql we do not need the zone id, since can hit the key directly
  def getRecordSet(zoneId: String, recordSetId: String): IO[Option[RecordSet]] =
    monitor("repo.RecordSet.getRecordSet") {
      IO {
        DB.readOnly { implicit s =>
          FIND_BY_ID.bindByName('id -> recordSetId).map(toRecordSet).single().apply()
        }
      }
    }
```

nimaeskandary: @rebstar6 added that check back

britneywright:
we should clarify this in the docs as well https://www.vinyldns.io/api/update-recordset.html

PR #657, opened 2019-05-28
manual review flag

nimaeskandary:
Batch change input has a manual review flag that defaults to true

In a later pr, a batch change create, if it passes initial input validations and manual review is set to true, will be saved as a pending manual review batch change

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
+++ b/modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
@@ -127,7 +127,4 @@ object VinylDNSConfig {
     } else List()
 
   lazy val maxZoneSize: Int = vinyldnsConfig.as[Option[Int]]("max-zone-size").getOrElse(60000)
-
-  lazy val batchChangeManualReview: Boolean =
-    vinyldnsConfig.as[Option[Boolean]]("batch-change-manual-review").getOrElse(false)
 }
--- a/modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
+++ b/modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
@@ -127,4 +127,7 @@ object VinylDNSConfig {
     } else List()
 
   lazy val maxZoneSize: Int = vinyldnsConfig.as[Option[Int]]("max-zone-size").getOrElse(60000)
+
+  lazy val batchChangeManualReviewEnabled: Boolean =
+    vinyldnsConfig.as[Option[Boolean]]("batch-change-manual-review-enabled").getOrElse(false)
 }
--- a/modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
+++ b/modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
@@ -127,4 +127,7 @@ object VinylDNSConfig {
     } else List()
 
   lazy val maxZoneSize: Int = vinyldnsConfig.as[Option[Int]]("max-zone-size").getOrElse(60000)
+
+  lazy val batchChangeManualReview: Boolean =
+    vinyldnsConfig.as[Option[Boolean]]("batch-change-manual-review").getOrElse(false)
 }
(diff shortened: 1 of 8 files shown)


modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
```diff
@@ -116,4 +116,7 @@ object VinylDNSConfig {
     } else List()
 
   lazy val maxZoneSize: Int = vinyldnsConfig.as[Option[Int]]("max-zone-size").getOrElse(60000)
+
+  lazy val batchChangeManualReview: Boolean =
```
rebstar6: what are we doing with this config?

nimaeskandary: Made before I decided this pr doesn't need anything feature flagged yet, can remove for now

modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
```diff
@@ -116,4 +116,7 @@ object VinylDNSConfig {
...
+    vinyldnsConfig.as[Option[Boolean]]("batch-change-manual-review").getOrElse(false)
```
mitruly: If this is an enabled feature flag, can we make this clearer by naming this `batch-change-manual-review-enabled`?

modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
nimaeskandary: removed

modules/api/src/test/scala/vinyldns/api/route/BatchChangeJsonProtocolSpec.scala
```diff
@@ -241,6 +241,28 @@ class BatchChangeJsonProtocolSpec
...
     }
 
+    "make manualReview true if it does not exist" in {
+      val result = BatchChangeInputSerializer.fromJson(addBatchChangeInputWithComment).value
+
+      result.manualReview shouldBe true
+    }
+
+    "successfully serialize manualReview if it is false" in {
+      val input = addBatchChangeInputWithComment.copy(
```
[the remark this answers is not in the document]
nimaeskandary: changed

modules/api/src/main/scala/vinyldns/api/route/BatchChangeJsonProtocol.scala
```diff
@@ -52,6 +53,7 @@ trait BatchChangeJsonProtocol extends JsonValidation {
       (
         (js \ "comments").optional[String],
         changeList,
+        (js \ "manualReview").default(VinylDNSConfig.batchChangeManualReviewEnabled),
```
rebstar6: need more than this. its not just a default, it needs to be forced in the case that this is false

nimaeskandary: Is there a point then in having in the request payload then? Could be set in the BatchChangeInput apply method based on config in that case

nimaeskandary: @rebstar6 Misunderstood, changed it so that if the manual review flag is enabled in config, then the user can set this to be what they want, and it defaults to true if not there. If not enabled in config, will force to false

britneywright: we don't want users changing this right? It should just go by whatever the config setting is.

nimaeskandary: I believe the discussion earlier concluded that we will force to false if the config is false, and allow the user to set it if the config is true

nimaeskandary:
Change in logic discussed, will default to manual review if enabled in config and soft validations pass. So it no longer is planned to be part of the HTTP route

PR #658, opened 2019-05-29
WIP - bring in optional single change fields

rebstar6:
working towards pending state

### DIFF
--- a/modules/core/src/main/scala/vinyldns/core/domain/batch/SingleChange.scala
+++ b/modules/core/src/main/scala/vinyldns/core/domain/batch/SingleChange.scala
@@ -28,6 +28,7 @@ sealed trait SingleChange {
   val systemMessage: Option[String]
   val typ: RecordType
   val inputName: String
+  val recordSetId: Option[String]
 
   def withFailureMessage(error: String): SingleChange = this match {
     case add: SingleAddChange =>
@@ -35,8 +36,9 @@ sealed trait SingleChange {
     case delete: SingleDeleteChange =>
       delete.copy(status = SingleChangeStatus.Failed, systemMessage = Some(error))
     case unapprovedAdd: UnapprovedSingleAddChange =>
-      // TODO not messing with status for now
-      unapprovedAdd.copy(systemMessage = Some(error))
+      unapprovedAdd.copy(status = SingleChangeStatus.Failed, systemMessage = Some(error))
+    case unapprovedDelete: UnapprovedSingleDeleteChange =>
+      unapprovedDelete.copy(status = SingleChangeStatus.Failed, systemMessage = Some(error))
   }
 
   // TODO technically withProcessingError/complete should only work on ApprovedSingleChange types
@@ -73,12 +75,37 @@ sealed trait SingleChange {
 
 sealed trait ApprovedSingleChange extends SingleChange {
   val recordChangeId: Option[String]
-  val recordSetId: Option[String]
   val zoneId: String
   val recordName: String
   val zoneName: String
   val recordKey = RecordKey(zoneId, recordName, typ)
 }
+
+final case class UnapprovedSingleAddChange(
+    zoneId: Option[String],
+    zoneName: Option[String],
+    recordName: Option[String],
+    inputName: String,
+    typ: RecordType,
+    ttl: Long,
+    recordData: RecordData,
+    status: SingleChangeStatus,
+    systemMessage: Option[String],
+    recordSetId: Option[String],
+    id: String = UUID.randomUUID().toString)
+    extends SingleChange
+
+final case class UnapprovedSingleDeleteChange(
+    zoneId: Option[String],
+    zoneName: Option[String],
+    recordName: Option[String],
+    inputName: String,
+    typ: RecordType,
+    status: SingleChangeStatus,
+    systemMessage: Option[String],
+    recordSetId: Option[String],
+    id: String = UUID.randomUUID().toString)
+    extends SingleChange
 
 final case class SingleAddChange(
     zoneId: String,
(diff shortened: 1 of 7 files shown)


modules/core/src/main/scala/vinyldns/core/protobuf/BatchChangeProtobufConversions.scala
```diff
@@ -87,55 +122,69 @@ trait BatchChangeProtobufConversions extends ProtobufConversions {
         fromPB(t, change)
       }
 
-  def toPB(change: SingleAddChange): Either[Throwable, VinylDNSProto.SingleChange] =
+  def toPB(change: SingleChange): Either[Throwable, VinylDNSProto.SingleChange] =
```
rebstar6: ** carefully review this stuff **

the outcome should be the same for the existing SingleAdd/SingleDelete types, just refactored so the common stuff is in a shared function

modules/core/src/main/scala/vinyldns/core/domain/batch/SingleChange.scala
```diff
@@ -26,22 +26,23 @@ sealed trait SingleChange {
...
     case delete: SingleDeleteChange =>
       delete.copy(status = SingleChangeStatus.Failed, systemMessage = Some(error))
+    case unapprovedAdd: UnapprovedSingleAddChange =>
+      unapprovedAdd.copy(status = SingleChangeStatus.Failed, systemMessage = Some(error))
+    case unapprovedDelete: UnapprovedSingleDeleteChange =>
+      unapprovedDelete.copy(status = SingleChangeStatus.Failed, systemMessage = Some(error))
   }
 
+  // TODO technically withProcessingError/complete should only work on ApprovedSingleChange types
+  // but it's pretty baked in at this point
```
rebstar6: this is a little meh. the RSCommandHandler calls withProcessingError/complete on the single changes in the batch. A workaround would be to similarly split out an `ApprovedBatchChange` from `BatchChange`. Working that through though, it feels like overkill

modules/core/src/main/scala/vinyldns/core/protobuf/BatchChangeProtobufConversions.scala
```diff
@@ -17,29 +17,36 @@
...
     case _: SingleDeleteChange => SingleChangeType.SingleDeleteType
+    case _: UnapprovedSingleAddChange => SingleChangeType.UnapprovedSingleAddType
+    case _: UnapprovedSingleDeleteChange => SingleChangeType.UnapprovedSingleDeleteType
   }
 }
 
 trait BatchChangeProtobufConversions extends ProtobufConversions {
 
-  /* Currently, we only support the add change type.  When we support additional change we will add them here */
+  val protoUnknownEmpty = ""
```
mitruly: I have reservations about mixing empty strings with `NULL` in SQL since they are different despite having some similar behavior, though am fine with this if everyone else is.

modules/core/src/main/scala/vinyldns/core/protobuf/BatchChangeProtobufConversions.scala
mitruly: I checked these scenarios and it seems that there is parity with existing protobuf serialization.

modules/core/src/main/scala/vinyldns/core/domain/batch/SingleChange.scala
```diff
@@ -67,9 +69,44 @@ sealed trait SingleChange {
         status = SingleChangeStatus.Complete,
         recordChangeId = Some(completeRecordChangeId),
         recordSetId = Some(recordSetId))
+    case unapproved => unapproved
   }
 }
 
+sealed trait ApprovedSingleChange extends SingleChange {
```
pauljamescleary: Does "Approved" represent the status here, especially for individual changes.  These are "unprocessed" or "pending" changes.

modules/core/src/main/scala/vinyldns/core/domain/batch/SingleChange.scala
```diff
@@ -67,9 +69,44 @@ sealed trait SingleChange {
...
+    typ: RecordType,
+    ttl: Long,
+    recordData: RecordData,
+    status: SingleChangeStatus,
+    systemMessage: Option[String],
+    recordSetId: Option[String],
+    id: String = UUID.randomUUID().toString)
+    extends SingleChange
+
+final case class UnapprovedSingleDeleteChange(
```
pauljamescleary: Can't we make the required fields optional on the existing `SingleChange` types?  Seems like we are duplicating a lot of things here.

rebstar6: it makes the calling functions more annoying. when we work with changes, we group by that recordKey, which assumes the zoneId/recordName is known. also other stuff assumes we know those fields.

most notibly. the BatchChangeConverter works exclusively with things that have passed the approval point

ill try to mess with it more and see if i cant merge a bit more of it

modules/core/src/main/scala/vinyldns/core/domain/batch/SingleChange.scala
```diff
@@ -26,22 +26,23 @@ sealed trait SingleChange {
...
-  val zoneName: String
-  val recordKey = RecordKey(zoneId, recordName, typ)
+  val recordSetId: Option[String]
 
   def withFailureMessage(error: String): SingleChange = this match {
     case add: SingleAddChange =>
       add.copy(status = SingleChangeStatus.Failed, systemMessage = Some(error))
     case delete: SingleDeleteChange =>
       delete.copy(status = SingleChangeStatus.Failed, systemMessage = Some(error))
+    case unapprovedAdd: UnapprovedSingleAddChange =>
```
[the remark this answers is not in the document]
rebstar6: Unapproved singleChanges live in a batch that hasnt yet passed approval point

PR #663, opened 2019-05-29
Single changes with optional fields

rebstar6:
this will fail, POC


modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeConverter.scala
```diff
@@ -126,6 +126,7 @@ class BatchChangeConverter(batchChangeRepo: BatchChangeRepository, messageQueue:
       existingRecordSets: ExistingRecordSets,
       userId: String,
       ownerGroupId: Option[String]): List[RecordSetChange] = {
+    // TODO: note, this also assumes we are past approval and know the zone/record split at this point
     val grouped = changes.groupBy(_.recordKey)
```
rebstar6: may mess with this and take recordKey out of the type itself

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeConverter.scala
```diff
@@ -126,16 +126,19 @@ class BatchChangeConverter(batchChangeRepo: BatchChangeRepository, messageQueue:
...
-
-    grouped.toList.flatMap {
-      case (_, groupedChanges) =>
-        /* TODO note: the flatmap means if we couldnt get a zone here, we wouldn't be warned
-        that is fine as we will only be passed this change from the service if the zone exists in that map.
-        If we move to getting zones from the DB in the converter, we should report status back on changes
-        where we can no longer find the zone (edge case - means someone submitted batch and then zone was deleted)
-         */
-        combineChanges(groupedChanges, existingZones, existingRecordSets, userId, ownerGroupId)
+    // NOTE: this also assumes we are past approval and know the zone/record split at this point
```
rebstar6: the main tradeoff with this approach (making the fields in the existing types optional) vs making new types for SingleChanges pre and post approval.

the batch converter is *only* dealing with changes post-approval, meaning all changes passed to it have known zone and record info. since thats not baked into the type though, we have to do some weird filtering and such to pull out what we need

modules/core/src/main/scala/vinyldns/core/protobuf/BatchChangeProtobufConversions.scala
```diff
@@ -39,6 +39,10 @@ object SingleChangeType extends Enumeration {
 
 trait BatchChangeProtobufConversions extends ProtobufConversions {
 
+  val protoUnknownEmpty = ""
```
mitruly: Main concern with this is that empty string and `NULL` are different though they exhibit similar behavior. If we're fine with using this, then we don't really have to make `zone_id` `NULL`able in `SingleChangePendingNullFields.sql`

pauljamescleary: I think the issue is how to make the protobuf backward compatible now that there are optional fields.

This option sets things to an empty string, which is unfortunate.

The alternative would be to create NEW types, and have these Single Changes represent already processed single changes.  You would have to switch of the status to know which to load.

pauljamescleary: Update: I think we can change the protobuf to optional.  Should probably test to make sure that we do not break existing protobufs, but it sounds possible.

pauljamescleary: Update: Confirmed!  Should be able to change the protobuf making required fields optional.  I wrote a little test:

1. Save a batch change using the existing protobuf
2. Modify the protobuf to make zoneId `optional`
3. Retrieve the batch change using new protobuf

Everything worked.

Let's do that on this end.

mitruly: Tested this locally and reached similar results. These are the changes that I made:
- Update `zoneId`, `recordName` and `zoneName` to be `optional` in the protobuf
- Update `protoUnknownEmpty` to represent `null` instead of `""`

rebstar6: updated to optional

PR #665, opened 2019-05-30
Add reject batch change endpoint

mitruly:
Add endpoint for feature work related to allowing rejection of batch changes in a pending approval state.

Changes in this pull request:
- Add explicit endpoint for `/zones/batchrecordchanges/:batchrecordchangeid/reject` to reject changes
- Add deserialization for optional body for comments
- Add validation for comment limit based on `review_comment` field in batch change attribute PR (https://github.com/vinyldns/vinyldns/pull/650)
- Add unit and functional tests

### DIFF
--- a/docker/functest/run.sh
+++ b/docker/functest/run.sh
@@ -15,6 +15,13 @@ else
   TEST_PATTERN="-k ${TEST_PATTERN}"
 fi
 
+# Assume manual batch review disabled if not specified
+if [ -z "$MANUAL_BATCH_REVIEW" ]; then
+  MANUAL_BATCH_REVIEW=
+else
+  MANUAL_BATCH_REVIEW="--run-manual-batch-review-tests=true"
+fi
+
 echo "Waiting for API to be ready at ${VINYLDNS_URL} ..."
 DATA=""
 RETRY=60
@@ -46,9 +53,9 @@ cd /app
 # If PROD_ENV is not true, we are in a local docker environment so do not skip anything
 if [ "${PROD_ENV}" = "true" ]; then
     # -m plays havoc with -k, using variables is a headache, so doing this by hand
-    echo "./run-tests.py live_tests -m \"not skip_production\" -v --url=${VINYLDNS_URL} --dns-ip=${DNS_IP} ${TEST_PATTERN}"
-    ./run-tests.py live_tests -v -m "not skip_production" --url=${VINYLDNS_URL} --dns-ip=${DNS_IP} ${TEST_PATTERN}
+    echo "./run-tests.py live_tests -m \"not skip_production\" -v --url=${VINYLDNS_URL} --dns-ip=${DNS_IP} ${MANUAL_BATCH_REVIEW} ${TEST_PATTERN}"
+    ./run-tests.py live_tests -v -m "not skip_production" --url=${VINYLDNS_URL} --dns-ip=${DNS_IP} ${MANUAL_BATCH_REVIEW} ${TEST_PATTERN}
 else
-    echo "./run-tests.py live_tests -v --url=${VINYLDNS_URL} --dns-ip=${DNS_IP} ${TEST_PATTERN}"
-    ./run-tests.py live_tests -v --url=${VINYLDNS_URL} --dns-ip=${DNS_IP} ${TEST_PATTERN}
+    echo "./run-tests.py live_tests -v --url=${VINYLDNS_URL} --dns-ip=${DNS_IP} ${MANUAL_BATCH_REVIEW} ${TEST_PATTERN}"
+    ./run-tests.py live_tests -v --url=${VINYLDNS_URL} --dns-ip=${DNS_IP} ${MANUAL_BATCH_REVIEW} ${TEST_PATTERN}
 fi
--- a/docker/functest/run.sh
+++ b/docker/functest/run.sh
@@ -16,10 +16,10 @@ else
 fi
 
 # Assume manual batch review disabled if not specified
-if [ -z "$MANUAL_BATCH_REVIEW" ]; then
-  MANUAL_BATCH_REVIEW=
+if [ -z "$MANUAL_BATCH_REVIEW_FLAG" ]; then
+  MANUAL_BATCH_REVIEW_OPTS=
 else
-  MANUAL_BATCH_REVIEW="--run-manual-batch-review-tests=true"
+  MANUAL_BATCH_REVIEW_OPTS="--run-manual-batch-review-tests=${MANUAL_BATCH_REVIEW_FLAG}"
 fi
 
 echo "Waiting for API to be ready at ${VINYLDNS_URL} ..."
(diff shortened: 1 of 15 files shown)


modules/api/src/main/scala/vinyldns/api/route/BatchChangeRouting.scala
```diff
@@ -64,11 +65,28 @@ trait BatchChangeRoute extends Directives {
...
+        monitor("Endpoint.rejectBatchChange") {
+          entity(as[RejectBatchChangeInput]) { rejectBatchChangeInput =>
+            handleRejections(rejectionHandler) {
+              validate(
+                rejectBatchChangeInput.comments.forall(c =>
+                  c.length > 0 && c.length <= MAX_COMMENT_LENGTH),
+                s"Comment length must be between 1 and $MAX_COMMENT_LENGTH characters, inclusive."
+              ) {
+                // TODO: Tie into batch change service with auth validation and rejection process
+                complete(StatusCodes.OK)
```
nimaeskandary: In most if not all of our routes we return the altered model on a success, could do that there and verify the status as a step in the func test

modules/api/src/main/scala/vinyldns/api/route/BatchChangeRouting.scala
```diff
@@ -64,11 +65,28 @@ trait BatchChangeRoute extends Directives {
...
             }
         }
+      } ~
+      (post & path("zones" / "batchrecordchanges" / Segment / "reject")) { _ =>
+        monitor("Endpoint.rejectBatchChange") {
+          entity(as[RejectBatchChangeInput]) { rejectBatchChangeInput =>
+            handleRejections(rejectionHandler) {
+              validate(
+                rejectBatchChangeInput.comments.forall(c =>
+                  c.length > 0 && c.length <= MAX_COMMENT_LENGTH),
```
nimaeskandary: Are empty strings that bad?

modules/api/src/main/scala/vinyldns/api/route/BatchChangeRouting.scala
mitruly: Yeah, that will have to be done when this is hooked up. I'll add a comment.

modules/api/src/main/scala/vinyldns/api/route/BatchChangeRouting.scala
```diff
@@ -64,11 +65,28 @@ trait BatchChangeRoute extends Directives {
               }
             }
         }
+      } ~
+      (post & path("zones" / "batchrecordchanges" / Segment / "reject")) { _ =>
```
nimaeskandary: PUT possibly? Not sure, kinda altering the existing model

modules/api/src/main/scala/vinyldns/api/route/BatchChangeRouting.scala
mitruly: It depends on how we want to handle this, as an empty string is technically a string; I know that there were issues with DynamoDB and empty string inserts. I see what you mean about clients, though we are doing something similar to maintain compatibility with other clients such as Postman where no entity is also a valid scenario.

nimaeskandary: I think we should investigate are standard around that since we are moving to SQL. Still think it should be an Option, but since we have so many clients that have to implement this wondering if there is harm in letting in empty strings if thats what the client sends

modules/api/src/main/scala/vinyldns/api/route/BatchChangeRouting.scala
mitruly: This is a good discussion point - was thinking about endpoints like posting `processingDisabled`  to our `/status` endpoint; if we are doing an update, shouldn't there be other characteristics of the batch change model be passed in as well?

modules/api/src/main/scala/vinyldns/api/route/BatchChangeRouting.scala
britneywright: this doesn't seem like the right spot for this validation. it seems like validations here are intended for query parameters.

mitruly: @britneywright Good point - I'll move this into the `BatchChangeJsonProtocol`

modules/api/src/main/scala/vinyldns/api/route/BatchChangeRouting.scala
pauljamescleary: `PUT` carries idempotent semantics, you can think of it like inserting an item in a Hashmap.  If you do the same thing 1000 times, the result will be the same.  Use `PUT` for modifying an entity as another example.

I don't think we want that here.  `POST` carries side-effecting or EXECUTE semantics.

Because the path is `/reject`, you are saying "execute this verb on this batch change".  `POST` is appropriate here.

modules/api/src/main/scala/vinyldns/api/route/BatchChangeRouting.scala
pauljamescleary: +1 this should be in the JsonProtocol

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeProtocol.scala
```diff
@@ -67,3 +67,5 @@ object ChangeInputType extends Enumeration {
   type ChangeInputType = Value
   val Add, DeleteRecordSet = Value
 }
+
+final case class RejectBatchChangeInput(comments: Option[String])
```
britneywright: think it would be clearer to call this review_comment to distinguish from the other `comments` batch change field.

modules/api/src/main/scala/vinyldns/api/route/BatchChangeRouting.scala
```diff
@@ -64,6 +64,15 @@ trait BatchChangeRoute extends Directives {
...
+        monitor("Endpoint.rejectBatchChange") {
+          entity(as[Option[RejectBatchChangeInput]]) { input =>
+            // TODO: Tie into batch change service with auth validation and rejection process
+            complete(StatusCodes.OK, input)
```
rebstar6: feature flag on the route im thinking (unless you want it in the service)

until the validations and backend are ready, I dont want this route doing anything

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeProtocol.scala
rebstar6: `reviewComment` is what it will be stored on the batchChange as. so i think that would make sense

docker/functest/run.sh
```diff
@@ -15,6 +15,13 @@ else
   TEST_PATTERN="-k ${TEST_PATTERN}"
 fi
 
+# Assume manual batch review disabled if not specified
+if [ -z "$MANUAL_BATCH_REVIEW" ]; then
+  MANUAL_BATCH_REVIEW=
+else
+  MANUAL_BATCH_REVIEW="--run-manual-batch-review-tests=true"
```
rebstar6: this is a little weird, if you set to false it will also set to true.

rebstar6: its just checking existence not value

modules/api/src/main/resources/reference.conf
```diff
@@ -106,4 +106,6 @@ vinyldns {
   }
 
   batch-change-limit = 1000
+
+  manual-batch-review-enabled = true
```
rebstar6: ref.conf at true but the config at `.getOrElse(false)` is kinda weird.

i kinda think leave out of ref.conf and just add to application.conf for running/testing, no? unless you think theres a good reason for it here

PR #666, opened 2019-05-30
WIP - ttl defaults, do not update TTL if excluded

rebstar6:
- make batch ttl optional
- if not provided, add should go to default, update should remain unchanged


modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -178,6 +178,23 @@ class BatchChangeService(
...
+    changes.mapValid {
+      case add: AddChangeForValidation =>
+        existingRecordSets
+          .get(add.recordKey)
+          .map { rs =>
+            val ttl = add.inputChange.ttl.getOrElse(rs.ttl)
+            val changeWithTtl = add.inputChange.copy(ttl = Some(ttl))
+            add.copy(inputChange = changeWithTtl)
+          }
+          .getOrElse(add)
```
rebstar6: doing this all with copy is a little yucky. the alternative would be to add another trait/type for addchanges that is identical to AddChangeForValidation but has a required ttl. it's feels like a lot of overhead

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -178,6 +178,23 @@ class BatchChangeService(
...
 
+  def doTtlMapping(
+      changes: ValidatedBatch[ChangeForValidation],
+      existingRecordSets: ExistingRecordSets): ValidatedBatch[ChangeForValidation] =
+    changes.mapValid {
+      case add: AddChangeForValidation =>
+        existingRecordSets
+          .get(add.recordKey)
+          .map { rs =>
+            val ttl = add.inputChange.ttl.getOrElse(rs.ttl)
```
rebstar6: ^^
if ttl was explicitly set on the input, use that
if not, and theres a record in the DB matching this add (meaning it must be an update), use the record's ttl
if none of the above, it will remain None which later gets converted to the system default

britneywright:
I know this still a WIP, but noting that this currently doesn't save the default TTL on the single change model if a TTL isn't provided. Testing in the portal shows an empty TTL field in the batch change summary view. 

batch change summary view with empty TTL field

rebstar6:
closing for #670

PR #667, opened 2019-05-30
add high value domain info to docs

### LINKED ISSUE #489: documentation for high value domains
**Motivation**
We implemented high value domains in #376 but didn't update the docs to include references to it where needed.

**Scope of change**
- Add the error messages related to high value domains where applicable
- explain how to configure high value domains

britneywright:
Fixes #489.

Changes in this pull request:
- add error message information in batch change errors doc
- add example in api config docs
- noticed `shared-approved-types` was incorrectly nested in api config example so fixed that.

### DIFF
--- a/modules/docs/src/main/tut/operator/config-api.md
+++ b/modules/docs/src/main/tut/operator/config-api.md
@@ -603,7 +603,7 @@ akka.http {
 }
 
 # types of unowned records that users can access in shared zones
-shared-approved-types = ["A", "AAAA", "CNAME", "PTR"]
+shared-approved-types = ["A", "AAAA", "CNAME", "PTR", "TXT"]
     
 # FQDNs / IPs that cannot be modified via VinylDNS
 # regex-list: list of regular expressions matching any FQDN that are not allowed to be modified by this VinylDNS instance
--- a/modules/docs/src/main/tut/operator/config-api.md
+++ b/modules/docs/src/main/tut/operator/config-api.md
@@ -612,7 +612,6 @@ high-value-domains = {
     regex-list = [
       "high-value-domain.*"
     ip-list = [
-      # using reverse zones in the vinyldns/bind9 docker image for testing
       "192.0.2.252",
       "192.0.2.253",
       "fd69:27cc:fe91:0:0:0:0:ffff",
--- a/modules/docs/src/main/tut/operator/config-api.md
+++ b/modules/docs/src/main/tut/operator/config-api.md
@@ -603,7 +603,7 @@ akka.http {
 }
 
 # types of unowned records that users can access in shared zones
-shared-approved-types = ["A", "AAAA", "CNAME", "PTR", "TXT"]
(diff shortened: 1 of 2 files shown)


modules/docs/src/main/tut/operator/config-api.md
```diff
@@ -600,9 +600,24 @@ akka.http {
...
   }
-
-   # types of unowned records that users can access in shared zones
-    shared-approved-types = ["A", "AAAA", "CNAME", "PTR"]
 }
 
-```
\ No newline at end of file
+# types of unowned records that users can access in shared zones
+shared-approved-types = ["A", "AAAA", "CNAME", "PTR", "TXT"]
```
mitruly: It makes sense to keep the old example config value for `shared-approved-types` since that is what matches `/main/resources/application.conf`.

modules/docs/src/main/tut/operator/config-api.md
```diff
@@ -600,9 +600,24 @@ akka.http {
...
+    
+# FQDNs / IPs that cannot be modified via VinylDNS
+# regex-list used for all record types except PTR
+# ip-list used exclusively for PTR records
+high-value-domains = {
+    regex-list = [
+      "high-value-domain.*" # for testing
+    ]
+    ip-list = [
+      # using reverse zones in the vinyldns/bind9 docker image for testing
```
pauljamescleary: can remove this comment.  I think the `ip-list` comment above should be a little more like

`# ip-list - list of IP addresses that cannot be modified by this VinylDNS instance`

Similarly, change `regex-list`

`# regex-list - list of regular expressions matching any FQDN that are not allowed to be modified by this VinylDNS instance`

modules/docs/src/main/tut/operator/config-api.md
```diff
@@ -600,9 +600,24 @@ akka.http {
...
\ No newline at end of file
+# types of unowned records that users can access in shared zones
+shared-approved-types = ["A", "AAAA", "CNAME", "PTR"]
+    
+# FQDNs / IPs that cannot be modified via VinylDNS
+# regex-list used for all record types except PTR
+# ip-list used exclusively for PTR records
+high-value-domains = {
+    regex-list = [
+      "high-value-domain.*" # for testing
```
pauljamescleary: remove the `# for testing`

britneywright: we might want to leave that, because it would actually cause the func tests to fail if someone removes it.

pauljamescleary: for documentation purposes (which this md is all about), it is more important to document what each value means and maybe possible values or examples.  None of the other config examples indicate "needed for testing"

modules/docs/src/main/tut/operator/config-api.md
rebstar6: +1 to pauls comments, otherwise im good with this PR

britneywright: oops missed the remove part.

PR #669, opened 2019-05-30
Adopt Structured Logging using Extended Elastic ECS format

deepakmohanakrishnan1984:
This PR is complete and ready to review.

Changes in this pull request:
- Use [logstash-logback-encoder](https://github.com/logstash/logstash-logback-encoder) library to emit structured JSON logs.
- Configure `logback.xml` files for API and Portal modules to use `net.logstash.logback.encoder.LoggingEventCompositeJsonEncoder` to emit logs as JSON documents.

With the current change using the library, the JSON formatted logs would like as shown in the sample below (but in a single line - formatted here for easy human readability):
```json
{
  "@timestamp": "2019-05-30T20:26:19.619+00:00",
  "message": "monitor='queue.SQS.receive' millis=1022 fail=0",
  "labels": {
    "application": "vinyldns-api",
    "env": "dev"
  },
  "ecs": {
    "version": "1.0.0"
  },
  "log": {
    "level": "INFO",
    "logger_name": "vinyldns.core.route.Monitor",
    "caller_class_name": "vinyldns.core.route.Monitor",
    "caller_method_name": "capture",
    "caller_file_name": "Monitor.scala",
    "caller_line_number": 97
  },
  "process": {
    "thread": {
      "name": "pool-3-thread-2"
    }
  }
}
```

-  Adopt [Elastic's ECS](https://www.elastic.co/guide/en/ecs/current/index.html) for JSON structured logs
1.1. Field mappings extend from ECS schema for VinylDNS specific fields
1.2. Code and configuration change to emit structured logs in ECS schema. For e.g.: The KV format in the message field in the above sample would be included as fields in JSON document under an appropriate context.


modules/api/src/main/resources/test/logback.xml
```diff
@@ -1,8 +1,50 @@
 <configuration>
-    <!-- Test configuration, log to console so we can get the docker logs -->
     <appender name="CONSOLE" class="ch.qos.logback.core.ConsoleAppender">
-        <encoder class="ch.qos.logback.classic.encoder.PatternLayoutEncoder">
-            <pattern>%d [test] %-5p | \(%logger{4}:%line\) | %msg %n</pattern>
+        <encoder class="net.logstash.logback.encoder.LoggingEventCompositeJsonEncoder">
```
pauljamescleary: if we really want to do this, we would have to update a lot more `logback.xml` files :)

I think specifically we should consider leaving `test` logback files alone, and only update the others.

This should also get into the portal.  Also, don't forget about src/universal

deepakmohanakrishnan1984: Sure. Update the PR to apply logbook config changes to both API and Portal modules.

modules/api/src/main/resources/logback.xml
```diff
@@ -1,8 +1,51 @@
...
+                        <globalCustomFields>
+                            <customFields>{"application": "vinyldns-api", "env": "dev"}</customFields>
+                        </globalCustomFields>
+                    </providers>
+                </nestedField>
+                <nestedField>
+                    <fieldName>ecs</fieldName>
+                    <providers>
+                        <globalCustomFields>
+                            <customFields>{"version": "1.0.0"}</customFields>
```
nimaeskandary: Is this meant to be the config version or the application version?

deepakmohanakrishnan1984: It is used to store the elastic common schema(ecs) version. Elastic would provide the migration tools based on this ecs version number for elasticsearch index mapping, kibana dashboards etc. 

PS: sorry for the delay in responding to your comments. Will continue work on this activity and watch the comments going forward.

modules/core/src/main/scala/vinyldns/core/logging/StructuredArgs.scala
```diff
@@ -0,0 +1,195 @@
...
+  /**
+    * handler function to extract acceptable fields from domain objects to include to log statements.
+    *
+    * @param t domain object of type T
+    * @tparam T type construct for domain object
+    * @return Map containing the acceptable fields
+    */
+  private def fields[T](t: T): Map[_, _] = t match {
+    //core domains
+    case u: User => Map(_Id -> u.id, Name -> u.userName, Type -> "user")
```
pauljamescleary: I think structured logging for the most part the way you have it here is good.  I struggle with the helpers, little library you used to try and make it more convenient.  For example, in here, what would happen if I decide do send a `SingleAddChange` in?  Will I get an exception?  Will there be a compile time error?

Certainly logging cannot throw an Exception, as that violates the rest of the codebase

modules/core/src/main/scala/vinyldns/core/logging/StructuredArgs.scala
```diff
@@ -0,0 +1,195 @@
...
+    case s: Seq[_] => Map("arr" -> s.map(fields).toArray)
+    case m: Map[_, _] => m
+    case x: String => Map(Value -> x)
+  }
+
+  private def detailedFields[T](t: T): Map[_, _] =
+    t match {
+      case dzc: ZoneChange =>
+        Map(
+          Id -> dzc.id,
```
pauljamescleary: Similarly to my above statement, I think it is worthwhile to do a comprehensive review of the log statements that we have and optimize them.  They were kinda mangled together, so building a framework to simplify those at this point maybe premature.

Probably need to get a SPIKE on our backlog to do a comprehensive logging review, and we should also consider distributed tracing, or some kind of correlation id

modules/core/src/main/scala/vinyldns/core/logging/StructuredArgs.scala
deepakmohanakrishnan1984: Thats a very valid point. The idea is to have a pattern match for every single domain model that we want to be logged. To avoid exceptions, we should raise a compilation error for `MatchError` missing pattern for new domain models. If not, we can have a catch all pattern match which is not my favorite though.

modules/core/src/main/scala/vinyldns/core/logging/StructuredArgs.scala
deepakmohanakrishnan1984: That would be a great start to introduce distributed tracing. It would definitely helps for event correlation, between your logs, metrics and traces using a correlation id concepts. I would love to participate in such discussion. Please feel free to include me for such discussions.

pauljamescleary:
Closing as PR is stale, can be reopened later

PR #670, opened 2019-05-31
Make TTL optional (again)

rebstar6:
Changes in this pull request:
- make batch ttl optional
- if not provided, add should go to default, update should remain unchanged

PR #671, opened 2019-05-31
make batch change TTL optional in the portal

britneywright:
builds on #670

Changes in this pull request:
- don't make TTL field required, but still validate minimum and maximum values if there's an input
- add tooltip indicating that the field is optional and what the system default is
- don't include the TTL value by default when adding rows, instead show the default value as placeholder text in the TTL field

### DIFF
--- a/modules/portal/app/models/Meta.scala
+++ b/modules/portal/app/models/Meta.scala
@@ -17,12 +17,17 @@
 package models
 import play.api.Configuration
 
-case class Meta(version: String, sharedDisplayEnabled: Boolean, batchChangeLimit: Int)
+case class Meta(
+    version: String,
+    sharedDisplayEnabled: Boolean,
+    batchChangeLimit: Int,
+    defaultTtl: Long)
 object Meta {
   def apply(config: Configuration): Meta =
     Meta(
       config.getOptional[String]("vinyldns.version").getOrElse("unknown"),
       config.getOptional[Boolean]("shared-display-enabled").getOrElse(false),
-      config.getOptional[Int]("batch-change-limit").getOrElse(1000)
+      config.getOptional[Int]("batch-change-limit").getOrElse(1000),
+      config.getOptional[Long]("default-ttl").getOrElse(7200L)
     )
 }
--- a/modules/portal/app/views/batchChanges/batchChangeNew.scala.html
+++ b/modules/portal/app/views/batchChanges/batchChangeNew.scala.html
@@ -67,11 +67,7 @@
                                             <span class="fa fa-question-circle"></span>
                                         </span>
                                     </th>
-                                    <th class="col-md-1">TTL
-                                        <span data-toggle="tooltip" data-placement="top" title="Optional. Default is {{defaultTtl}}.">
-                                            <span class="fa fa-question-circle"></span>
-                                        </span>
-                                    </th>
+                                    <th class="col-md-1">TTL (optional)</th>
                                     <th>Record Data</th>
                                     <th></th>
                                 </tr>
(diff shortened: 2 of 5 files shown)


modules/portal/app/views/batchChanges/batchChangeNew.scala.html
```diff
@@ -67,7 +67,11 @@ <h4>Changes</h4>
                                             <span class="fa fa-question-circle"></span>
                                         </span>
                                     </th>
-                                    <th class="col-md-1">TTL</th>
+                                    <th class="col-md-1">TTL
+                                        <span data-toggle="tooltip" data-placement="top" title="Optional. Default is {{defaultTtl}}.">
```
mitruly: May make more sense to clarify that the default of `7200` is for new record creation; not providing TTL for an existing record will persist the existing record's TTL value.

britneywright: that's in the documentation, i don't want to get overly explanatory in the tooltip. the main point is indicating it's an optional field.

modules/portal/app/views/batchChanges/batchChangeNew.scala.html
```diff
@@ -67,7 +67,11 @@ <h4>Changes</h4>
                                             <span class="fa fa-question-circle"></span>
                                         </span>
                                     </th>
-                                    <th class="col-md-1">TTL</th>
+                                    <th class="col-md-1">TTL
```
rebstar6: can you make this `TTL (optional)` or otherwise indicate that it can be left empty? i see it in the tooltip, but i dont know that people are actually gonna do that hover

modules/portal/app/models/Meta.scala
```diff
@@ -17,12 +17,17 @@
...
+    batchChangeLimit: Int,
+    defaultTtl: Long)
 object Meta {
   def apply(config: Configuration): Meta =
     Meta(
       config.getOptional[String]("vinyldns.version").getOrElse("unknown"),
       config.getOptional[Boolean]("shared-display-enabled").getOrElse(false),
-      config.getOptional[Int]("batch-change-limit").getOrElse(1000)
+      config.getOptional[Int]("batch-change-limit").getOrElse(1000),
+      config.getOptional[Long]("default-ttl").getOrElse(7200L)
```
rebstar6: may not need this, depends on what you want in the tooltip

were not using it outside of the tooltip really, its fine to have the example display show whatever

modules/portal/app/views/batchChanges/batchChangeNew.scala.html
britneywright: took out the tooltip and put in the optional text instead. don't want things to get too junky and verbose up there.

nimaeskandary: Even though the API already adds it if its not there I feel like its clearer to fill it out in the portal to whatever it is set to in config and not try to explain the optional thing

britneywright: we can't fill it in because we don't know at this point if it's an add or an update, and the API now will either use the previous TTL if it's an update or the default TTL if it's an add.

nimaeskandary: I just don't feel like it's intuitive right now for people who are not familiar with DNS, if they knew what TTL was then they would change the pre filled value to be what they want in the update, and if they didn't know then the pre filled default would work for them

rebstar6: @nimaeskandary ultimately we want them to enter nothing if they dont know what ttl is. if you think there is better wording/display we can use for that, let us know. pre-populating the default will make TTL changes happen that we dont want to happen

PR #673, opened 2019-06-03
disable multi record batch update

nimaeskandary:
There can be multiple records in a RecordSet. Adds a check to block updates of these records via the batch change interface. Can be disabled via config

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/DomainValidationErrors.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/DomainValidationErrors.scala
@@ -17,7 +17,7 @@
 package vinyldns.api.domain
 
 import vinyldns.api.domain.batch.SupportedBatchChangeRecordTypes
-import vinyldns.core.domain.record.RecordType
+import vinyldns.core.domain.record.{RecordSet, RecordType}
 import vinyldns.core.domain.record.RecordType.RecordType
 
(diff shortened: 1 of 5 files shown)


modules/api/src/main/scala/vinyldns/api/domain/DomainValidationErrors.scala
```diff
@@ -149,4 +149,12 @@ final case class MissingOwnerGroupId(recordName: String, zoneName: String)
   def message: String =
     s"""Zone "$zoneName" is a shared zone, so owner group ID must be specified for record "$recordName"."""
 }
+
+final case class MultipleRecordsInRecordSet(record: RecordSet) extends DomainValidationError {
+  def message: String =
+    s"""RecordSet with name ${record.name} and type ${record.typ.toString} cannot be updated in a single Batch Change
+       |because it contains multiple DNS records (${record.records.length}). If this is expected, issue a
```
pauljamescleary: probably `because it contains ${record.records.length} DNS records`

pauljamescleary: I wouldn't recommend telling them what to do here.  I think "multi record updates in a batch change are not allowed"

rebstar6:
replacing with #677

PR #674, opened 2019-06-03
Add email notifier

dscleaver:
Changes in this pull request:
- Add email notifier
- Provide email on batch change to the requesting user

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -36,6 +36,6 @@ services:
   mail:
     image: flaviovs/mock-smtp
     ports:
-      - "25:25"
+      - "19025:25"
     volumes:
       - ../target:/var/lib/mock-smtp
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -32,3 +32,10 @@ services:
       - "9324:9324"
     volumes:
       - ./elasticmq/custom.conf:/etc/elasticmq/elasticmq.conf
+
+  mail:
+    image: flaviovs/mock-smtp
+    ports:
+      - "25:25"
+    volumes:
+      - ../target:/var/lib/mock-smtp
--- a/modules/api/src/it/scala/vinyldns/api/notifier/email/EmailNotifierIntegrationSpec.scala
+++ b/modules/api/src/it/scala/vinyldns/api/notifier/email/EmailNotifierIntegrationSpec.scala
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2018 Comcast Cable Communications Management, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package vinyldns.api.notifier.email
 
 import com.typesafe.config.{Config, ConfigFactory}
--- a/modules/api/src/it/scala/vinyldns/api/notifier/email/EmailNotifierIntegrationSpec.scala
+++ b/modules/api/src/it/scala/vinyldns/api/notifier/email/EmailNotifierIntegrationSpec.scala
@@ -27,7 +27,7 @@ import vinyldns.core.domain.record.AData
 import org.joda.time.DateTime
 import vinyldns.core.TestMembershipData._
 import java.nio.file.{Files, Path, Paths}
-import cats.effect.IO
+import cats.effect.{IO, Resource}
 import scala.collection.JavaConverters._
 import org.scalatest.BeforeAndAfterEach
 import cats.implicits._
@@ -100,12 +100,10 @@ class EmailNotifierIntegrationSpec
     } yield ()
 
   def retrieveEmailFiles(path: Path): IO[List[Path]] =
-    IO(Files.newDirectoryStream(path, "*.eml")).bracket { s =>
+    Resource.fromAutoCloseable(IO(Files.newDirectoryStream(path, "*.eml"))).use { s =>
       IO {
         s.iterator.asScala.toList
       }
-    } { s =>
-      IO(s.close)
     }
 
 }
--- a/modules/api/src/it/scala/vinyldns/api/notifier/email/EmailNotifierIntegrationSpec.scala
+++ b/modules/api/src/it/scala/vinyldns/api/notifier/email/EmailNotifierIntegrationSpec.scala
@@ -13,6 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 package vinyldns.api.notifier.email
 
 import com.typesafe.config.{Config, ConfigFactory}
--- a/modules/api/src/it/scala/vinyldns/api/notifier/email/EmailNotifierIntegrationSpec.scala
+++ b/modules/api/src/it/scala/vinyldns/api/notifier/email/EmailNotifierIntegrationSpec.scala
@@ -1,0 +1,95 @@
+package vinyldns.api.notifier.email
+
+import com.typesafe.config.{Config, ConfigFactory}
+import vinyldns.core.notifier._
+import vinyldns.api.MySqlApiIntegrationSpec
+import vinyldns.mysql.MySqlIntegrationSpec
+import org.scalatest.{Matchers, WordSpecLike}
+import vinyldns.core.domain.batch._
+import vinyldns.core.domain.record.RecordType
+import vinyldns.core.domain.record.AData
+import org.joda.time.DateTime
+import vinyldns.core.TestMembershipData._
+import java.nio.file.{Files, Path, Paths}
+import cats.effect.IO
+import scala.collection.JavaConverters._
+import org.scalatest.BeforeAndAfterEach
+import cats.implicits._
+
+class EmailNotifierIntegrationSpec
+    extends MySqlApiIntegrationSpec
+    with MySqlIntegrationSpec
+    with Matchers
+    with WordSpecLike
+    with BeforeAndAfterEach {
+
+  import vinyldns.api.domain.DomainValidations._
+
+  val emailConfig: Config = ConfigFactory.load().getConfig("vinyldns.email.settings")
+
+  val targetDirectory = Paths.get("../../target")
+
+  override def beforeEach: Unit =
+    deleteEmailFiles(targetDirectory).unsafeRunSync()
+
+  override def afterEach: Unit =
+    deleteEmailFiles(targetDirectory).unsafeRunSync()
+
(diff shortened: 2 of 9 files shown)


modules/api/src/main/scala/vinyldns/api/notifier/email/EmailNotifier.scala
```diff
@@ -0,0 +1,125 @@
...
+  def sendBatchChangeNotification(bc: BatchChange): IO[Unit] =
+    userRepository.getUser(bc.userId).flatMap {
+      case Some(UserWithEmail(email)) =>
+        send(email) { message =>
+          message.setSubject(s"VinylDNS Batch change ${bc.id} results")
+          message.setContent(formatBatchChange(bc), "text/html")
+          message
+        }
+      case None => IO { logger.warn(s"Unable to find user: ${bc.userId}") }
+      case _ => IO.unit
```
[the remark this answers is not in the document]
dscleaver: My thought is that automation users probably don’t have emails and aren’t likely to want emails and we probably also don’t want to log them. 

I wasn’t concerned with users with bad emails. If you care about the emails then you’ll complain... not sure how much pro activity is needed.

modules/api/src/main/scala/vinyldns/api/notifier/email/EmailNotifierConfig.scala
```diff
@@ -0,0 +1,58 @@
...
+
+  implicit val smtpPropertiesReader: ConfigReader[Properties] = {
+    ConfigReader[ConfigObject].map { config =>
+      val props = new Properties()
+      convertToProperties("mail.smtp", config, props)
+      props
+    }
+  }
+
+  def convertToProperties(baseKey: String, config: ConfigObject, props: Properties): Unit =
```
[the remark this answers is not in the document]
dscleaver: I’m not convinced that the implicit class is necessary. I think I am going to move this function in as a helper though. So maybe that can take care of it.

As for the Null pointer I think I would rather explicitly check for valueType != NULL

modules/api/src/main/scala/vinyldns/api/notifier/email/EmailNotifierProvider.scala
```diff
@@ -0,0 +1,39 @@
...
+
+import vinyldns.core.notifier.{Notifier, NotifierConfig, NotifierProvider}
+import vinyldns.core.domain.membership.UserRepository
+import pureconfig.module.catseffect.loadConfigF
+import cats.effect.IO
+import javax.mail.Session
+
+class EmailNotifierProvider extends NotifierProvider {
+
+  import EmailNotifierConfig._
```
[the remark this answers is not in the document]
dscleaver: It imports the implicits from the helper. Pretty sure it needs to be there

modules/api/src/main/scala/vinyldns/api/notifier/email/EmailNotifierConfig.scala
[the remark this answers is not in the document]
dscleaver: See what you think of this new version.

modules/api/src/main/scala/vinyldns/api/notifier/email/EmailNotifier.scala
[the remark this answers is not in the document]
mitruly: I think I agree with not doing anything if no e-mail address is provided, but think we should just log if the e-mail resolution fails. While we wouldn't attempt to send an e-mail if that step fails, it is potentially unexpected behavior from a user perspective.

modules/api/src/main/scala/vinyldns/api/notifier/email/EmailNotifier.scala
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+  def formatBatchChange(bc: BatchChange): String =
```
pauljamescleary: can we get a screen shot of the email with the embedded HTML?

modules/api/src/main/scala/vinyldns/api/notifier/email/EmailNotifier.scala
```diff
@@ -0,0 +1,125 @@
...
+    s"""<h1>Batch Change Results</h1>
+      | <b>Id:</b> ${bc.id}<br/>
+      | <b>Comments:</b> ${bc.comments.getOrElse("")}</br>
```
pauljamescleary: so it is possible for the batch change to be rejected (although not for a while).  We should hard code the status to be "Implemented" for the time being.

we should also add the `userName` of the user that submitted the change, along with the date/time it was implemented.  The reason is that this information is used as kind of an "audit trail" outside of VinylDNS.  You can imagine someone saving this email and looking at it later.

modules/api/src/main/scala/vinyldns/api/notifier/email/EmailNotifierConfig.scala
```diff
@@ -0,0 +1,60 @@
...
+import pureconfig.ConfigReader
+import scala.util.Try
+import pureconfig.error.CannotConvert
+import java.util.Properties
+import com.typesafe.config.{ConfigObject, ConfigValue}
+import com.typesafe.config.ConfigValueType
+
+object EmailNotifierConfig {
+
+  implicit val smtpPropertiesReader: ConfigReader[Properties] = {
```
pauljamescleary: I am curious as to why you chose to use a custom `ConfigReader` as opposed to having `case class` and then using pureconfig?

dscleaver: There are a lot of mail.smtp properties and rather than capture each one, this will handle any that are placed under smtp

modules/api/src/main/scala/vinyldns/api/notifier/email/EmailNotifier.scala
pauljamescleary: `<b>Stats: </b> Implemented`

modules/api/src/it/scala/vinyldns/api/notifier/email/EmailNotifierIntegrationSpec.scala
```diff
@@ -0,0 +1,110 @@
...
+  import vinyldns.api.domain.DomainValidations._
+
+  val emailConfig: Config = ConfigFactory.load().getConfig("vinyldns.email.settings")
+
+  val targetDirectory = Paths.get("../../target")
+
+  override def beforeEach: Unit =
+    deleteEmailFiles(targetDirectory).unsafeRunSync()
+
+  override def afterEach: Unit =
```
nimaeskandary: Could this be afterAll?

dscleaver: I mean there is only one test... I was anticipating more, and it's easier to check that one email happened in each test.

docker/docker-compose.yml
```diff
@@ -32,3 +32,10 @@ services:
       - "9324:9324"
     volumes:
       - ./elasticmq/custom.conf:/etc/elasticmq/elasticmq.conf
+
+  mail:
+    image: flaviovs/mock-smtp
+    ports:
+      - "25:25"
```
nimaeskandary: Should this be using the default smtp port on the host machine?

dscleaver: Didn't cause an issue locally or on travis... I'm open to changing it though

modules/api/src/it/scala/vinyldns/api/notifier/email/EmailNotifierIntegrationSpec.scala
```diff
@@ -0,0 +1,111 @@
...
+  def deleteEmailFiles(path: Path): IO[Unit] =
+    for {
+      files <- retrieveEmailFiles(path)
+      _ <- files.traverse { file =>
+        IO(Files.delete(file))
+      }
+    } yield ()
+
+  def retrieveEmailFiles(path: Path): IO[List[Path]] =
+    IO(Files.newDirectoryStream(path, "*.eml")).bracket { s =>
```
mitruly: We should be able to use `Resource` since `DirectoryStream<T>` is a sub-interface of `AutoCloseable` (https://docs.oracle.com/javase/8/docs/api/java/lang/AutoCloseable.html):
```
    Resource
      .fromAutoCloseable(IO(Files.newDirectoryStream(path, "*.eml")))
      .use(directoryStream => IO(directoryStream.iterator.asScala.toList))
```

dscleaver: Thanks changed it to use Resource

docker/docker-compose.yml
pauljamescleary: Right, you may have port access issues on 25, I would recommend using 19xxx port externally

modules/api/src/main/scala/vinyldns/api/notifier/email/EmailNotifier.scala
```diff
@@ -0,0 +1,134 @@
...
+      case None => IO { logger.warn(s"Unable to find user: ${bc.userId}") }
+      case _ => IO.unit
+    }
+
+  def formatBatchChange(bc: BatchChange): String =
+    s"""<h1>Batch Change Results</h1>
+      | <b>Submitter:</b> ${bc.userName} <br/>
+      | <b>Created:</b> ${bc.createdTimestamp.toString(DateTimeFormat.fullDateTime)} <br/>
+      | <b>Id:</b> ${bc.id}<br/>
+      | <b>Status:</b> Implemented<br/>
```
britneywright: this should be the status of the batch change based on the statuses of the single changes. Completed, Failed or PartialFailure.

modules/api/src/main/scala/vinyldns/api/notifier/email/EmailNotifier.scala
```diff
@@ -0,0 +1,134 @@
...
+      | <b>Comments:</b> ${bc.comments.getOrElse("")}</br>
```
britneywright: Can we change this from "Comments" to "Description" and move it under the submitter. We call this "Description" in the portal. Can we also not show it at all if there isn't a value, rather than showing the label with no text.

modules/api/src/main/scala/vinyldns/api/notifier/email/EmailNotifier.scala
dscleaver: I think Paul wanted it to eventually be about approval. See above.

rebstar6: we not have a BatchChange `approvalStatus` in addition to the normal status. Pipe that (or both) in.

modules/api/src/main/scala/vinyldns/api/notifier/email/EmailNotifier.scala
dscleaver: Due to whitelisting required in the comcast mail relay server, we have to wait until we get this deployed somewhere.

modules/api/src/main/scala/vinyldns/api/notifier/email/EmailNotifier.scala
pauljamescleary: @britneywright I agree with "Description".  I would always show it, even if it has no value.

PR #675, opened 2019-06-03
Fix LDAP lookup

mitruly:
Our current LDAP authenticator and higher-level functions (eg. `authenticate` and `lookup`) return `UserDoesNotExistException` as a response when all domain authentications/look ups have been exhausted despite possible issues with the LDAP connectivity. This was fine before when we simply used LDAP authentication as a means to grant access to VinylDNS, but we need to clean up the logic if we are going to enable user locking.

Changes in this pull request:
- Limit the scenarios in which `UserDoesNotExistException` is returned to make sure that the user truly does not exist

### DIFF
--- a/modules/portal/app/controllers/LdapAuthenticator.scala
+++ b/modules/portal/app/controllers/LdapAuthenticator.scala
@@ -219,8 +219,9 @@ class LdapAuthenticator(
               "Unable to successfully perform search in at least one LDAP domain"))
       case h :: t =>
         f(h).recoverWith {
-          case _: LdapServiceException => findUserDetails(t, userName, f, false)
-          case _ => findUserDetails(t, userName, f, allDomainConnectionsUp)
+          case _: UserDoesNotExistException =>
+            findUserDetails(t, userName, f, allDomainConnectionsUp)
+          case _ => findUserDetails(t, userName, f, false)
         }
     }
 
--- a/modules/portal/app/controllers/LdapAuthenticator.scala
+++ b/modules/portal/app/controllers/LdapAuthenticator.scala
@@ -113,8 +113,7 @@ object LdapAuthenticator {
     def searchContext(
         dirContext: DirContext,
         organization: String,
-        lookupUserName: String,
-        lookupFailureMessage: String): Either[LdapException, LdapUserDetails] =
+        lookupUserName: String): Either[LdapException, LdapUserDetails] =
       try {
         val searchControls = new SearchControls()
         searchControls.setSearchScope(2)
@@ -126,7 +125,7 @@ object LdapAuthenticator {
         )
 
         if (result.hasMore) Right(LdapUserDetails(result.next()))
-        else Left(UserDoesNotExistException(lookupFailureMessage))
+        else Left(UserDoesNotExistException(s"[$lookupUserName] LDAP entity does not exist"))
       } catch {
         case unexpectedError: Throwable => Left(LdapServiceException(unexpectedError.getMessage))
       } finally {
@@ -141,8 +140,7 @@ object LdapAuthenticator {
         searchContext(
           context,
           searchDomain.organization,
-          username,
-          s"[$username] can authenticate but LDAP entity does not exist")
+          username)
       }
 
     private[controllers] def lookup(
@@ -154,8 +152,7 @@ object LdapAuthenticator {
           searchContext(
             context,
             searchDomain.organization,
-            user,
-            s"[$user] LDAP entity does not exist")
+            user)
         }
   }
 
@@ -180,17 +177,13 @@ object LdapAuthenticator {
   }
 }
 
-trait LdapException extends Exception {
-  val message: String
-}
+sealed abstract class LdapException(message: String) extends Exception(message)
 
-case class UserDoesNotExistException(message: String) extends LdapException
-case class LdapServiceException(errorMessage: String) extends LdapException {
-  val message: String = s"Encountered error communicating with LDAP service: $errorMessage"
-}
-case class InvalidCredentials(username: String) extends LdapException {
-  val message: String = s"Provided credentials were invalid for user [$username]."
-}
+final case class UserDoesNotExistException(message: String) extends LdapException(message)
+final case class LdapServiceException(errorMessage: String) extends LdapException(
+  s"Encountered error communicating with LDAP service: $errorMessage")
+final case class InvalidCredentials(username: String) extends LdapException(
+  s"Provided credentials were invalid for user [$username].")
 
 /**
   * Top level ldap authenticator that tries authenticating on multiple domains. Authentication is
@@ -206,8 +199,9 @@ class LdapAuthenticator(
 
   /**
     * Attempts to search for user in specified LDAP domains. Attempts all LDAP domains that are specified in order; in
-    * the event that user details are not found in any of the domains, returns an error based on whether at least one
-    * LDAP search was successful (to distinguish between user not existing vs LDAP service being unreachable)
+    * the event that user details are not found in any of the domains, returns an error based on whether all
+    * LDAP search domains had successful connections (to distinguish between user not existing vs LDAP service being
+    * unreachable)
     *
     * @param domains List of domains in LDAP to lookup user
     * @param userName Username to lookup
--- a/modules/portal/app/controllers/LdapAuthenticator.scala
+++ b/modules/portal/app/controllers/LdapAuthenticator.scala
@@ -137,10 +137,7 @@ object LdapAuthenticator {
         username: String,
         password: String): Either[LdapException, LdapUserDetails] =
       createContext(s"${searchDomain.organization}\\$username", password).flatMap { context =>
-        searchContext(
-          context,
-          searchDomain.organization,
-          username)
+        searchContext(context, searchDomain.organization, username)
       }
 
     private[controllers] def lookup(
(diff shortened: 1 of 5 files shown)


modules/portal/app/controllers/LdapAuthenticator.scala
```diff
@@ -189,34 +204,57 @@ class LdapAuthenticator(
...
+      case Nil =>
+        if (ldapSearchSuccessful)
+          Left(UserDoesNotExistException(s"[$userName] LDAP entity does not exist"))
+        else
+          Left(
+            LdapServiceException(
+              "Unable to successfully perform search in at least one LDAP domain"))
+      case h :: t =>
+        f(h).recoverWith {
+          case _: LdapServiceException => findUserDetails(t, userName, f, ldapSearchSuccessful)
```
mitruly: `LdapServiceException` is the catch-all exception for something bad happening with the LDAP authentication.

modules/portal/app/controllers/LdapAuthenticator.scala
```diff
@@ -104,53 +110,52 @@ object LdapAuthenticator {
       .map(searchDomain ⇒ searchDomain.organization → searchDomain.domainName)
       .toMap
 
+    def searchContext(
```
mitruly: Abstracted from `authenticate` and `lookup` methods.

modules/portal/app/controllers/LdapAuthenticator.scala
```diff
@@ -104,53 +110,52 @@ object LdapAuthenticator {
...
+        dirContext: DirContext,
+        organization: String,
+        lookupUserName: String,
+        lookupFailureMessage: String): Either[LdapException, LdapUserDetails] =
```
pauljamescleary: Do we need to pass in the lookupFailure here, the two failure messages are very similar I think we can just bring them together.

modules/portal/app/controllers/LdapAuthenticator.scala
```diff
@@ -175,7 +180,17 @@ object LdapAuthenticator {
   }
 }
 
-class UserDoesNotExistException(message: String) extends Exception(message: String)
+trait LdapException extends Exception {
```
pauljamescleary: better to use `sealed abstract class LdapException(message: String) extends Exception(message)`

modules/portal/app/controllers/LdapAuthenticator.scala
```diff
@@ -175,7 +180,17 @@ object LdapAuthenticator {
...
+  val message: String
+}
+
+case class UserDoesNotExistException(message: String) extends LdapException
```
pauljamescleary: Maybe I didn't look far enough, but `case class` should always be marked `final` to ensure that pattern matches are total.

modules/portal/app/controllers/LdapAuthenticator.scala
```diff
@@ -189,34 +204,58 @@ class LdapAuthenticator(
     serviceAccount: ServiceAccount)
     extends Authenticator {
 
+  /**
+    * Attempts to search for user in specified LDAP domains. Attempts all LDAP domains that are specified in order; in
+    * the event that user details are not found in any of the domains, returns an error based on whether at least one
```
pauljamescleary: this comment is hard to follow "returns an error based on whether at least one LDAP search was successful"

modules/portal/app/controllers/LdapAuthenticator.scala
```diff
@@ -189,34 +204,58 @@ class LdapAuthenticator(
...
+    */
   private def findUserDetails(
       domains: List[LdapSearchDomain],
       userName: String,
-      f: LdapSearchDomain => Try[LdapUserDetails]): Try[LdapUserDetails] = domains match {
-    case Nil => Failure(new UserDoesNotExistException(s"[$userName] LDAP entity does not exist"))
-    case h :: t => f(h).recoverWith { case _ => findUserDetails(t, userName, f) }
-  }
+      f: LdapSearchDomain => Either[LdapException, LdapUserDetails],
+      allDomainConnectionsUp: Boolean): Either[LdapException, LdapUserDetails] =
```
pauljamescleary: what is `allDomainConnectionsUp` for?  You can do a pattern match on `case h :: Nil` on the last item in the list, and if that fails then you know that you have exhausted all domains

mitruly: @pauljamescleary `allDomainConnectionsUp` is to only return `UserDoesNotExistException` if all LDAP `DirContext`s succeed. Updated the comment to be clearer.

pauljamescleary: @mitruly right, I think that is what I was getting.  If you were to add a pattern match on `case h :: Nil` you would know that you were "at the end of the list"

```scala
domains match {
  case Nil =>
    // Catch just in case someone STARTS with an empty list
    Left(UserDoesNotExistException)
  case h :: Nil =>
    // We are on the last item in the list
    f(h).recoverWith(xxx)
  case h :: t =>
    // f(h). and do recursion
}
```

pauljamescleary: So, now that everything is an either, why can't we just scan through the list, if we find the user we exit, otherwise we keep trying?

mitruly: @pauljamescleary That makes sense - it'd be nice to get this cleaner if possible.

There were a few attributes that I was hoping to achieve with the updated `findUserDetails` behavior:
- If we initially encounter an `LdapServiceError` (ie. unexpected and non-permissible) error, we want to keep checking in the event that another domain search succeeds (existing behavior)
- If we exhaust all domain searches and didn't encounter any unexpected errors, we want to return `UserDoesNotExistException` (existing behavior)
- If we exhaust all domain searches and encountered _at least one_ unexpected error, we don't want to say the user doesn't exist; rather, we want to indicate that there was some sort of error in the lookup process (modified behavior)

pauljamescleary: Does this work...

```scala
domains.collectFirst { case d => f(d).toOption }.getOrElse(UserDoesNotExistException(""))
```

So, we keep trying each domain, as the result of `f` will be either a Left with a failure (which we don't really care about) or a Right with the actual user.

When we hit the first user, we would stop looking and return success

If we get to the end of the list, we would wind up with `None`, so we just say `UserDoesNotExist`

pauljamescleary: @mitruly I see, you want to not necessarily fall back to `UserDoesNotExistException` if there was a failure.

So, you want to stop as soon as you find the user OR you get an ldap service exception (no need to keep going as your one error may be the one that has the user).

```
domains.collectFirst { case d =>
  f(d) match {
    case LdapServiceException => Some(Left(LdapServiceException))
    case user => Some(Right(user))
    case _ => None
  }
}
```

A little sloppy, but you can still use `collectFirst`.  As soon as it yields a `Some` you stop

pauljamescleary: @mitruly I am good with your change after tinkering with it and understanding better what you are doing.

Going to resolve this conversation 👍

modules/portal/app/controllers/LdapAuthenticator.scala
```diff
@@ -189,34 +191,59 @@ class LdapAuthenticator(
...
 
-  def authenticate(username: String, password: String): Try[LdapUserDetails] =
-    findUserDetails(searchBase, username, authenticator.authenticate(_, username, password))
+  def authenticate(username: String, password: String): Either[LdapException, LdapUserDetails] =
+    findUserDetails(searchBase, username, authenticator.authenticate(_, username, password), true)
 
-  def lookup(username: String): Try[LdapUserDetails] =
-    findUserDetails(searchBase, username, authenticator.lookup(_, username, serviceAccount))
+  def lookup(username: String): Either[LdapException, LdapUserDetails] =
+    findUserDetails(searchBase, username, authenticator.lookup(_, username, serviceAccount), true)
```
rebstar6: is it possible to start up vinyl with an empty list for LDAP auths?

if so, this feels risky when you get to locking (and maybe handled in next pr, if so thats fine). My worry is that if you use oidc and this gets triggered with no ldap config info, all users will get locked.

this said, `true` is accurate here, i just wonder if we should have one more error that is strictly "NoLdapDomainsConfigured" for the non-configged case

modules/portal/app/controllers/LdapAuthenticator.scala
```diff
@@ -189,34 +191,59 @@ class LdapAuthenticator(
...
+        if (allDomainConnectionsUp)
+          Left(UserDoesNotExistException(s"[$userName] LDAP entity does not exist"))
+        else
+          Left(
+            LdapServiceException(
+              "Unable to successfully perform search in at least one LDAP domain"))
+      case h :: t =>
+        f(h).recoverWith {
+          case _: LdapServiceException => findUserDetails(t, userName, f, false)
+          case _ => findUserDetails(t, userName, f, allDomainConnectionsUp)
```
rebstar6: should the defaulting here not go the other way? seems to me like `allDomainConnectionsUp` should only stay true in the case that you get `UserDoesNotExistException` and fall to false in all others.

thinking about errors from config again, if your service account doesnt have access, createContext fails with InvalidCredentials...which has nothing to do with the user being looked up and shouldnt trigger locking

mitruly: @rebstar6 That's a good point; updated to only have `UserDoesNotExistException` maintain that all domain connections are up.

modules/portal/app/controllers/LdapAuthenticator.scala
mitruly: I can definitely handle this in https://github.com/vinyldns/vinyldns/pull/524

PR #677, opened 2019-06-04
Config disabling of multirecord batch updates

rebstar6:
in batch:
- disallow updates in batch that will update/delete an existing recordset with multiple recorddata fields
- disallow updates in batch that will create new recordsets with multiple recorddata fields

all of ^ config driven

### DIFF
--- a/modules/api/functional_test/live_tests/batch/create_batch_change_test.py
+++ b/modules/api/functional_test/live_tests/batch/create_batch_change_test.py
@@ -2891,3 +2891,102 @@ def test_create_batch_update_record_in_shared_zone_for_unassociated_user_in_owne
         if create_rs:
             delete_rs = shared_client.delete_recordset(shared_zone['id'], create_rs['recordSet']['id'], status=202)
             shared_client.wait_until_recordset_change_status(delete_rs, 'Complete')
+
+
+@pytest.mark.skip_production
+def test_create_batch_duplicates_add_check(shared_zone_test_context):
+    """
+    Test recordsets with multiple records cannot be added in batch (relies on config, skip-prod)
+    """
+    client = shared_zone_test_context.ok_vinyldns_client
+
+    batch_change_input = {
+        "comments": "this is optional",
+        "changes": [
+            get_change_A_AAAA_json("multi.ok.", address="1.2.3.4"),
+            get_change_A_AAAA_json("multi.ok.", address="4.5.6.7"),
+            get_change_PTR_json("192.0.2.44", ptrdname="multi.test"),
+            get_change_PTR_json("192.0.2.44", ptrdname="multi2.test"),
+            get_change_TXT_json("multi-txt.ok.", text="some-multi-text"),
+            get_change_TXT_json("multi-txt.ok.", text="more-multi-text"),
+            get_change_MX_json("multi-mx.ok.", preference=0),
+            get_change_MX_json("multi-mx.ok.", preference=1000, exchange="bar.foo."),
+
+            # adding an HVD so this will fail if accidentally run against wrong config
+            get_change_A_AAAA_json("high-value-domain")
+        ]
+    }
+
+    response = client.create_batch_change(batch_change_input, status=400)
+    def err(name, type):
+        return 'Multi-record recordsets are not enabled for this instance of VinylDNS. ' \
+               'There are duplicate changes with name {} and type {} in this batch.'.format(name, type)
+
+    assert_error(response[0], error_messages=[err("multi.ok.", "A")])
+    assert_error(response[1], error_messages=[err("multi.ok.", "A")])
+    assert_error(response[2], error_messages=[err("192.0.2.44", "PTR")])
+    assert_error(response[3], error_messages=[err("192.0.2.44", "PTR")])
+    assert_error(response[4], error_messages=[err("multi-txt.ok.", "TXT")])
+    assert_error(response[5], error_messages=[err("multi-txt.ok.", "TXT")])
+    assert_error(response[6], error_messages=[err("multi-mx.ok.", "MX")])
+    assert_error(response[7], error_messages=[err("multi-mx.ok.", "MX")])
+
+
+@pytest.mark.skip_production
+def test_create_batch_duplicates_update_check(shared_zone_test_context):
+    """
+    Test recordsets with multiple records cannot be edited in batch (relies on config, skip-prod)
+    """
+    client = shared_zone_test_context.ok_vinyldns_client
+    ok_zone = shared_zone_test_context.ok_zone
+
+    # record sets to setup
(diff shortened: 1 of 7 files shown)


modules/api/src/test/scala/vinyldns/api/domain/batch/BatchChangeValidationsSpec.scala
```diff
@@ -1493,4 +1496,145 @@ class BatchChangeValidationsSpec
 
     result(0) shouldBe valid
   }
+
```
rebstar6: not func tested, it relies on config, which isnt great

modules/api/src/main/scala/vinyldns/api/domain/DomainValidationErrors.scala
```diff
@@ -149,4 +150,20 @@ final case class MissingOwnerGroupId(recordName: String, zoneName: String)
...
+  def message: String =
+    s"""RecordSet with name $fqdn and type ${record.typ.toString} cannot be updated in a single Batch Change
+       |because it contains multiple DNS records (${record.records.length}).""".stripMargin
+      .replaceAll("\n", " ")
+}
+
+final case class NewMultiRecordError(change: AddChangeForValidation) extends DomainValidationError {
+  def message: String =
+    s"""Multi-record recordsets are not enabled for this instance of VinylDNS.
+       |There are duplicate changes with name ${change.inputChange.inputName} and type ${change.inputChange.typ}
```
pauljamescleary: "duplicate changes" here is confusing.

We are trying to say "Cannot create a new record set with multiple records for inputName and typ" I think

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
```diff
@@ -172,16 +175,36 @@ class BatchChangeValidations(changeLimit: Int, accessValidation: AccessValidatio
...
+      change: ChangeForValidation,
+      recordSet: RecordSet): SingleValidation[Unit] =
+    if (!multiRecordEnabled & recordSet.records.length > 1) {
+      ExistingMultiRecordError(change.inputChange.inputName, recordSet).invalidNel
+    } else ().validNel
+
+  def newRecordSetIsNotMulti(
+      change: AddChangeForValidation,
+      changeGroups: ChangeForValidationMap): SingleValidation[Unit] =
+    if (!multiRecordEnabled && changeGroups
```
pauljamescleary: having the `changeGroups.getList.collect.length` in an `if` is aggressive and hard to follow.

I would have that separately as a `val` so it is easier to follow

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
```diff
@@ -199,14 +222,17 @@ class BatchChangeValidations(changeLimit: Int, accessValidation: AccessValidatio
...
+      existingRecordSets.get(change.recordKey) match {
         case Some(rs) =>
           userCanUpdateRecordSet(change, auth, rs.ownerGroupId) |+|
             ownerGroupProvidedIfNeeded(
               change,
               existingRecordSets.get(change.zone.id, change.recordName, change.inputChange.typ),
-              batchOwnerGroupId)
-        case None => RecordDoesNotExist(change.inputChange.inputName).invalidNel
+              batchOwnerGroupId) |+|
+            newRecordSetIsNotMulti(change, changeGroups) |+|
```
pauljamescleary: kinda strange to put "multiRecordSet" checks inside `authAndOwnerGroupValidations`

Probably makes sense to create a new local val and then update the whole of validations below

modules/api/src/test/scala/vinyldns/api/domain/batch/BatchChangeValidationsSpec.scala
```diff
@@ -1493,4 +1496,166 @@ class BatchChangeValidationsSpec
...
+  property(
+    "validateChangesWithContext: succeed update/delete to a multi record existing RecordSet if multi enabled") {
+    val existing = List(
```
pauljamescleary: More of an observation here.  We seem to be doing multiple tests in one (if I follow).  Like here we are attempting an update of a multi-record and a delete of the multi-record

The test setup and assertions become difficult.  I was wondering "why are there so many existing recordsets"?

If the test scope was narrower "succeed delete to a multi-record record set if multi enabled", then the existing should just be a single multi-record record set, and then the actual test should be a single DeleteRecordSet change.  The assertion would be a single test.

With all the things going on, the test becomes hard to follow.

There is likely a lot of precedence here and 1500 lines :scream: , just something to keep in mind that unit tests should as best as possible test one thing.

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
rebstar6: moved out anyway, i's typed really, cname already has its own non-configged duplicate check

modules/api/src/test/scala/vinyldns/api/domain/batch/BatchChangeValidationsSpec.scala
rebstar6: i know, its a balance. this test file is already a PITA because you cant nest the property tests (we should eventually make this a wordspec methinks.

i kinda like the grouped tests so long as it's one concept 🤷‍♀

modules/api/functional_test/live_tests/batch/create_batch_change_test.py
```diff
@@ -2947,3 +2891,102 @@ def test_create_batch_update_record_in_shared_zone_for_unassociated_user_in_owne
         if create_rs:
             delete_rs = shared_client.delete_recordset(shared_zone['id'], create_rs['recordSet']['id'], status=202)
             shared_client.wait_until_recordset_change_status(delete_rs, 'Complete')
+
+
+@pytest.mark.skip_production
```
mitruly: Should we run these tests in all environments if no record changes get made?

rebstar6: nah, depending on whats done with the config value, these will fail. should be left off and they can be manually run if wanted

britneywright:
should add to the API configuration guide: https://www.vinyldns.io/operator/config-api
and the batch change information: https://www.vinyldns.io/api/batchchange-model.html#batch-change-information-

modules/api/functional_test/live_tests/batch/create_batch_change_test.py
mitruly: Ooh, that's right.

rebstar6:
@britneywright making a docs PR right now as followup

PR #678, opened 2019-06-04
validate batch change review rejection

britneywright:
Rebased off #665 

Changes in this pull request:
- Adds validations to the reject batch change endpoint to check if a batch change is PendingReview and if the reviewer is a super user or support user

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeErrors.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeErrors.scala
@@ -26,9 +26,6 @@ sealed trait BatchChangeErrorResponse
 final case class InvalidBatchChangeInput(errors: List[DomainValidationError])
     extends BatchChangeErrorResponse
 
-final case class InvalidBatchChangeReview(errors: List[BatchChangeErrorResponse])
-    extends BatchChangeErrorResponse
-
 // This separates error by change requested
 final case class InvalidBatchChangeResponses(
     changeRequests: List[ChangeInput],
--- a/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeErrors.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeErrors.scala
@@ -26,6 +26,9 @@ sealed trait BatchChangeErrorResponse
 final case class InvalidBatchChangeInput(errors: List[DomainValidationError])
     extends BatchChangeErrorResponse
 
+final case class InvalidBatchChangeReview(errors: List[BatchChangeErrorResponse])
+    extends BatchChangeErrorResponse
+
 // This separates error by change requested
 final case class InvalidBatchChangeResponses(
     changeRequests: List[ChangeInput],
@@ -44,3 +47,8 @@ final case class BatchConversionError(change: SingleChange) extends BatchChangeE
        |and type "${change.typ}"""".stripMargin
 }
 final case class UnknownConversionError(message: String) extends BatchChangeErrorResponse
+
+final case class BatchChangeNotPendingApproval(id: String) extends BatchChangeErrorResponse {
+  def message: String =
+    s"""Batch change $id is not pending approval, so it cannot be rejected."""
+}
--- a/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeInterfaces.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeInterfaces.scala
@@ -27,6 +27,7 @@ object BatchChangeInterfaces {
   type SingleValidation[A] = ValidatedNel[DomainValidationError, A]
   type ValidatedBatch[A] = List[ValidatedNel[DomainValidationError, A]]
   type BatchResult[A] = EitherT[IO, BatchChangeErrorResponse, A]
+  type BatchApproval[A] = ValidatedNel[BatchChangeErrorResponse, A]
 
   implicit class IOBatchResultImprovements[A](theIo: IO[A]) {
     def toBatchResult: BatchResult[A] = EitherT.liftF(theIo)
@@ -70,6 +71,11 @@ object BatchChangeInterfaces {
 
   implicit class SingleValidationImprovements[A](validation: SingleValidation[A]) {
     def asUnit: SingleValidation[Unit] =
+      validation.map(_ => ())
+  }
+
+  implicit class BatchApprovalImprovements[A](validation: BatchApproval[A]) {
+    def asUnit: BatchApproval[Unit] =
       validation.map(_ => ())
   }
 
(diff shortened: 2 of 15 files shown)


britneywright:
I propose removing the func test for the reject batch change endpoint since we don't test any support or super user functionality we'd only be able to test that a regular user attempting to reject a batch change fails.

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeErrors.scala
```diff
@@ -26,6 +26,9 @@ sealed trait BatchChangeErrorResponse
 final case class InvalidBatchChangeInput(errors: List[DomainValidationError])
     extends BatchChangeErrorResponse
 
+final case class InvalidBatchChangeReview(errors: List[BatchChangeErrorResponse])
```
mitruly: This is a bit weird to have `InvalidBatchChangeReview` both take in a collection of `BatchChangeErrorResponse` and extend `BatchChangeErrorResponse`; should probably take in a `List[DomainValidationError]` if we care about returning both possible errors at once.

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeInterfaces.scala
```diff
@@ -27,6 +27,7 @@ object BatchChangeInterfaces {
   type SingleValidation[A] = ValidatedNel[DomainValidationError, A]
   type ValidatedBatch[A] = List[ValidatedNel[DomainValidationError, A]]
   type BatchResult[A] = EitherT[IO, BatchChangeErrorResponse, A]
+  type BatchApproval[A] = ValidatedNel[BatchChangeErrorResponse, A]
```
mitruly: Don't think that this or the `implicit class` are needed based on other comments and don't see usage in code.

britneywright:
this is passing in Travis, not sure why it's showing as Pending here.

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
```diff
@@ -91,6 +95,28 @@ class BatchChangeValidations(changeLimit: Int, accessValidation: AccessValidatio
         else NotAMemberOfOwnerGroup(groupId, authPrincipal.signedInUser.userName).invalidNel
     }
 
+  def validateRejectedBatchChange(
```
mitruly: Nitpick: should rename this to something like `validateRejectBatchChangeInput`; `validateRejectedBatchChange` sounds like we are validating a batch change that has already been rejected.

britneywright: I could name it validateRejectBatchChangeAttempt perhaps? validateRejectBatchChangeInput would be inaccurate because the RejectBatchChangeInput is not what is being validated, it’s the batch change and the user attempting to reject it.

mitruly: Maybe `validateBatchChangeRejection` worked better here? We don't want to get too verbose and want the naming to not be too confusing, either.

britneywright: works for me, will update

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -85,6 +85,15 @@ class BatchChangeService(
         batchChangeInput.ownerGroupId)
     } yield conversionResult.batchChange
 
+  def rejectBatchChange(
+      batchChangeId: String,
+      authPrincipal: AuthPrincipal,
+      rejectionComment: Option[RejectBatchChangeInput]): BatchResult[BatchChange] =
```
mitruly: Should this be `reviewComment` to maintain consistency with the `batch_change` SQL changes?

britneywright: i don't think so, reviewComment is a parameter of this, open to other suggestions though.

mitruly: Right, the `RejectBatchChangeInput`'s input (which is also used for the JSON deserialization) was changed from `comment` to `reviewComment`; not really a big deal but inconsistent in not keeping the name throughout (and I'm fine as long as others are).

On another note, should either:
- change `rejectionComment` to an `Option[String]` here and pass in `input.flatMap(_.reviewComment)`
- change the name to something else that indicates that it is the payload and not just the comment (in the event that we extend the input model to have include more than just an optional comment)

britneywright: i changed the name to rejectBatchChangeInput, same as what we do for authPrincipal

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
```diff
@@ -94,6 +98,28 @@ class BatchChangeValidations(
...
+      batchChange: BatchChange): Either[BatchChangeErrorResponse, Unit] =
+    batchChange.approvalStatus match {
+      case BatchChangeApprovalStatus.PendingApproval => ().asRight
+      case _ => BatchChangeNotPendingApproval(batchChange.id).asLeft
+    }
+
+  def validateAuthorizedReviewer(
+      auth: AuthPrincipal,
+      batchChange: BatchChange): Either[BatchChangeErrorResponse, Unit] =
+    if (auth.canReadAll) {
```
nimaeskandary: `canReadAll` doesn't fit this context, can that be refactored to something like `isSystemAdmin`

modules/api/src/main/scala/vinyldns/api/route/BatchChangeRouting.scala
```diff
@@ -68,12 +68,13 @@ trait BatchChangeRoute extends Directives {
       }
 
     val manualBatchReviewRoutes =
-      (post & path("zones" / "batchrecordchanges" / Segment / "reject")) { _ =>
+      (post & path("zones" / "batchrecordchanges" / Segment / "reject")) { id =>
         monitor("Endpoint.rejectBatchChange") {
           entity(as[Option[RejectBatchChangeInput]]) { input =>
-            // TODO: Tie into batch change service with auth validation and rejection process
```
nimaeskandary: That second half of the TODO on rejection process still applies, can that be left?

modules/api/src/main/scala/vinyldns/api/route/BatchChangeRouting.scala
```diff
@@ -99,6 +100,8 @@ trait BatchChangeRoute extends Directives {
       case Left(cnf: BatchChangeNotFound) => complete(StatusCodes.NotFound, cnf.message)
       case Left(una: UserNotAuthorizedError) => complete(StatusCodes.Forbidden, una.message)
       case Left(uct: BatchConversionError) => complete(StatusCodes.BadRequest, uct)
+      case Left(bcnpa: BatchChangeNotPendingApproval) =>
```
nimaeskandary: interesting precedent, weird var names

modules/api/src/main/scala/vinyldns/api/route/BatchChangeRouting.scala
britneywright: there's nothing left to here for the process.

britneywright:
@rebstar6 can I get a review on this one.

britneywright:
this is passing in Travis, not sure why Github is showing Pending, so I'll merge.

PR #679, opened 2019-06-04
Better sync logging

pauljamescleary:
* Cleaning up logging around sync to make it easier to query
* Added a log statement in the main health check to better inspect health check failures

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/engine/ZoneSyncHandler.scala
+++ b/modules/api/src/main/scala/vinyldns/api/engine/ZoneSyncHandler.scala
@@ -108,8 +108,8 @@ object ZoneSyncHandler extends DnsConversions with Monitored {
             changesWithUserIds
               .filter { chg =>
                 chg.recordSet.name != zone.name && chg.recordSet.name.contains(".") &&
-                chg.recordSet.typ != RecordType.SRV && chg.recordSet.typ != RecordType.TXT &&
-                chg.recordSet.typ != RecordType.NAPTR
+                  chg.recordSet.typ != RecordType.SRV && chg.recordSet.typ != RecordType.TXT &&
+                  chg.recordSet.typ != RecordType.NAPTR
               }
               .map(_.recordSet.name)
               .grouped(1000)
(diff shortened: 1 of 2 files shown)


modules/api/src/main/scala/vinyldns/api/engine/ZoneSyncHandler.scala
```diff
@@ -83,11 +83,12 @@ object ZoneSyncHandler extends DnsConversions with Monitored {
       vinyldnsLoader: (Zone, RecordSetRepository) => VinylDNSZoneViewLoader =
         VinylDNSZoneViewLoader.apply): IO[ZoneChange] =
     monitor("zone.sync") {
-      time(s"zone.sync(${zoneChange.zoneId})") {
+      time(s"zone.sync; zoneName='${zoneChange.zone.name}'") {
```
rebstar6: can you throw the change ID up here somewhere too?

modules/api/src/main/scala/vinyldns/api/engine/ZoneSyncHandler.scala
```diff
@@ -142,7 +145,7 @@ object ZoneSyncHandler extends DnsConversions with Monitored {
     }.attempt
       .map {
         case Left(e: Throwable) =>
-          logger.error(s"Encountered error syncing zone ${zoneChange.zone.name}", e)
+          logger.error(s"Encountered error syncing ; zoneName='${zoneChange.zone.name}'", e)
```
rebstar6: change id in this one too

modules/core/src/main/scala/vinyldns/core/health/HealthCheck.scala
```diff
@@ -17,17 +17,21 @@
...
 import cats.effect.IO
+import org.slf4j.LoggerFactory
 
 object HealthCheck {
 
   type HealthCheck = IO[Either[HealthCheckError, Unit]]
 
   case class HealthCheckError(message: String) extends Throwable(message)
 
+  private val logger = LoggerFactory.getLogger("HealthCheck")
```
rebstar6: not for now, but i pulled in https://github.com/lightbend/scala-logging to the syncer and its nicer on logger creation. you just `val logger = Logger("name")` and go on your way

rebstar6: for now - generally dont we name loggers by full path? `LoggerFactory.getLogger("vinyldns.core.health.HealthCheck")`

pauljamescleary: We have to somehow differentiate between an `object` and a `class` for logging :/

I have traditionally omitted the full class name for `object`.  I will double check but we should follow that precedent here

pauljamescleary: Yep, like boot `private val logger = LoggerFactory.getLogger("Boot")`

rebstar6: ok thats fine then :)

PR #681, opened 2019-06-10
Reverse zone discovery bug

rebstar6:
bug fix.

in short, zone discovery for ipv4 relies on `endswith` to address delegated zone challenges. As a result, if you are looking for the zone for `1.2.3.0`, that will be transformed into a base of `3.2.1.in-addr.arpa.`, and we get all ending with that (so `3.2.1.in-addr.arpa.` and `0/30.3.2.1.in-addr.arpa.` (good), AND`13.2.1.in-addr.arpa.` (bad!))

this fixes that problem

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/ReverseZoneHelpers.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/ReverseZoneHelpers.scala
@@ -47,8 +47,9 @@ object ReverseZoneHelpers {
     }
   }
 
-  // NOTE: this function assumes record/zone interface as the record name is strictly the base string within
-  // zone context
+  // NOTE: this function assumes record/zone interface. Unless prior checks have been made, it should not
+  // be used to check if an FQDN is within a reverse zone unless we know the rest of the zone matches the higher
+  // octets
   def ptrIsInClasslessDelegatedZone(zone: Zone, recordName: String): Either[Throwable, Unit] =
     if (zone.isIPv4) {
       handleIpv4RecordValidation(zone: Zone, recordName)
--- a/modules/api/src/main/scala/vinyldns/api/domain/ReverseZoneHelpers.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/ReverseZoneHelpers.scala
@@ -34,6 +34,7 @@ object ReverseZoneHelpers {
       Try(CIDR.valueOf(mask).contains(ipAddr)).getOrElse(false)
     }
 
+  // NOTE: this will not work for zones with less than 3 octets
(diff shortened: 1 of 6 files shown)


rebstar6:
hmm, still missing something here, closing, though likely will reopen

modules/api/src/main/scala/vinyldns/api/domain/ReverseZoneHelpers.scala
```diff
@@ -35,19 +34,30 @@ object ReverseZoneHelpers {
...
-          handleIpv4RecordValidation(zone: Zone, recordName)
-        } else if (zone.isIPv6) {
-          handleIpv6RecordValidation(zone: Zone, recordName)
-        } else {
-          InvalidRequest(
-            s"RecordSet $recordName does not specify a valid IP address in zone ${zone.name}").asLeft
-        }
-      }
-      case _ => ().asRight
+  // NOTE: this will not work for zones with less than 3 octets
```
rebstar6: this note is a pain, though this is true across our entire batch interface. if we want to expand to support 2 octet zones, that will be a larger task throughout

modules/api/src/test/scala/vinyldns/api/domain/ReverseZoneHelpersSpec.scala
```diff
@@ -240,68 +240,93 @@ class ReverseZoneHelpersSpec
...
           error shouldBe a[InvalidRequest]
         }
       }
     }
+    "ipIsInIpv4ReverseZone" should {
+      "return true for the base zone" in {
+        val zone = Zone("1.2.3.in-addr.arpa.", "email")
+        ReverseZoneHelpers.ipIsInIpv4ReverseZone(zone, "3.2.1.10") shouldBe true
+      }
+      "return false for a zone ending in the base zone" in {
```
nimaeskandary: zone `not` ending in the base zone?

rebstar6: oops

rebstar6: actually no oops. this is ending in the base zone, as in 21.2.3.in-addr.arpa. endswith 1.2.3.in-addr.arpa., but this is a bad match.

this is the exact case this bug is fixing

nimaeskandary: I see, was reading it assuming an implicit dot in `.1.2.3.in-addr.arpa.`

PR #683, opened 2019-06-11
sync-banned zones list

rebstar6:
allows you to configure a list of zones that we will disallow syncs in. At the moment, this does not stop the sync that occurs on initial connection

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
+++ b/modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
@@ -27,6 +27,7 @@ import net.ceedubs.ficus.Ficus._
 import net.ceedubs.ficus.readers.EnumerationReader._
 import vinyldns.api.domain.zone.ZoneRecordValidations
 import vinyldns.core.domain.record.RecordType
+import vinyldns.core.domain.DomainHelpers.ensureTrailingDot
 
 import scala.collection.JavaConverters._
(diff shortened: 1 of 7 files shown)


modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
```diff
@@ -79,6 +80,9 @@ object VinylDNSConfig {
   lazy val highValueIpList: List[IpAddress] =
     getOptionalStringList("high-value-domains.ip-list").flatMap(ip => IpAddress(ip))
 
+  lazy val syncBannedZones: List[String] =
```
britneywright: can we call this something like `disabledSyncList`

britneywright:
can we use the word "disabled" rather than "banned"

modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
```diff
@@ -79,6 +80,9 @@ object VinylDNSConfig {
...
+    getOptionalStringList("sync-banned-zones").map(zn => ensureTrailingDot(zn.toLowerCase()))
```
britneywright: "disabled-sync-list" sounds better, "sync-banned-zones" sounds like we are syncing zones that are banned, rather than not allowing specific zones to be synced.

rebstar6:
closing for now

PR #685, opened 2019-06-12
update 'Zone already exists' verbiage

### LINKED ISSUE #79: Update verbiage and docs around "Zone already exists"
**Motivation**
Presently, when a user attempts to connect to a new zone that already exists in VinylDNS, we return a message with "Zone already exists."  This is confusing to end users.

**Recommendation**
1. Update the message returned from the API server to say "The zone xxx is already registered in VinylDNS.  

If the user just does not have access, add to the return message...

```
The zone admin group is yyy.  Please contact <group name> <group email> for access to the zone.
```

It is possible that the user does have access, just is not in view.  If so, then add to the return message...

```
The zone can be found by entering <zone name here> in the search field.
```

2. Update the FAQ with information on what to do when that message is received.

britneywright:
Fixes #79.

Changes in this pull request:
- add "Please contact <zone email> for access to the zone." to error message when a user tries to connect to a zone that already exists in VinylDNS.

 I don't think we've used zone email for anything to this point, it's there so we should use it. If it's not as useful as the admin group name and admin group email we could change to that in the future, but this seems like it meets the objective of putting contact information in the message in the least complex manner.

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneService.scala
@@ -221,7 +221,7 @@ class ZoneService(
         case Some(existingZone) if existingZone.status != ZoneStatus.Deleted =>
           ZoneAlreadyExistsError(
             s"Zone with name $zoneName already exists. " +
-              s"Please contact ${existingZone.email} for access to the zone.").asLeft
+              s"Please contact ${existingZone.email} to request access to the zone.").asLeft
         case _ => ().asRight
       }
       .toResult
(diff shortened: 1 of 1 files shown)


modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneService.scala
```diff
@@ -219,7 +219,9 @@ class ZoneService(
       .getZoneByName(zoneName)
       .map {
         case Some(existingZone) if existingZone.status != ZoneStatus.Deleted =>
-          ZoneAlreadyExistsError(s"Zone with name $zoneName already exists").asLeft
+          ZoneAlreadyExistsError(
+            s"Zone with name $zoneName already exists. " +
+              s"Please contact ${existingZone.email} for access to the zone.").asLeft
```
mitruly: Does `Please contact <email> to request access to the zone` flow better? Contacting technically doesn't guarantee that access will be granted, heh

britneywright: updated

Issue #642, opened 2019-05-22
Allow all users to see any existing group


britneywright:
**Is your feature request related to a problem? Please describe.**
We want to open up read-only access of data in VinylDNS. This scope of this card is limited to individual groups. A separate card will be created for the list groups endpoint.

**Describe the solution you'd like**
Allow any user to see any existing group they request.

They should be able to see the specified group as well as the members and admins, but they should not be able to update or delete the group.

**Additional context**
I believe the endpoints this effects are:
- get group
- list group admins
- list group members
- list group activity

we may want to include and return an access level check that notes if the requesting user is a super user, group admin, group member, or none.

nspadaccino:
Duplicate of https://github.com/vinyldns/vinyldns/issues/1088
Resolved by https://github.com/vinyldns/vinyldns/pull/1097

PR #689, opened 2019-06-14
Expand Zone access

britneywright:
With the addition of shared zones, some users are getting confused about what they should expect to see in the Zones section of VinylDNS. They also aren't sure how to know if a shared zone they want to work with exists in VinylDNS. Currently the "zones" API endpoint and Zones page in the portal only return zones a user is in the owner group for or has an ACL rule for. We return all zones to super and support users. We should make clear to users what zones they own but also allow them to see what zones exist in VinylDNS. This PR does not change their ability to access the specific zones though.

Changes in this pull request:
**API:**
- add `listAll` boolean parameter to the listZones API endpoint. Default is false. 
- update `canSeeZone` method in `AccessValidations.scala` to include shared zones
- add `accessLevel` to ZoneInfo and ZoneSummaryInfo, used `canSeeZone` and `canChangeZone` methods to determine a user's access level.
- users can see all zones in the listZones response but can only access shared zones, and private zones they own or have access to through ACL rules.

**Portal:**
- in the portal add tabs in the Zones page, one for "My Zones" (existing "zones" API call) and one for "All Zones" ("zones" API call with "listAll" set to true)
- change the ""No zones found." message in the portal to "No zones found. You do not own any zones in VinylDNS. You can manage records in shared zones through Batch Changes."
- Use "accessLevel" to determine what zones in the All Zones tab have a "View" button. Must have accessLevel Delete or Read.
- Use "accessLevel" to determine which zones show the "Manage Zone" tab. Only accessLevel Delete.
-  Use "accessLevel" to determine if "Sync Zone" button shows
- Use "accessLevel" and zone shared status to determine if "Create Recordset" button shows.

**note:** This does not change logic around super and support users, although i propose we do stop returning all zones by default for those users so they can distinguish what they manage personally vs what they manage in a system administrator capacity.

Screen Shot 2019-06-19 at 12 00 53 PM
Screen Shot 2019-06-19 at 12 00 45 PM

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/AccessValidations.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/AccessValidations.scala
@@ -126,12 +126,6 @@ object AccessValidations extends AccessValidationAlgebra {
       }
     }
 
-  def getZonesAccess(auth: AuthPrincipal, zones: List[Zone]): List[ZoneSummaryInfo] =
-    zones.map { zn =>
-      val accessLevel = getZoneAccess(auth, zn)
-      ZoneSummaryInfo(zn, "", accessLevel)
-    }
-
   def getZoneAccess(auth: AuthPrincipal, zone: Zone): AccessLevel =
     if (canChangeZone(auth, zone.name, zone.adminGroupId).isRight)
       AccessLevel.Delete
--- a/modules/api/src/main/scala/vinyldns/api/domain/AccessValidations.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/AccessValidations.scala
@@ -30,7 +30,7 @@ object AccessValidations extends AccessValidationAlgebra {
   def canSeeZone(auth: AuthPrincipal, zone: Zone): Either[Throwable, Unit] =
     ensuring(
       NotAuthorizedError(s"User ${auth.signedInUser.userName} cannot access zone '${zone.name}'"))(
-      zone.shared || auth.isSystemAdmin || auth
+      auth.isSystemAdmin || auth
         .isGroupMember(zone.adminGroupId) || userHasAclRules(auth, zone))
 
   def canChangeZone(
@@ -132,13 +132,12 @@ object AccessValidations extends AccessValidationAlgebra {
       ZoneSummaryInfo(zn, "", accessLevel)
     }
 
-  def getZoneAccess(auth: AuthPrincipal, zone: Zone): AccessLevel = {
+  def getZoneAccess(auth: AuthPrincipal, zone: Zone): AccessLevel =
     if (canChangeZone(auth, zone.name, zone.adminGroupId).isRight)
       AccessLevel.Delete
     else if (canSeeZone(auth, zone).isRight)
       AccessLevel.Read
     else AccessLevel.NoAccess
-  }
 
   /* Non-algebra methods */
   def getAccessFromAcl(
--- a/modules/api/src/main/scala/vinyldns/api/domain/AccessValidations.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/AccessValidations.scala
@@ -133,12 +133,11 @@ object AccessValidations extends AccessValidationAlgebra {
     }
 
   def getZoneAccess(auth: AuthPrincipal, zone: Zone): AccessLevel = {
-    var accessLevel = AccessLevel.NoAccess
     if (canChangeZone(auth, zone.name, zone.adminGroupId).isRight)
-      accessLevel = AccessLevel.Delete
+      AccessLevel.Delete
     else if (canSeeZone(auth, zone).isRight)
-      accessLevel = AccessLevel.Read
-    accessLevel
+      AccessLevel.Read
+    else AccessLevel.NoAccess
   }
 
   /* Non-algebra methods */
--- a/modules/api/src/main/scala/vinyldns/api/domain/AccessValidations.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/AccessValidations.scala
@@ -98,22 +98,8 @@ object AccessValidations extends AccessValidationAlgebra {
     if (auth.isSuper || auth.isGroupMember(zone.adminGroupId))
       recordSets.map(RecordSetListInfo(_, AccessLevel.Delete))
     else {
-      val rulesForUser = zone.acl.rules.filter(ruleAppliesToUser(auth, _))
-
-      def getAccessFromUserAcls(recordName: String, recordType: RecordType): AccessLevel = {
-        // user filter has already been applied
-        val validRules = rulesForUser.filter { rule =>
-          ruleAppliesToRecordType(recordType, rule) && ruleAppliesToRecordName(
-            recordName,
-            recordType,
-            zone,
-            rule)
-        }
-        getPrioritizedAccessLevel(recordType, validRules)
-      }
-
       recordSets.map { rs =>
-        val aclAccessLevel = getAccessFromUserAcls(rs.name, rs.typ)
+        val aclAccessLevel = getAccessFromAcl(auth, rs.name, rs.typ, zone)
         val accessLevel = {
           if ((aclAccessLevel == AccessLevel.NoAccess) && auth.isSystemAdmin)
             AccessLevel.Read
--- a/modules/api/src/main/scala/vinyldns/api/domain/AccessValidations.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/AccessValidations.scala
@@ -30,7 +30,7 @@ object AccessValidations extends AccessValidationAlgebra {
   def canSeeZone(auth: AuthPrincipal, zone: Zone): Either[Throwable, Unit] =
     ensuring(
       NotAuthorizedError(s"User ${auth.signedInUser.userName} cannot access zone '${zone.name}'"))(
-      auth.isSystemAdmin || auth.isGroupMember(zone.adminGroupId) || userHasAclRules(auth, zone))
+      zone.shared || auth.isSystemAdmin || auth.isGroupMember(zone.adminGroupId) || userHasAclRules(auth, zone))
 
   def canChangeZone(
       auth: AuthPrincipal,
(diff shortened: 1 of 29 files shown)


modules/portal/public/lib/controllers/controller.zones.js
```diff
@@ -20,12 +20,14 @@ angular.module('controller.zones', [])
...
     $scope.alerts = [];
     $scope.zonesLoaded = false;
+    $scope.allZonesLoaded = false;
     $scope.hasZones = false; // Re-assigned each time zones are fetched without a query
 
     $scope.query = "";
 
     // Paging status for zone sets
     var zonesPaging = pagingService.getNewPagingParams(100);
+    var allZonesPaging = pagingService.getNewPagingParams(100);
```
mitruly: I experienced strange behavior with paging where the **Previous** and **Next** buttons would appear but not page as expected.

Tested by connecting to two different zones using two different users on a local instance of VinylDNS and setting paging params to `1`. Paging would work correctly on `My Zones` tab, but not the new `All Zones` tab.

modules/portal/public/lib/controllers/controller.zones.js
```diff
@@ -66,37 +68,64 @@ angular.module('controller.zones', [])
...
-            zonesPaging.next = response.data.nextId;
-            updateZoneDisplay(response.data.zones);
-            if (!$scope.query.length) {
-                $scope.hasZones = response.data.zones.length > 0;
-            }
-        }
+        allZonesPaging = pagingService.resetPaging(allZonesPaging);
 
-        return zonesService
+        zonesService
```
mitruly: This seems to be a valid use case for having two service calls since we're not guaranteed that zones that the user has access to will be returned when grabbing all zones.

britneywright:
with what I've got currently there would be issues with checking zone access in the "all zones" tab in the portal if the user has access to more than 100 zones. It's only checking the zones in "All Zones" against the ID's of the first page of results in "My Zones". have to chew on this.

modules/api/src/main/scala/vinyldns/api/domain/AccessValidations.scala
```diff
@@ -116,6 +117,9 @@ object AccessValidations extends AccessValidationAlgebra {
         val accessLevel = {
           if ((aclAccessLevel == AccessLevel.NoAccess) && auth.canReadAll)
             AccessLevel.Read
+          else if (zone.shared && (rs.ownerGroupId == None || auth.isGroupMember(
```
mitruly: A more concise form that is functionally the same:
```
          else if (zone.shared && rs.ownerGroupId.forall(auth.isGroupMember))
            AccessLevel.Delete
```

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneService.scala
```diff
@@ -120,43 +121,65 @@ class ZoneService(
...
 
-  def zoneAdminGroupMapping(zones: List[Zone], groups: Set[Group]): List[ZoneSummaryInfo] =
+  def getZonesAccess(auth: AuthPrincipal, zones: List[Zone]): List[ZoneSummaryInfo] =
+    zones.map { zn =>
+      var accessLevel = AccessLevel.NoAccess
+      if (canChangeZone(auth, zn.name, zn.adminGroupId).isRight)
+        accessLevel = AccessLevel.Delete
+      else if (canSeeZone(auth, zn).isRight)
+        accessLevel = AccessLevel.Read
+      ZoneSummaryInfo(zn, "", accessLevel)
```
mitruly: Maybe should pass in the zone admin group name instead of having an empty string here.

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneService.scala
```diff
@@ -242,6 +265,15 @@ class ZoneService(
     }
   }.toResult
 
+  def getZoneAccess(auth: AuthPrincipal, zone: Zone): Result[AccessLevel] = {
```
mitruly: This should probably get moved to `AccessValidations`. Can also have this just return an `AccessLevel` and have `def getZonesAccess(auth: AuthPrincipal, zones: List[Zone])` above call this.

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneService.scala
britneywright: at this point we don't have the zone admin group name, that happens next in `zoneAdminGroupMapping`

pauljamescleary:
Is the plan to make Groups read only as well, so users can link to the admin group for the zone to contact them?

britneywright:
yes, there's a separate issue for that https://github.com/vinyldns/vinyldns/issues/642.

pauljamescleary:
We should probably show the `Connect Zone` button on `All Zones` as it contains both mine and everything else.

pauljamescleary:
I don't think we want to hide everything in `Manage Zone`.  Perhaps we should rename the tabs and just make that one read only.

There is some useful information in there (admin group, status, id, last sync time, created).  Perhaps show that info as read only and hide the connection info and acl rules.

modules/api/src/main/scala/vinyldns/api/domain/AccessValidations.scala
```diff
@@ -116,13 +117,30 @@ object AccessValidations extends AccessValidationAlgebra {
...
     }
 
+  def getZonesAccess(auth: AuthPrincipal, zones: List[Zone]): List[ZoneSummaryInfo] =
+    zones.map { zn =>
+      val accessLevel = getZoneAccess(auth, zn)
+      ZoneSummaryInfo(zn, "", accessLevel)
+    }
+
+  def getZoneAccess(auth: AuthPrincipal, zone: Zone): AccessLevel = {
+    var accessLevel = AccessLevel.NoAccess
```
pauljamescleary: you can add an `else` and remove the `var` here to avoid the extra variable

```scala
if (canChangeZone) AccessLevel.Delete
else if (canSeeZone) AccessLevel.Read
else AccessLevel.NoAccess
```

britneywright:
pulled this back a bit. Originally added shared zones to the the `canSeeZone` method, removed that at this time. So users will still see the shared zones in the listZones response, but as it is currently they still will not have access to shared zones directly via the `getZone` or `getZoneByName` API calls. We'll discuss in the future what users should and shouldn't have access to in a zone before opening that up more.

modules/api/src/main/scala/vinyldns/api/domain/AccessValidations.scala
```diff
@@ -30,7 +30,8 @@ object AccessValidations extends AccessValidationAlgebra {
   def canSeeZone(auth: AuthPrincipal, zone: Zone): Either[Throwable, Unit] =
     ensuring(
       NotAuthorizedError(s"User ${auth.signedInUser.userName} cannot access zone '${zone.name}'"))(
-      auth.isSystemAdmin || auth.isGroupMember(zone.adminGroupId) || userHasAclRules(auth, zone))
+      auth.isSystemAdmin || auth
+        .isGroupMember(zone.adminGroupId) || userHasAclRules(auth, zone))
```
mitruly: `userHasAclRules` doesn't work as intended since it will return `true` if you create a `NoAccess` ACL rule for a user who doesn't otherwise have access to the zone. We probably want to use `getAccessFromAcl` and check that it is not `NoAccess`.

modules/api/src/main/scala/vinyldns/api/domain/AccessValidations.scala
```diff
@@ -116,13 +117,28 @@ object AccessValidations extends AccessValidationAlgebra {
...
+          else if (zone.shared && rs.ownerGroupId.forall(auth.isGroupMember))
+            AccessLevel.Delete
           else
             aclAccessLevel
         }
         RecordSetListInfo(rs, accessLevel)
       }
     }
 
+  def getZonesAccess(auth: AuthPrincipal, zones: List[Zone]): List[ZoneSummaryInfo] =
```
mitruly: I think we should get rid of this helper since: 1. it is a helper function that is returning a list of incomplete `ZoneSummaryInfo` (in that it doesn't have the owner group info, which is added later on in the service), and 2. it is calling `getZoneAccess`, which is actually retrieving the zone access.

We can just have `zoneAdminGroupMapping` first call `getZoneAccess`, then find the owner group name and return a complete `ZoneSummaryInfo`.

modules/api/src/main/scala/vinyldns/api/domain/AccessValidations.scala
```diff
@@ -116,13 +117,28 @@ object AccessValidations extends AccessValidationAlgebra {
         val accessLevel = {
           if ((aclAccessLevel == AccessLevel.NoAccess) && auth.isSystemAdmin)
             AccessLevel.Read
+          else if (zone.shared && rs.ownerGroupId.forall(auth.isGroupMember))
```
mitruly: Is it possible to get rid of `rulesForUser` and `getAccessFromUserAcls` above and use the `getAccessFromAcl` helper method?

modules/portal/public/lib/controllers/controller.zones.js
```diff
@@ -66,37 +68,71 @@ angular.module('controller.zones', [])
         return $scope.myGroupIds.indexOf(groupId) > -1;
     };
 
+    $scope.canAccessZone = function(accessLevel) {
+        switch(accessLevel) {
+            case 'noAccess':
```
mitruly: Think this should be `NoAccess`.

modules/portal/public/lib/controllers/controller.zones.js
```diff
@@ -66,37 +68,71 @@ angular.module('controller.zones', [])
...
+                return false;
+            case 'Read':
```
mitruly: Should we also care about `Write`? Can just have the `NoAccess` case and then:
```
default:
     return true;
```

modules/portal/public/lib/controllers/controller.zones.js
britneywright: this should be working correctly again

modules/portal/public/lib/controllers/controller.zones.js
britneywright: We don't currently have a Write access for Zone, but changing this to be an if statement to catch all non Read and Delete access.

britneywright:
I'd prefer to leave the changing the way we handle ACL rules for a separate issue/PR @mitruly since that's existing behavior. I'll take a look at your other comments.

mitruly:
🍻

PR #692, opened 2019-06-17
Repository preparation for LDAP sync

mitruly:
Precursor to #524.

Changes in this pull request:
- Create `task` repo in MySQL
- Create helper functions in `user` MySQL
- Deprecate DynamoDB `user` repo as unimplemented
- Update core and API modules with repo configurations

### DIFF
--- a/modules/core/src/main/scala/vinyldns/core/task/TaskRepository.scala
+++ b/modules/core/src/main/scala/vinyldns/core/task/TaskRepository.scala
@@ -23,7 +23,7 @@ import scala.concurrent.duration.FiniteDuration
 
 trait TaskRepository extends Repository {
 
-  def fetchAndClaimTask(name: String, pollingInterval: FiniteDuration): IO[Boolean]
+  def claimTask(name: String, pollingInterval: FiniteDuration): IO[Boolean]
 
   def releaseTask(name: String): IO[Unit]
 }
--- a/modules/core/src/main/scala/vinyldns/core/task/TaskRepository.scala
+++ b/modules/core/src/main/scala/vinyldns/core/task/TaskRepository.scala
@@ -23,7 +23,7 @@ import scala.concurrent.duration.FiniteDuration
 
 trait TaskRepository extends Repository {
 
-  def fetchAndClaimTask(name: String, pollingInterval: FiniteDuration): IO[Int]
+  def fetchAndClaimTask(name: String, pollingInterval: FiniteDuration): IO[Boolean]
 
   def releaseTask(name: String): IO[Unit]
 }
--- a/modules/core/src/main/scala/vinyldns/core/task/TaskRepository.scala
+++ b/modules/core/src/main/scala/vinyldns/core/task/TaskRepository.scala
@@ -23,9 +23,7 @@ import scala.concurrent.duration.FiniteDuration
 
 trait TaskRepository extends Repository {
 
-  def unclaimedTaskExists(name: String, pollingInterval: FiniteDuration): IO[Boolean]
-
-  def claimTask(name: String): IO[Unit]
+  def fetchAndClaimTask(name: String, pollingInterval: FiniteDuration): IO[Int]
 
   def releaseTask(name: String): IO[Unit]
 }
--- a/modules/core/src/main/scala/vinyldns/core/task/TaskRepository.scala
+++ b/modules/core/src/main/scala/vinyldns/core/task/TaskRepository.scala
@@ -14,13 +14,18 @@
  * limitations under the License.
  */
 
-package vinyldns.core.repository
+package vinyldns.core.task
 
-trait Repository
+import cats.effect.IO
+import vinyldns.core.repository.Repository
 
-object RepositoryName extends Enumeration {
-  type RepositoryName = Value
-  val user, group, membership, groupChange, recordSet, recordChange, zoneChange, zone, batchChange,
-  userChange, task =
-    Value
+import scala.concurrent.duration.FiniteDuration
+
+trait TaskRepository extends Repository {
+
+  def unclaimedTaskExists(name: String, pollingInterval: FiniteDuration): IO[Boolean]
+
+  def claimTask(name: String): IO[Unit]
+
+  def releaseTask(name: String): IO[Unit]
 }
--- a/modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlTaskRepositoryIntegrationSpec.scala
+++ b/modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlTaskRepositoryIntegrationSpec.scala
@@ -68,45 +68,11 @@ class MySqlTaskRepositoryIntegrationSpec extends WordSpec with BeforeAndAfterAll
     }
   }
 
-  def fetchAndClaimTaskWithSleep(name: String, pollingInterval: FiniteDuration): IO[Unit] =
-    IO {
-      val pollingExpirationHours = pollingInterval.toHours * 2
-      val currentTime = DateTime.now
-      DB.localTx { implicit s =>
-        val statement =
-          sql"""
-            |START TRANSACTION;
-            |SELECT *
-            |  FROM task
-            | WHERE (in_flight = 0
-            |    OR updated IS NULL
-            |    OR updated < {updatedTimeComparison})
-            |   AND name = {taskName} FOR UPDATE;
-            |SELECT SLEEP(10);
-            |UPDATE task
-            |   SET in_flight = 1, updated = {currentTime}
-            | WHERE (in_flight = 0
-            |    OR updated IS NULL
-            |    OR updated < {updatedTimeComparison})
-            |   AND name = {taskName};
-            |COMMIT;
-            """.stripMargin
-
-        statement
-          .bindByName(
-            'currentTime -> currentTime,
-            'taskName -> name,
-            'updatedTimeComparison -> currentTime.minusHours(pollingExpirationHours.toInt))
-          .update()
-          .apply()
-      }
-    }
-
-  "fetchAndClaimTask" should {
+  "claimTask" should {
     "return true if non-in-flight task exists and updated time is null" in {
       val f = for {
         _ <- insertTask(0, startDateTime, startDateTime)
-        unclaimedTaskExists <- repo.fetchAndClaimTask(TASK_NAME, 1.hour)
+        unclaimedTaskExists <- repo.claimTask(TASK_NAME, 1.hour)
       } yield unclaimedTaskExists
 
       f.unsafeRunSync() shouldBe true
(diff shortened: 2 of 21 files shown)


nimaeskandary:
Is this our first time deprecating a repo? Iffy on just gutting it right away in case there is an installation out there that uses dynamo. Could we have a dedicated pr for that with steps on how to migrate and also upload our dynamo -> mysql user script in some blog like format? (just in the pr prob works too)

rebstar6:
@nimaeskandary we did it with the recordset repo too fwiw

modules/mysql/src/main/resources/db/migration/V3.15__Tasks.sql
```diff
@@ -0,0 +1,17 @@
...
+
+USE ${dbName};
+
+CREATE TABLE task (
+  name VARCHAR(255) NOT NULL,
+  in_flight BIT(1) NOT NULL,
+  created DATETIME NOT NULL,
+  updated DATETIME,
+  PRIMARY KEY (name),
+  INDEX name_index (name),
```
nimaeskandary: primary key is already an index no?

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlTaskRepository.scala
```diff
@@ -0,0 +1,79 @@
...
+    """.stripMargin
+
+  // Fetch unclaimed task, with the exception of the expiration interval is exceeded
+  // `updated IS NULL` case is for the first run where the seeded data does not have an updated time set
+  private val FETCH_UNCLAIMED =
+    sql"""
+       |SELECT *
+       |  FROM task
+       | WHERE (in_flight = 0
+       |    OR updated IS NULL
```
nimaeskandary: could it be initialized with NOW() instead?

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlTaskRepository.scala
```diff
@@ -0,0 +1,79 @@
...
+       |    OR updated < {updatedTimeComparison})
+       |   AND name = {taskName}
```
rebstar6: will this repo be unique by task name? if so, i cant imagine we need any other indexes

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlTaskRepository.scala
```diff
@@ -0,0 +1,79 @@
...
+            'taskName -> name)
+          .map(_.string(1))
+          .first()
+          .apply()
+          .nonEmpty
+      }
+    }
+
+  def claimTask(name: String): IO[Unit] = IO {
+    DB.localTx { implicit s =>
```
nimaeskandary: curious what mutexes are planned for the implementation, should a read of the value be done first in this localtx and have the claim fail if in flight is already 1?

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlTaskRepository.scala
mitruly: The issue with using `NOW()` is timing differences between the application and MySQL server.

modules/mysql/src/main/resources/db/migration/V3.15__Tasks.sql
mitruly: Yes, I'll remove this.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlTaskRepository.scala
mitruly: Yes, task name is unique; can remove the other index.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlTaskRepository.scala
mitruly: The idea is to do a read before the claim (in the next PR), or you can look get a preview in https://github.com/vinyldns/vinyldns/pull/524/files#diff-8e3b44c876bc7dee09201b501f158a24.

It also shouldn't be problematic for two nodes to run a sync task in the worst-case scenario.

nimaeskandary: Right, my issue with that is it doesn't protect against race conditions

mitruly: Valid points, @nimaeskandary. Added two SQL statements for acquiring and releasing a table lock, which we would ideally do before/after updating values (eg. `inFlight` status, `updated` time).

rebstar6: woah, locking the whole table feels aggressive. if anything, you are now using a lock on something you are treating as a lock.

if you within the same transaction, get the item, check that in_flight == 0, and then claim it if not, you should be able to avoid locking the whole table

modules/core/src/main/scala/vinyldns/core/task/TaskRepository.scala
```diff
@@ -0,0 +1,31 @@
...
+import cats.effect.IO
+import vinyldns.core.repository.Repository
+
+import scala.concurrent.duration.FiniteDuration
+
+trait TaskRepository extends Repository {
+
+  def unclaimedTaskExists(name: String, pollingInterval: FiniteDuration): IO[Boolean]
+
+  def claimTask(name: String): IO[Unit]
```
rebstar6: per convo in the repo itself, it probably makes sense transaction-wise to merge unclaimedTaskExists and claimTask into a `claimTaskIfAvailable: IO[Boolean]` that will, within 1 transaction, check the state and claim if it can (true if yes, false if no).

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlTaskRepository.scala
rebstar6: also the issue with locking the table from a caller outside of the database is that your table could get locked forever if the node goes down at a bad time. no bueno. better off dealing in transactions

pauljamescleary: I agree that lock table is aggressive.

Have to look at the transaction isolation level.  The method runs inside of a transaction, and the MySql database protects against things like dirty reads.  Basically you should not read a value that someone else is currently writing to.  

It is worthwhile to look at the default transaction level and make sure we are good.

pauljamescleary: Looking at this, may want to just set the transaction level to serializable for the task repository: https://mydbops.wordpress.com/2018/06/22/back-to-basics-isolation-levels-in-mysql/

pauljamescleary: Also, nice little write up on locking and mysql using just sql - https://medium.com/urbanclap-engineering/pessimistic-locking-for-a-distributed-system-part-1-cc85c755c357

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlTaskRepository.scala
```diff
@@ -0,0 +1,108 @@
...
+   *
+   * `updated IS NULL` case is for the first run where the seeded data does not have an updated time set
+   */
+
+  private val START_TRANSACTION =
+    sql"""
+      |START TRANSACTION;
+      """.stripMargin
+
+  private val FETCH_UNCLAIMED_TASK =
```
mitruly: These are defined as separate statements since we need to execute the update separately to return back its value.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlTaskRepository.scala
```diff
@@ -0,0 +1,108 @@
...
+      |   SET in_flight = 0, updated = {currentTime}
+      | WHERE name = {name}
+      """.stripMargin
+
+  def fetchAndClaimTask(name: String, pollingInterval: FiniteDuration): IO[Boolean] =
+    IO {
+      val pollingExpirationHours = pollingInterval.toHours * 2
+      val currentTime = DateTime.now
+      DB.localTx { implicit s =>
+        START_TRANSACTION.execute()
```
mitruly: Opting for transaction rather than setting isolation level to `SERIALIZABLE` since AWS Aurora does not support it (source: https://mydbops.wordpress.com/2018/06/22/back-to-basics-isolation-levels-in-mysql/).

`START TRANSACTION` commences a transaction until `COMMIT` or `ROLLBACK`. Don't think we need to explicitly handle a timeout of blocking on `FOR UPDATE` call since this is wrapped within an `IO`.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlTaskRepository.scala
```diff
@@ -0,0 +1,108 @@
...
+            'updatedTimeComparison -> currentTime.minusHours(pollingExpirationHours.toInt),
+            'taskName -> name,
+            'currentTime -> currentTime)
+          .first()
+          .update()
+          .apply()
+
+        COMMIT.execute()
+
+        updateResult == 1
```
mitruly: Checking that there was an update and converting to `Boolean`.

modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlTaskRepositoryIntegrationSpec.scala
```diff
@@ -0,0 +1,177 @@
...
+        val statement =
+          sql"""
+            |START TRANSACTION;
+            |SELECT *
+            |  FROM task
+            | WHERE (in_flight = 0
+            |    OR updated IS NULL
+            |    OR updated < {updatedTimeComparison})
+            |   AND name = {taskName} FOR UPDATE;
+            |SELECT SLEEP(10);
```
mitruly: Created this with an embedded `SELECT SLEEP` statement so we can test the blocking nature of `FOR UPDATE` and that upon release, a waiting read should yield the updated values.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlTaskRepository.scala
```diff
@@ -0,0 +1,108 @@
...
+      """.stripMargin
+
+  def fetchAndClaimTask(name: String, pollingInterval: FiniteDuration): IO[Boolean] =
+    IO {
+      val pollingExpirationHours = pollingInterval.toHours * 2
+      val currentTime = DateTime.now
+      DB.localTx { implicit s =>
+        START_TRANSACTION.execute()
+
+        FETCH_UNCLAIMED_TASK
```
rebstar6: we dont need the fetch it seems. youre not doing anything with it...

given that, you are probably fine without the transaction stuff, just do the update and return based on whether the update was made :)

mitruly: The fetch performs a row lock so that the table is not accessible by another session until the current transaction ends; that said just doing an update is probably sufficient enough.

PR #693, opened 2019-06-18
Reject batch change

mitruly:
Updates `BatchChangeService` rejection to fail a valid `BatchChange` and its associated `SingleChange`s.

Changes in this pull request:
- Add `update` method to batch change repo
- Update `BatchChangeService` to perform rejection in repository
- Update tests

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
@@ -92,7 +92,9 @@ class BatchChangeService(
     for {
       batchChange <- getExistingBatchChange(batchChangeId)
       _ <- validateBatchChangeRejection(batchChange, authPrincipal).toBatchResult
-      rejectedBatchChange <- rejectBatchChange(batchChange, rejectBatchChangeInput.flatMap(_.reviewComment),
+      rejectedBatchChange <- rejectBatchChange(
+        batchChange,
+        rejectBatchChangeInput.flatMap(_.reviewComment),
         authPrincipal.signedInUser.id)
     } yield rejectedBatchChange
 
@@ -362,8 +364,10 @@ class BatchChangeService(
       listWithGroupNames = listResults.copy(batchChanges = summariesWithGroupNames)
     } yield listWithGroupNames
 
-  def rejectBatchChange(batchChange: BatchChange, reviewComment: Option[String], reviewerId: String):
-    BatchResult[BatchChange] = {
+  def rejectBatchChange(
+      batchChange: BatchChange,
+      reviewComment: Option[String],
+      reviewerId: String): BatchResult[BatchChange] = {
     // Store single changes as Failed status
     val rejectedSingleChanges = batchChange.changes.map {
       case sad: SingleAddChange => sad.copy(status = SingleChangeStatus.Failed)
@@ -371,10 +375,19 @@ class BatchChangeService(
     }
 
     // Update rejection attributes and single changes for batch change
-    val rejectedBatch = batchChange.copy(approvalStatus = BatchChangeApprovalStatus.ManuallyRejected,
-      reviewerId = Some(reviewerId), reviewComment = reviewComment, reviewTimestamp = Some(DateTime.now),
-      changes = rejectedSingleChanges)
+    val rejectedBatch = batchChange.copy(
+      approvalStatus = BatchChangeApprovalStatus.ManuallyRejected,
+      reviewerId = Some(reviewerId),
+      reviewComment = reviewComment,
+      reviewTimestamp = Some(DateTime.now),
+      changes = rejectedSingleChanges
+    )
 
-    batchChangeRepo.save(rejectedBatch).toBatchResult
+    val saveChanges = for {
+      _ <- batchChangeRepo.updateBatch(rejectedBatch)
+      _ <- batchChangeRepo.updateSingleChanges(rejectedSingleChanges)
+    } yield rejectedBatch
+
+    saveChanges.toBatchResult
   }
 }
--- a/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
@@ -92,7 +92,9 @@ class BatchChangeService(
     for {
       batchChange <- getExistingBatchChange(batchChangeId)
       _ <- validateBatchChangeRejection(batchChange, authPrincipal).toBatchResult
-    } yield batchChange
+      rejectedBatchChange <- rejectBatchChange(batchChange, rejectBatchChangeInput.flatMap(_.reviewComment),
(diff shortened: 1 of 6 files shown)


modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -359,4 +363,31 @@ class BatchChangeService(
...
+      approvalStatus = BatchChangeApprovalStatus.ManuallyRejected,
+      reviewerId = Some(reviewerId),
+      reviewComment = reviewComment,
+      reviewTimestamp = Some(DateTime.now),
+      changes = rejectedSingleChanges
+    )
+
+    val saveChanges = for {
+      _ <- batchChangeRepo.updateBatch(rejectedBatch)
+      _ <- batchChangeRepo.updateSingleChanges(rejectedSingleChanges)
```
rebstar6: we discussed queueing this in case of failure. that may be overkill for now (ill let @pauljamescleary weigh in if he feels strongly).

if we dont queue, probably update the single changes 1st. that way, if it fails for some reason mid-way, you can still resubmit the rejection without issue

modules/api/src/test/scala/vinyldns/api/domain/batch/BatchChangeServiceSpec.scala
```diff
@@ -310,14 +310,25 @@ class BatchChangeServiceSpec
...
+            .rejectBatchChange(
+              batchChange.id,
+              supportUserAuth,
+              Some(RejectBatchChangeInput(Some("review comment"))))
+            .value)
 
-      result shouldBe batchChange
+      result.status shouldBe BatchChangeStatus.Failed
+      result.approvalStatus shouldBe BatchChangeApprovalStatus.ManuallyRejected
+      result.changes.foreach(_.status shouldBe SingleChangeStatus.Failed)
```
britneywright: Should we have a different status for Single Changes, like Rejected? To distinguish between a Single Change that fails processing and a Single Change that is rejected before processing.

mitruly: We should be able to get that information from the `approvalStatus` on the `BatchChange`, so not sure if it's worthwhile to create a new enumeration value specifically for rejected.

pauljamescleary: Yea, this is something we haven't gone into depth on.

The batch change itself is rejected.  There is no means of rejecting an individual / single change, so what is it's status? 

Spitballing a bit, but I don't like setting it to "Failed", as "Failed" means something went wrong (we already have seen an example of that).  Something other than Failed for sure.

I think having a "ManuallyRejected" on the SingleChange seems reasonable.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlBatchChangeRepository.scala
```diff
@@ -47,11 +47,19 @@ class MySqlBatchChangeRepository
...
     sql"""
          |INSERT INTO batch_change(id, user_id, user_name, created_time, comments, owner_group_id,
-         |                          approval_status, reviewer_id, review_comment, review_timestamp)
+         |                         approval_status, reviewer_id, review_comment, review_timestamp)
          |     VALUES ({id}, {userId}, {userName}, {createdTime}, {comments}, {ownerGroupId},
-         |              {approvalStatus}, {reviewerId}, {reviewComment}, {reviewTimestamp})
+         |            {approvalStatus}, {reviewerId}, {reviewComment}, {reviewTimestamp})
         """.stripMargin
 
+  private final val UPDATE_BATCH_CHANGE =
```
pauljamescleary: curious as to whether we should be deviating from put semantics for this?

PUT semantics gives us last-write wins, and won't error if the the thing does not exist.

Can't say I have a terrifically strong opinion on the matter; only if we carry an UPDATE, we should rename `PUT_BATCH_CHANGE` to `INSERT_BATCH_CHANGE`

pauljamescleary: Ugh, yea, I do have an opinion.  The method on the repo is called `save` which _implies_ `PUT` semantics.

I would be curious why we don't just REPLACE INTO in `PUT_BATCH_CHANGE`?

mitruly: @pauljamescleary Updated the `PUT_BATCH_CHANGE` query to use `ON DUPLICATE KEY UPDATE` rather than `REPLACE INTO` for performance reasons.

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -359,4 +363,26 @@ class BatchChangeService(
...
+    // Update rejection attributes and single changes for batch change
+    val rejectedBatch = batchChange.copy(
+      approvalStatus = BatchChangeApprovalStatus.ManuallyRejected,
+      reviewerId = Some(reviewerId),
+      reviewComment = reviewComment,
+      reviewTimestamp = Some(DateTime.now),
+      changes = rejectedSingleChanges
+    )
+
+    batchChangeRepo.save(rejectedBatch).toBatchResult
```
mitruly: One of the disadvantages about using `save` as a catch-all query is that `saveBatchChanges` requires that `PUT_BATCH_CHANGE` is called before `PUT_SINGLE_CHANGE` due to the foreign key constraint.

From @rebstar6's earlier comment (https://github.com/vinyldns/vinyldns/pull/693#discussion_r294898127), we should probably be saving the updated single changes first when rejecting the batch change. This was circumvented in commit https://github.com/vinyldns/vinyldns/pull/693/commits/5969adbb28e48d570076815491ce7224d3c34765 since we had explicit `UPDATE` queries and could control the ordering.

modules/api/src/test/scala/vinyldns/api/domain/batch/BatchChangeServiceSpec.scala
mitruly: Think there should be some further discussion before implementing this since there are a couple of items that I can see that are slightly inhibiting:
- Our `single_change` table currently uses `VARCHAR 10` for `status`; we would need to at least have a migration to add `ManuallyRejected`
- Need to discuss logic for setting `BatchChangeStatus` in the `BatchChange` model. This is most likely adding another check for `SingleChangeStatus` of `ManuallyRejected`, but need to confirm.

rebstar6: im starting on a single change statuses PR now. will incorporate this in

rebstar6:
rejected status was added/merged in; need to rebase taht out

rebstar6:
closing for #719

Issue #660, opened 2019-05-29
Consistent paging nextId when table is empty


nimaeskandary:
**Motivation**
This came up in https://github.com/vinyldns/vinyldns/pull/640#discussion_r286585404. Making a card for discussion. Gist is, not all paging functions, between all our repos and dynamo vs sql, return None if you end right on the table boundary and you will page an empty page for the last one

My viewpoints are this doesn't really matter, as clients will get an empty list and a nextId of None for the next one, so don't see it breaking anything

**Scope of change**
All paging implementations could do a lookahead of one, and decide if the nextId is needed

nspadaccino:
Resolved in https://github.com/vinyldns/vinyldns/pull/694

PR #694, opened 2019-06-18
correct mysql listZones paging

britneywright:
part of #660

Changes in this pull request:
- retrieve one more item than the given maxItems, use that to determine if there should be a nextID

### DIFF
--- a/modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneRepository.scala
+++ b/modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneRepository.scala
@@ -239,7 +239,7 @@ class MySqlZoneRepository extends ZoneRepository with ProtobufConversions with M
           }
 
           sb.append(s" GROUP BY z.name ")
-          sb.append(s" LIMIT $maxItems")
+          sb.append(s" LIMIT ${maxItems + 1}")
 
           val query = sb.toString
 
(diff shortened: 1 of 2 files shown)


modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneRepository.scala
```diff
@@ -249,10 +249,13 @@ class MySqlZoneRepository extends ZoneRepository with ProtobufConversions with M
             .list()
             .apply()
 
-          val nextId = if (results.size < maxItems) None else results.lastOption.map(_.name)
+          val newResults =
+            if (results.size > maxItems) results.init else results
```
pauljamescleary: results.init throws an exception, use `dropRight(1)` instead

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneRepository.scala
```diff
@@ -249,10 +249,13 @@ class MySqlZoneRepository extends ZoneRepository with ProtobufConversions with M
             .list()
             .apply()
 
-          val nextId = if (results.size < maxItems) None else results.lastOption.map(_.name)
+          val newResults =
```
mitruly: Could actually combine these two so we only need to do the check once, maintaining the same functionality:
```
          val (newResults, nextId) =
            if (results.size > maxItems) (results.dropRight(1), results.lastOption.map(_.name))
            else (results, None)
```

britneywright: like it

PR #699, opened 2019-06-19
update zone shared status in the portal

### LINKED ISSUE #698: Toggle zone access in the portal
Currently the only way to update a zone from private to shared is through the API. We should add a toggle in the portal that allows system administrators to change it.

britneywright:
Fixes #698.

Changes in this pull request:
- change Access value in Manage Zone to be a drop down menu with Private and Shared. Default value is the zone's current shared value. false = Private, true = Shared.
- hide Abandon Zone button when confirming zone update. We already hide the Update Zone button at this point, just forgot to hide Abandon Zone as well in an earlier PR.

zone access drop down

not displaying abandon zone button when confirming a zone update

### DIFF
--- a/modules/portal/app/views/zones/zoneTabs/manageZone.scala.html
+++ b/modules/portal/app/views/zones/zoneTabs/manageZone.scala.html
@@ -51,7 +51,14 @@
                                     <label class="col-md-3 control-label">Access</label>
                                     <div class="col-md-9">
                                         <div>
-                                            <p class="form-control-static">{{ updateZoneInfo.shared ? "Shared" : "Private" }}</p>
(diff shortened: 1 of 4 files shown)


mitruly:
Noticed that initially going to the zone, there's a blank option that appears for the shared dropdown:
Screen Shot 2019-06-21 at 11 04 24 AM

britneywright:
Yeah that's default AngularJS behavior, can see it here https://docs.angularjs.org/api/ng/directive/select#simple-select-elements-with-static-options. Doesn't impact the data.

mitruly:
Also noticed that toggling the access value from and back to the original value doesn't disable the `Update Zone` button (though it _is_ disabled with the zone admin group field).

Issue #659, opened 2019-05-29
Batch Change Review Process


britneywright:
RFC: https://github.com/vinyldns/rfcs/pull/9

Planned work:

- add field to the Batch Change model. These fields are **NOT** to be set by users creating the batch change:

name          | type        | description |
 ------------ | :---------- | :---------- |
approvalStatus        | string      | The approval status of the batch change. Existing batch changes will default to "AutoApproved". Other approval statuses are "PendingApproval", "ManuallyApproved", "ManuallyRejected" |
reviewerId | string | ID of the user who reviews the pending batch change
reviewComment | string | Optional comment the reviewer may include when approving or rejecting the batch change. Max length is 1024.
reviewTimestamp | dateTime | The time the reviewer approved or rejected the batch change

The only new field that will be explicitly set by a user is the `reviewComment`, that should only be set at the time an authorized reviewer is approving or rejecting the batch change.

- add endpoints:
  - reject batch change
  - approve batch change

- add "Manual Review" flag config setting. 
If the config value is set to true it enables the batch change review process/workflow if the batch change has permissible errors. If the config value is set to false the batch change will fail if it has any errors and no manual review process takes place.

britneywright:
closing this since we implemented.

PR #700, opened 2019-06-20
allow super and support users to list all batch changes

britneywright:
This is in support of implementing the batch change review process (#659).

Currently the `listBatchChange` is restricted to only showing the batch changes of the user making the request. We want to allow super and support users the ability to view all batch changes, regardless of user. 

Changes in this pull request:
- Add `listAll` Boolean parameter to the `listBatchChange` endpoint, default is false.
- If `'listAll` is set to true and the user is a super user or support user all batch changes in the system are returned (bearing in mind the current maxItems parameter, etc)
- if `listAll` is set to true but the user is not a super user or support user then only their batch changes will be returned.

### DIFF
--- a/modules/api/src/test/scala/vinyldns/api/repository/InMemoryBatchChangeRepository.scala
+++ b/modules/api/src/test/scala/vinyldns/api/repository/InMemoryBatchChangeRepository.scala
@@ -113,9 +113,7 @@ class InMemoryBatchChangeRepository extends BatchChangeRepository {
       userId: Option[String],
       startFrom: Option[Int] = None,
       maxItems: Int = 100): IO[BatchChangeSummaryList] = {
-    val userBatchChanges =
-      if (userId.isDefined) batches.values.toList.filter(_.userId == userId.get)
-      else batches.values.toList
+    val userBatchChanges = batches.values.toList.filter(b => userId.forall(_ == b.userId))
     val batchChangeSummaries = for {
       sc <- userBatchChanges
       ids = sc.changes
--- a/modules/api/src/test/scala/vinyldns/api/repository/InMemoryBatchChangeRepository.scala
+++ b/modules/api/src/test/scala/vinyldns/api/repository/InMemoryBatchChangeRepository.scala
@@ -109,11 +109,13 @@ class InMemoryBatchChangeRepository extends BatchChangeRepository {
     IO.pure(changes)
   }
 
-  def getBatchChangeSummariesByUserId(
-      userId: String,
+  def getBatchChangeSummaries(
+      userId: Option[String],
       startFrom: Option[Int] = None,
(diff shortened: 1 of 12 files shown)


britneywright:
we use `validateSharedZoneAuthorized`  in the ZoneService to check if a super or support user is trying to change the zone shared status. 

https://github.com/vinyldns/vinyldns/blob/master/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneValidations.scala#L80-L87

britneywright:
I'd prefer to leave the changing the way we handle ACL rules for a separate issue/PR @mitruly since that's existing behavior. I'll take a look at your other comments.

modules/api/src/test/scala/vinyldns/api/repository/InMemoryBatchChangeRepository.scala
```diff
@@ -109,11 +109,13 @@ class InMemoryBatchChangeRepository extends BatchChangeRepository {
...
   }
 
-  def getBatchChangeSummariesByUserId(
-      userId: String,
+  def getBatchChangeSummaries(
+      userId: Option[String],
       startFrom: Option[Int] = None,
       maxItems: Int = 100): IO[BatchChangeSummaryList] = {
-    val userBatchChanges = batches.values.toList.filter(_.userId == userId)
+    val userBatchChanges =
```
mitruly: Pattern matching is more pleasing than using `.get` on `Option`:
```
      userId match {
        case Some(id) => batches.values.toList.filter(_.userId == id)
        case None => batches.values.toList
      }
```

rebstar6: can even do `batches.values.toList.filter(b => id.forall(_ == b.userId))` if youre feeling fancy :P

mitruly: So fancy

britneywright: 🍸

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlBatchChangeRepository.scala
```diff
@@ -87,7 +87,7 @@ class MySqlBatchChangeRepository
          |         FROM single_change sc
          |         JOIN batch_change bc
          |           ON sc.batch_change_id = bc.id
-         |        WHERE bc.user_id={userId}
+         |        WHERE bc.user_id LIKE {userId}
```
rebstar6: it seems better to only add the `=` in the case that a userId has been entered rather than always have this WHERE in the mix

PR #702, opened 2019-06-21
Allow updating and deleting dotted host records in batch change

britneywright:
VinylDNS doesn't allow the creation of dotted hosts, but legacy zones that are imported into Vinyl may have dotted hosts. We want to give users the ability to update or delete those existing dotted hosts. 

Changes in this pull request:
- affects batch change only.
- This only includes the following record types: A, AAAA, CNAME, MX (TXT was already allowed)
- Does not allowed creating new dotted hosts through batch change for any record type besides TXT

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/DomainValidationErrors.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/DomainValidationErrors.scala
@@ -166,13 +166,4 @@ final case class NewMultiRecordError(change: AddChangeForValidation) extends Dom
        |type ${change.inputChange.typ}.""".stripMargin
       .replaceAll("\n", " ")
 }
-
-final case class CreateDottedHostError(
-    recordSetName: String,
-    zoneName: String,
-    recordType: RecordType)
-    extends DomainValidationError {
-  def message: String =
-    s"""Cannot create dotted host "$recordSetName" with record type $recordType in the zone $zoneName"""
-}
 // $COVERAGE-ON$
--- a/modules/api/src/main/scala/vinyldns/api/domain/DomainValidationErrors.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/DomainValidationErrors.scala
@@ -167,4 +167,9 @@ final case class NewMultiRecordError(change: AddChangeForValidation) extends Dom
       .replaceAll("\n", " ")
 }
 
+final case class CreateDottedHostError(recordSetName: String, zoneName: String)
+    extends DomainValidationError {
+  def message: String =
+    s"Record with name $recordSetName is a dotted host which is illegal in the zone: $zoneName"
+}
 // $COVERAGE-ON$
--- a/modules/api/src/main/scala/vinyldns/api/domain/DomainValidationErrors.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/DomainValidationErrors.scala
@@ -167,9 +167,12 @@ final case class NewMultiRecordError(change: AddChangeForValidation) extends Dom
       .replaceAll("\n", " ")
 }
 
-final case class CreateDottedHostError(recordSetName: String, zoneName: String)
+final case class CreateDottedHostError(
+    recordSetName: String,
+    zoneName: String,
+    recordType: RecordType)
     extends DomainValidationError {
   def message: String =
-    s"Record with name $recordSetName is a dotted host which is illegal in the zone: $zoneName"
+    s"""Cannot create dotted host "$recordSetName" with record type $recordType in the zone $zoneName"""
 }
 // $COVERAGE-ON$
--- a/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
@@ -125,25 +125,6 @@ class BatchChangeService(
       }
       .toBatchResult
 
-  // zone name possibilities for all non-PTR changes
-  def getPossibleNonPtrZoneNames(nonPtr: List[ChangeInput]): Set[String] = {
-    def getZonesForNonDottedTypes(nonDotted: ChangeInput): Set[String] = {
-      val apexZoneName = nonDotted.inputName
-      val nonApexZoneName = getZoneFromNonApexFqdn(apexZoneName)
-
-      Set(apexZoneName, nonApexZoneName).filterNot(_ == "")
-    }
-
-    nonPtr
-      .map {
-        case record if List(A, AAAA, MX, PTR, TXT).contains(record.typ) =>
-          getAllPossibleZones(record.inputName).toSet
-        case otherForward => getZonesForNonDottedTypes(otherForward)
-      }
-      .toSet
-      .flatten
-  }
-
   def getZonesForRequest(changes: ValidatedBatch[ChangeInput]): IO[ExistingZones] = {
     // ipv4 search will be by filter, NOT by specific name because of classless reverse zone delegation
     def getPtrIpv4ZoneFilters(ipv4ptr: List[ChangeInput]): Set[String] =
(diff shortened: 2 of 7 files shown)


modules/api/src/main/scala/vinyldns/api/domain/DomainValidationErrors.scala
```diff
@@ -166,4 +166,10 @@ final case class NewMultiRecordError(change: AddChangeForValidation) extends Dom
        |type ${change.inputChange.typ}.""".stripMargin
       .replaceAll("\n", " ")
 }
+
+final case class CreateDottedHostError(recordSetName: String, zoneName: String)
+    extends DomainValidationError {
+  def message: String =
+    s"Record with name $recordSetName is a dotted host which is illegal in the zone: $zoneName"
```
mitruly: I think we want to rework this message a bit since this scenario is a bit tricky. There are currently two scenarios that we are allowing dotted hosts (as of this PR):
- Dotted host already exists within a zone
- Dotted host is of an allowable type (eg. `TXT`)

Since we're also throwing this error specifically in a forbidden create, we should pass in a type to the message and reword to something like:
```
s"Cannot create dotted host [$recordSetName] with type $recordType in zone [$zoneName]"
```

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -122,7 +122,8 @@ class BatchChangeService(
 
     nonPtr
       .map {
-        case txt if txt.typ == TXT => getAllPossibleZones(txt.inputName).toSet
+        case record if List(A, AAAA, MX, PTR, TXT).contains(record.typ) =>
```
mitruly: I think it wouldn't be too complicated to include `CNAME` in this; basically the `CNAME` discovery just attempts apex and non-apex, and throws an error if it matches on apex. Since `dottedZoneDiscovery` is used a `collectFirst` to get the most specific match, we could potentially use `dottedZoneDiscovery` and map that into some logic that checks whether the record would be an apex record for `CNAME`.

modules/api/src/main/scala/vinyldns/api/domain/DomainValidationErrors.scala
```diff
@@ -166,4 +166,13 @@ final case class NewMultiRecordError(change: AddChangeForValidation) extends Dom
...
       .replaceAll("\n", " ")
 }
+
+final case class CreateDottedHostError(
+    recordSetName: String,
+    zoneName: String,
+    recordType: RecordType)
+    extends DomainValidationError {
+  def message: String =
+    s"""Cannot create dotted host "$recordSetName" with record type $recordType in the zone $zoneName"""
```
rebstar6: hmm, our communication is around ZoneDiscovery error. im hesitant to add this new one in. ultimately what people need to do in this case is connect a new zone...need to think about the proper wording

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
rebstar6: also should have no PTR in there. just getAllPossibleZones for all the forward types

rebstar6: and yeah, pass CNAME through standard zone discovery just like the other forward types. when you check for dotted, also check that it's not apex for a cname at that point

modules/api/src/main/scala/vinyldns/api/domain/DomainValidationErrors.scala
britneywright: Do we want to add another line that suggests the record name and zone? But then we'd either have to validate the zone exists and their ability to use it, or just make a suggestion with no validations that the zone even exists.

britneywright: I think we should just put a more detailed explanation in the docs and leave it to the user to decide what course of action they want to take to remedy the issue. They might decide to use hyphens in the record name, request a new zone, etc.

rebstar6: yeah i think thats fine, i just wonder if we should still return ZoneDiscoveryError in this case

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
rebstar6: this comment still stands. not sure why were doing any matching in here anymore

rebstar6: the whole `def getPossibleNonPtrZoneNames` is i think just now `nonPtr.flatMap(chg => getAllPossibleZones(chg.inputName)).toSet`

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -245,20 +245,6 @@ class BatchChangeService(
   def standardZoneDiscovery(
```
rebstar6: merge all forward zone discovery into 
```
  def standardZoneDiscovery(
      change: ChangeInput,
      zoneMap: ExistingZones): SingleValidation[ChangeForValidation] = {

    // getAllPossibleZones is ordered most to least specific, so 1st match is right
    val zone = getAllPossibleZones(change.inputName).map(zoneMap.getByName).collectFirst {
      case Some(zn) => zn
    }

    zone match {
      case Some(zn) if zn.name == change.inputName && change.typ == CNAME =>
        RecordAlreadyExists(change.inputName).invalidNel
      case Some(zn) =>
        ChangeForValidation(zn, relativize(change.inputName, zn.name), change).validNel
      case None => ZoneDiscoveryError(change.inputName).invalidNel
    }
  }
```

rebstar6: and while youre at it rename to forward

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
```diff
@@ -231,6 +231,15 @@ class BatchChangeValidations(
     else ().validNel
   }
 
+  def newRecordSetIsNotDotted(change: AddChangeForValidation): SingleValidation[Unit] =
```
rebstar6: this is only called from validateAddWithContext, which is already split by type. no reason to have any type matching in here

also, i dont see it being called from the CNAME case in validateAddWithContext

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -245,20 +245,6 @@ class BatchChangeService(
...
-    val nonApexZone = zoneMap.getByName(nonApexName)
-
-    apexZone.orElse(nonApexZone) match {
-      case Some(zn) =>
-        ChangeForValidation(zn, relativize(change.inputName, zn.name), change).validNel
-      case None => ZoneDiscoveryError(change.inputName).invalidNel
-    }
-  }
-
-  def dottedZoneDiscovery(
```
rebstar6: similar to the last PR, you have a bunch of changes in BatchChangeService and BatchChangeValidation but no unit tests in those files. those need adding

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -266,27 +235,14 @@ class BatchChangeService(
     }
 
     zone match {
+      case Some(zn) if (zn.name == change.inputName && change.typ == CNAME) =>
+        RecordAlreadyExists(change.inputName).invalidNel
```
rebstar6: weird that were using RecordAlreadyExists for this. i see that we were before, so thats ok. i wonder why though? really this should be something about how you cant have a CNAME at a zone apex

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
```diff
@@ -231,6 +231,17 @@ class BatchChangeValidations(
...
+    change.inputChange.typ match {
+      case A | AAAA | MX
+          if change.recordName == change.zone.name || !change.recordName.contains(".") =>
+        ().validNel
+      case CNAME if !change.recordName.contains(".") =>
```
rebstar6: dont really need to split out this case, the apex check is elsewhere


  def newRecordSetIsNotDotted(change: AddChangeForValidation): SingleValidation[Unit] = {
    if (change.recordName.contains(".") && change.recordName != change.zone.name) {
      // a dotted host is functionally a zone discovery issue; we expect names to be non-dotted
      // within a zone
      ZoneDiscoveryError(change.inputChange.inputName).invalidNel
    } else {
      ().validNel
    }
  }

PR #703, opened 2019-06-24
allow updates for dotted host records in update recordset

britneywright:
in the same vein as #702 we want to allow users to update existing recordsets with dotted hosts.

This pertains to the update recordset route.

Changes in this pull request:
- users can modify dotted host records with the following exceptions:
  - users cannot create new dotted hosts (except TXT record types)
  - users cannot change the name of a dotted host to another dotted host (except TXT record types)

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetValidations.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetValidations.scala
@@ -95,7 +95,7 @@ object RecordSetValidations {
       existingRecordSet: Option[RecordSet] = None): Either[Throwable, Unit] =
     newRecordSet.typ match {
       case CNAME => cnameValidations(newRecordSet, existingRecordsWithName, zone)
-      case NS => nsValidations(newRecordSet, zone)
+      case NS => nsValidations(newRecordSet, zone, existingRecordSet)
       case SOA => soaValidations(newRecordSet, zone)
       case PTR => ptrValidations(newRecordSet, zone)
       case SRV | TXT | NAPTR => ().asRight // SRV, TXT and NAPTR do not go through dotted host check
--- a/modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetValidations.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetValidations.scala
@@ -77,52 +77,30 @@ object RecordSetValidations {
       !existingRecordsWithName.exists(rs => rs.id != newRecordSet.id && rs.typ == newRecordSet.typ)
     )
 
-  def isNotDotted(recordSet: RecordSet, zone: Zone): Either[Throwable, Unit] =
-    ensuring(InvalidRequest(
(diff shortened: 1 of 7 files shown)


modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetValidations.scala
```diff
@@ -149,6 +162,31 @@ object RecordSetValidations {
 
   }
 
+  def cnameEditValidations(
+      newRecordSet: RecordSet,
+      existingRecordsWithName: List[RecordSet],
+      zone: Zone,
+      oldRecordSetName: String): Either[Throwable, Unit] = {
```
mitruly: Not sure if this is overkill, but may be worthwhile to refactor this to make `oldRecordSetName` an `Option[String]` and have `isNotNewDotted` also take in an `oldRecordSetName: Option[String]`. We can then remove `isNotDotted` and have `isNotNewDotted` handle allowing existing dotted record changes but disallow new dotted names.

rebstar6: to michaels point, i think this whole PR can probably be achieved by changing isNotDotted to
```
  def isNotDotted(recordSet: RecordSet, zone: Zone, existingRecordset: Option[RecordSet]): Either[Throwable, Unit] =
    ensuring(InvalidRequest(
      s"Record with name ${recordSet.name} is a dotted host which is illegal in this zone ${zone.name}"))(
      recordSet.name == zone.name ||
        !recordSet.name.contains(".") ||
        existingRecordset.exists(_.name == recordSet.name)
    )
```
and passing in the existingRecordset on an update

rebstar6: i prefer that to duplicating whats basically the same method for cname validation

modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetValidations.scala
```diff
@@ -77,39 +77,30 @@ object RecordSetValidations {
...
-  def typeSpecificAddValidations(
+  def typeSpecificValidations(
       newRecordSet: RecordSet,
       existingRecordsWithName: List[RecordSet],
-      zone: Zone): Either[Throwable, Unit] =
+      zone: Zone,
+      existingRecordSet: Option[RecordSet] = None): Either[Throwable, Unit] =
     newRecordSet.typ match {
       case CNAME => cnameValidations(newRecordSet, existingRecordsWithName, zone)
       case NS => nsValidations(newRecordSet, zone)
```
rebstar6: nsValidations in edit passed in the old. `typeSpecificValidations ` should have oldRecordSet: Option[RS] passed in and you can pass that through.

seems the rest are the same

rebstar6: oh lol, existingRecordSet. yeah just pass that into nsValidations too :)

PR #704, opened 2019-06-24
add approve pending batch change route

britneywright:
Part of Batch Change Review Process (#659)

Changes in this pull request:
- add route for approving a pending batch change
- include basic validations that zone is pending and reviewer is authorized

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
@@ -94,6 +94,15 @@ class BatchChangeService(
       _ <- validateBatchChangeRejection(batchChange, authPrincipal).toBatchResult
     } yield batchChange
 
+  def approveBatchChange(
+      batchChangeId: String,
+      authPrincipal: AuthPrincipal,
+      approveBatchChangeInput: Option[ApproveBatchChangeInput]): BatchResult[BatchChange] =
+    for {
+      batchChange <- getExistingBatchChange(batchChangeId)
+      _ <- validateBatchChangeApproval(batchChange, authPrincipal).toBatchResult
+    } yield batchChange
+
   def getBatchChange(id: String, auth: AuthPrincipal): BatchResult[BatchChangeInfo] =
     for {
       batchChange <- getExistingBatchChange(id)
--- a/modules/api/src/main/scala/vinyldns/api/route/BatchChangeJsonProtocol.scala
+++ b/modules/api/src/main/scala/vinyldns/api/route/BatchChangeJsonProtocol.scala
@@ -42,7 +42,8 @@ trait BatchChangeJsonProtocol extends JsonValidation {
     BatchChangeSerializer,
     BatchChangeErrorListSerializer,
     BatchChangeErrorSerializer,
-    RejectBatchChangeInputSerializer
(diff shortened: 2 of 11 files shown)


modules/api/src/test/scala/vinyldns/api/route/BatchChangeRoutingSpec.scala
```diff
@@ -284,6 +284,18 @@ class BatchChangeRoutingSpec
...
+    def approveBatchChange(
+        batchChangeId: String,
+        authPrincipal: AuthPrincipal,
+        rejectionComment: Option[ApproveBatchChangeInput])
+      : EitherT[IO, BatchChangeErrorResponse, BatchChange] =
+      (batchChangeId, authPrincipal.isSystemAdmin) match {
+        case ("pendingBatchId", true) => EitherT(IO.pure(genericValidResponse.asRight))
+        case ("pendingBatchId", false) =>
+          EitherT(IO.pure(UserNotAuthorizedError("notAuthedID").asLeft))
+        case (_, _) => EitherT(IO.pure(BatchChangeNotPendingApproval("batchId").asLeft))
```
mitruly: This can actually just be `case _ =>` since we're not using either value of the tuple.

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -94,6 +94,15 @@ class BatchChangeService(
       _ <- validateBatchChangeRejection(batchChange, authPrincipal).toBatchResult
     } yield batchChange
 
+  def approveBatchChange(
```
rebstar6: expect this tested in batchchangeservicespec

modules/api/src/test/scala/vinyldns/api/route/BatchChangeRoutingSpec.scala
```diff
@@ -568,4 +580,55 @@ class BatchChangeRoutingSpec
...
+
+    "return OK if comments are not provided, batch change is PendingApproval, and reviewer is authorized" in {
+      Post("/zones/batchrecordchanges/pendingBatchId/approve").withEntity(
+        HttpEntity(ContentTypes.`application/json`, compact(render("")))) ~>
+        batchChangeRoute(supportUserAuth) ~> check {
+        status shouldBe OK
+      }
+    }
+
+    "return Forbidden if user is not a super or support admin" in {
```
rebstar6: route tests are exercising the serialization and possible responses from the service, but not the service method itself since its going to the fake `def approveBatchChange` you have in this file, not the real one in the service.

this test is valid, you want to check that if the service responds with UserNotAuthorizedError, the route maps to Forbidden...but because of that, you need separate tests in the service for what that returns errors for

modules/api/src/main/scala/vinyldns/api/route/BatchChangeJsonProtocol.scala
```diff
@@ -195,6 +196,16 @@ trait BatchChangeJsonProtocol extends JsonValidation {
         .map(RejectBatchChangeInput)
   }
 
+  case object ApproveBatchChangeInputSerializer
```
rebstar6: we do also have a BatchChangeJsonProtocolSpec. probably want a test or 2 in there

PR #706, opened 2019-06-25
WIP: starter code for manual approval flow

rebstar6:
no tests yet, expect failures

the general idea is to split out soft failures, and if all errors are in that bucket, save the batch change.
have not yet actually assigned any errors as soft or decided what to save for those on the single change (use a list of error codes? save something in system message? something else?). These will all be followup

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/DomainValidationErrors.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/DomainValidationErrors.scala
@@ -107,7 +107,7 @@ final case class InvalidBatchRecordType(param: String) extends DomainValidationE
       s"${SupportedBatchChangeRecordTypes.get}."
 }
 
-final case class ZoneDiscoveryError(name: String) extends SoftBatchError {
+final case class ZoneDiscoveryError(name: String) extends DomainValidationError {
   def message: String =
     s"""Zone Discovery Failed: zone for "$name" does not exist in VinylDNS. """ +
       "If zone exists, then it must be connected to in VinylDNS."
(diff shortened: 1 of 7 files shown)


modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeProtocol.scala
```diff
@@ -28,16 +31,46 @@ final case class BatchChangeInput(
...
 final case class AddChangeInput(
     inputName: String,
     typ: RecordType,
     ttl: Option[Long],
     record: RecordData)
-    extends ChangeInput
+    extends ChangeInput {
 
-final case class DeleteChangeInput(inputName: String, typ: RecordType) extends ChangeInput
+  def asNewStoredChange(errors: List[SoftBatchError]): SingleChange = {
```
rebstar6: TODO actually save the errors in some way. thinking about doing so as a list of error codes???

modules/api/src/main/scala/vinyldns/api/domain/DomainValidationErrors.scala
```diff
@@ -24,6 +24,13 @@ import vinyldns.core.domain.record.RecordType.RecordType
 // $COVERAGE-OFF$
 sealed trait DomainValidationError {
   def message: String
+
+  val isSoftFailure: Boolean = false
+}
+
+sealed trait SoftBatchError extends DomainValidationError {
```
pauljamescleary: The question that comes to my mind is, should the characteristic of being "soft" be on the error itself, or external to the error?

In my head, I imagined that under certain circumstances an error would be soft, but others the same error would be hard.  That could be configuration driven, or even depending on the zone, type of zone, or type of record.

Should the model be ...

```scala
final case class SoftError(error: DomainValidationError)
```

instead?  From a developer ergonomic standpoint, it is easy to pattern match, and using composition instead of inheritance gives more flexibility.

pauljamescleary: could even be

```scala
final case class BatchWarning(error: DomainValidationError, isFatal: Boolean = true)
```

rebstar6:
closing, this work has been pulled into other PRs

PR #708, opened 2019-06-25
update Zones docs

britneywright:
Changes in this pull request:
- Add `listAll` parameter in Zone API docs
- Add `accessLevel` attributes in Zone API docs
- Add a page for zone search in the portal docs that gives examples of the wildcard search
- Add information about My Zones and All Zones tabs in the Zones page in the portal


modules/docs/src/main/tut/api/list-zones.md
```diff
@@ -19,6 +19,7 @@ name          | type          | required?   | description |
 nameFilter    | string        | no          | Characters that are part of the zone name to search for.  The wildcard character `*` is supported, for example `www*`.  Omit the wildcard character when searching for an exact zone name. |
 startFrom     | *any*         | no          | In order to advance through pages of results, the startFrom is set to the `nextId` that is returned on the previous response.  It is up to the client to maintain previous pages if the client wishes to advance forward and backward.   If not specified, will return the first page of results |
 maxItems      | int           | no          | The number of items to return in the page.  Valid values are 1 - 100. Defaults to 100 if not provided. |
+listAll       | boolean       | no          | Returns all zones, regardless of requesting user's zone ownership or access. Defaults to false if not provided. |
```
rebstar6: all is confusing with "return all" as in, no pagination. maybe this:
```
If false, only zones the user has explicit ownership or ACL access to will be returned. If true, all zones in the system will be returned (still paginated)
```

modules/docs/src/main/tut/api/zone-model.md
```diff
@@ -36,6 +36,7 @@ acl           | ZoneACL     | The access control rules governing the zone.  See
 id            | string      | The unique identifier for this zone
 latestSync    | date-time   | The last date and time the zone was synced
 isTest        | boolean     | Defaults to **false**. Used for restricted access during VinylDNS testing, can be ignored by clients
+accessLevel   | string      | Access level of the user requesting the zone. Current levels are Delete (full access), Read and NoAccess.
```
pauljamescleary: So, is this now returned on _every_ response?  The `ZoneModel` conveys more than the search endpoint, also `getZone` for example

britneywright: that's correct

PR #710, opened 2019-06-25
Save a batch needing approval

rebstar6:
split off of https://github.com/vinyldns/vinyldns/pull/706

save a batch that needs approval. will need minor updates (commented) when we have the right singleChange status options

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/DomainValidationErrors.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/DomainValidationErrors.scala
@@ -24,14 +24,9 @@ import vinyldns.core.domain.record.RecordType.RecordType
 // $COVERAGE-OFF$
 sealed trait DomainValidationError {
   def message: String
-
-  val isSoftFailure: Boolean = false
 }
 
-sealed trait SoftBatchError extends DomainValidationError {
-  override val isSoftFailure: Boolean = true
-  // TODO envisioning error codes, display options in here
-}
+sealed trait SoftBatchError extends DomainValidationError
 
 // The request itself is invalid in this case, so we fail fast
 final case class ChangeLimitExceeded(limit: Int) extends DomainValidationError {
--- a/modules/api/src/main/scala/vinyldns/api/domain/DomainValidationErrors.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/DomainValidationErrors.scala
@@ -173,4 +173,5 @@ final case class NewMultiRecordError(change: AddChangeForValidation) extends Dom
        |type ${change.inputChange.typ}.""".stripMargin
       .replaceAll("\n", " ")
 }
+
 // $COVERAGE-ON$
--- a/modules/api/src/main/scala/vinyldns/api/domain/DomainValidationErrors.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/DomainValidationErrors.scala
@@ -22,9 +22,8 @@ import vinyldns.core.domain.record.{RecordSet, RecordType}
 import vinyldns.core.domain.record.RecordType.RecordType
 
 // $COVERAGE-OFF$
-sealed abstract class DomainValidationError(isFatalError: Boolean = true) {
+sealed abstract class DomainValidationError(val isFatal: Boolean = true) {
   def message: String
-  def isFatal: Boolean = isFatalError
 }
 
 // The request itself is invalid in this case, so we fail fast
--- a/modules/api/src/main/scala/vinyldns/api/domain/DomainValidationErrors.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/DomainValidationErrors.scala
@@ -22,11 +22,10 @@ import vinyldns.core.domain.record.{RecordSet, RecordType}
 import vinyldns.core.domain.record.RecordType.RecordType
 
 // $COVERAGE-OFF$
-sealed trait DomainValidationError {
+sealed abstract class DomainValidationError(isFatalError: Boolean = true) {
   def message: String
+  def isFatal: Boolean = isFatalError
 }
-
-sealed trait SoftBatchError extends DomainValidationError
 
 // The request itself is invalid in this case, so we fail fast
 final case class ChangeLimitExceeded(limit: Int) extends DomainValidationError {
--- a/modules/api/src/main/scala/vinyldns/api/domain/DomainValidationErrors.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/DomainValidationErrors.scala
@@ -24,6 +24,13 @@ import vinyldns.core.domain.record.RecordType.RecordType
 // $COVERAGE-OFF$
 sealed trait DomainValidationError {
   def message: String
+
+  val isSoftFailure: Boolean = false
+}
+
+sealed trait SoftBatchError extends DomainValidationError {
+  override val isSoftFailure: Boolean = true
+  // TODO envisioning error codes, display options in here
 }
 
 // The request itself is invalid in this case, so we fail fast
--- a/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeProtocol.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeProtocol.scala
@@ -16,7 +16,10 @@
 
 package vinyldns.api.domain.batch
 
+import vinyldns.api.VinylDNSConfig
+import vinyldns.api.domain.SoftBatchError
 import vinyldns.core.domain.DomainHelpers.ensureTrailingDot
+import vinyldns.core.domain.batch._
 import vinyldns.core.domain.record.RecordData
 import vinyldns.core.domain.record.RecordType._
 
@@ -28,6 +31,7 @@ final case class BatchChangeInput(
 sealed trait ChangeInput {
   val inputName: String
   val typ: RecordType
+  def asNewStoredChange(errors: List[SoftBatchError]): SingleChange
 }
 
 final case class AddChangeInput(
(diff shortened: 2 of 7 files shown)


modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -331,9 +343,56 @@ class BatchChangeService(
...
+        .map(_.batchChange)
+    case PendingApproval if manualReviewEnabled =>
+      // save the change, will need to return to it later on approval
+      batchChangeRepo.save(batchChange).toBatchResult
+    case _ =>
+      // this should not be called with a rejected change (or if manual review is off)!
+      logger.error(
+        s"convertOrSave called with a rejected batch change;" +
+          s"batchChangeId='${batchChange.id}; manualReviewEnabled=$manualReviewEnabled'")
+      UnknownConversionError("Cannot convert a rejected batch change").toLeftBatchResult
```
britneywright: Should we have separate error messages for a rejected batch change when manual review is enabled vs other statuses when a manual review is disabled?

rebstar6: i thought about it, though ultimately its a bug if we ever get here, just need it for the case class matching

modules/api/src/main/scala/vinyldns/api/domain/DomainValidationErrors.scala
```diff
@@ -24,6 +24,13 @@ import vinyldns.core.domain.record.RecordType.RecordType
 // $COVERAGE-OFF$
 sealed trait DomainValidationError {
   def message: String
+
+  val isSoftFailure: Boolean = false
+}
+
+sealed trait SoftBatchError extends DomainValidationError {
+  override val isSoftFailure: Boolean = true
```
nimaeskandary: val override seems weird to me, possibly just make this a def too? I'm no scala pro though if this doesn't matter then ignore

rebstar6: its required:

[error] /Users/rstar201/Desktop/Vinyl/vinyldns/modules/api/src/main/scala/vinyldns/api/domain/DomainValidationErrors.scala:32:7: overriding value isSoftFailure in trait DomainValidationError of type Boolean;
[error]  value isSoftFailure needs `override' modifier
[error]   val isSoftFailure: Boolean = true

rebstar6: removed anyway

modules/api/src/test/scala/vinyldns/api/domain/batch/BatchChangeInputSpec.scala
```diff
@@ -46,4 +48,44 @@ class BatchChangeInputSpec extends WordSpec with Matchers {
...
+      val asAdd = converted.asInstanceOf[SingleAddChange]
+
+      asAdd.zoneId shouldBe None
+      asAdd.zoneName shouldBe None
+      asAdd.recordName shouldBe None
+      asAdd.inputName shouldBe "some.test.com."
+      asAdd.typ shouldBe A
+      asAdd.ttl shouldBe VinylDNSConfig.defaultTtl
+      asAdd.recordData shouldBe AData("1.1.1.1")
+      // TODO this should change to NeedsApproval
```
britneywright: minor thing, in the BatchChangeProtocol your comment says this should be NeedsReview.

modules/api/src/test/scala/vinyldns/api/domain/batch/BatchChangeInputSpec.scala
```diff
@@ -46,4 +48,44 @@ class BatchChangeInputSpec extends WordSpec with Matchers {
...
+
+      converted shouldBe a[SingleDeleteChange]
+      val asDelete = converted.asInstanceOf[SingleDeleteChange]
+
+      asDelete.zoneId shouldBe None
+      asDelete.zoneName shouldBe None
+      asDelete.recordName shouldBe None
+      asDelete.inputName shouldBe "some.test.com."
+      asDelete.typ shouldBe A
+      // TODO this should change to NeedsApproval
```
britneywright: same comment here as above.

modules/api/src/test/scala/vinyldns/api/domain/batch/BatchChangeServiceSpec.scala
```diff
@@ -862,14 +886,105 @@ class BatchChangeServiceSpec
...
+        "some.test.delete.",
+        TXT,
+        SingleChangeStatus.Pending,
+        None,
+        None,
+        None,
+        result.changes(2).id
+      )
+    }
+    "return a BatchChangeErrorList if all data inputs are valid/soft failures and manual review is disabled" in {
```
britneywright: Is this the same test as the one that starts on 889?

rebstar6: hmm. apparently yes :P

modules/api/src/main/scala/vinyldns/api/domain/DomainValidationErrors.scala
```diff
@@ -22,8 +22,9 @@ import vinyldns.core.domain.record.{RecordSet, RecordType}
 import vinyldns.core.domain.record.RecordType.RecordType
 
 // $COVERAGE-OFF$
-sealed trait DomainValidationError {
+sealed abstract class DomainValidationError(isFatalError: Boolean = true) {
   def message: String
+  def isFatal: Boolean = isFatalError
```
pauljamescleary: don't need this def, can just do in the constructor...

`sealed abstract class DomainValidationError(val isFatalError: Boolean = true)`

rebstar6: ah, good call

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -331,9 +345,54 @@ class BatchChangeService(
...
+      batchChange: BatchChange,
+      existingZones: ExistingZones,
+      existingRecordSets: ExistingRecordSets,
+      ownerGroupId: Option[String]): BatchResult[BatchChange] = batchChange.approvalStatus match {
+    case AutoApproved =>
+      // send on to the converter, it will be saved there
+      batchChangeConverter
+        .sendBatchForProcessing(batchChange, existingZones, existingRecordSets, ownerGroupId)
+        .map(_.batchChange)
+    case ManuallyApproved if manualReviewEnabled =>
```
pauljamescleary: not thrilled with having the same thing happen in two case statements, can we combine these into one to DRY it up?

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -331,9 +345,54 @@ class BatchChangeService(
...
+        .sendBatchForProcessing(batchChange, existingZones, existingRecordSets, ownerGroupId)
+        .map(_.batchChange)
+    case PendingApproval if manualReviewEnabled =>
+      // save the change, will need to return to it later on approval
+      batchChangeRepo.save(batchChange).toBatchResult
+    case _ =>
+      // this should not be called with a rejected change (or if manual review is off)!
+      logger.error(
+        s"convertOrSave called with a rejected batch change;" +
+          s"batchChangeId='${batchChange.id}; manualReviewEnabled=$manualReviewEnabled'")
```
pauljamescleary: missing some `'` here; also, fwiw, we probably should not use single quotes (I know we use them elsewhere) as they do not get stripped out in splunk properly.  For example, we see in the cron job I created `db='vinyl'` because of single quotes.  Really, should be able to do `db=vinyl` instead.

We should use double quotes or omit quotes if we know there are no spaces in the name=value pair

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
rebstar6: yeah went back and forth, i think its more clear that just this one follows the flag. i can combine though if you have a strong pref

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeProtocol.scala
```diff
@@ -28,16 +32,48 @@ final case class BatchChangeInput(
...
     inputName: String,
     typ: RecordType,
     ttl: Option[Long],
     record: RecordData)
-    extends ChangeInput
+    extends ChangeInput {
 
-final case class DeleteChangeInput(inputName: String, typ: RecordType) extends ChangeInput
+  // TODO - once we have SingleChangeStatus.NeedsReview, this should use that
+  def asNewStoredChange(errors: NonEmptyList[DomainValidationError]): SingleChange = {
```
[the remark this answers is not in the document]
rebstar6: no but we will, this is part 1 of 3 or 4 for this card. i can take it out if you want but it will just be put right back in shortly

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
rebstar6: actually, with the IF this is better left split

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeProtocol.scala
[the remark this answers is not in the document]
pauljamescleary: ok

PR #715, opened 2019-06-28
Add approvalStatus query param to list batch changes

britneywright:
In support of #659 

Changes in this pull request:
- add `approvalStatus` query parameter on the list batch changes route to filter returned batch changes by their approval status. Parameter is available for any users.
- In conjunction with the `ignoreAccess` parameter this will allow super users to retrieve all batch changes in the system that are in a pending approval state.

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/route/BatchChangeRouting.scala
+++ b/modules/api/src/main/scala/vinyldns/api/route/BatchChangeRouting.scala
@@ -24,6 +24,7 @@ import cats.effect._
 import vinyldns.api.VinylDNSConfig
 import vinyldns.core.domain.auth.AuthPrincipal
 import vinyldns.api.domain.batch._
+import vinyldns.core.domain.batch.BatchChangeApprovalStatus
 
 trait BatchChangeRoute extends Directives {
   this: VinylDNSJsonProtocol with VinylDNSDirectives with JsonValidationRejection =>
@@ -32,6 +33,7 @@ trait BatchChangeRoute extends Directives {
 
   final private val MAX_ITEMS_LIMIT: Int = 100
 
+  case class BatchChangeApproval(approvalType: String) {}
   val batchChangeRoute: AuthPrincipal => server.Route = { authPrincipal: AuthPrincipal =>
     val standardBatchChangeRoutes = (post & path("zones" / "batchrecordchanges")) {
       monitor("Endpoint.postBatchChange") {
@@ -53,17 +55,27 @@ trait BatchChangeRoute extends Directives {
         parameters(
           "startFrom".as[Int].?,
           "maxItems".as[Int].?(MAX_ITEMS_LIMIT),
-          "ignoreAccess".as[Boolean].?(false)) {
-          (startFrom: Option[Int], maxItems: Int, ignoreAccess: Boolean) =>
+          "ignoreAccess".as[Boolean].?(false),
+          "approvalStatus".as[String].?) {
+          (
+              startFrom: Option[Int],
+              maxItems: Int,
+              ignoreAccess: Boolean,
+              approvalStatus: Option[String]) =>
             {
+              val convertApprovalStatus = BatchChangeApprovalStatus.fromString(approvalStatus)
               handleRejections(invalidQueryHandler) {
                 validate(
                   0 < maxItems && maxItems <= MAX_ITEMS_LIMIT,
                   s"maxItems was $maxItems, maxItems must be between 1 and $MAX_ITEMS_LIMIT, inclusive.") {
-                  execute(batchChangeService
-                    .listBatchChangeSummaries(authPrincipal, startFrom, maxItems, ignoreAccess)) {
-                    summaries =>
-                      complete(StatusCodes.OK, summaries)
+                  execute(
+                    batchChangeService.listBatchChangeSummaries(
+                      authPrincipal,
+                      startFrom,
+                      maxItems,
+                      ignoreAccess,
+                      convertApprovalStatus)) { summaries =>
+                    complete(StatusCodes.OK, summaries)
                   }
                 }
               }
--- a/modules/api/src/main/scala/vinyldns/api/route/BatchChangeRouting.scala
+++ b/modules/api/src/main/scala/vinyldns/api/route/BatchChangeRouting.scala
@@ -63,7 +63,8 @@ trait BatchChangeRoute extends Directives {
               ignoreAccess: Boolean,
               approvalStatus: Option[String]) =>
             {
-              val convertApprovalStatus = BatchChangeApprovalStatus.fromString(approvalStatus)
+              val convertApprovalStatus = BatchChangeApprovalStatus.find(approvalStatus.mkString)
+
               handleRejections(invalidQueryHandler) {
                 validate(
                   0 < maxItems && maxItems <= MAX_ITEMS_LIMIT,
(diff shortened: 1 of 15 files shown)


modules/core/src/main/scala/vinyldns/core/domain/batch/BatchChange.scala
```diff
@@ -69,6 +69,19 @@ object BatchChangeStatus extends Enumeration {
 object BatchChangeApprovalStatus extends Enumeration {
   type BatchChangeApprovalStatus = Value
   val AutoApproved, PendingApproval, ManuallyApproved, ManuallyRejected = Value
+
+  def fromString(status: Option[String]): Option[BatchChangeApprovalStatus] =
+    status match {
```
pauljamescleary: should do...

```scala
stats.map(_.toLowerCase) match { ...
```

this way you don't have to have the `if x.toLowerCase()` repeated below

pauljamescleary: An alternative implementation could be...

```scala
object BatchChangeApprovalStatus extends Enumeration {
  type BatchChangeApprovalStatus = Value
  val AutoApproved, PendingApproval, ManuallyApproved, ManuallyRejected = Value

  private val valueMap = BatchChangeApprovalStatus.values.map(v => v.toString.toLowerCase -> v).toMap

  def find(name: String): Option[BatchChangeApprovalStatus] = valueMap.get(name.toLowerCase)
}
```

The benefit is that you build the map when the enumeration is defined

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlBatchChangeRepository.scala
```diff
@@ -221,15 +221,25 @@ class MySqlBatchChangeRepository
...
-          userId.foreach(uid => sb.append(s"WHERE bc.user_id = '$uid' "))
+          (userId, approvalStatus) match {
+            case (Some(uid), Some(status)) =>
+              sb.append(
+                s"WHERE bc.user_id = '$uid' AND bc.approval_status = '${fromApprovalStatus(status)}'")
+            case (Some(uid), None) => sb.append(s"WHERE bc.user_id = '$uid'")
+            case (None, Some(status)) =>
+              sb.append(s"WHERE bc.approval_status = '${fromApprovalStatus(status)}'")
+            case _ => ()
+          }
```
pauljamescleary: You can simplify this using `mkString` which will auto separate everything...

```scala
val uid = userId.map(u => s"bc.user_id = '$u'")
val as = approvalStatus.map(a => s"bc.approval_status = '${fromApprovalStatus(a)}'")
val opts = uid ++ as // returns a List of each option, if the option is None it is omitted

sb.append("WHERE").append(opts.mkString(" AND "))
```

modules/core/src/main/scala/vinyldns/core/domain/batch/BatchChange.scala
britneywright: I wasn't sure how strict we want to be with the parameter values passed in. Should we be flexible and allow a user to pass in "pending" to get batch changes with the status PendingApproval, for example?

pauljamescleary: They should match, and be case insensitive.  For the most part, we control the client ecosystem, so it isn't a stretch to expect clients to behave especially if our docs are current and convey the appropriate values

modules/api/src/main/scala/vinyldns/api/route/BatchChangeRouting.scala
```diff
@@ -53,17 +55,28 @@ trait BatchChangeRoute extends Directives {
         parameters(
           "startFrom".as[Int].?,
           "maxItems".as[Int].?(MAX_ITEMS_LIMIT),
-          "ignoreAccess".as[Boolean].?(false)) {
-          (startFrom: Option[Int], maxItems: Int, ignoreAccess: Boolean) =>
+          "ignoreAccess".as[Boolean].?(false),
+          "approvalStatus".as[String].?) {
```
pauljamescleary: So, I just realized that `EnumNameSerializer` should take care of this, i.e. you should be able to do...

`"approvalStatus".as[BatchChangeApprovalStatus]`

The downside is that `EnumNameSerializer` is case sensitive.

If we made `EnumNameSerializer` case INsensitive, then the above would work and you could eliminate the `BatchChangeApprovalStatus.find` all together.

britneywright:
we can't override the Enumeration `withName` method which seems to be where we'd want to compare on lowercase values if we were trying to handle this with our serializers.

modules/api/src/main/scala/vinyldns/api/route/BatchChangeRouting.scala
```diff
@@ -53,17 +55,28 @@ trait BatchChangeRoute extends Directives {
...
+          (
+              startFrom: Option[Int],
+              maxItems: Int,
+              ignoreAccess: Boolean,
+              approvalStatus: Option[String]) =>
             {
+              val convertApprovalStatus = BatchChangeApprovalStatus.find(approvalStatus.mkString)
```
rebstar6: are we case sensitive with this search? do we want to be? just curious.

britneywright: yes, we compare the official status names with the given string, both lowercase 
 here: https://github.com/vinyldns/vinyldns/pull/715/files#diff-db7fcdf09cc4f6be4635a4351a706f56R79

modules/api/src/main/scala/vinyldns/api/route/BatchChangeRouting.scala
```diff
@@ -32,6 +33,7 @@ trait BatchChangeRoute extends Directives {
 
   final private val MAX_ITEMS_LIMIT: Int = 100
 
+  case class BatchChangeApproval(approvalType: String) {}
```
rebstar6: is this case class used?

modules/api/src/main/scala/vinyldns/api/route/BatchChangeRouting.scala
rebstar6: the way we pass this in feels weird to me bc the option

`val convertApprovalStatus = approvalStatus.flatMap(BatchChangeApprovalStatus.find)` makes more sense

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlBatchChangeRepository.scala
```diff
@@ -256,7 +263,14 @@ class MySqlBatchChangeRepository
               .apply()
           val maxQueries = queryResult.take(maxItems)
           val nextId = if (queryResult.size <= maxItems) None else Some(startValue + maxItems)
-          BatchChangeSummaryList(maxQueries, startFrom, nextId, maxItems)
+          val ignoreAccess = userId.isDefined
```
rebstar6: isnt this the opposite? if userId is defined, we are not ignoring access?

modules/api/src/main/scala/vinyldns/api/route/BatchChangeRouting.scala
britneywright: oops think that was a mistake

PR #716, opened 2019-06-28
Add Rejected, NeedsReview status

rebstar6:
Needs review for a change that did not pass validations and is being saved in pending.
Rejected will be used when ManualRejection is called on a batch


modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeProtocol.scala
```diff
@@ -53,7 +53,7 @@ final case class AddChangeInput(
       typ,
       knownTtl,
       record,
-      SingleChangeStatus.Pending,
+      SingleChangeStatus.NeedsReview,
```
britneywright: can take out the todo comment on line 45 now

modules/core/src/main/scala/vinyldns/core/domain/batch/SingleChange.scala
```diff
@@ -105,10 +105,12 @@ final case class SingleDeleteChange(
  - Pending has not yet been processed
  - Complete has been processed - must have recordChangeId
  - Failed had some error (see systemMessage) - may have recordChangeId, not required
+ - NeedsReview means there was a validation error and it needs manual review
+ - Rejected means the reviewer has rejected this batch change
```
britneywright: don't think these comments are needed now

britneywright: oops never mind,  thought they were part of the other file

PR #717, opened 2019-06-28
WIP - Generic Task Scheduler

pauljamescleary:
WIP - creates a more general task scheduler so it can be reused for other things (like scheduling batch changes).


modules/core/src/main/scala/vinyldns/core/task/TaskScheduler.scala
```diff
@@ -0,0 +1,65 @@
...
+          }
+        }
+        .handleErrorWith { error =>
+          // Make sure that we restart our stream on failure
+          logger.error(s"""Unexpected error is task; taskName="${task.name}" """, error)
+          scheduledTaskStream()
+        }
+
+    // Return the Fiber so the caller can eventually cancel it
+    scheduledTaskStream().compile.drain.start
```
rebstar6: this is just a style question mostly - in the commandhandler, we do the `start` call in Boot rather than in the run function itself. any reason to have it here vs there?

pauljamescleary: I changed this up, and made the return a `Stream` instead, leaving it up to the caller to get the `Fiber` so they can cancel.

modules/core/src/main/scala/vinyldns/core/task/TaskScheduler.scala
```diff
@@ -0,0 +1,80 @@
...
+              .releaseTask(t.name)
+              .as(logger.info(s"""Released task; taskName="${task.name}" """)))
+        .getOrElse(IO.unit)
+    }
+
+    def runTask(maybeTask: Option[Task]): IO[Unit] = maybeTask.map(_.run()).getOrElse(IO.unit)
+
+    // Acquires a task, runs it, and makes sure it is cleaned up, swallows the error via a log
+    def runOnceSafely(task: Task): IO[Unit] =
+      claimTask().bracket(runTask)(releaseTask).handleError { error =>
```
pauljamescleary: `.bracket` makes sure that the finalizer (releaseTask) gets called no matter what.

We want to swallow the error as well

modules/core/src/main/scala/vinyldns/core/task/TaskScheduler.scala
```diff
@@ -0,0 +1,80 @@
...
+        logger.error(s"""Unexpected error is task; taskName="${task.name}" """, error)
+      }
+
+    // Awake on the interval provided
+    Stream.awakeEvery[IO](task.runEvery).evalMap(_ => runOnceSafely(task))
```
pauljamescleary: Return a stream, allow the caller (usually `Boot`) to be able to start up and shut down the stream.

Usually, this will be `schedule(...).compile.drain.start`

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlTaskRepository.scala
```diff
@@ -52,12 +54,11 @@ class MySqlTaskRepository extends TaskRepository {
...
   def claimTask(name: String, pollingInterval: FiniteDuration): IO[Boolean] =
     IO {
-      val pollingExpirationHours = pollingInterval.toHours * 2
-      val currentTime = DateTime.now
+      val currentTime = Instant.now
       DB.localTx { implicit s =>
         val updateResult = CLAIM_UNCLAIMED_TASK
           .bindByName(
-            'updatedTimeComparison -> currentTime.minusHours(pollingExpirationHours.toInt),
+            'updatedTimeComparison -> currentTime.minusMillis(pollingInterval.toMillis),
```
pauljamescleary: will rename `pollingInterval` to be `taskTimeout` instead

rebstar6: wow, hours to millis is a big jump. do we need that much specificity?

pauljamescleary: @rebstar6 Ultimately it is up to the caller / Task implementer to make sure the taskTimeout is a sane value based on their Task implementation.  For example, UserSyncTask uses 24.hours, but I could use this for scheduled batch changes and use 5 minutes.    

`FiniteDuration` is the type of argument passed in, but before was doing `Hours` because of the half generic half coupled to UserSync.  Because of `FiniteDuration` we cannot make assumptions of what the user intended to pass in.  This fixes the assumption of "hours" while making the task scheduler generic.

modules/core/src/main/scala/vinyldns/core/task/TaskScheduler.scala
```diff
@@ -0,0 +1,100 @@
...
+    * Schedules a task to be run.  Will insert the task into the TaskRepository if it is not
+    * already present.  The result is a Stream that the caller has to manage the lifecycle for.
+    *
+    * @example
+    *          val scheduledStream = TaskScheduler.schedule(...)
+    *          val handle = scheduledStream.compile.drain.start.unsafeRunSync()
+    *          ...
+    *          // once everything is done you can cancel it via the handle
+    *          handle.cancel.unsafeRunSync()
+    * @return a Stream that when run will awake on the interal defined on the Task provided, and
```
mitruly: `interal` → `interval`

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlTaskRepository.scala
```diff
@@ -50,14 +52,19 @@ class MySqlTaskRepository extends TaskRepository {
       | WHERE name = {name}
       """.stripMargin
 
-  def claimTask(name: String, pollingInterval: FiniteDuration): IO[Boolean] =
+  private val PUT_TASK =
+    sql"""
+         |INSERT IGNORE INTO task(name, in_flight, created, updated)
+         |VALUES ({taskName}, 0, NOW(), NULL)
```
mitruly: Main concern here is mixing `DateTime.now` with SQL `NOW()`, which could have a time difference depending on server configurations.

pauljamescleary: I agree, I am making _everything_ use `NOW()` and the datetime in MySQL instead of passing it around.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlTaskRepository.scala
pauljamescleary: actually this went away, so resolving

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlTaskRepository.scala
```diff
@@ -70,7 +75,15 @@ class MySqlTaskRepository extends TaskRepository {
 
   def releaseTask(name: String): IO[Unit] = IO {
     DB.localTx { implicit s =>
-      UNCLAIM_TASK.bindByName('currentTime -> DateTime.now, 'name -> name).update().apply()
+      UNCLAIM_TASK.bindByName('taskName -> name).update().apply()
+    }
+  }
+
+  // Save the task, do not overwrite if it is already there
+  def saveTask(name: String): IO[Unit] = IO {
```
mitruly: If we are using `saveTask` to perform the initial task insertion, we should remove this from `V3.15__Tasks.sql`.

modules/core/src/main/scala/vinyldns/core/task/TaskScheduler.scala
```diff
@@ -0,0 +1,103 @@
...
+        case true =>
+          logger.info(s"""Successfully found and claimed task; taskName="${task.name}" """)
+          Some(task)
+        case false =>
+          logger.info(s"""No task claimed; taskName="${task.name}" """)
+          None
+      }
+    }
+
+    def releaseTask(maybeTask: Option[Task]): IO[Unit] = IO.suspend {
```
rebstar6: why is this in IO.suspend but runTask isnt?

pauljamescleary: This is because a bug in our cats effect version...

https://github.com/typelevel/cats-effect/pull/421

I could also upgrade our cats effect version, but was a little hesitant on that

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlTaskRepository.scala
pauljamescleary: @mitruly one can never modify migrations that have already been run.  Will leave it there as it really doesn't hurt.  Always the friendly reminder _never modify database migrations once they have been run_

PR #718, opened 2019-06-28
Ldap Syncer using Generic TaskScheduler

### LINKED ISSUE #523: Sync users against LDAP

pauljamescleary:
closes #523


modules/portal/app/modules/VinylDNSModule.scala
```diff
@@ -56,12 +62,28 @@ class VinylDNSModule(environment: Environment, configuration: Configuration)
...
+              new UserAccountAccessor(
+                repositories.userRepository,
+                repositories.userChangeRepository),
+              auth,
+              settings.ldapSyncPollingInterval),
+            repositories.taskRepository
+          )
+          .compile
+          .drain
+          .start
```
pauljamescleary: We now use the `TaskScheduler` to schedule the task itself

modules/portal/app/tasks/UserSyncTask.scala
```diff
@@ -0,0 +1,48 @@
...
+
+import cats.effect.IO
+import controllers.{Authenticator, UserAccountAccessor}
+import org.slf4j.{Logger, LoggerFactory}
+import vinyldns.core.domain.membership.LockStatus
+import vinyldns.core.task.Task
+
+import scala.concurrent.duration._
+
+class UserSyncTask(
```
pauljamescleary: All this code was copied from @mitruly PR, just framed it in a `Task` implementation

modules/portal/test/tasks/UserSyncTaskSpec.scala
```diff
@@ -0,0 +1,66 @@
...
+  val mockUserAccountAccessor: UserAccountAccessor = {
+    val mockObject = mock[UserAccountAccessor]
+    mockObject.getAllUsers.returns(IO(List(notAuthUser)))
+    mockObject
+      .lockUsers(List(notAuthUser))
+      .returns(IO(List(lockedNotAuthUser)))
+    mockObject
+  }
+
+  "SyncUserTask" should {
```
pauljamescleary: had to add some verifications in here as @mitruly specs weren't testing much

britneywright:
Is there a scenario where we'd want to unlock a user that was previously locked if they turn up in a subsequent ldap sync? If not should we exclude those users from the subsequent ldap syncs? I believe this searches across all users every time, not just the ones that are currently unlocked.

rebstar6:
@britneywright noticed the same on MLYs PR - we protobuf users, and locked is within the protobuf. as a result, without a migration, we cant search the DB just for unlocked users :/

britneywright:
We should also update the Operator Guide either in this PR or make a new issue so we don't forgot to in a separate PR https://www.vinyldns.io/operator/setup-ldap

PR #719, opened 2019-07-01
Reject batch change [updated]

britneywright:
rebase of #693 while @mitruly is on vacation 🍹😎

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
@@ -397,8 +397,10 @@ class BatchChangeService(
     } yield listWithGroupNames
   }
 
-  def rejectBatchChange(batchChange: BatchChange, reviewComment: Option[String], reviewerId: String):
-    BatchResult[BatchChange] = {
+  def rejectBatchChange(
+      batchChange: BatchChange,
+      reviewComment: Option[String],
+      reviewerId: String): BatchResult[BatchChange] = {
     val rejectedSingleChanges = batchChange.changes.map {
       case sad: SingleAddChange => sad.copy(status = SingleChangeStatus.Rejected)
       case sdc: SingleDeleteChange => sdc.copy(status = SingleChangeStatus.Rejected)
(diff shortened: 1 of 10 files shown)


modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -432,4 +436,25 @@ class BatchChangeService(
         ignoreAccess = ignoreAccess)
     } yield listWithGroupNames
   }
+
+  def rejectBatchChange(
+      batchChange: BatchChange,
+      reviewComment: Option[String],
+      reviewerId: String): BatchResult[BatchChange] = {
+    val rejectedSingleChanges = batchChange.changes.map {
+      case sad: SingleAddChange => sad.copy(status = SingleChangeStatus.Rejected)
```
pauljamescleary: I don't think you need the pattern match here.

```scala
val rejectedChanges = batchChange.changes.map(_.copy(status = SingleChangeStatus.Rejected))
```

britneywright: actually that results in an error ` value copy is not a member of vinyldns.core.domain.batch.SingleChange`

pauljamescleary: Ugh, that is annoying.  Ok.

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -432,4 +436,25 @@ class BatchChangeService(
...
+      case sdc: SingleDeleteChange => sdc.copy(status = SingleChangeStatus.Rejected)
```
pauljamescleary: Could this be on the `SingleChange` trait itself?

```
sealed trait SingleChange {
  def reject: SingleChange = this match { 
    case sad: SingleAddChange => sad.copy(status = SingleChangeStatus.Rejected)
    case sdc: SingleDeleteChange => sdc.copy(status = SingleChangeStatus.Rejected)
  }
}

...

val rejectedSingleChanges = batchChange.changes.map(_.reject)
```

I also wonder if it is better just to have a `reject` on the `BatchChange` itself that takes care of this?

PR #723, opened 2019-07-02
Apex cname error message

britneywright:
branched off of #702 

Changes in this pull request:
- create CnameAtApexError message and use it instead of RecordAlreadyExists error message when a CNAME record creation is attempted at a zone apex.

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/DomainValidationErrors.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/DomainValidationErrors.scala
@@ -167,7 +167,7 @@ final case class NewMultiRecordError(change: AddChangeForValidation) extends Dom
       .replaceAll("\n", " ")
 }
 
-final case class CnameAtZoneApexError() extends DomainValidationError {
-  def message: String = "CNAME at the zone apex is not allowed."
+final case class CnameAtZoneApexError(zoneName: String) extends DomainValidationError {
+  def message: String = s"CNAME cannot be the same name as zone $zoneName"
 }
 // $COVERAGE-ON$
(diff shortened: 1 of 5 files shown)


modules/api/src/main/scala/vinyldns/api/domain/DomainValidationErrors.scala
```diff
@@ -167,4 +167,7 @@ final case class NewMultiRecordError(change: AddChangeForValidation) extends Dom
       .replaceAll("\n", " ")
 }
 
+final case class CnameAtZoneApexError() extends DomainValidationError {
+  def message: String = "CNAME at the zone apex is not allowed."
```
pauljamescleary: I would like to re-word this for people who don't know DNS that well.

`CNAME cannot be the same name as the zone it lives in`?

The wording is tricky, but we want to convey that the CNAME cannot be the same as the name of the zone.

Also, should include the name of the CNAME or zone name in the message

modules/api/src/main/scala/vinyldns/api/domain/DomainValidationErrors.scala
```diff
@@ -166,4 +166,8 @@ final case class NewMultiRecordError(change: AddChangeForValidation) extends Dom
        |type ${change.inputChange.typ}.""".stripMargin
       .replaceAll("\n", " ")
 }
+
+final case class CnameAtZoneApexError(zoneName: String) extends DomainValidationError {
+  def message: String = s"CNAME cannot be the same name as zone $zoneName"
```
mitruly: Nitpick: Would like to be more consistent with other error messages and have `$zoneName` enclosed in double quotes or something. It would also help with indicating that the zone name is dot-terminated.

PR #727, opened 2019-07-02
Fixing task scheduler running too often

pauljamescleary:
* `TaskRepository.claimTask` - modified to take a polling interval in addition to the task timeout
* `MySqlTaskRepository` - this is where the bug lies.  Before we were claiming the task if it WAS NOT in_flight and if the task had not expired.  This meant that every node if it started at a different time would run the task.  Changed the SQL so that we check if the task is NOT in_flight that it also hasn't been executed within the last pollingInterval.  *double check this guy*
* `TaskScheduler` - small fix to auto execute the task on task scheduler start up.  This is useful for tasks with long polling intervals.

### DIFF
--- a/modules/core/src/main/scala/vinyldns/core/task/TaskScheduler.scala
+++ b/modules/core/src/main/scala/vinyldns/core/task/TaskScheduler.scala
@@ -59,7 +59,7 @@ object TaskScheduler extends Monitored {
       cs: ContextShift[IO]): Stream[IO, Unit] = {
 
     def claimTask(): IO[Option[Task]] =
-      taskRepository.claimTask(task.name, task.timeout).map {
+      taskRepository.claimTask(task.name, task.timeout, task.runEvery).map {
         case true =>
           logger.info(s"""Successfully found and claimed task; taskName="${task.name}" """)
           Some(task)
@@ -90,10 +90,11 @@ object TaskScheduler extends Monitored {
       }
 
     // We must first schedule the task in the repository and then create our stream
+    // After it exists, we run it once when we first start the stream
     // Semantics of repo.scheduleTask are idempotent, if the task already exists we are ok
     // Then we run our scheduled task with awakeEvery
     Stream
-      .eval(taskRepository.saveTask(task.name))
+      .eval(taskRepository.saveTask(task.name).flatMap(_ => runOnceSafely(task)))
       .flatMap { _ =>
         Stream
           .awakeEvery[IO](task.runEvery)
--- a/modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlTaskRepository.scala
+++ b/modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlTaskRepository.scala
@@ -39,8 +39,8 @@ class MySqlTaskRepository extends TaskRepository {
     sql"""
       |UPDATE task
       |   SET in_flight = 1, updated = NOW()
-      | WHERE (in_flight = 0 OR updated < DATE_SUB(NOW(),INTERVAL {timeoutSeconds} SECOND))
-      |   AND (updated IS NULL OR updated < DATE_SUB(NOW(),INTERVAL {pollingInterval} SECOND))
+      | WHERE (in_flight = 0 AND (updated IS NULL OR updated < DATE_SUB(NOW(),INTERVAL {pollingInterval} SECOND)))
+      |   OR (updated < DATE_SUB(NOW(),INTERVAL {timeoutSeconds} SECOND))
       |   AND name = {taskName};
       """.stripMargin
 
--- a/modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlTaskRepository.scala
+++ b/modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlTaskRepository.scala
@@ -62,6 +62,8 @@ class MySqlTaskRepository extends TaskRepository {
     * Note - the column in MySQL is datetime with no fractions, so the best we can do is seconds
     * If taskTimeout is less than one second, this will never claim as
     * FiniteDuration.toSeconds results in ZERO OL for something like 500.millis
+    * Enforces the invariant that pollingInterval < taskTimeout.  If pollingInterval > taskTimeout, the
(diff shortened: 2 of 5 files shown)


modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlTaskRepository.scala
```diff
@@ -28,16 +28,19 @@ class MySqlTaskRepository extends TaskRepository {
...
    */
   private val CLAIM_UNCLAIMED_TASK =
     sql"""
       |UPDATE task
       |   SET in_flight = 1, updated = NOW()
-      | WHERE (in_flight = 0
-      |    OR updated IS NULL
-      |    OR updated < DATE_SUB(NOW(),INTERVAL {timeoutSeconds} SECOND))
+      | WHERE (in_flight = 0 OR updated < DATE_SUB(NOW(),INTERVAL {timeoutSeconds} SECOND))
+      |   AND (updated IS NULL OR updated < DATE_SUB(NOW(),INTERVAL {pollingInterval} SECOND))
```
pauljamescleary: Oh, wow, what!

Ok, here it goes... 

1. `updated IS NULL` will be true ONLY after the initial save and before the first claim.  Because `updated` is NULL, cannot do any of the other date comparisons, so we MUST have that in there

2. `updated < DATE_SUB(...timeoutSeconds...)` - this situation means that the task timeout has expired.  In other words, `in_flight = 1` BUT the task is taking WAYYYY TOOO LONG to finish.  We assume in this situation that the task that was running it died.

3. `updated < DATE_SUB(...pollingInterval...)` - THIS situation is true only if the polling interval is exceeded.

Let's truth table this jawn out

| in_flight | updated | {timeout} | {pollingInterval} | outcome |
| ------------- | ------------- | ------------- | ------------- | ------------- |
| 0 | NULL | 10 seconds | 2 seconds | TRUE (easy) |
| 0 | 1 seconds ago | 10 seconds | 2 seconds | FALSE - polling interval has not elapsed |
| 0 | 4 seconds ago | 10 seconds | 2 seconds | TRUE - polling interval has elapsed |
| 0 | 11 seconds ago | 10 seconds | 2 seconds | TRUE- BOTH intervals have elapsed |
| 1 | 11 seconds ago | 10 seconds | 2 seconds | TRUE - timeout has elapsed |
| 1 | 9 seconds ago | 10 seconds | 2 seconds | FALSE - timeout has not elapsed yet |

modules/core/src/main/scala/vinyldns/core/task/TaskScheduler.scala
```diff
@@ -90,10 +90,11 @@ object TaskScheduler extends Monitored {
       }
 
     // We must first schedule the task in the repository and then create our stream
+    // After it exists, we run it once when we first start the stream
     // Semantics of repo.scheduleTask are idempotent, if the task already exists we are ok
     // Then we run our scheduled task with awakeEvery
     Stream
-      .eval(taskRepository.saveTask(task.name))
+      .eval(taskRepository.saveTask(task.name).flatMap(_ => runOnceSafely(task)))
```
pauljamescleary: this little guy will do a `claim / do / release` cycle once on startup before entering the polling.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlTaskRepository.scala
```diff
@@ -28,16 +28,19 @@ class MySqlTaskRepository extends TaskRepository {
...
+   *  Claims the task if the last time it was updated is older than the polling interval AND in_flight=0 or expired
    */
   private val CLAIM_UNCLAIMED_TASK =
     sql"""
       |UPDATE task
       |   SET in_flight = 1, updated = NOW()
-      | WHERE (in_flight = 0
-      |    OR updated IS NULL
-      |    OR updated < DATE_SUB(NOW(),INTERVAL {timeoutSeconds} SECOND))
+      | WHERE (in_flight = 0 OR updated < DATE_SUB(NOW(),INTERVAL {timeoutSeconds} SECOND))
```
rebstar6: we dont allow config for timeoutSeconds anywhere. probably should. Also what is the relationship between timeoutSeconds and pollingInterval? do we always want pollingInterval <= timeoutSeconds?

rebstar6: logically, if pollingInterval is ever > timeoutSeconds, in_flight no longer matters as `updated < DATE_SUB(NOW(),INTERVAL {timeoutSeconds} SECOND)` must be true if `updated < DATE_SUB(NOW(),INTERVAL {pollingInterval} SECOND)`

pauljamescleary: Yes, we do not enforce any invariants on the pollingInterval to timeoutSeconds.  That would have to be enforced on task creation.  We could force pollingInterval = timeoutSeconds if pollingInterval >= timeoutSeconds.

pauljamescleary: Added 2 test cases, force pollingInterval <= taskTimeout

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlTaskRepository.scala
rebstar6: it may be easier to follow if you did:

WHERE (in_flight = 0 AND (updated IS NULL OR updated < DATE_SUB(NOW(),INTERVAL {pollingInterval} SECOND))
OR updated < DATE_SUB(NOW(),INTERVAL {timeoutSeconds}

not sure if there are any mysql implications

mitruly: Agree with @rebstar6 that breaking out the conditions makes it more easily readable.

pauljamescleary: OK

PR #728, opened 2019-07-03
[WIP / DO NOT REVIEW] Store batch errors

rebstar6:
pushing up to review one piece of this, still needs changes

### DIFF
--- a/modules/core/src/main/scala/vinyldns/core/domain/DomainValidationErrors.scala
+++ b/modules/core/src/main/scala/vinyldns/core/domain/DomainValidationErrors.scala
@@ -14,16 +14,31 @@
  * limitations under the License.
  */
 
-package vinyldns.api.domain
+package vinyldns.core.domain
 
-import vinyldns.api.domain.batch.BatchTransformations.AddChangeForValidation
-import vinyldns.api.domain.batch.SupportedBatchChangeRecordTypes
 import vinyldns.core.domain.record.{RecordSet, RecordType}
 import vinyldns.core.domain.record.RecordType.RecordType
 
 // $COVERAGE-OFF$
 sealed abstract class DomainValidationError(val isFatal: Boolean = true) {
   def message: String
+}
+
+// Note: any errors deemed nonFatal will need to have a defined error code
+// also note, a code mapping CANNOT be changed once created
+object DomainValidationErrorCode extends Enumeration {
+  type DomainValidationErrorCode = Value
+  val Undefined, DVE01 = Value
+
+  def fromError(err: DomainValidationError): DomainValidationErrorCode = err match {
+    case _: ZoneDiscoveryError => DVE01
+    case _ => Undefined
+  }
+
+  def toErrorText(code: DomainValidationErrorCode): String = code match {
+    case DVE01 => "ZoneDiscoveryError"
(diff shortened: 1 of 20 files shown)


modules/core/src/main/scala/vinyldns/core/domain/DomainValidationErrors.scala
```diff
@@ -26,6 +24,23 @@ sealed abstract class DomainValidationError(val isFatal: Boolean = true) {
   def message: String
 }
 
+// Note: any errors deemed nonFatal will need to have a defined error code
```
rebstar6: ignore everything in this PR for now except this - most changes were because storing this stuff on a SingleChange (core) means the errors must also be in core (theyre in api right now). i may just make that move a PR in its own right.


anywho, this is the trick here - we do not want to store the full error text on a change. Im thinking we have some kind of enum for error codes. however, its only a one-way map since all of these error have parameters that cant be recovered. An alternative approach is to just have a `def errorCode` in the sealed abstract class and every error will have to define their code, but then youre possibly at risk of duplicating codes? that still also has the issue where you cant recover the original error bc of the params.

im torn.

pauljamescleary: My gut reaction is that we should move things to core and serialize.  tbf, everything from the "Service layer" on down can be in core.  The only thing that shouldn't be in core would be JSON particulars, Akka-http things, Mysql things, etc.  We are in this "move as we go" kind of mentality.  The end state should be all services, domain objects, validations, and repositories live in core.

Serializing the actual errors, some errors will need additional information to properly deserialize (like the zone name, change limit, etc), unless we store the entire message.  For example:

```scala
final case class ChangeLimitExceeded(limit: Int) extends DomainValidationError {
  def message: String = s"Cannot request more than $limit changes in a single batch change request"
}
```

Since the message is a `def` it doesn't have to be saved, just the `ChangeLimitExceeded` class and the limit (which will be pretty small).

Giving this about 10 minutes of thought, I am inclined  to move the errors to core and add them to the protobuf.

rebstar6: do we need to be able to serialize out? i was thinking of having SingleChange maintain a list of DomainValidationErrorCode rather than DomainValidationError. could go either way though.

britneywright: one thing i think we need is for Zone Discovery errors we'd probably want to be reporting the zone name that wasn't found, the other errors don't seem like they really require additional data beyond the error code/class name.

pauljamescleary: My hesitation with the error code approach is that there is an additional mapping layer between the saved state and the objects.  I am sure there are pros to this, the downside is that the disconnect presents a class of bugs where inadvertent code refactoring, moving around, or modification can inadvertently break the mapping.  Also, some errors (either now or in the future) may require saving of additional state, which the error code mapping could not support.  From my past experience in policy admin, saving state with errors was the norm, and is why I lean toward that eventuality.

rebstar6:
replacing with #743

PR #729, opened 2019-07-03
[outdated] Domain errors to core

rebstar6:
SingleChange is in core. will need these guys over there too if we want to be able to store anything related to the errors on a single change.

Functional changes for storing will be in followup PR


britneywright:
closing this. rebased and addressed the failing tests in #734

PR #730, opened 2019-07-03
add backend ID info to docs

### LINKED ISSUE #623: add zone backend ID info to documentation
Need to update the documentation to include the optional zone backend ID attribute implemented in: https://github.com/vinyldns/vinyldns/pull/522

britneywright:
Fixes #623.

Changes in this pull request:
- add zone backend ID info to the zone model docs
- add zone backend ID info to the configuration example

### DIFF
--- a/modules/docs/src/main/tut/api/zone-model.md
+++ b/modules/docs/src/main/tut/api/zone-model.md
@@ -31,7 +31,7 @@ account       | string      | **DEPRECATED** The account that created the zone |
 email         | string      | The distribution email for the zone |
 backendId     | string      | Optional. Recommended over `connection` and `transferConnection`. The configuration ID of the DNS backend server for the zone. If not provided, default keys will be used unless connection and transfer connection are provided. |
 connection    | ZoneConnection | Optional. The connection used to issue DDNS updates to the backend zone.  If not provided, default keys will be used unless backendId is provided.  See the [Zone Connection Attributes](#zone-conn-attr) for more information |
-transferConnection | ZoneConnection | Optional. The connection that is used to sync the zone with the DNS backend.  This can be different than the update connection.  If not provided, default keys will be used |
+transferConnection | ZoneConnection | Optional. The connection that is used to sync the zone with the DNS backend.  This can be different than the update connection.  If not provided, default keys will be used unless backendId is provided. |
 shared        | boolean     | An indicator that the zone is shared with anyone. At this time only VinylDNS administrators can set this to true.|
 acl           | ZoneACL     | The access control rules governing the zone.  See the [Zone ACL Rule Attributes](#zone-acl-rule-attr) for more information
 id            | string      | The unique identifier for this zone
@@ -231,6 +231,14 @@ The **IPv6** ACL Rule
 Will give Read permissions to PTR Record Sets 1000:1000:1000:1000:0000:0000:0000:0000 to 1000:1000:1000:1000:FFFF:FFFF:FFFF:FFFF, as 64 bits is half of an IPv6 address.
 
 #### ZONE CONNECTION ATTRIBUTES <a id="zone-conn-attr"></a>
+In order for VinylDNS to make updates in DNS, it needs key information for every zone. There are 3 ways to specify that key information; ask your VinylDNS admin which is appropriate for your zone based on the configuration of the service:
+
+1. Leave connection, transfer connection, and backendId blank: In this case, the default VinylDNS keys will be used
+2. Specify a backendID on the zone: if multiple backends are configured for your instance of VinylDNS, you can specify a backendID on the zone and the keys associated with that backend will be used.
+3. Specify ZoneConnection and TransferConnection on the zone itself: See below for details
+
+Note that if both a backendID and specific connection keys are included on a zone, the specific connection keys will be used.
+
 Zone Connection specifies the connection information to the backend DNS server.
 
 field        | type        | description |
--- a/modules/docs/src/main/tut/api/zone-model.md
+++ b/modules/docs/src/main/tut/api/zone-model.md
@@ -190,7 +190,7 @@ ACL rules can be applied to specific record types and can include record masks t
 apply to. These record masks apply to record names, but because PTR record names are part their reverse zone ip, the use of regular
 expressions for record masks are not supported.
 <br><br>
-Instead PTR record masks must be CIDR rules, which will denote a range of ip addresses that the rule will apply to.
+Instead PTR record masks must be CIDR rules, which will denote a range of IP addresses that the rule will apply to.
 While more information and useful CIDR rule utility tools can be found online, CIDR rules describe how many bits of an ip address' binary representation
 must be the same for a match.
 
--- a/modules/docs/src/main/tut/api/zone-model.md
+++ b/modules/docs/src/main/tut/api/zone-model.md
@@ -235,7 +235,7 @@ Zone Connection specifies the connection information to the backend DNS server.
 
 field        | type        | description |
 ------------ | :---------- | :---------- |
-primaryServer | string      | The ip address or host that is connected to.  This can take a port as well `127.0.0.1:5300`.  If no port is specified, 53 will be assumed. |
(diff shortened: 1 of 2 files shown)


modules/docs/src/main/tut/api/zone-model.md
```diff
@@ -249,6 +230,27 @@ The **IPv6** ACL Rule
 
 Will give Read permissions to PTR Record Sets 1000:1000:1000:1000:0000:0000:0000:0000 to 1000:1000:1000:1000:FFFF:FFFF:FFFF:FFFF, as 64 bits is half of an IPv6 address.
 
+#### ZONE CONNECTION ATTRIBUTES <a id="zone-conn-attr"></a>
```
britneywright: Do we need some type of backend ID vs zone connection explainer?

modules/docs/src/main/tut/api/zone-model.md
```diff
@@ -249,6 +230,27 @@ The **IPv6** ACL Rule
...
+Zone Connection specifies the connection information to the backend DNS server.
+
+field        | type        | description |
+------------ | :---------- | :---------- |
+primaryServer | string      | The ip address or host that is connected to.  This can take a port as well `127.0.0.1:5300`.  If no port is specified, 53 will be assumed. |
```
mitruly: `ip` → `IP`.

modules/docs/src/main/tut/api/zone-model.md
```diff
@@ -29,8 +29,9 @@ adminGroupId  | string      | The id of the administrators group for the zone |
 created       | date-time   | The time when the zone was first created |
 account       | string      | **DEPRECATED** The account that created the zone |
 email         | string      | The distribution email for the zone |
-connection    | ZoneConnection | The connection used to issue DDNS updates to the backend zone.  If not provided, default keys will be used.  See the [Zone Connection Attributes](#zone-conn-attr) for more information |
-transferConnection | ZoneConnection | The connection that is used to sync the zone with the DNS backend.  This can be different than the update connection.  If not provided, default keys will be used |
+backendId     | string      | Optional. Recommended over `connection` and `transferConnection`. The configuration ID of the DNS backend server for the zone |
```
mitruly: I think it would be beneficial to have a breakdown (eg. in a simple table) of the priority order for determining the DNS keys (source: `ZoneConnectionValidator.scala`) based on existence:
- Primary/transfer connection
- DNS backend connection
- Default connection

pauljamescleary: I agree, but probably not right here.  Perhaps as a separate section linked in here

modules/docs/src/main/tut/api/zone-model.md
pauljamescleary: Yep, I think that we should cover the different ways to connect:

1. Global default
2. BackendID
3. Zone level override

modules/docs/src/main/tut/api/zone-model.md
```diff
@@ -29,8 +29,9 @@ adminGroupId  | string      | The id of the administrators group for the zone |
...
+connection    | ZoneConnection | Optional. The connection used to issue DDNS updates to the backend zone.  If not provided, default keys will be used.  See the [Zone Connection Attributes](#zone-conn-attr) for more information |
```
mitruly: Based on the comment above (https://github.com/vinyldns/vinyldns/pull/730/files#r300079707), the following comment is not entirely correct:
> If not provided, default keys will be used

If this is not provided, VinylDNS will attempt to get the `backendId`. Agree with Paul that it might not be good to have the order tabled out, but we should be clear on this.

modules/docs/src/main/tut/operator/config-api.md
```diff
@@ -631,4 +631,23 @@ high-value-domains = {
       "fd69:27cc:fe91:0:0:0:ffff:0"
     ]
 }
+
+# Zone Connection Data
+backends = [
+    {
+      id = "func-test-backend"
```
rebstar6: use "test-backend-id" to name this in the docs

modules/docs/src/main/tut/api/zone-model.md
rebstar6: agreed on ^; add that generally the Global default or BackendID approach is recommended. from an operators perspective, Global default is required (bc legacy), BackendID support is not required

britneywright: Is global default required? In the operator guide currently it sounds as though it's optional: https://www.vinyldns.io/operator/config-api#default-zone-connections. I'm assuming this is the area we'd want to explain global default and backend IDs.

mitruly: Yes, global default is required; if not provided, an error will be encountered upon startup.

modules/docs/src/main/tut/api/zone-model.md
mitruly: This comment still stands.

britneywright: You mean if the backendId is provided? otherwise it would be the default keys. I'll update.

modules/docs/src/main/tut/api/zone-model.md
```diff
@@ -29,8 +29,9 @@ adminGroupId  | string      | The id of the administrators group for the zone |
...
-connection    | ZoneConnection | The connection used to issue DDNS updates to the backend zone.  If not provided, default keys will be used.  See the [Zone Connection Attributes](#zone-conn-attr) for more information |
-transferConnection | ZoneConnection | The connection that is used to sync the zone with the DNS backend.  This can be different than the update connection.  If not provided, default keys will be used |
+backendId     | string      | Optional. Recommended over `connection` and `transferConnection`. The configuration ID of the DNS backend server for the zone. If not provided, default keys will be used unless connection and transfer connection are provided. |
+connection    | ZoneConnection | Optional. The connection used to issue DDNS updates to the backend zone.  If not provided, default keys will be used unless backendId is provided.  See the [Zone Connection Attributes](#zone-conn-attr) for more information |
+transferConnection | ZoneConnection | Optional. The connection that is used to sync the zone with the DNS backend.  This can be different than the update connection.  If not provided, default keys will be used |
```
mitruly: The `ZoneConnection` > `backendId` > `defaultConnection` hierarchy also applies to `transferConnection`.

modules/docs/src/main/tut/api/zone-model.md
```diff
@@ -88,31 +89,11 @@ accessLevel   | string      | Access level of the user requesting the zone. Curr
   },
   "id": "9cbdd3ac-9752-4d56-9ca0-6a1a14fc5562",
   "latestSync": "2016-12-16T15:27:26Z",
+  "backendId":"func-test-backend",
```
rebstar6: hmm, actually this is kinda a bad example. in this case we have both backendId and zoneConnection stuff specified, which means this would be ignored. maybe give 2 examples?

modules/docs/src/main/tut/api/zone-model.md
```diff
@@ -249,6 +230,27 @@ The **IPv6** ACL Rule
 
 Will give Read permissions to PTR Record Sets 1000:1000:1000:1000:0000:0000:0000:0000 to 1000:1000:1000:1000:FFFF:FFFF:FFFF:FFFF, as 64 bits is half of an IPv6 address.
 
+#### ZONE CONNECTION ATTRIBUTES <a id="zone-conn-attr"></a>
+Zone Connection specifies the connection information to the backend DNS server.
```
rebstar6: I would just set this up with the context. something like:

In order for VinylDNS to make updates in DNS, it needs key information for every zone. There are 3 ways to specify that key information; ask your VinylDNS admin which is appropriate for your zone based on the configuration of the service:
1. Leave connection, transfer connection, and backendId blank: In this case, the default VinylDNS keys will be used
2. Specify a backendID on the zone: if multiple backends are configured for your instance of VinylDNS, you can specify a backendID on the zone and the keys associated with that backend will be used.
3. Specify ZoneConnection and TransferConnection on the zone itself: See below for details

Note that if both a backendID and specific connection keys are included on a zone, the specific connection keys will be used,

PR #735, opened 2019-07-10
Approve pending batch change flow

britneywright:
Part of #659 

When an authorized user approves a batch change we should update the batch change approval status and related fields, run the batch change through validations again and then process the validations. 

**Note:** The current scope of this PR is assuming the happy path, no new validation errors. Depending on feedback I can tackle the error handling here or break it out into another PR. Think it might also depend on what we're doing in #728 
 
Changes in this pull request:
- Update approved batch change's approvalStatus (ManuallyApproved), reviewerId, reviewComment and reviewTimestamp
- convert the SingleChanges in the batch change into ChangeInputs to rerun through validations and collect the the existing recordset and zone information for processing.
- assumed happy path, we send the batch change for processing


modules/core/src/test/scala/vinyldns/core/domain/batch/BatchChangeInputSpec.scala
```diff
@@ -63,7 +61,7 @@ class BatchChangeInputSpec extends WordSpec with Matchers {
       asAdd.recordName shouldBe None
       asAdd.inputName shouldBe "some.test.com."
       asAdd.typ shouldBe A
-      asAdd.ttl shouldBe VinylDNSConfig.defaultTtl
+      asAdd.ttl shouldBe 7200L
```
britneywright: Not sure where to put the config value to use in core

modules/core/src/main/scala/vinyldns/core/domain/batch/BatchChangeProtocol.scala
```diff
@@ -43,7 +42,7 @@ final case class AddChangeInput(
     extends ChangeInput {
 
   def asNewStoredChange(errors: NonEmptyList[DomainValidationError]): SingleChange = {
-    val knownTtl = ttl.getOrElse(VinylDNSConfig.defaultTtl)
+    val knownTtl = ttl.getOrElse(7200L)
```
britneywright: Not sure where to put the config value to use in core

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -83,14 +83,23 @@ class BatchChangeService(
...
-      changeForConversion <- buildResponse(batchChangeInput, validatedSingleChanges, auth).toBatchResult
+      changeForConversion <- buildOrGetBatchChange(batchChangeInput, validatedSingleChanges, auth)
       serviceCompleteBatch <- convertOrSave(
         changeForConversion,
         zoneMap,
         recordSets,
         batchChangeInput.ownerGroupId)
     } yield serviceCompleteBatch
 
+  def convertToBatchChangeInput(batchChange: BatchChange): BatchChangeInput = {
```
mitruly: `BatchChangeInput` is the input JSON deserialized into an internal data structure; it would be preferable to move the existing data validations in `applyBatchChange` logic into a shared function and create a new method that takes in a `BatchChange` instead of `BatchChangeInput` (since the batch itself will already exist in a pending state). `approveBatchChange` can then be changed to do something like:
```
  reValidatedBatchChange <- applyBatchChange(updatedBatchChange, authPrincipal)
```

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -287,6 +313,16 @@ class BatchChangeService(
     }
   }
 
+  def buildOrGetBatchChange(
```
mitruly: It seems like passing this in will skip `buildResponse` which will return a response based off of whether any of the single validations failed in the batch.

I think we want to be calling `buildResponse` in `applyBatchChange` regardless of whether we are creating a new batch or approving an existing one.

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -83,14 +83,23 @@ class BatchChangeService(
...
       serviceCompleteBatch <- convertOrSave(
         changeForConversion,
         zoneMap,
         recordSets,
         batchChangeInput.ownerGroupId)
     } yield serviceCompleteBatch
 
+  def convertToBatchChangeInput(batchChange: BatchChange): BatchChangeInput = {
+    val convertedChanges = batchChange.changes.map(_.asChangeInput)
+    BatchChangeInput(
```
rebstar6: its unclear to me why we are doing this

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -107,7 +116,24 @@ class BatchChangeService(
...
+      convertBatchChange = convertToBatchChangeInput(updatedBatchChange)
+      reValidatedBatchChange <- applyBatchChange(convertBatchChange, authPrincipal)
+    } yield reValidatedBatchChange
+
+  def approveBatchChange(
+      batchChange: BatchChange,
+      auth: AuthPrincipal,
+      review: Option[ApproveBatchChangeInput]): BatchResult[BatchChange] = {
+    val reviewComment = review.flatMap(b => b.reviewComment)
+    val updatedBatchChange =
```
rebstar6: dont flip this yet. it has to pass through all the validations before we can say it was manually approved. the flow should be:

- re-pass through the validations with some flag indicating that the batch is passing through with approval in there
- figure out how soft errors should be passed - for example, with that flag, a shared approved failure would be skipped, but zone discovery still needs to run
- if it makes it though everything, it becomes manuallyApproved, otherwise it stays pending.

this flow is tough, may need to rethink this one

rebstar6:
i think this will be replaced with #748 or related sets of PRs out of that WIP. Can probably close

PR #738, opened 2019-07-11
make zone discovery case insensitive

britneywright:
This is an issue brought to our attention by user. In batch change if the zone name in the input name is capitalized in any way it will fail zone discovery. record name is case insensitive and the zone should de as well. 

example of error through the portal:
Screen Shot 2019-07-11 at 11 30 19 AM

Changes in this pull request:
- downcase the entire ChangeInput input name before sending through validations

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeProtocol.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeProtocol.scala
@@ -83,7 +83,7 @@ object AddChangeInput {
       case PTR => inputName
       case _ => ensureTrailingDot(inputName)
     }
-    new AddChangeInput(transformName.toLowerCase, typ, ttl, record)
+    new AddChangeInput(transformName, typ, ttl, record)
   }
 }
 
@@ -93,7 +93,7 @@ object DeleteChangeInput {
       case PTR => inputName
       case _ => ensureTrailingDot(inputName)
     }
-    new DeleteChangeInput(transformName.toLowerCase, typ)
+    new DeleteChangeInput(transformName, typ)
   }
 }
 
--- a/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeProtocol.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeProtocol.scala
@@ -83,7 +83,7 @@ object AddChangeInput {
       case PTR => inputName
       case _ => ensureTrailingDot(inputName)
     }
-    new AddChangeInput(transformName, typ, ttl, record)
+    new AddChangeInput(transformName.toLowerCase, typ, ttl, record)
   }
 }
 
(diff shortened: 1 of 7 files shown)


modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeProtocol.scala
```diff
@@ -83,7 +83,7 @@ object AddChangeInput {
       case PTR => inputName
       case _ => ensureTrailingDot(inputName)
     }
-    new AddChangeInput(transformName, typ, ttl, record)
+    new AddChangeInput(transformName.toLowerCase, typ, ttl, record)
```
mitruly: While this does achieve a similar effect, the one subtle difference is that the zone name is always going to be stored as lower case.

To achieve the desired behavior and retain current behavior where input name is matched against the exact zone casing, updating `ExistingZones`' `zoneMap` and `getByName` to convert to lowercase may be a better approach.

britneywright: We should be saving the values lower cased.

rebstar6:
Is this the right solution here? i would argue we should do all of our comparisons case-insensitive (lowercase those), but ultimately save/apply what the user has given up

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeProtocol.scala
mitruly: I agree with @rebstar6 (https://github.com/vinyldns/vinyldns/pull/738#issuecomment-510888379) regarding this.

mitruly: This comment still stands.

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -222,9 +222,9 @@ class BatchChangeService(
     changes.mapValid { change =>
       change.typ match {
         case A | AAAA | CNAME | MX | TXT => forwardZoneDiscovery(change, zoneMap)
-        case PTR if validateIpv4Address(change.inputName).isValid =>
+        case PTR if validateIpv4Address(change.inputName.toLowerCase()).isValid =>
```
mitruly: Tested this out locally and this approach doesn't work since it is assuming that the zone name is stored as lowercase, which is not necessarily true:
- Spin up local instance of VinylDNS and connect to zone `OK.`
- Create a new batch change with A record pointing to `lowercase.OK.`
- A zone discovery error is encountered upon submission

The suggested implementation at https://github.com/vinyldns/vinyldns/pull/738#discussion_r303189761 will circumvent this since the `zoneMap` is a map between the lowercase version of the zone name while preserving the zone name casing. It is also backwards-compatible and requires fewer `.toLowerCase` changes throughout existing code.

PR #739, opened 2019-07-11
Properly reject requests sent to non-existent routes

### LINKED ISSUE #388: GET /not-supported returns unauthorized
**Describe the bug**
Go to `/not-here` in the api - it is passing along to our authorization as if this is an authorized route, and you get a 401 with `Authentication Failed: Authorization header not found`

This should respond with whatever the default not found is for akka http (404?), and the request should not be passed along to the AuthPrinciple logic

mitruly:
Presently, we wrap all of our authenticated routes (from `VinylDNSService`) with a directive that verifies authentication and processes the result before route resolution. This causes users to encounter a `403 Forbidden` response even if navigating to a non-existent route. This PR delegates to the authentication step after route resolution.

Fixes #388.

Changes in this pull request:
- Refactor existing routes. Combined authenticate step with existing `execute` function. Created two separate helpers: `authenticateAndExecute` and `authenticateAndExecuteWithEntity` for authenticating routes without and with a request body, respectively.
- Convert standard `404 Not Found` response to `405 MethodNotAllowed` to avoid confusion with existing API response when retrieving data that may not exist.


modules/api/src/main/scala/vinyldns/api/route/BatchChangeRouting.scala
```diff
@@ -130,4 +129,24 @@ trait BatchChangeRoute extends Directives {
         complete(StatusCodes.BadRequest, bcnpa.message)
       case Left(uce: UnknownConversionError) => complete(StatusCodes.InternalServerError, uce)
     }
+
+  /**
+    * Helpers for handling authentication, invoking service call and then generating a response back to user
+    */
+  private def authenticateAndExecute[A](f: AuthPrincipal => BatchResult[A])(g: A => Route): Route =
```
mitruly: This is copied to the four routing files since `BatchChangeRouting` doesn't utilize a `Throwable`, meaning that we have a generic helper defined in `VinylDNSService`. Seemed possibly overkill to create shared, generic methods but would be interested in feedback about this.

Flow:
- Perform authentication checks. Upon failures, handle rejections; otherwise, invoke service call.
- Hand off service call result to response handler (which maps `Either` to a response).

modules/api/src/main/scala/vinyldns/api/route/BatchChangeRouting.scala
```diff
@@ -130,4 +129,24 @@ trait BatchChangeRoute extends Directives {
...
+    handleRejections(validationRejectionHandler)(authenticate { authPrincipal =>
+      onSuccess(f(authPrincipal).value.unsafeToFuture()) { result =>
+        sendResponse(result, g)
+      }
+    })
+
+  private def authenticateAndExecuteWithEntity[A, B](f: (AuthPrincipal, B) => BatchResult[A])(
+      g: A => Route)(implicit um: FromRequestUnmarshaller[B]): Route =
+    handleRejections(validationRejectionHandler)(authenticate { authPrincipal =>
```
mitruly: Flow:
- Perform authentication checks. Upon failures, handle rejections; otherwise, deserialize request entity into expected JSON.
- If unmarshalling succeeds, pass deserialized entity into service call.
- Hand off service call result to response handler (which maps Either to a response).

modules/api/src/main/scala/vinyldns/api/route/MembershipRouting.scala
```diff
@@ -32,156 +33,139 @@ trait MembershipRoute extends Directives {
...
-            execute(membershipService.deleteGroup(groupId, authPrincipal)) { group =>
-              complete(StatusCodes.OK, GroupInfo(group))
-            }
+      }
+  } ~
+    path("groups") {
+      post {
+        monitor("Endpoint.createGroup") {
+          authenticateAndExecuteWithEntity[Group, CreateGroupInput] { (authPrincipal, input) =>
+            val group = Group(
```
mitruly: We did not need to have a `Group.build` method since JSON unmarshalling and verification is handled by spray-json4s.

Opted to pull `CreateGroupInput` into `core` library since it requires `UserInfo` which is strictly used for the API JSON response.

modules/api/src/main/scala/vinyldns/api/route/VinylDNSService.scala
```diff
@@ -158,9 +157,30 @@ class VinylDNSService(
...
+      .newBuilder()
+      .handle {
+        case MalformedRequestContentRejection(msg, MappingException(_, _)) =>
+          complete(
+            HttpResponse(
+              status = StatusCodes.BadRequest,
+              entity = HttpEntity(ContentTypes.`application/json`, msg)
+            ))
+      }
+      .handleNotFound {
```
mitruly: Converts `404 NotFound` to `405 MethodNotAllowed`.

britneywright: Why are we changing the 404 to 405?

mitruly: @britneywright This is from a recent comment on the issue: https://github.com/vinyldns/vinyldns/issues/388#issuecomment-498777933. The reasoning is that we want to distinguish a non-existent route from a `NotFound` response for something like fetching a zone.

britneywright:
I don't see a reason to use `405` over `404`s. 404 is the standard response code for any request that isn't found.

britneywright:
I also found this fancy diagram if we want to discuss response codes more https://raw.githubusercontent.com/for-GET/http-decision-diagram/master/httpdd.png

modules/api/src/main/scala/vinyldns/api/route/JsonValidation.scala
```diff
@@ -30,26 +30,9 @@ import org.json4s.ext._
 import org.json4s.jackson.JsonMethods._
 
 import scala.reflect.ClassTag
-import cats.data._
-import cats.implicits._
 
 case class JsonErrors(errors: List[String])
 
-trait JsonValidationRejection extends Directives {
```
mitruly: This got pulled into `VinylDNSDirectives` since there's only one default validation rejection handler.

modules/api/src/main/scala/vinyldns/api/route/VinylDNSDirectives.scala
```diff
@@ -124,4 +148,84 @@ trait VinylDNSDirectives extends Directives {
       case Invalid(errors) =>
         reject(ValidationRejection(compact(render("errors" -> errors.toList.toSet))))
     }
+
+  /**
+    * Helpers to handle route authentication flow for routing. Implementing classes/objects
+    * must provide sendResponse implementation.
+    */
+  trait AuthenticationResultImprovements {
```
mitruly: Pulled authentication directive into `VinylDNSDirectives` as a trait so that `sendResponse` can be implemented individually for each route.

modules/api/src/main/scala/vinyldns/api/route/VinylDNSDirectives.scala
```diff
@@ -124,4 +148,84 @@ trait VinylDNSDirectives extends Directives {
...
+      authenticate { authPrincipal =>
+        entity(as[B]) { deserializedEntity =>
+          onSuccess(f(authPrincipal, deserializedEntity).value.unsafeToFuture()) { result =>
+            sendResponse(result, g)
+          }
+        }
+      }
+  }
+
+  trait AuthenticationBatchResultImprovements {
```
rebstar6: why is this a trait if its only used by the batch route

modules/api/src/main/scala/vinyldns/api/route/MembershipRouting.scala
rebstar6: i dont love this in the route, membershipService.createGroup should really just take a CreateGroupInput

thats likely not part of this PR though

modules/api/src/main/scala/vinyldns/api/route/VinylDNSDirectives.scala
mitruly: I can remove the trait and just have the object in the batch change route.

modules/api/src/test/scala/vinyldns/api/route/GenericRoutingSpec.scala
```diff
@@ -0,0 +1,44 @@
...
+
+package vinyldns.api.route
+import akka.http.scaladsl.model.StatusCodes
+import akka.http.scaladsl.server.Route
+import akka.http.scaladsl.testkit.ScalatestRouteTest
+import org.scalatest.mockito.MockitoSugar
+import org.scalatest.{Matchers, WordSpec}
+import vinyldns.api.domain.zone.ZoneServiceAlgebra
+
+class GenericRoutingSpec
```
rebstar6: probably can put this in VinylDNSDirectivesSpec

modules/api/src/main/scala/vinyldns/api/route/MembershipRouting.scala
mitruly: Yeah, I initially tried that but realized it involved changing `UserInfo` and some other stuff so I opted for parity.

modules/api/src/main/scala/vinyldns/api/route/VinylDNSDirectives.scala
rebstar6: scratch that, i see it. i think you can add a type here and it's generic:

AuthenticationResultImprovements[E] where most of them are AuthenticationResultImprovements[Throwable]. you would have to switch Result to be EitherT[IO, E, A] so it will work with both

modules/api/src/main/scala/vinyldns/api/route/VinylDNSDirectives.scala
```diff
@@ -33,7 +36,25 @@ import scala.util.control.NonFatal
...
+        case MalformedRequestContentRejection(msg, MappingException(_, _)) =>
+          complete(
+            HttpResponse(
+              status = StatusCodes.BadRequest,
+              entity = HttpEntity(ContentTypes.`application/json`, msg)
+            ))
+      }
+      .handleNotFound {
+        extractUnmatchedPath { p =>
+          complete((StatusCodes.NotFound, s"The requested path [$p] does not exist."))
```
mitruly: Changing from `405 MethodNotAllowed` to `404 NotFound` since it seems that is the most appropriate for this scenario. Some information from airbrake.io:
> a `405 Method Not Allows` indicates that the user agent (the web browser, in most cases) has requested a valid resource using an invalid HTTP method.
(source: https://airbrake.io/blog/http-errors/405-method-not-allowed)

modules/api/functional_test/live_tests/authentication_test.py
```diff
@@ -27,3 +28,13 @@ def test_request_succeeds_when_user_is_found_and_not_locked():
     client = VinylDNSClient(VinylDNSTestContext.vinyldns_url, 'okAccessKey', 'okSecretKey')
 
     client.list_batch_change_summaries(status=200)
+
+def test_request_fails_when_accessing_non_existent_route():
+    """
+    Test request fails with MethodNotAllowed (405) when route cannot be resolved, regardless of authentication
```
britneywright: update the status code in this description

pauljamescleary:
This seems like a ton of changes to work around changing the outcome of the route.

Each route can wind up with a successful response, or a List of Rejections.  The `~` chains routes together, and will pass along the list of rejections from the prior route.  This typically will be method or resource not found.  If the downstream route can handle the route, it will wipe the rejections and return the response.  It can also wipe the rejections if it has a super-ceding rejection.

Seems like we should be able to do something with `handleRejections` where we pattern match on the rejection itself.  There is also "seal route".

@mitruly have you explored alternatives that take a simpler approach to managing rejections?  Is it impossible to do?

pauljamescleary:
I think this could be the way we are building our authenticated routes.  For example, we authenticate **before** we match on the path `get & path("zones")`; which seems backwards.

We should be matching on the path first, and then doing the authentication.  That will cause the proper rejection of `NotFound` or `MethodNotAllowed`

Instead of ...

```scala
val zoneRoute = { authPrincipal: AuthPrincipal =>
    (post & path("zones") & monitor("Endpoint.createZone")) {
      entity(as[CreateZoneInput]) { createZoneInput =>
        execute(zoneService.connectToZone(encrypt(createZoneInput), authPrincipal)) { chg =>
          complete(StatusCodes.Accepted, chg)
        }
      }
```

we should have

```scala
val zoneRoute = {
    (post & path("zones") & monitor("Endpoint.createZone")) { 
      authenticate { authPrincipal: AuthPrincipal =>
        entity(as[CreateZoneInput]) { createZoneInput =>
          execute(zoneService.connectToZone(encrypt(createZoneInput), authPrincipal)) { chg =>
            complete(StatusCodes.Accepted, chg)
          }
        }
  }
```

the subtle change nests `authenticate` inside the `post & path` and will never get invoked if the outer path does not match

pauljamescleary:
As a final note (for now), we should review the ordering and behavior of the directives.

For example:

`get & path("/zones")` - this should fail with a `MethodNotAllowed` if we only support `post` for that path, and should fail with a `NotFound` if `get` is accepted but the path is `path("/foo")`

rebstar6:
@pauljamescleary the flip is actually in there - if you look at one route:
```
path("groups" / Segment) { groupId =>
    get {
      monitor("Endpoint.getGroup") {
        authenticateAndExecute(membershipService.getGroup(groupId, _)) { group =>
          complete(StatusCodes.OK, GroupInfo(group))
        }
      }
    } ~
...
```
the auth is happening in authenticateAndExecute, which is a helper for 
```
  def authenticateAndExecute[A](f: AuthPrincipal => EitherT[IO, E, A])(g: A => Route): Route =
    authenticate { authPrincipal =>
      onSuccess(f(authPrincipal).value.unsafeToFuture()) { result =>
        sendResponse(result, g)
      }
    }
```


to your point, there is definitely a lot more refactoring in the PR than maybe needed, though at it's core the auth flip to be inside is happening

pauljamescleary:
@rebstar6 yes, that is what I am thinking. I think there is a lot more going on than is needed.

mitruly:
@pauljamescleary I don't think we need to explicitly seal the route (eg. `Route.seal(...)`) since we have an implicit rejection handler at the top-level scope (which we're including in `VinylDNSService` or `VinylDNSDirectives`)

(source: https://doc.akka.io/docs/akka-http/current/routing-dsl/routes.html#route-seal-method-to-modify-httpresponse)

I think the earlier discussions with @rebstar addresses the authentication order - there really is not a workaround to performing the route resolution first before the authentication; the issue before was that the authentication would fail, the authentication rejection handler would exit with a `401 Unauthorized` response, and that would bubble up to the top-level rejection handler. This refactor includes converging rejection handlers, which makes developing on these routes easier moving forward.

We can definitely implement the correct `405 MethodNotFound` for an invalid HTTP method on a valid resource.

modules/api/src/main/scala/vinyldns/api/route/VinylDNSDirectives.scala
```diff
@@ -124,4 +126,44 @@ trait VinylDNSDirectives extends Directives {
...
+    * Flow:
+    * - Authenticate user. Proceed if successful; otherwise return unauthorized error to user.
+    * - Deserialize request entity into expected data structure. Proceed if successful; otherwise
+    * return error to user.
+    * - Invoke service call, f, and return the response to the user.
+    */
+  def authenticateAndExecuteWithEntity[A, B](f: (AuthPrincipal, B) => EitherT[IO, E, A])(
+      g: A => Route)(implicit um: FromRequestUnmarshaller[B]): Route =
+    authenticate { authPrincipal =>
+      entity(as[B]) { deserializedEntity =>
```
pauljamescleary: probably should have `entity` first before `authenticate`

modules/api/src/main/scala/vinyldns/api/route/VinylDNSDirectives.scala
```diff
@@ -31,7 +33,7 @@ import scala.concurrent.duration._
 import scala.util.Failure
 import scala.util.control.NonFatal
 
-trait VinylDNSDirectives extends Directives {
+trait VinylDNSDirectives[E] extends Directives {
```
pauljamescleary: having this parameterized on the Error type is strange

modules/api/src/main/scala/vinyldns/api/route/VinylDNSDirectives.scala
mitruly: This can be changed if we feel strongly about this; the rationale was that if a user does not have proper authorization, we may not want to expose info about the endpoint.

modules/api/src/main/scala/vinyldns/api/route/VinylDNSDirectives.scala
mitruly: This is because `BatchChangeErrorResponse` is a trait which does not extend `Throwable` - a workaround could be to have `BatchChangeErrorResponse` extend `Throwable`.

modules/api/src/main/scala/vinyldns/api/route/RecordSetRouting.scala
```diff
@@ -35,115 +35,126 @@ case class ListRecordSetsResponse(
...
+      case Left(RecordSetAlreadyExists(msg)) => complete(StatusCodes.Conflict, msg)
+      case Left(ZoneInactiveError(msg)) => complete(StatusCodes.BadRequest, msg)
+      case Left(NotAuthorizedError(msg)) => complete(StatusCodes.Forbidden, msg)
+      case Left(ZoneUnavailableError(msg)) => complete(StatusCodes.Conflict, msg)
+      case Left(RecordSetNotFoundError(msg)) => complete(StatusCodes.NotFound, msg)
+      case Left(InvalidRequest(msg)) => complete(StatusCodes.UnprocessableEntity, msg)
+      case Left(PendingUpdateError(msg)) => complete(StatusCodes.Conflict, msg)
+      case Left(RecordSetChangeNotFoundError(msg)) => complete(StatusCodes.NotFound, msg)
+      case Left(InvalidGroupError(msg)) => complete(StatusCodes.UnprocessableEntity, msg)
+      case Left(e) => failWith(e)
```
pauljamescleary: we have this "catch all", that does not need to be implemented by all `sendResponse` implementations as long as the return type is a `PartialFunction`, then when it runs you can do `sendResponse(...).orElse(defaultHandling)` (if my syntax is correct, maybe `andThen`

modules/api/src/main/scala/vinyldns/api/route/VinylDNSDirectives.scala
```diff
@@ -124,4 +126,44 @@ trait VinylDNSDirectives extends Directives {
       case Invalid(errors) =>
         reject(ValidationRejection(compact(render("errors" -> errors.toList.toSet))))
     }
+
+  /**
+    * Helpers to handle route authentication flow for routing. Implementing classes/objects
+    * must provide sendResponse implementation.
+    */
+  def sendResponse[A](either: Either[E, A], f: A => Route): Route
```
pauljamescleary: can add the error type parameter to here instead

modules/api/src/main/scala/vinyldns/api/route/VinylDNSDirectives.scala
```diff
@@ -124,4 +126,44 @@ trait VinylDNSDirectives extends Directives {
...
+  /**
+    * Authenticate user and execute service call without request entity
+    *
+    * Flow:
+    * - Authenticate user. Proceed if successful; otherwise return unauthorized error to user.
+    * - Invoke service call, f, and return the response to the user.
+    */
+  def authenticateAndExecute[A](f: AuthPrincipal => EitherT[IO, E, A])(g: A => Route): Route =
+    authenticate { authPrincipal =>
+      onSuccess(f(authPrincipal).value.unsafeToFuture()) { result =>
```
pauljamescleary: Yea, I suppose `sendResponse` could be just the error handlers, then in here you could do...

```scala
case Right(a) => g(a)
case Left(e) => handleErrors(e).orElse(defaultErrorHandler)
```

That would simplify the `sendResponse` requirement on users to just do `def handleErrors[E](e: E): PartialFunction[E, Route]`

Just thinking that the requirement on implementers for `sendResponse` could be more DRY

modules/api/src/main/scala/vinyldns/api/route/VinylDNSDirectives.scala
mitruly: If we move the error type parameter from the `trait`, we would also have to specify error type on `authenticateAndExecute` and `authenticateAndExecuteWithEntity` despite them being the same type. Is there an alternative to not having to provide the types for the authenticate helpers if we move the error type to the `handleErrors` helper?

modules/api/functional_test/live_tests/authentication_test.py
```diff
@@ -27,3 +28,23 @@ def test_request_succeeds_when_user_is_found_and_not_locked():
...
+    """
+    client = VinylDNSClient(VinylDNSTestContext.vinyldns_url, 'unknownAccessKey', 'anyAccessSecretKey')
+    url = urljoin(VinylDNSTestContext.vinyldns_url, u'/no-existo')
+    _, data = client.make_request(url, u'GET', client.headers, status=404)
+
+    assert_that(data, is_("The requested path [/no-existo] does not exist."))
+
+def test_request_fails_with_unsupported_http_method():
+    """
+    Test request fails with MethodNotAllowed (405) when HTTP Method is not supported via any route
```
mitruly: @pauljamescleary `405 MethodNotAllowed` is already supported by default, but works a bit differently than expected in Akka HTTP in that it will only be returned if an HTTP method that is not supported by any of the routes is used.

pauljamescleary: Realistically, we should return a `405 MethodNotAllowed` if someone is doing a `POST` at a URL that doesn't support `POST`.  We should get a `404 NotFound` for `/foobar` for example

modules/api/src/main/scala/vinyldns/api/route/VinylDNSDirectives.scala
```diff
@@ -124,4 +126,51 @@ trait VinylDNSDirectives extends Directives {
...
         reject(ValidationRejection(compact(render("errors" -> errors.toList.toSet))))
     }
+
+  /**
+    * Helpers to handle route authentication flow for routing. Implementing classes/objects
+    * must provide handleErrors implementation.
+    */
+  def handleErrors(e: E): PartialFunction[E, Route]
+
+  def handleEither[A](either: Either[E, A]): Route = either match {
```
pauljamescleary: I don't see this used

modules/api/src/main/scala/vinyldns/api/route/VinylDNSDirectives.scala
pauljamescleary: ok

modules/api/src/main/scala/vinyldns/api/route/VinylDNSDirectives.scala
```diff
@@ -31,7 +33,7 @@ import scala.concurrent.duration._
 import scala.util.Failure
 import scala.util.control.NonFatal
 
-trait VinylDNSDirectives extends Directives {
+trait VinylDNSDirectives[E <: Throwable] extends Directives {
```
pauljamescleary: why do we need to be parameterized on the `trait`?  Can we not just parameterize `def handleErrors[E]` instead?

modules/api/src/main/scala/vinyldns/api/route/VinylDNSDirectives.scala
mitruly: Whoops, this is not used

modules/api/src/main/scala/vinyldns/api/route/VinylDNSDirectives.scala
pauljamescleary: I recommend parameterizing on the methods, not the trait

rebstar6: authenticateAndExecute and authenticateAndExecuteWithEntity also rely on that parameter.

if you are making the batch error case also throwable though, it should need no param anymore and those functions can just be on Throwable i think

modules/api/src/main/scala/vinyldns/api/route/MembershipRouting.scala
```diff
@@ -18,170 +18,163 @@ package vinyldns.api.route
...
+              input.name,
+              input.email,
+              input.description,
+              memberIds = input.members.map(_.id),
```
rebstar6: group.build combined `members ++ admins` into the memberids. watch that :P

mitruly: Good catch @rebstar6

modules/api/functional_test/live_tests/authentication_test.py
britneywright: don't think we really need this test 🤷🏽‍♀️

modules/api/functional_test/live_tests/authentication_test.py
```diff
@@ -27,3 +28,23 @@ def test_request_succeeds_when_user_is_found_and_not_locked():
...
+
+    assert_that(data, is_("The requested path [/no-existo] does not exist."))
+
+def test_request_fails_with_unsupported_http_method_for_route():
+    """
+    Test request fails with MethodNotAllowed (405) when HTTP Method is not supported for specified route
+    """
+    client = VinylDNSClient(VinylDNSTestContext.vinyldns_url, 'unknownAccessKey', 'anyAccessSecretKey')
+    url = urljoin(VinylDNSTestContext.vinyldns_url, u'/zones')
+    _, data = client.make_request(url, u'PUT', client.headers, status=405)
```
pauljamescleary: I would argue that `PUT /zones` could be a valid route some day.

Perhaps use one of the unauthenticated routes, like `POST /health`?

modules/api/functional_test/live_tests/membership/create_group_test.py
```diff
@@ -219,3 +219,26 @@ def test_create_group_no_members(shared_zone_test_context):
     finally:
         if result:
             client.delete_group(result['id'], status=(200,404))
+
+def test_create_group_adds_admins_to_member_list(shared_zone_test_context):
```
pauljamescleary: why is this in there?

modules/api/functional_test/live_tests/membership/update_group_test.py
```diff
@@ -614,3 +614,31 @@ def test_update_group_not_authorized(shared_zone_test_context):
     finally:
         if saved_group:
             ok_client.delete_group(saved_group['id'], status=(200,404))
+
+def test_update_group_adds_admins_to_member_list(shared_zone_test_context):
```
pauljamescleary: and this?

modules/api/src/main/scala/vinyldns/api/route/VinylDNSDirectives.scala
```diff
@@ -124,4 +126,48 @@ trait VinylDNSDirectives extends Directives {
...
+  def handleErrors(e: Throwable): PartialFunction[Throwable, Route]
+
+  /**
+    * Authenticate user and execute service call without request entity
+    *
+    * Flow:
+    * - Authenticate user. Proceed if successful; otherwise return unauthorized error to user.
+    * - Invoke service call, f, and return the response to the user.
+    */
+  def authenticateAndExecute[A, E <: Throwable](f: AuthPrincipal => EitherT[IO, E, A])(
```
pauljamescleary: Is there a reason on the type bounds to `Throwable`, can we just leave it as `E`?

modules/api/functional_test/live_tests/membership/create_group_test.py
mitruly: @rebstar6 brought up a good point that we didn't have existing tests for this as I removed the `Group.build` stuff.

modules/api/src/main/scala/vinyldns/api/route/VinylDNSDirectives.scala
mitruly: Binding to `Throwable` allows us to use `failWith`. The tricky thing is that the `handleErrors` partial function actually captures all cases for `BatchChangeErrorResponse`, but we still need to make these helpers generic for other routes/future error types.

pauljamescleary: When organizing code, it is good practice to not let concerns of one part of your code leak into another.

In this case, if the requirement is to use `failWith`, there is no reason to force that akka-http need down to our domain.

I would recommend wrapping `E` in a wrapper exception...

```scala
@ final case class UnexpectedError[E](e: E) extends Throwable 
defined class UnexpectedError

@ case class MyError(b: Int) 
defined class MyError

@ val err = MyError(2) 
err: MyError = MyError(2)

@ val endGame = UnexpectedError(err) 
endGame: UnexpectedError[MyError] = UnexpectedError(MyError(2))

@ val normal = UnexpectedError(new RuntimeException("fail")) 
normal: UnexpectedError[RuntimeException] = UnexpectedError(
  java.lang.RuntimeException: fail
)
...

pauljamescleary: Think of the codebase in layers:

Route -- uses --> Domain --> uses --> Repository

Concerns from the Route should not leak into the Domain, nor should concerns with MySQL leak into the Domain

modules/api/src/main/scala/vinyldns/api/route/VinylDNSDirectives.scala
pauljamescleary: Yea, we have to do this because of type parameter things with `handleError` ... resolving

PR #740, opened 2019-07-11
add errors to record owner group field

britneywright:
With the implementation of shared zones some batch change requests require a record owner group. Users see the error message when relevant on the single line, but they still seem to be confused about how to resolve it, they aren't recognizing it's referring to the Record Owner Group field.

### Current portal view with error messages:
Screen Shot 2019-07-11 at 2 57 07 PM

Some enhancements I want to try are highlighting the record owner group field if any of the changes error because there is no owner group and adding more messaging that the field is required.

### Updated portal view error messages:
Screen Shot 2019-07-11 at 1 48 00 PM

### DIFF
--- a/modules/portal/public/lib/batch-change/batch-change-new.controller.js
+++ b/modules/portal/public/lib/batch-change/batch-change-new.controller.js
@@ -34,6 +34,7 @@
             $scope.newBatch = {comments: "", changes: [{changeType: "Add", type: "A+PTR"}]};
             $scope.alerts = [];
             $scope.batchChangeErrors = false;
+            $scope.ownerGroupError = false;
             $scope.formStatus = "pendingSubmit";
 
             $scope.addSingleChange = function() {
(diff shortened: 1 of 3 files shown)


modules/portal/public/lib/batch-change/batch-change-new.controller.js
```diff
@@ -93,6 +94,9 @@
                         } else {
                             $scope.newBatch.changes = error.data;
                             $scope.batchChangeErrors = true;
+                            $scope.ownerGroupError = error.data.filter(d => d.errors)
```
mitruly: This can be simplified to:
```
$scope.ownerGroupError = error.data
     .flatMap(d => d.errors)
     .some(e => e.includes('owner group ID must be specified for record'));
```

rebstar6:
ive been wondering more generally if it's worth just making that record owner group field required in the portal, and fail before they even submit. Would keep the api as is.

in either case, lets try this way 1st and then if there are still issues, we can make stricter

Issue #720, opened 2019-07-02
Move user lock status out of protobuf


britneywright:
This will allow us to update the ldap syncer to only check against unlocked users and potentially add a route that that lists all locked users.

remerle:
This is part of a larger issue of pulling indexable values out of the blobs.  Closing this in favor of #975

PR #741, opened 2019-07-12
update group detail page

britneywright:
Screen Shot 2019-07-12 at 9 27 19 AM

Changes in this pull request:
- add group email and description to the group detail page in the portal. Only show description if one is provided.
- ~~add lock status in the in the group members table. Currently getting from mapping the member info in group info (has member ID and lock status) with the members list info (has member ID and rest of info). Will change down the road when we have lock status on the user details #720~~ 
- updated the response data in the `list group members` call.`list group members` will now include user lock status.

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipProtocol.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipProtocol.scala
@@ -75,7 +75,7 @@ case class UserInfo(
     lastName: Option[String] = None,
     email: Option[String] = None,
     created: Option[DateTime] = None,
-    lockStatus: LockStatus = LockStatus.Unlocked
+    lockStatus: Option[LockStatus] = None
 )
 object UserInfo {
   def apply(user: User): UserInfo =
@@ -86,7 +86,7 @@ object UserInfo {
       lastName = user.lastName,
       email = user.email,
       created = Some(user.created),
-      lockStatus = user.lockStatus
+      lockStatus = Some(user.lockStatus)
     )
 }
 
@@ -97,7 +97,8 @@ case class MemberInfo(
     lastName: Option[String] = None,
     email: Option[String] = None,
     created: Option[DateTime] = None,
-    isAdmin: Boolean = false
+    isAdmin: Boolean = false,
+    lockStatus: LockStatus
 )
 
 object MemberInfo {
@@ -109,7 +110,8 @@ object MemberInfo {
       lastName = user.lastName,
       email = user.email,
       created = Some(user.created),
-      isAdmin = group.adminUserIds.contains(user.id)
+      isAdmin = group.adminUserIds.contains(user.id),
+      lockStatus = user.lockStatus
     )
 }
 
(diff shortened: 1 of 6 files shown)


modules/portal/public/lib/controllers/controller.membership.js
```diff
@@ -179,8 +179,11 @@ angular.module('controller.membership', []).controller('MembershipController', f
             function getGroupMemberListSuccess(result) {
                 $log.log('refresh::getGroupMemberList-success', result);
                 //update groups
-                $scope.membership.members = result.members;
+                var membersWithLockStatus = result.members
+                    .map(member => Object.assign(member, $scope.membership.group.members.find(gm => gm.id == member.id)));
```
rebstar6: im a little confused, can you explain what this line is doing?

britneywright: There's two sets of data we use on the group page, there's group info and member list info. 

Group info looks like: 
```
{"id":"group-id","name":"nice","email":"efonwe","description":"fonfwfon","created":"2019-07-12T01:44:24Z","status":"Active","members":[{"id":"user-id","lockStatus":"Unlocked"}],"admins":[{"id":"user-id","lockStatus":"Unlocked"}]}
```

Member list info looks like:
```
{"members":[{"id":"user-id","userName":"test-user","firstName":"Test","lastName":"User","email":"test@test.com","created":"2019-07-12T01:40:37Z","isAdmin":true}],"maxItems":100}
```
 
I'm coming the "members" in group info with the "members" in the members list to get the user lockStatus. This is a small workaround until we put lockStatus in the member list. I didn't look into doing that yet, since I was keeping this to just portal changes for the first commit.

rebstar6: wow, that is wild. i wonder why we bother returning lockedStatus in the group call

it definitely should be added to the MemberInfo type; see how much work that actually is, id rather just do that

britneywright: fixed it now, turned out to be pretty minor changes on the API side.

modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipProtocol.scala
```diff
@@ -75,7 +75,7 @@ case class UserInfo(
     lastName: Option[String] = None,
     email: Option[String] = None,
     created: Option[DateTime] = None,
-    lockStatus: LockStatus = LockStatus.Unlocked
+    lockStatus: Option[LockStatus] = None
```
rebstar6: why did this become optional?

modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipProtocol.scala
```diff
@@ -97,7 +97,8 @@ case class MemberInfo(
     lastName: Option[String] = None,
     email: Option[String] = None,
     created: Option[DateTime] = None,
-    isAdmin: Boolean = false
+    isAdmin: Boolean = false,
+    lockStatus: LockStatus
```
rebstar6: default to LockStatus.Unlocked?

rebstar6: mm actually probably doesnt matter since we use apply anyway

modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipProtocol.scala
britneywright: to not have the lock status visible in the member response in the get group call, only the IDs.

britneywright: I thought that was intentionally the structure of UserInfo since username is optional here even though it's required in the User class.

PR #742, opened 2019-07-12
allow updates to dotted CNAME records

britneywright:
Addendum to #703

I missed allowing updates to CNAME dotted hosts in my previous PR.

Changes in this pull request:
- remove initial JSON validation that doesn't allow CNAMEs to contain a dot
- Pass existing recordset through the CNAME validations so we can compare the new and existing recordset names in the `isNotDotted` method.

### DIFF
--- a/docker/bind9/zones/ok.hosts
+++ b/docker/bind9/zones/ok.hosts
@@ -12,3 +12,5 @@ test             IN  A   3.3.3.3
 test             IN  A   4.4.4.4
 @                IN  A   5.5.5.5
 already-exists   IN  A   6.6.6.6
+dotted.a         IN  A   7.7.7.7
+dottedc.name     IN CNAME test.example.com
--- a/modules/api/functional_test/live_tests/recordsets/create_recordset_test.py
+++ b/modules/api/functional_test/live_tests/recordsets/create_recordset_test.py
@@ -657,23 +657,23 @@ def test_create_dotted_a_record_apex_with_trailing_dot_succeeds(shared_zone_test
             delete_result = client.delete_recordset(apex_a_rs['zoneId'], apex_a_rs['id'], status=202)
             client.wait_until_recordset_change_status(delete_result, 'Complete')
 
-def test_create_dotted_cname_record_apex_fails(shared_zone_test_context):
+def test_create_dotted_cname_record_fails(shared_zone_test_context):
     """
-    Test that creating a CNAME record set with record name matching dotted apex returns an error.
+    Test that creating a CNAME record set with dotted record name returns an error.
     """
     client = shared_zone_test_context.ok_vinyldns_client
     zone = shared_zone_test_context.parent_zone
 
     apex_cname_rs = {
         'zoneId': zone['id'],
-        'name': zone['name'].rstrip('.'),
+        'name': 'dot.ted',
         'type': 'CNAME',
         'ttl': 500,
-        'records': [{'cname': 'foo'}]
+        'records': [{'cname': 'foo.bar.'}]
     }
 
-    errors = client.create_recordset(apex_cname_rs, status=400)['errors']
-    assert_that(errors[0], is_("CNAME data must be absolute"))
+    error = client.create_recordset(apex_cname_rs, status=422)
+    assert_that(error, is_("Record with name dot.ted and type CNAME is a dotted host which is not allowed in zone parent.com."))
 
 def test_create_cname_with_multiple_records(shared_zone_test_context):
     """
@@ -699,6 +699,23 @@ def test_create_cname_with_multiple_records(shared_zone_test_context):
     errors = client.create_recordset(new_rs, status=400)['errors']
     assert_that(errors[0], is_("CNAME record sets cannot contain multiple records"))
 
+def test_create_dotted_cname_record_apex_fails(shared_zone_test_context):
+    """
+    Test that creating a CNAME record set with record name matching dotted apex returns an error.
+    """
+    client = shared_zone_test_context.ok_vinyldns_client
+    zone = shared_zone_test_context.parent_zone
+
+    apex_cname_rs = {
+        'zoneId': zone['id'],
+        'name': zone['name'].rstrip('.'),
+        'type': 'CNAME',
+        'ttl': 500,
+        'records': [{'cname': 'foo.bar.'}]
+    }
+
+    error = client.create_recordset(apex_cname_rs, status=422)
+    assert_that(error, is_("CNAME RecordSet cannot have name '@' because it points to zone origin"))
 
 def test_create_cname_pointing_to_origin_symbol_fails(shared_zone_test_context):
     """
--- a/modules/api/functional_test/live_tests/recordsets/create_recordset_test.py
+++ b/modules/api/functional_test/live_tests/recordsets/create_recordset_test.py
@@ -657,9 +657,9 @@ def test_create_dotted_a_record_apex_with_trailing_dot_succeeds(shared_zone_test
             delete_result = client.delete_recordset(apex_a_rs['zoneId'], apex_a_rs['id'], status=202)
             client.wait_until_recordset_change_status(delete_result, 'Complete')
 
-def test_create_dotted_cname_record_fails(shared_zone_test_context):
+def test_create_cname_record_apex_fails(shared_zone_test_context):
     """
-    Test that creating a CNAME record set with dotted record name returns an error.
+    Test that creating a CNAME record set with record name matching zone name returns an error.
     """
     client = shared_zone_test_context.ok_vinyldns_client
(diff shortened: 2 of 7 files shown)


modules/api/src/main/scala/vinyldns/api/route/DnsJsonProtocol.scala
```diff
@@ -162,11 +162,7 @@ trait DnsJsonProtocol extends JsonValidation {
         (js \ "zoneId").required[String]("Missing RecordSet.zoneId"),
         (js \ "name")
           .required[String]("Missing RecordSet.name")
-          .check("Record name must not exceed 255 characters" -> checkDomainNameLen)
```
mitruly: There is also another check in `cnameValidations` (`RecordSetValidations`) that is checking `isNotDotted`.

britneywright:
where are you seeing the creation of dotted CNAMEs, in record set, batch change, or both?

modules/api/src/main/scala/vinyldns/api/route/DnsJsonProtocol.scala
rebstar6: add a dotted update func test, this will validate it isnt being stopped anywhere

modules/api/functional_test/live_tests/recordsets/update_recordset_test.py
```diff
@@ -1854,8 +1854,38 @@ def test_update_dotted_cname_record_apex_fails(shared_zone_test_context):
...
-        errors = client.update_recordset(create_rs, status=400)['errors']
-        assert_that(errors[0],is_("Record name cannot contain '.' with given type"))
+        error = client.update_recordset(create_rs, status=422)
+        assert_that(error,is_("CNAME RecordSet cannot have name '@' because it points to zone origin"))
+
+    finally:
+        delete_response = client.delete_recordset(zone['id'],create_rs['id'], status=202)['status']
+        client.wait_until_recordset_deleted(delete_response, 'Complete')
+
+def test_update_cname_as_dotted_host_fails(shared_zone_test_context):
```
mitruly: Think this PR is supposed to allow updating/deleting existing dotted CNAME records, just not new creates.

modules/api/functional_test/live_tests/recordsets/update_recordset_test.py
```diff
@@ -1854,8 +1854,38 @@ def test_update_dotted_cname_record_apex_fails(shared_zone_test_context):
...
+
+    apex_cname_rs = {
+        'zoneId': zone['id'],
+        'name': 'link',
+        'type': 'CNAME',
+        'ttl': 500,
+        'records': [{'cname': 'got.reference'}]
+    }
+
+    create_response = client.create_recordset(apex_cname_rs, status=202)
```
mitruly: This will fail because we can't create; need to create via DNS backend and sync in the change.

modules/api/functional_test/live_tests/recordsets/update_recordset_test.py
britneywright: this is confirming that you can't update a record that doesn't have a dotted name with a dotted name, and that it fails with the correct error.

modules/api/functional_test/live_tests/recordsets/update_recordset_test.py
britneywright: not sure why you think this will fail when the name isn't dotted.

britneywright:
We don't currently test for successfully updating dotted host records in the func tests because we can't create dotted host records in the func tests. Do we want to test those, and if so do we want to add dotted records in the zone host files or in the TestDataLoader?

modules/api/functional_test/live_tests/recordsets/update_recordset_test.py
rebstar6: do we have another test somewhere confirming that you can update a dotted cname?

britneywright: that's what I referred to in this comment:

> We don't currently test for successfully updating dotted host records in the func tests because we can't create dotted host records in the func tests. Do we want to test those, and if so do we want to add dotted records in the zone host files or in the TestDataLoader?

rebstar6: ah i see, yeah throw one in a zone file :)

modules/api/functional_test/live_tests/recordsets/update_recordset_test.py
mitruly: Thought I saw this as creating a dotted host; nvm on this comment

modules/api/functional_test/live_tests/recordsets/update_recordset_test.py
mitruly: We should clarify that this is updating an existing non-dotted CNAME to dotted.

britneywright:
expect this to have one failing func test from adding the new dotted records. Need to fix that before making the new tests.

modules/api/functional_test/live_tests/recordsets/create_recordset_test.py
```diff
@@ -673,7 +673,7 @@ def test_create_dotted_cname_record_apex_fails(shared_zone_test_context):
     }
 
     errors = client.create_recordset(apex_cname_rs, status=400)['errors']
-    assert_that(errors[0], is_("Record name cannot contain '.' with given type"))
+    assert_that(errors[0], is_("CNAME data must be absolute"))
```
rebstar6: this test is no longer testing what we want. this error is on the cname target. make the data`'records': [{'cname': 'foo.bar.'}]`  so it validates that you cant create a dotted cname

docker/bind9/zones/ok.hosts
```diff
@@ -12,3 +12,5 @@ test             IN  A   3.3.3.3
 test             IN  A   4.4.4.4
 @                IN  A   5.5.5.5
 already-exists   IN  A   6.6.6.6
+dotted.a         IN  A   7.7.7.7
+dottedc.name     IN CNAME test.example.com
```
mitruly: Can this be `dotted.cname`?

modules/api/functional_test/live_tests/recordsets/create_recordset_test.py
```diff
@@ -699,6 +699,23 @@ def test_create_cname_with_multiple_records(shared_zone_test_context):
     errors = client.create_recordset(new_rs, status=400)['errors']
     assert_that(errors[0], is_("CNAME record sets cannot contain multiple records"))
 
+def test_create_dotted_cname_record_apex_fails(shared_zone_test_context):
```
mitruly: The test name is a bit misleading, since we don't allow apex CNAME records, regardless of whether they are top-level or not. Could just have this named `test_create_cname_record_apex_fails`.

docker/bind9/zones/ok.hosts
britneywright: no, at least one of the tests in list recordsets is looks for all records that contain "cname" and it would mess with the cleanup, tried to make these additions as minimally disruptive as possible.

modules/api/functional_test/live_tests/recordsets/create_recordset_test.py
```diff
@@ -657,23 +657,23 @@ def test_create_dotted_a_record_apex_with_trailing_dot_succeeds(shared_zone_test
             delete_result = client.delete_recordset(apex_a_rs['zoneId'], apex_a_rs['id'], status=202)
             client.wait_until_recordset_change_status(delete_result, 'Complete')
 
-def test_create_dotted_cname_record_apex_fails(shared_zone_test_context):
+def test_create_cname_record_apex_fails(shared_zone_test_context):
```
rebstar6: this isnt apex, this is testing dotted

britneywright: no this is apex, i made a separate test for dotted.

britneywright: scratch that, mixed up the test names

modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetValidations.scala
```diff
@@ -94,7 +94,7 @@ object RecordSetValidations {
       zone: Zone,
       existingRecordSet: Option[RecordSet] = None): Either[Throwable, Unit] =
     newRecordSet.typ match {
-      case CNAME => cnameValidations(newRecordSet, existingRecordsWithName, zone)
+      case CNAME => cnameValidations(newRecordSet, existingRecordsWithName, zone, existingRecordSet)
```
rebstar6: yeesh. bad on us

PR #743, opened 2019-07-12
[WIP] Store batch errors take 2

rebstar6:
messing with storing these errors, proto changes

### DIFF
--- a/modules/core/src/main/protobuf/VinylDNSProto.proto
+++ b/modules/core/src/main/protobuf/VinylDNSProto.proto
@@ -197,6 +197,7 @@ message SingleChange {
   optional string recordChangeId = 10;
   optional string recordSetId = 11;
   optional SingleChangeData changeData = 12;
+  repeated DomainValidationError validationErrors = 13;
 }
 
 message User {
@@ -221,4 +222,13 @@ message UserChange {
   required string id = 4;
   required string changeType = 5;
   optional User oldUser = 6;
+}
+
+message DomainValidationError {
+  required int64 errorType = 1;
+  optional bytes data = 2;
+}
+
+message DomainValidationErrorOneString {
+  required string stringOne = 1;
 }
--- a/modules/core/src/main/protobuf/VinylDNSProto.proto
+++ b/modules/core/src/main/protobuf/VinylDNSProto.proto
@@ -225,7 +225,7 @@ message UserChange {
 }
 
 message DomainValidationError {
-  required int64 errorType = 1;
+  required string errorType = 1;
   optional bytes data = 2;
 }
 
--- a/modules/core/src/main/scala/vinyldns/core/protobuf/BatchChangeProtobufConversions.scala
+++ b/modules/core/src/main/scala/vinyldns/core/protobuf/BatchChangeProtobufConversions.scala
@@ -17,6 +17,7 @@
 package vinyldns.core.protobuf
 
 import cats.syntax.all._
+import vinyldns.core.domain.{DomainValidationError, ZoneDiscoveryError}
 import vinyldns.core.domain.batch.{
   SingleAddChange,
   SingleChange,
@@ -26,6 +27,7 @@ import vinyldns.core.domain.batch.{
 import vinyldns.core.domain.record.RecordType
 import vinyldns.core.protobuf.SingleChangeType.{SingleAddType, SingleChangeType, SingleDeleteType}
 import vinyldns.proto.VinylDNSProto
+import scala.collection.JavaConverters._
 
 object SingleChangeType extends Enumeration {
   type SingleChangeType = Value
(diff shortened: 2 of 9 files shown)


modules/core/src/main/protobuf/VinylDNSProto.proto
```diff
@@ -197,6 +197,7 @@ message SingleChange {
   optional string recordChangeId = 10;
   optional string recordSetId = 11;
```
rebstar6: this is the file to look at for now

modules/core/src/main/protobuf/VinylDNSProto.proto
```diff
@@ -222,3 +223,12 @@ message UserChange {
...
 }
+
+message DomainValidationError {
+  required string errorType = 1;
+  optional bytes data = 2;
+}
+
+message DomainValidationErrorOneString {
+  required string stringOne = 1;
+}
```
rebstar6: as an alternate, can actually make separate types for the data stored in each of these errors. felt a little silly given that most of them just have a string

modules/core/src/main/scala/vinyldns/core/protobuf/BatchChangeProtobufConversions.scala
```diff
@@ -37,10 +39,22 @@ object SingleChangeType extends Enumeration {
   }
 }
 
+object DomainValidationErrorCode extends Enumeration {
+  type DomainValidationErrorType = Value
+  // NOTE: once defined, an error code cannot be changed!
```
rebstar6: this is tricky. note that once we have an error coded in the protobuf, we cant change it (though we can edit the message...)

modules/core/src/main/scala/vinyldns/core/protobuf/BatchChangeProtobufConversions.scala
```diff
@@ -17,6 +17,7 @@
 package vinyldns.core.protobuf
```
rebstar6: this is the other file worth looking at right now

modules/core/src/main/protobuf/VinylDNSProto.proto
pauljamescleary: a lot of errors have multiple data points in them.

Question is: What is it that we are trying to accomplish?

Our primary requirement is that we want to be able to store and retrieve error information.  Given that a batch has a size of 1000, worst case we have 1000 errors in the batch.

The other requirement is that we know what the error type is (soft vs. hard), and there are rules to how we handle each kind (i.e. business logic), so the actual type is important.

So, how can we ensure that all current errors and possible future errors are serialized / de-serialized properly so that we know the type and accompanying information?

If we choose one path over another, what happens to backward compatibility?

What are the pros and cons to each approach?

rebstar6: we only need to store soft errors, keep that in mind to start.

that said, in theory and DomainValidationError can be a soft error, which means, in theory, we need some kind of protobuf way to store any DomainValidationError. I think any way we do this is not limiting for future errors, but does have some considerations for code readability/changability

Option 1 (in this PR): DomainValidationError by errorType (errorCode?); each type according to parameters
```
message DomainValidationError {
  required string errorType = 1;
  optional bytes data = 2;
}
message DomainValidationErrorOneString {
  required string stringOne = 1;
}
``` 
In this case, I have a code for each type, and then (eventually) a series of generic parameter holders (oneString, oneInt, etc). This mostly avoids us having to outline every error to this point in protobuf. DomainValidationErrorOneString handles ZoneDiscoveryError, RecordAlreadyExists, RecordDoesNotExist, etc that take in one string, and we can build others as we need them.

Pros: not a crazy amount of protobuf changes needed right now; gives you all the info you need (can serialize/deserialize fully to/from the error). Cons: once an error is stored, its pretty much set. this is probably less obvious in this setup than other options. For example, if we decided to change ZoneDiscoveryError for whatever reason, you would have to account for anything currently stored as DomainValidationErrorOneString, which isnt ideal.

rebstar6: an alternate would be to fully define all the types in the serializer. a bunch of work we dont actually need right now. the pro here is its more clear that things cant be changed

rebstar6: another alternate is to just store the error codes with no context. this means you are only able to return the error type/code, but would not be able to recreate the error out of the protobuf. This is the smallest to store and easiest for now, but does mean you are a little stuck down the line if you want more info. that said, i dont hate this approach (List[DomainValidationErrorCode] on the SingleChange) since i think it gives all the info we need for now, and we can deal with the more info question at a later point if we decide we need it

modules/core/src/main/scala/vinyldns/core/protobuf/BatchChangeProtobufConversions.scala
```diff
@@ -37,10 +39,22 @@ object SingleChangeType extends Enumeration {
...
+    case _: ZoneDiscoveryError => ZDE
+    case _ => Unknown
+  }
+}
+
 trait BatchChangeProtobufConversions extends ProtobufConversions {
   /* Currently, we only support the add change type.  When we support additional change we will add them here */
   def fromPB(
       changeType: SingleChangeType,
+      errors: List[DomainValidationError],
```
mitruly: Wonder if `errors: Option[NonEmptyList[DomainValidationError]]` would be overkill here.

modules/core/src/main/scala/vinyldns/core/protobuf/BatchChangeProtobufConversions.scala
mitruly: Is the plan to support all `DomainValidationError`s here (since other instances may want error `X` to be configured as a soft error)?

modules/core/src/main/protobuf/VinylDNSProto.proto
```diff
@@ -222,3 +223,12 @@ message UserChange {
   required string changeType = 5;
   optional User oldUser = 6;
 }
+
+message DomainValidationError {
+  required string errorType = 1;
```
mitruly: I'm typically in favor of not using a string (unless the string is more of the error type, since it is subject to change) for the data mapping.

How about having `errorType` be some numeric value that ties to a `DomainValidationError` and `string` be the error message at the time that it is stored? The rationale behind this is that our error messages tend to change and this would offer some form of backwards compatibility.

rebstar6:
closing for #745

PR #745, opened 2019-07-15
Store errors on single changes

rebstar6:
Just storage in this PR. Still todo in future PRs:
- Deal with response serialization (JsonProtocol)
- Bring in an actual soft failure

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetValidations.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetValidations.scala
@@ -126,7 +126,7 @@ object RecordSetValidations {
       ensuring(
         RecordSetAlreadyExists(s"RecordSet with name ${newRecordSet.name} already " +
           s"exists in zone ${zone.name}, CNAME record cannot use duplicate name"))(
-        !existingRecordsWithName.exists(_.id != newRecordSet.id)
+        existingRecordsWithName.forall(_.id == newRecordSet.id)
       )
     }
 
--- a/modules/core/src/main/scala/vinyldns/core/domain/DomainValidationErrors.scala
+++ b/modules/core/src/main/scala/vinyldns/core/domain/DomainValidationErrors.scala
@@ -24,6 +24,17 @@ sealed abstract class DomainValidationError(val isFatal: Boolean = true) {
   def message: String
 }
 
+object DomainValidationErrorType extends Enumeration {
+  type DomainValidationErrorType = Value
+  // NOTE: once defined, an error code cannot be changed!
+  val ZDE, Unknown = Value
+
+  def from(error: DomainValidationError): DomainValidationErrorType = error match {
+    case _: ZoneDiscoveryError => ZDE
+    case _ => Unknown
+  }
+}
+
 // The request itself is invalid in this case, so we fail fast
 final case class ChangeLimitExceeded(limit: Int) extends DomainValidationError {
   def message: String = s"Cannot request more than $limit changes in a single batch change request"
--- a/modules/core/src/main/scala/vinyldns/core/domain/DomainValidationErrors.scala
+++ b/modules/core/src/main/scala/vinyldns/core/domain/DomainValidationErrors.scala
@@ -24,17 +24,6 @@ sealed abstract class DomainValidationError(val isFatal: Boolean = true) {
   def message: String
 }
 
-object DomainValidationErrorType extends Enumeration {
-  type DomainValidationErrorType = Value
-  // NOTE: once defined, an error code cannot be changed!
-  val ZDE, Unknown = Value
-
-  def from(error: DomainValidationError): DomainValidationErrorType = error match {
-    case _: ZoneDiscoveryError => ZDE
-    case _ => Unknown
-  }
-}
-
 // The request itself is invalid in this case, so we fail fast
 final case class ChangeLimitExceeded(limit: Int) extends DomainValidationError {
   def message: String = s"Cannot request more than $limit changes in a single batch change request"
--- a/modules/core/src/main/scala/vinyldns/core/domain/DomainValidationErrors.scala
+++ b/modules/core/src/main/scala/vinyldns/core/domain/DomainValidationErrors.scala
@@ -97,7 +97,8 @@ final case class InvalidBatchRecordType(param: String, supported: Set[RecordType
     s"""Invalid Batch Record Type: "$param", valid record types for batch changes include $supported."""
 }
 
-final case class ZoneDiscoveryError(name: String) extends DomainValidationError {
+final case class ZoneDiscoveryError(name: String, fatal: Boolean = true)
+    extends DomainValidationError(fatal) {
   def message: String =
     s"""Zone Discovery Failed: zone for "$name" does not exist in VinylDNS. """ +
       "If zone exists, then it must be connected to in VinylDNS."
--- a/modules/core/src/main/scala/vinyldns/core/domain/DomainValidationStoredError.scala
+++ b/modules/core/src/main/scala/vinyldns/core/domain/DomainValidationStoredError.scala
@@ -1,0 +1,37 @@
+/*
+ * Copyright 2018 Comcast Cable Communications Management, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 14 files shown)


modules/core/src/main/scala/vinyldns/core/domain/DomainValidationErrors.scala
```diff
@@ -24,6 +24,17 @@ sealed abstract class DomainValidationError(val isFatal: Boolean = true) {
   def message: String
 }
 
+object DomainValidationErrorType extends Enumeration {
+  type DomainValidationErrorType = Value
+  // NOTE: once defined, an error code cannot be changed!
+  val ZDE, Unknown = Value
+
+  def from(error: DomainValidationError): DomainValidationErrorType = error match {
```
mitruly: I think I'm fine with having the type used. May be worthwhile to store the error message string at the time that the error was associated with the `SingleChange`.

I don't think it's worth trying to store parameters for the error types since some of our error input parameters have changed in the past, and this would require migration scripts each time these are updated.

modules/core/src/main/scala/vinyldns/core/domain/DomainValidationStoredError.scala
```diff
@@ -0,0 +1,37 @@
...
+
+object DomainValidationStoredError {
+  def apply(error: DomainValidationError): DomainValidationStoredError =
+    new DomainValidationStoredError(DomainValidationErrorType.from(error), error.message)
+}
+
+object DomainValidationErrorType extends Enumeration {
+  type DomainValidationErrorType = Value
+  // NOTE: once defined, an error code cannot be changed!
+  val ZDE, Unknown = Value
```
rebstar6: will build this out with all the types rather than an unknown

modules/core/src/main/scala/vinyldns/core/protobuf/BatchChangeProtobufConversions.scala
```diff
@@ -117,21 +122,37 @@ trait BatchChangeProtobufConversions extends ProtobufConversions {
...
-      change.recordName.foreach(x => sc.setRecordName(x))
-      change.systemMessage.foreach(x => sc.setSystemMessage(x))
-      change.recordChangeId.foreach(x => sc.setRecordChangeId(x))
-      change.recordSetId.foreach(x => sc.setRecordSetId(x))
 
+      addCommonSingleChangeFields(sc, change)
       sc.build()
     }
 
+  def addCommonSingleChangeFields(
```
rebstar6: consolidating. unit because these functions are actually modifying the builder itself

modules/core/src/main/scala/vinyldns/core/domain/DomainValidationStoredError.scala
```diff
@@ -0,0 +1,69 @@
...
+  // NOTE: once defined, an error code type cannot be changed!
+  val ChangeLimitExceeded, BatchChangeIsEmpty, GroupDoesNotExist, NotAMemberOfOwnerGroup,
+  InvalidDomainName, InvalidLength, InvalidEmail, InvalidRecordType, InvalidPortNumber,
+  InvalidIpv4Address, InvalidIpv6Address, InvalidIPAddress, InvalidTTL, InvalidMxPreference,
+  InvalidBatchRecordType, ZoneDiscoveryError, RecordAlreadyExists, RecordDoesNotExist,
+  CnameIsNotUniqueError, UserIsNotAuthorized, RecordNameNotUniqueInBatch, RecordInReverseZoneError,
+  HighValueDomainError, MissingOwnerGroupId, ExistingMultiRecordError, NewMultiRecordError,
+  CnameAtZoneApexError = Value
+
+  def from(error: DomainValidationError): DomainValidationErrorType =
```
rebstar6: at a certain point, does it make sense to just store `error.getClass.getSimpleName` as a string and not enumerate? its less safe in the sense that DomainValidationStoredError would then just be strings, but also this feels perhaps silly

modules/core/src/main/scala/vinyldns/core/domain/DomainValidationStoredError.scala
```diff
@@ -0,0 +1,37 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package vinyldns.core.domain
+
+import vinyldns.core.domain.DomainValidationErrorType.DomainValidationErrorType
+
+case class DomainValidationStoredError(errorType: DomainValidationErrorType, message: String)
```
pauljamescleary: I know naming is hard, but "Stored" feels inconsequential to the domain, our implementation leaking into the domain model.  Imagine communicating this to Dave Back or Joe Crowe, what is a "DomainValidationStoredError", what words would be more suitable?

We are indicating "these are errors that happened as part of evaluation that need to be fixed".  There is also this "temporary" nature to them as they can be resolved and essentially disappear.

We are also conflating the set of "all validation errors" with "those that may need resolution", the latter is a small subset of the former.

Coming up with an alternative name is not easy as things conflate, "BatchChangeIssue" or "BatchChangeProblem" or "BatchChangeHalt" come to mind.

modules/core/src/main/scala/vinyldns/core/protobuf/BatchChangeProtobufConversions.scala
```diff
@@ -73,18 +78,28 @@ trait BatchChangeProtobufConversions extends ProtobufConversions {
...
             change.getId
           )
-
       }
     }
 
+  def fromPB(errors: List[VinylDNSProto.DomainValidationStoredError])
+    : Either[Throwable, List[DomainValidationStoredError]] = Either.catchNonFatal {
+    errors.map { e =>
+      val errorType = DomainValidationErrorType.withName(e.getErrorType)
```
pauljamescleary: I think `withName` will throw if there is not a match

modules/core/src/main/scala/vinyldns/core/domain/DomainValidationStoredError.scala
pauljamescleary: not in love with `ZDE`, as TLAs naturally diverge from some standard and you get FLAs or variations thereof

modules/core/src/main/scala/vinyldns/core/domain/DomainValidationStoredError.scala
pauljamescleary: Also "FailureReason"

rebstar6: none of this is yet hitting the user as we specify the SingleAddChangeSerializer/SingleDeleteChangeSerializer rather than relying on any default values, so just keep that in mind. In either case, I will change.

modules/core/src/main/scala/vinyldns/core/protobuf/BatchChangeProtobufConversions.scala
rebstar6: wont that get caught by the `Either.catchNonFatal`?

modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetValidations.scala
```diff
@@ -126,7 +126,7 @@ object RecordSetValidations {
       ensuring(
         RecordSetAlreadyExists(s"RecordSet with name ${newRecordSet.name} already " +
           s"exists in zone ${zone.name}, CNAME record cannot use duplicate name"))(
-        !existingRecordsWithName.exists(_.id != newRecordSet.id)
+        existingRecordsWithName.forall(_.id == newRecordSet.id)
```
rebstar6: this is unrelated, i was looking at it and the double negative confused me and i forgot i changed it until i saw it here

modules/core/src/main/scala/vinyldns/core/domain/DomainValidationErrors.scala
```diff
@@ -97,7 +97,8 @@ final case class InvalidBatchRecordType(param: String, supported: Set[RecordType
     s"""Invalid Batch Record Type: "$param", valid record types for batch changes include $supported."""
 }
 
-final case class ZoneDiscoveryError(name: String) extends DomainValidationError {
+final case class ZoneDiscoveryError(name: String, fatal: Boolean = true)
```
rebstar6: still leaving at fatal, needed to specify for testing (cant mock a final case class, and DomainValidationError is a sealed trait...means i cant mock out something that will pass the match in DomainValidationErrorType

rebstar6: mocking the fatal response back. was just explaining why this change was needed

britneywright:
Looks like there's only test coverage on ZoneDiscoveryError and HighValueDomain error in the SingleChangeError `from` method, so coverage of that file is only 13.97%

PR #748, opened 2019-07-17
[WIP - do not review] Approval flow for pending

rebstar6:
Fixes #.

Changes in this pull request:
-
-

PR #750, opened 2019-07-18
pass RejectBatchChangeInput and ApproveBatchChangeInput from routes

britneywright:
splitting up #748 into smaller PRs.

Changes in this pull request:
- in batch change routing turn the optional body on the reject batch change and approve batch change endpoints into instances of RejectBatchChangeInput and ApproveBatchChangeInput, respectively.
- We have to keep the entity as an option since it's only attribute is optional. Otherwise an `RequestEntityExpectedRejection` is raised.

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/route/BatchChangeRouting.scala
+++ b/modules/api/src/main/scala/vinyldns/api/route/BatchChangeRouting.scala
@@ -87,8 +87,11 @@ trait BatchChangeRoute extends Directives {
       (post & path("zones" / "batchrecordchanges" / Segment / "reject")) { id =>
         monitor("Endpoint.rejectBatchChange") {
           entity(as[Option[RejectBatchChangeInput]]) { input =>
-            execute(batchChangeService.rejectBatchChange(id, authPrincipal, input)) { chg =>
-              complete(StatusCodes.OK, chg)
+            execute(
+              batchChangeService
+                .rejectBatchChange(id, authPrincipal, input.getOrElse(RejectBatchChangeInput()))) {
+              chg =>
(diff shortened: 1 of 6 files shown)


modules/api/src/main/scala/vinyldns/api/route/BatchChangeRouting.scala
```diff
@@ -87,8 +87,11 @@ trait BatchChangeRoute extends Directives {
       (post & path("zones" / "batchrecordchanges" / Segment / "reject")) { id =>
         monitor("Endpoint.rejectBatchChange") {
           entity(as[Option[RejectBatchChangeInput]]) { input =>
-            execute(batchChangeService.rejectBatchChange(id, authPrincipal, input)) { chg =>
-              complete(StatusCodes.OK, chg)
+            execute(
+              batchChangeService
```
rebstar6: id think this should just be
```
entity(as[RejectBatchChangeInput]) { input =>
            execute(
              batchChangeService
                .rejectBatchChange(id, authPrincipal, input)) { chg =>
              complete(StatusCodes.OK, chg)
            }
```

britneywright: If I do `as[RejectBatchChangeInput]` instead of `as[Option[RejectBatchChangeInput]]` and there's no comment given then the body of the post request is empty which results in an error. If we require the comment then we'd also expect a body.

rebstar6: mm gotcha. it still smells a little weird, but i dont know a better way and its not worth holding it up for that. approving

PR #751, opened 2019-07-18
Batch change rejection email notifications

mitruly:
Ties in e-mail notifications into batch change rejection (for pending batch changes). Notifications are already triggered upon implementation by the `RecordSetChangeHandler`.

Changes in this pull request:
- Notify upon e-mail rejection
- Update e-mail content so that non-`AutoApproved` batch changes include review info.

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/notifier/email/EmailNotifier.scala
+++ b/modules/api/src/main/scala/vinyldns/api/notifier/email/EmailNotifier.scala
@@ -58,7 +58,7 @@ class EmailNotifier(config: EmailNotifierConfig, session: Session, userRepositor
     message.saveChanges()
     val transport = session.getTransport("smtp")
     transport.connect()
-    transport.sendMessage(message, message.getAllRecipients())
+    transport.sendMessage(message, message.getAllRecipients)
     transport.close()
   }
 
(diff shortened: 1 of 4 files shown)


mitruly:
Sample rejected e-mail format:
image

modules/api/src/main/scala/vinyldns/api/notifier/email/EmailNotifier.scala
```diff
@@ -79,19 +79,35 @@ class EmailNotifier(config: EmailNotifierConfig, session: Session, userRepositor
...
+    // For manually reviewed e-mails, add additional info; e-mails are not sent for pending batch changes
+    if (bc.approvalStatus != AutoApproved) {
+      bc.reviewerId.foreach(reviewerId => sb.append(s"<b>Reviewer:</b> $reviewerId <br/>"))
+      bc.reviewComment.foreach(reviewComment =>
+        sb.append(s"<b>Review comment:</b> $reviewComment <br/>"))
+      bc.reviewTimestamp.foreach(reviewTimestamp =>
+        sb.append(
+          s"<b>Time reviewed:</b> ${reviewTimestamp.toString(DateTimeFormat.fullDateTime)} <br/>"))
+    }
+
```
britneywright: can we add an additional line break here so the single changes table isn't right against the summary details?

modules/api/src/main/scala/vinyldns/api/notifier/email/EmailNotifier.scala
```diff
@@ -79,19 +79,35 @@ class EmailNotifier(config: EmailNotifierConfig, session: Session, userRepositor
       case _ => IO.unit
     }
 
-  def formatBatchChange(bc: BatchChange): String =
-    s"""<h1>Batch Change Results</h1>
+  def formatBatchChange(bc: BatchChange): String = {
```
britneywright: The PR i'm working on now #752 adds reviewer userName to BatchChangeInfo and BatchChangeSummary, if we change this BatchChangeInfo we could use that reviewerUserName instead of ID. I also just realized we don't have owner group info in this email and that name is also in BatchChangeInfo.

pauljamescleary: I agree we should definitely use the reviewer name

rebstar6: this is a bigger change, id expect on approval it will just rely on the RecordSetChangeHandler to notify, which will just be sending a BatchChange (it doesnt have the other metadata).

I could be convinced otherwise, but I actually think we just get this done with the info we have right now, and followup with improvements to add more info. It may make sense, for example, to just store the userName along with userId for the reviewer on the change itself to accomplish this

rebstar6:
can you screenshot an example with a review comment?

modules/api/src/main/scala/vinyldns/api/notifier/email/EmailNotifier.scala
rebstar6: *actually* the notifier already has the user repo! does it make sense to lookup the reviewer in here?

modules/api/src/main/scala/vinyldns/api/notifier/email/EmailNotifier.scala
mitruly: Sure can do. FWIW, current e-mails are being send without an additional line break between the data table and header info.

modules/api/src/main/scala/vinyldns/api/notifier/email/EmailNotifier.scala
pauljamescleary: We definitely need the reviewer info, the GUID is rather useless to the people getting these emails.

rebstar6: yeah, just a question of how we get it. I dont think passing in the batchChangeInfo is a good idea because that is a return type and this notifier is called way back in our record change processing, which doesn't have all that info. one way to get around that is to have the notifier do its own lookups for the info it wants to add

mitruly:
@rebstar6 This is a sample e-mail with `reviewComment` provided (which doesn't include the added line break before the single changes table):
Screen Shot 2019-07-22 at 10 46 08 AM

britneywright:
Noting that we'll plan to add reviewer user name/id and owner group name/id in a future PR once we decide how to pull that info in for the email.

PR #752, opened 2019-07-18
add approval fields to batch responses

britneywright:
Part of #659 

Changes in this pull request:
- add batch change approval fields to batch change responses. Those fields are:
  - approvalStatus
  - reviewerId (optional)
  - reviewerName (optional)
  - reviewComment (optional)
  - reviewTimestamp (optional)

These values will return for Get BatchChange requests and List Batch Change requests. Approve and Reject batch changes and Create Batch Change already return all of the above fields except reviewerName which is consistent with not returning ownerGroupName in those responses.

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
@@ -33,7 +33,13 @@ import vinyldns.core.domain.batch.BatchChangeApprovalStatus.BatchChangeApprovalS
 import vinyldns.core.domain.batch._
 import vinyldns.core.domain.batch.BatchChangeApprovalStatus._
 import vinyldns.core.domain.{CnameAtZoneApexError, ZoneDiscoveryError}
-import vinyldns.core.domain.membership.{Group, GroupRepository, User, UserRepository}
+import vinyldns.core.domain.membership.{
+  Group,
+  GroupRepository,
+  ListUsersResults,
+  User,
+  UserRepository
+}
 import vinyldns.core.domain.record.RecordType._
 import vinyldns.core.domain.record.RecordSetRepository
 import vinyldns.core.domain.zone.ZoneRepository
@@ -455,12 +461,22 @@ class BatchChangeService(
       summary.copy(ownerGroupName = groupName)
     }
 
+  def addReviewerUserNamesToSummaries(
+      summaries: List[BatchChangeSummary],
+      reviewers: ListUsersResults): List[BatchChangeSummary] =
+    summaries.map { summary =>
+      val userName =
+        summary.reviewerId.flatMap(userId => reviewers.users.find(_.id == userId).map(_.userName))
+      summary.copy(reviewerName = userName)
+    }
+
   def listBatchChangeSummaries(
       auth: AuthPrincipal,
       startFrom: Option[Int] = None,
       maxItems: Int = 100,
       ignoreAccess: Boolean = false,
-      approvalStatus: Option[BatchChangeApprovalStatus] = None): BatchResult[BatchChangeSummaryList] = {
+      approvalStatus: Option[BatchChangeApprovalStatus] = None)
+    : BatchResult[BatchChangeSummaryList] = {
     val userId = if (ignoreAccess && auth.isSystemAdmin) None else Some(auth.userId)
     for {
       listResults <- batchChangeRepo
@@ -471,8 +487,13 @@ class BatchChangeService(
       summariesWithGroupNames = addOwnerGroupNamesToSummaries(
         listResults.batchChanges,
         rsOwnerGroups)
+      reviewerIds = listResults.batchChanges.flatMap(_.reviewerId).toSet
+      reviewerUserNames <- userRepository.getUsers(reviewerIds, None, Some(maxItems)).toBatchResult
+      summariesWithReviewerUserNames = addReviewerUserNamesToSummaries(
+        summariesWithGroupNames,
+        reviewerUserNames)
       listWithGroupNames = listResults.copy(
-        batchChanges = summariesWithGroupNames,
+        batchChanges = summariesWithReviewerUserNames,
         ignoreAccess = ignoreAccess,
         approvalStatus = approvalStatus)
     } yield listWithGroupNames
--- a/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
@@ -33,7 +33,7 @@ import vinyldns.core.domain.batch.BatchChangeApprovalStatus.BatchChangeApprovalS
 import vinyldns.core.domain.batch._
 import vinyldns.core.domain.batch.BatchChangeApprovalStatus._
 import vinyldns.core.domain.{CnameAtZoneApexError, ZoneDiscoveryError}
-import vinyldns.core.domain.membership.{Group, GroupRepository}
+import vinyldns.core.domain.membership.{Group, GroupRepository, User, UserRepository}
 import vinyldns.core.domain.record.RecordType._
 import vinyldns.core.domain.record.RecordSetRepository
 import vinyldns.core.domain.zone.ZoneRepository
@@ -54,6 +54,7 @@ object BatchChangeService {
       batchChangeValidations,
       dataAccessor.batchChangeRepository,
       batchChangeConverter,
+      dataAccessor.userRepository,
       manualReviewEnabled,
       authProvider,
       notifiers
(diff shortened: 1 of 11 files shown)


britneywright:
think we should change Reject and Approve to return BatchChangeInfo instead of BatchChange, so reviewer name and owner group name can be used in email. thoughts?

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlBatchChangeRepository.scala
```diff
@@ -261,7 +261,13 @@ class MySqlBatchChangeRepository
                   BatchChangeStatus
                     .calculateBatchStatus(approvalStatus, pending > 0, failed > 0, complete > 0),
                   Option(res.string("owner_group_id")),
-                  res.string("id")
+                  res.string("id"),
+                  None,
+                  approvalStatus,
+                  Option(res.string("reviewer_id")),
```
pauljamescleary: use `res.stringOpt` instead of `Option` around it

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlBatchChangeRepository.scala
```diff
@@ -261,7 +261,13 @@ class MySqlBatchChangeRepository
...
+                  None,
+                  Option(res.string("review_comment")),
```
pauljamescleary: same here

britneywright:
> Don't we also have to add it to `extractBatchChange` for the `get` responses?

`extractBatchChange` was already updated

britneywright:
Updated @pauljamescleary

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlBatchChangeRepository.scala
```diff
@@ -261,7 +261,13 @@ class MySqlBatchChangeRepository
                   BatchChangeStatus
                     .calculateBatchStatus(approvalStatus, pending > 0, failed > 0, complete > 0),
                   Option(res.string("owner_group_id")),
-                  res.string("id")
+                  res.string("id"),
+                  None,
+                  approvalStatus,
+                  res.stringOpt("reviewer_id"),
+                  None,
```
mitruly: I think `BatchChangeService` needs to update `listBatchChangeSummaries` to get the reviewer username; otherwise it doesn't make sense to add a field that is not being used.

modules/core/src/main/scala/vinyldns/core/domain/batch/BatchChangeSummary.scala
```diff
@@ -31,7 +31,12 @@ case class BatchChangeSummary(
     status: BatchChangeStatus,
     ownerGroupId: Option[String],
     id: String = UUID.randomUUID().toString,
-    ownerGroupName: Option[String] = None)
+    ownerGroupName: Option[String] = None,
+    approvalStatus: BatchChangeApprovalStatus,
+    reviewerId: Option[String],
```
rebstar6: do we need all this info in the summary view?

definitely would want approvalStatus, curious if we see ourselves using the detailed review info in the other views

rebstar6:
just discussed - we need to also bring in the errors on SingleChange to getBatchChange. can be followup PR though

britneywright:
I was able to add the validation errors to the SingleAddChangeSerializer and SingleDeleteChangeSerializer so they appear in the batch change responses now

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -412,8 +439,13 @@ class BatchChangeService(
       summariesWithGroupNames = addOwnerGroupNamesToSummaries(
         listResults.batchChanges,
         rsOwnerGroups)
+      reviewerIds = listResults.batchChanges.flatMap(_.reviewerId).toSet
+      reviewerUserNames <- userRepository.getUsers(reviewerIds, None, Some(maxItems)).toBatchResult
```
mitruly: I'm hesitant to repurpose an API endpoint for this. Would rather do something that we already do for existing lookups like:
```
reviewerUserNames <- reviewerIds.map(userRepository.getUser).parSequence.toBatchResult
```

Also, instead of simply having a set for `reviewerUserNames`, we can have a map between `reviewerId` -> `reviewerUserName` so we can do a direct lookup instead of using `.find` on a `List`.

rebstar6: is this resolved? getUsers as one call is better than a bunch of individual getUser

modules/api/src/test/scala/vinyldns/api/route/BatchChangeJsonProtocolSpec.scala
```diff
@@ -335,6 +336,7 @@ class BatchChangeJsonProtocolSpec
         ("systemMessage" -> "systemMessage") ~
         ("recordChangeId" -> decompose(None)) ~
         ("recordSetId" -> decompose(None)) ~
+        ("validationErrors" -> decompose(List())) ~
```
rebstar6: should probably have a test with actual errors

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -395,6 +413,15 @@ class BatchChangeService(
       summary.copy(ownerGroupName = groupName)
     }
 
+  def addReviewerUserNamesToSummaries(
```
mitruly: I'm not overly ecstatic about the existing precedent of `BatchChangeRepository.getBatchChangeSummaries` returning an `IO[BatchChangeSummaryList]`, with the list being incomplete and the service adding information to it later. Honestly not sure of what a good approach would be, since it may be a bigger hassle of managing intermediate types; however with Scala having cheap types, I feel like this is approaching more OO code where we have to chase down where `X` information is actually being populated in `BatchChangeSummaryList`.

Would appreciate feedback from @rebstar6 and @pauljamescleary on this.

pauljamescleary: If I follow along, the concern is how objects are being constructed?

The Batch change repository does not have access to the user repository (and shouldn't), so it cannot do a lookup of the username.  The alternative would be to save the reviewer name along with the reviewer user id.

Some food for thought:

1. What is the downside of saving the reviewer user name in the database?  What is the upside?

2. What is the downside of making the calls to populate the reviewer name here?  What is the upside?

britneywright: I know we don't save the group name in the database for batch changes because those could change at any given time. I don't think it's possible for the username to change, at least not at this time.

rebstar6: this is how we handle all name-adding.

i do agree, it would probably be better to have a different type coming out of the DB that doesnt include the data the DB isnt giving (read: everything but the names), and then the service constructs another type that is BatchChangeSummaries. that said, since this is how we do it elsewhere, im fine with it here

modules/core/src/main/scala/vinyldns/core/domain/batch/BatchChange.scala
```diff
@@ -111,7 +119,12 @@ object BatchChangeInfo {
       id,
       status,
       ownerGroupName,
-      scheduledTime
+      approvalStatus,
+      reviewerId,
+      reviewerUserName,
+      reviewComment,
+      reviewTimestamp,
+      scheduledTime: Option[DateTime]
```
rebstar6: this is off

britneywright: corrected!

modules/api/src/test/scala/vinyldns/api/domain/batch/BatchChangeServiceSpec.scala
```diff
@@ -1209,6 +1254,9 @@ class BatchChangeServiceSpec
 
       result.batchChanges.length shouldBe 1
       result.batchChanges(0).createdTimestamp shouldBe batchChange.createdTimestamp
+      result.batchChanges(0).ownerGroupId shouldBe None
+      result.batchChanges(0).approvalStatus shouldBe BatchChangeApprovalStatus.ManuallyApproved
```
mitruly: We should also validate `reviewerId`, `reviewComment` and `reviewTimestamp` here.

britneywright: added!

PR #753, opened 2019-07-19
Support building docker images locally and without SBT

### LINKED ISSUE #744: Move jar creation inside Dockerfile
(Probably related to #26)

I don't have sbt installed which prevents me from running the bin/docker-*.sh-Scripts. 

If the api.jar was built inside a docker container, there would be not sbt tool needed on the host.

**Additional context**
I've come up with the following Dockerfile which builds the api docker image from "scratch" - needs only this file, run.sh and docker.conf. This looks much cleaner to me than the current solution.
 
```
FROM hseeberger/scala-sbt:8u212_1.2.8_2.12.8 as builder

RUN git clone -b 'v0.9.2' --single-branch --depth 1 https://github.com/vinyldns/vinyldns.git /vinyldns

RUN cd /vinyldns ; sbt api/assembly


FROM openjdk:8u201-jdk-alpine3.9

RUN apk add --update --no-cache netcat-openbsd bash

COPY --from=builder /vinyldns/modules/api/target/scala-2.12/vinyldns.jar /app/vinyldns-server.jar

COPY run.sh /app/run.sh
RUN chmod a+x /app/run.sh

COPY docker.conf /app/docker.conf

EXPOSE 9000
EXPOSE 2551

# set the entry point for the container to start vinyl, specify the config resource
ENTRYPOINT ["/app/run.sh"]

```

pauljamescleary:
Fixes #744 

Overhauls our docker release process to run releases of docker outside of SBT.  This was tested through the signing process deploying SNAPSHOT images to docker hub.

All images do a git clone to pull vinyldns, and perform the build of the container based on that git clone.

- A new `build` directory contains the things to be built, for this PR only the docker images are here.  Anticipate possibly pulling in other artifacts in the future.
- `build/docker` contains all of the docker setup
- `build/docker/test` contains a func test docker image that can be used to test any VinylDNS API instance
- `build/docker/test-bind9` contains our bind9 setup
- `build/docker/api` is a MySQL only API distribution
- `build/docker/portal` is a MySQL only Portal distribution

In addition, a few files are added to make the build easier:

- `build/start.sh` starts a given vinyldns version
- `build/stop.sh` stops the current running vinyldns setup
- `build/release.sh` this is the **MAIN** guy that performs the build.  Lots of arguments that are useful for testing.  Can provide a repository (for publishing to a docker register that is not docker hub), for building clean vs. assuming docker images.  Besides building everything, also runs func tests (smoke tests) against the docker image before completing.

Added a `README.md` for others to follow along with

To test..

1. Make sure you increase your docker machine settings to 4GB (the default of 2GB failed)
2. Go to the `build` directory
3. From there, run `./release.sh -c -t 111` to only build the docker images, this should build `0.9.4-b111`
4. Start it up, run `./start.sh -v 0.9.4-b111`
5. Login to the portal at http://localhost:9001
6. When finished, run `./stop.sh` from the build directory
*If you see an error code 137 error, you need to increase your memory*


docker/build-api/Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM hseeberger/scala-sbt:8u171_2.12.6_1.1.6 as builder
```
pauljamescleary: We need to revisit our SBT version.  We had a bunch of issues with release after 1.1.6 so we stayed put.

If we can update that, we can use a later version of this image.

docker/build-api/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+ARG branch=master
```
pauljamescleary: Allows one to override the branch to pull for the build, for example, passing in a branch fo a tag v0.9.2

docker/build-api/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+RUN git clone -b ${branch} --single-branch --depth 1 https://github.com/vinyldns/vinyldns.git /vinyldns
+
+RUN cd /vinyldns ; sbt api/stage
```
pauljamescleary: `sbt stage` is the same thing used by sbt docker that we use to publish the docker images to docker hub.

While not _exactly_ the same as our public docker hub images, it is pretty darn close

docker/build-api-test/docker.conf
```diff
@@ -0,0 +1,241 @@
...
+      # see https://github.com/brettwooldridge/HikariCP
+      connection-timeout-millis = 1000
+      idle-timeout = 10000
+      max-lifetime = 600000
+      maximum-pool-size = 20
+      minimum-idle = 20
+      register-mbeans = true
+    }
+    # Repositories that use this data store are listed here
+    repositories {
```
rebstar6: not critical and maybe not the point of this PR, but is there any way to allow specification of which repos will be where (or even just a flag for a mysql-only startup?). would be nice to be able to test against alternate configs.

mitruly:
Going to dig a bit into this; ran into an issue building the portal due to tag issue. Possibly may need to specify a separate tag for portal:
```
ERROR: for portal  (<Service: portal>, "The command '/bin/sh -c apt-get update     && apt-get install -y     curl     gnupg     && curl -sL https://deb.nodesource.com/setup_6.x | bash -     && apt-get install -y nodejs     && npm install -g grunt-cli' returned a non-zero code: 100")
ERROR: Service 'portal' failed to build: The command '/bin/sh -c apt-get update     && apt-get install -y     curl     gnupg     && curl -sL https://deb.nodesource.com/setup_6.x | bash -     && apt-get install -y nodejs     && npm install -g grunt-cli' returned a non-zero code: 100
Error response from daemon: No such image: vinyldns/portal:0.9.4-b111
```

PR #755, opened 2019-07-19
Approval up front checks

rebstar6:
more broken off of the approval flow
- get the requester associated with the change
- convert the change into a BatchChangeInput

after this, need to pass ^ through the validations

### DIFF
--- a/modules/api/functional_test/live_tests/batch/approve_batch_change_test.py
+++ b/modules/api/functional_test/live_tests/batch/approve_batch_change_test.py
@@ -19,10 +19,10 @@ def test_approve_batch_change_with_comments_exceeding_max_length_fails(shared_zo
     """
 
     client = shared_zone_test_context.ok_vinyldns_client
-    reject_batch_change_input = {
+    approve_batch_change_input = {
         "reviewComment": "a"*1025
     }
-    errors = client.approve_batch_change("some-id", reject_batch_change_input, status=400)['errors']
+    errors = client.approve_batch_change("some-id", approve_batch_change_input, status=400)['errors']
     assert_that(errors, contains_inanyorder("Comment length must not exceed 1024 characters."))
 
 @pytest.mark.manual_batch_review
--- a/modules/api/functional_test/live_tests/batch/approve_batch_change_test.py
+++ b/modules/api/functional_test/live_tests/batch/approve_batch_change_test.py
@@ -9,7 +9,7 @@ def test_approve_batch_change_with_invalid_batch_change_id_fails(shared_zone_tes
 
     client = shared_zone_test_context.ok_vinyldns_client
 
-    error = client.reject_batch_change("some-id", status=404)
+    error = client.approve_batch_change("some-id", status=404)
     assert_that(error, is_("Batch change with id some-id cannot be found"))
 
 @pytest.mark.manual_batch_review
@@ -22,7 +22,7 @@ def test_approve_batch_change_with_comments_exceeding_max_length_fails(shared_zo
     reject_batch_change_input = {
         "reviewComment": "a"*1025
     }
-    errors = client.reject_batch_change("some-id", reject_batch_change_input, status=400)['errors']
+    errors = client.approve_batch_change("some-id", reject_batch_change_input, status=400)['errors']
     assert_that(errors, contains_inanyorder("Comment length must not exceed 1024 characters."))
 
 @pytest.mark.manual_batch_review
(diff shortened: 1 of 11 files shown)


modules/api/functional_test/live_tests/batch/approve_batch_change_test.py
```diff
@@ -9,7 +9,7 @@ def test_approve_batch_change_with_invalid_batch_change_id_fails(shared_zone_tes
 
     client = shared_zone_test_context.ok_vinyldns_client
 
-    error = client.reject_batch_change("some-id", status=404)
+    error = client.approve_batch_change("some-id", status=404)
```
rebstar6: oops lol. this whole file was testing reject instead of approve

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -111,6 +115,11 @@ class BatchChangeService(
     for {
       batchChange <- getExistingBatchChange(batchChangeId)
       _ <- validateBatchChangeApproval(batchChange, authPrincipal).toBatchResult
+      _ = BatchChangeInput(batchChange)
+      _ <- EitherT.fromOptionF[IO, BatchChangeErrorResponse, AuthPrincipal](
+        authProvider.getAuthPrincipalByUserId(batchChange.userId),
+        BatchRequesterNotFound(batchChange.userId)
```
rebstar6: these wont be _ in the next pr, need to feed them into the validations flow. _ for now so it doesnt yell at me for unused variables

modules/api/src/main/scala/vinyldns/api/Boot.scala
```diff
@@ -128,11 +129,16 @@ object Boot extends App {
           loaderResponse.healthChecks)
       val batchChangeConverter =
         new BatchChangeConverter(repositories.batchChangeRepository, messageQueue)
+      val authPrincipalProvider =
+        new MembershipAuthPrincipalProvider(
+          repositories.userRepository,
+          repositories.membershipRepository)
```
pauljamescleary: is it a small enough change to also have the auth principal provider passed into the VinylDNS service for Route authentication?  If so, we can follow up with a separate PR, I don't like initializing the same thing twice

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
pauljamescleary: why are you creating a `BatchChangeInput`?

modules/api/src/main/scala/vinyldns/api/Boot.scala
rebstar6: i did already, line 151 of this file :)

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
rebstar6: you can see it if you flip over to the wip - https://github.com/vinyldns/vinyldns/pull/748/files#diff-3d42091f87b70df0dd77babb5a2a5a8fR128

the full approval flow in one is a lot of code to review. this will be needed in the next PR, just wanted to split out some things to make it easier to look at

rebstar6: the actual answer though is that we are going to run it through the same set of validations that we run a create batch through. those functions all work off the BatchChangeInput, so the easiest way of accomplishing that same run-through is to recreate the input

pauljamescleary: ok

modules/api/functional_test/live_tests/batch/approve_batch_change_test.py
```diff
@@ -22,7 +22,7 @@ def test_approve_batch_change_with_comments_exceeding_max_length_fails(shared_zo
     reject_batch_change_input = {
         "reviewComment": "a"*1025
     }
-    errors = client.reject_batch_change("some-id", reject_batch_change_input, status=400)['errors']
+    errors = client.approve_batch_change("some-id", reject_batch_change_input, status=400)['errors']
```
pauljamescleary: change the variables like "reject_batch_change_input" if it should all be "approve" in here

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
pauljamescleary: we have the "userName" on the batch change, can we use that in the error sent back instead?

rebstar6: adding bath for good measure, the IDs can make splunking easier

modules/api/src/test/scala/vinyldns/api/route/BatchChangeRoutingSpec.scala
```diff
@@ -715,5 +716,14 @@ class BatchChangeRoutingSpec
         status shouldBe BadRequest
       }
     }
+
+    "return NotFound if the requesting user cant be found" in {
+      Post("/zones/batchrecordchanges/notFoundUser/approve").withEntity(HttpEntity(
```
britneywright: why is `notFoundUser` a part of this route? Isn't this returning notFound because this route isn't valid not because the user doesn't exist?

rebstar6: thats the fake id for this test. the route is /zones/batchrecordchanges/[batch-id]/approve. in my test classes higher up i match on "notFoundUser" to generate the fake response back

PR #756, opened 2019-07-21
Adding scheduled time to all the things

pauljamescleary:
Add scheduled time field to prepare for scheduled batch changes.

* `create_batch_change_test.py` - add a test that ensures posting and retrieving using the scheduled time field works as expected
* `api/src/main/resources/application.conf` - fix an issue where the task repo wasn't loaded on `reStart`
* `BatchChangeProtocol.scala` - add scheduledTime field to `BatchChangeInput`
* `BatchChangeService.scala` - modify the `buildResonse` method, it does some logic to produce the resulting `BatchChange` entity, and we need to ensure that the `scheduledTime` field propagates through that logic
* `BatchChangeJsonProtocol.scala` - make sure that `BatchChangeInputSerializer` takes in `scheduledTime`, make sure that `BatchChangeSerializer` converts `scheduledTime` to json
* `BatchChange.scala` - add `scheduledTime` field
* `BatchChangeSummary.scala` - add `scheduledTime` field
* `V3.18__ScheduledChange.sql` - add a `scheduled_time` field to the `batch_change` table; add an index on `scheduled_time` as well to support querying by scheduled time
* `MySqlBatchChangeRepository` - make sure that save and get methods support `scheduledTime`

### DIFF
--- a/modules/api/functional_test/live_tests/batch/create_batch_change_test.py
+++ b/modules/api/functional_test/live_tests/batch/create_batch_change_test.py
@@ -1,5 +1,6 @@
 from hamcrest import *
 from utils import *
+import datetime
 
 def does_not_contain(x):
     is_not(contains(x))
@@ -265,6 +266,32 @@ def test_create_batch_change_with_adds_success(shared_zone_test_context):
                       'records': [{'preference': 1000, 'exchange': 'bar.foo.'}]}
         verify_recordset(rs16, expected16)
 
+    finally:
+        clear_zoneid_rsid_tuple_list(to_delete, client)
+
+
+def test_create_batch_change_with_scheduled_time(shared_zone_test_context):
+    """
+    Test successfully creating a batch change with scheduled time set
+    """
+    client = shared_zone_test_context.ok_vinyldns_client
+    dt = datetime.datetime.now().strftime('%Y-%m-%dT%H:%M:%SZ')
+
+    batch_change_input = {
+        "comments": "this is optional",
+        "changes": [
+            get_change_A_AAAA_json("parent.com.", address="4.5.6.7"),
+        ],
+        "scheduledTime": dt
+    }
+
+    to_delete = []
+    try:
+        result = client.create_batch_change(batch_change_input, status=202)
+        completed_batch = client.wait_until_batch_change_completed(result)
+        record_set_list = [(change['zoneId'], change['recordSetId']) for change in completed_batch['changes']]
+        to_delete = set(record_set_list)
+        assert_that(completed_batch['scheduledTime'], dt)
     finally:
         clear_zoneid_rsid_tuple_list(to_delete, client)
 
--- a/modules/api/functional_test/live_tests/batch/create_batch_change_test.py
+++ b/modules/api/functional_test/live_tests/batch/create_batch_change_test.py
@@ -270,7 +270,7 @@ def test_create_batch_change_with_adds_success(shared_zone_test_context):
         clear_zoneid_rsid_tuple_list(to_delete, client)
 
 
-def test_create_batch_change_with_scheduled_time(shared_zone_test_context):
+def test_create_batch_change_with_scheduled_time_succeeds(shared_zone_test_context):
     """
     Test successfully creating a batch change with scheduled time set
     """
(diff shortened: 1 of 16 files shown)


modules/api/src/main/scala/vinyldns/api/route/BatchChangeJsonProtocol.scala
```diff
@@ -57,7 +58,8 @@ trait BatchChangeJsonProtocol extends JsonValidation {
       (
         (js \ "comments").optional[String],
         changeList,
-        (js \ "ownerGroupId").optional[String]
+        (js \ "ownerGroupId").optional[String],
+        (js \ "scheduledTime").optional[DateTime]
```
mitruly: It may be worthwhile to providing the acceptable format (`YYYY-MM-DDTHH:MM:SSZ`) for this error message; unsupported value just yields the following error message (using input `2019-07-22T14:16:28.360Z`):
```
    "errors": [
        "Invalid date format 2019-07-22T14:16:28.360Z"
    ]
```

pauljamescleary: Yea, that is a good idea, our invalid date message doesn't tell you how it should be done

modules/api/functional_test/live_tests/batch/create_batch_change_test.py
```diff
@@ -269,6 +270,32 @@ def test_create_batch_change_with_adds_success(shared_zone_test_context):
...
+        "scheduledTime": dt
+    }
+
+    to_delete = []
+    try:
+        result = client.create_batch_change(batch_change_input, status=202)
+        completed_batch = client.wait_until_batch_change_completed(result)
+        record_set_list = [(change['zoneId'], change['recordSetId']) for change in completed_batch['changes']]
+        to_delete = set(record_set_list)
+        assert_that(completed_batch['scheduledTime'], dt)
```
rebstar6: can you add an assert to another test to confirm it's not appearing as a field when empty? dont want people seeing this before its fully implemented

modules/api/src/main/scala/vinyldns/api/route/JsonValidation.scala
```diff
@@ -58,7 +58,8 @@ object VinylDateParser {
     format.dateFormat
       .parse(s)
       .map(_.getTime)
-      .getOrElse(throw new MappingException(s"Invalid date format $s"))
+      .getOrElse(throw new MappingException(
+        s"Invalid date format $s; provide the date format as YYYY-MM-DDTHH:MM:SSZ"))
```
rebstar6: thats how we take in dates? how do we know what time zone they are talking about?

modules/api/src/main/resources/application.conf
```diff
@@ -54,6 +54,9 @@ vinyldns {
       }
       record-set {
 
+      }
+      task {
```
rebstar6: why are we putting the task repo in the api?

modules/mysql/src/main/resources/db/migration/V3.18__ScheduledChange.sql
```diff
@@ -0,0 +1,7 @@
+CREATE SCHEMA IF NOT EXISTS ${dbName};
+
+USE ${dbName};
+
+ALTER TABLE batch_change ADD scheduled_time DATETIME NULL;
+
+CREATE INDEX scheduled_time_index ON batch_change (scheduled_time);
```
rebstar6: not 100% certain we need this index. things should only become scheduled if the approval status is NeedsApproval, so youre only ever filtering on that smaller set if youre doing it right

rebstar6:
sry merge conflicted you. you need to actually rebase and change `mapN(BatchChangeInput(_, _, _))` in the BatchChangeJsonProtocol

modules/mysql/src/main/resources/db/migration/V3.18__ScheduledChange.sql
pauljamescleary: Yea, I will drop the index, as this changed recently.

Instead of providing an endpoint that will allow one to query for elapsed changes, they will "get all in manual review that are open" and further filter for those that have a scheduledTime

modules/api/src/main/resources/application.conf
pauljamescleary: I was unable to do `reStart` without it

modules/api/src/main/scala/vinyldns/api/route/JsonValidation.scala
pauljamescleary: good point, need the timezone, should be in `ISO8601` format.  I will update the datetime parser and add a unit test

modules/api/functional_test/live_tests/batch/create_batch_change_test.py
pauljamescleary: Done

modules/api/src/main/resources/application.conf
rebstar6: thats...weird

modules/api/src/main/scala/vinyldns/api/route/JsonValidation.scala
pauljamescleary: @rebstar6 so "Z" means "Zulu time", which indicates UTC.  Basically we ONLY support second level resolution in UTC, so this is correct.

rebstar6: go figure. ok works for me

modules/api/src/main/resources/application.conf
rebstar6: just in this branch? im not seeing this issue in master

PR #757, opened 2019-07-22
the rest of the approval flow

rebstar6:
Note: this flow assumes that something happened outside the system (sync, zone create, etc) that makes the previously invalid change now invalid.

skipping checks because of approval (for something like lack or access) will come later when we need it

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
@@ -80,6 +80,23 @@ class BatchChangeService(
       auth: AuthPrincipal,
       allowManualReview: Boolean): BatchResult[BatchChange] =
     for {
+      validationOutput <- applyBatchChangeValidationFlow(batchChangeInput, auth)
+      changeForConversion <- buildResponse(
+        batchChangeInput,
+        validationOutput.validatedChanges,
+        auth,
+        allowManualReview).toBatchResult
+      serviceCompleteBatch <- convertOrSave(
+        changeForConversion,
+        validationOutput.existingZones,
+        validationOutput.existingRecordSets,
+        batchChangeInput.ownerGroupId)
+    } yield serviceCompleteBatch
+
+  def applyBatchChangeValidationFlow(
+      batchChangeInput: BatchChangeInput,
+      auth: AuthPrincipal): BatchResult[BatchValidationFlowOutput] =
+    for {
       existingGroup <- getOwnerGroup(batchChangeInput.ownerGroupId)
       _ <- validateBatchChangeInput(batchChangeInput, existingGroup, auth)
       inputValidatedSingleChanges = validateInputChanges(batchChangeInput.changes)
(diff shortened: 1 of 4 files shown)


modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -74,6 +74,22 @@ class BatchChangeService(
   def applyBatchChange(
       batchChangeInput: BatchChangeInput,
       auth: AuthPrincipal): BatchResult[BatchChange] =
+    for {
```
rebstar6: functionally, applyBatchChange is unchanged. split into 3 steps: validations, response building, and SQS/save. the 1st step here is shared with the approval flow

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -347,6 +371,33 @@ class BatchChangeService(
     }
   }
 
+  def buildResponseForApprover(
+      existingBatchChange: BatchChange,
+      batchChangeInput: BatchChangeInput,
+      transformed: ValidatedBatch[ChangeForValidation],
+      reviewInfo: BatchChangeReviewInfo): Either[BatchChangeErrorResponse, BatchChange] =
+    if (transformed.forall(_.isValid)) {
```
rebstar6: for the approver, there is no more pending state - either the change is now valid, or it's not.

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchTransformations.scala
```diff
@@ -115,6 +117,26 @@ object BatchTransformations {
       )
     }
 
+    def asStoredChangeWithId(changeId: String): SingleChange = {
```
mitruly: It may be worthwhile to just make `asNewStoredChange: SingleChange` accept an `Option[String]`.

mitruly:
Note to reviewers: when testing this, `ZoneDiscoveryError` will need to be updated
```
final case class ZoneDiscoveryError(name: String, fatal: Boolean = false)
```

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchTransformations.scala
rebstar6: i thought the same, and then i truly couldnt figure out how to make it work gracefully. the SingleChange types take id as a non-option with a default. so i would have to just have 2 cases anyway right? unless i make a change and `.copy` for the ID, which would work but is maybe less efficient?

lmk if im missing something

mitruly: I was thinking about using `getOrElse`:
```
final case class AddChangeForValidation(...) extends ChangeForValidation {
   def asNewStoredChange(changeId: Option[String]): SingleChange = {
      ...
      SingleAddChange(
      ...,
      changeId.getOrElse(UUID.randomUUID().toString)
      )
   }
}
```

Both approaches are functionally the same so I'm fine with either.

mitruly:
That codecov diff check 😱

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchTransformations.scala
rebstar6: yeah true, and pull in UUID. Ill swtich to this

PR #758, opened 2019-07-22
Add pending batch data in the portal

britneywright:
for #659, depends on new batch change statuses in  #768 and #771.

Changes in this pull request:
- add feature flags for manual batch review and scheduled batch change.
- add batch change data related to manual batch review to new batch change form and batch change detail views
- add batch change data related to scheduled batch changes to new batch change form, batch change detail and batch change summaries view
- change timestamps in batch change to display user's local time zone.

If a batch change that was in a pending state is approved, we will display the approval status in addition to the batch change status, since a batch change could be approved but it could still result in a Partial Failure or Failed during processing.

### DIFF
--- a/modules/portal/app/controllers/FrontendController.scala
+++ b/modules/portal/app/controllers/FrontendController.scala
@@ -100,7 +100,7 @@ class FrontendController @Inject()(
   }
 
   def index(): Action[AnyContent] = userAction.async { implicit request =>
-    Future(Ok(views.html.batchChanges.batchChanges(request.user.userName)))
+    Future(Ok(views.html.batchChanges.batchChanges(request.user.userName, request.user.isSuper)))
   }
 
   def viewAllGroups(): Action[AnyContent] = userAction.async { implicit request =>
@@ -121,7 +121,7 @@ class FrontendController @Inject()(
   }
 
   def viewAllBatchChanges(): Action[AnyContent] = userAction.async { implicit request =>
-    Future(Ok(views.html.batchChanges.batchChanges(request.user.userName)))
+    Future(Ok(views.html.batchChanges.batchChanges(request.user.userName, request.user.isSuper)))
   }
 
   def viewBatchChange(batchId: String): Action[AnyContent] = userAction.async { implicit request =>
--- a/modules/portal/app/controllers/FrontendController.scala
+++ b/modules/portal/app/controllers/FrontendController.scala
@@ -100,7 +100,9 @@ class FrontendController @Inject()(
   }
 
   def index(): Action[AnyContent] = userAction.async { implicit request =>
-    Future(Ok(views.html.batchChanges.batchChanges(request.user.userName, request.user.isSuper)))
+    Future(
+      Ok(views.html.batchChanges
+        .batchChanges(request.user.userName, request.user.isSuper, request.user.isSupport)))
   }
 
   def viewAllGroups(): Action[AnyContent] = userAction.async { implicit request =>
@@ -121,7 +123,9 @@ class FrontendController @Inject()(
   }
 
   def viewAllBatchChanges(): Action[AnyContent] = userAction.async { implicit request =>
-    Future(Ok(views.html.batchChanges.batchChanges(request.user.userName, request.user.isSuper)))
+    Future(
+      Ok(views.html.batchChanges
+        .batchChanges(request.user.userName, request.user.isSuper, request.user.isSupport)))
   }
 
   def viewBatchChange(batchId: String): Action[AnyContent] = userAction.async { implicit request =>
--- a/modules/portal/app/views/batchChanges/batchChangeDetail.scala.html
+++ b/modules/portal/app/views/batchChanges/batchChangeDetail.scala.html
@@ -37,6 +37,7 @@
         </div>
         <div class="row">
             <div class="col-md-12">
+                <p><strong>ID:</strong> {{batch.id}}</p>
                 <p><strong>Submitted:</strong> {{batch.createdTimestamp}}</p>
                 <p ng-if="batch.comments"><strong>Description:</strong> {{batch.comments}}</p>
                 @if(meta.sharedDisplayEnabled) {
--- a/modules/portal/app/views/batchChanges/batchChangeDetail.scala.html
+++ b/modules/portal/app/views/batchChanges/batchChangeDetail.scala.html
@@ -15,13 +15,12 @@
     <!-- PAGE TITLE -->
     <div class="page-title">
         <h3>
-            <span class="fa fa-list-ol"></span> Batch Change
             <span ng-if="batch.status == 'Complete'" class="label label-success">{{batch.status}}</span>
-            <span ng-if="batch.status == 'Pending' && batch.approvalStatus != 'PendingApproval'" class="label label-info">{{batch.status}}</span>
+            <span ng-if="batch.status == 'Failed'" class="label label-danger">{{batch.status}}</span>
             <span ng-if="batch.status == 'PartialFailure'" class="label label-warning">Partial Failure</span>
-            <span ng-if="batch.status == 'Failed' && batch.approvalStatus != 'ManuallyRejected'" class="label label-danger">{{batch.status}}</span>
-            <span ng-if="batch.approvalStatus == 'PendingApproval'" class="label label-warning">Pending Review</span>
-            <span ng-if="batch.approvalStatus == 'ManuallyRejected'" class="label label-danger">Rejected</span>
+            <span ng-if="batch.status == 'PendingProcessing'" class="label label-info">Pending Processing</span>
+            <span ng-if="batch.status == 'PendingReview'" class="label label-warning">Pending Review</span>
+            <span ng-if="batch.status == 'Rejected'" class="label label-danger">{{batchChange.status}}</span>
         </h3>
     </div>
     <!-- END PAGE TITLE -->
@@ -47,6 +46,7 @@
                             <span class="fa fa-warning"></span> Group deleted
                         </span>
                     </p>
+                }
                     <p ng-if="batch.approvalStatus == 'ManuallyApproved'"><strong>Review Status:</strong> Approved</p>
                     <p ng-if="batch.reviewerUserName"><strong>Reviewer:</strong> {{batch.reviewerUserName}}</p>
                     <p ng-if="batch.reviewerId && !batch.reviewerUserName">
@@ -58,7 +58,10 @@
                     </p>
                     <p ng-if="batch.reviewComment"><strong>Review Comment:</strong> {{batch.reviewComment}}</p>
                     <p ng-if="batch.reviewTimestamp"><strong>Review Time:</strong> {{batch.reviewTimestamp}}</p>
+                @if(meta.scheduledBatchChangesEnabled) {
+                    <p ng-if="batch.scheduledTime"><strong>Scheduled Time:</strong> {{batch.scheduledTime}}</p>
                 }
+
             </div>
         </div>
         <br>
--- a/modules/portal/app/views/batchChanges/batchChangeDetail.scala.html
+++ b/modules/portal/app/views/batchChanges/batchChangeDetail.scala.html
@@ -17,7 +17,7 @@
         <h3>
             <span class="fa fa-list-ol"></span> Batch Change
             <span ng-if="batch.status == 'Complete'" class="label label-success">{{batch.status}}</span>
-            <span ng-if="batch.status == 'Pending' && batch.approvalStatus !='PendingApproval'" class="label label-info">{{batch.status}}</span>
+            <span ng-if="batch.status == 'Pending' && batch.approvalStatus != 'PendingApproval'" class="label label-info">{{batch.status}}</span>
             <span ng-if="batch.status == 'PartialFailure'" class="label label-warning">Partial Failure</span>
             <span ng-if="batch.status == 'Failed' && batch.approvalStatus != 'ManuallyRejected'" class="label label-danger">{{batch.status}}</span>
             <span ng-if="batch.approvalStatus == 'PendingApproval'" class="label label-warning">Pending Review</span>
--- a/modules/portal/app/views/batchChanges/batchChangeDetail.scala.html
+++ b/modules/portal/app/views/batchChanges/batchChangeDetail.scala.html
@@ -15,6 +15,7 @@
     <!-- PAGE TITLE -->
     <div class="page-title">
         <h3>
+            <span class="fa fa-list-ol"></span> Batch Change
             <span ng-if="batch.status == 'Complete'" class="label label-success">{{batch.status}}</span>
             <span ng-if="batch.status == 'Failed'" class="label label-danger">{{batch.status}}</span>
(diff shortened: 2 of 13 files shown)


modules/portal/app/views/batchChanges/batchChangeDetail.scala.html
```diff
@@ -17,8 +17,11 @@
         <h3>
             <span class="fa fa-list-ol"></span> Batch Change
             <span ng-if="batch.status =='Complete'" class="label label-success">{{batch.status}}</span>
+            <span ng-if="batch.status =='Pending' && batch.approvalStatus !='PendingApproval'" class="label label-info">{{batch.status}}</span>
```
mitruly: Can we be consistent with whitespace formatting after `==`/`!=`? Personally I would prefer to have a space after since most languages will otherwise throw a compilation error if there is no space.

modules/portal/app/views/batchChanges/batchChangeDetail.scala.html
```diff
@@ -44,6 +47,17 @@ <h3>
                             <span class="fa fa-warning"></span> Group deleted
                         </span>
                     </p>
+                    <p ng-if="batch.approvalStatus == 'ManuallyApproved'"><strong>Review Status:</strong> Approved</p>
```
mitruly: Should we also show review status if the batch is `PendingApproval` or `ManuallyRejected`?

britneywright: No, if the batch change approval status is PendingApproval we're showing "Pending Review" as the batch change status. If the batch change approval status is "Manually Rejected" we're showing "Rejected" as the batch change status.

modules/portal/app/views/batchChanges/batchChangeDetail.scala.html
```diff
@@ -16,11 +16,11 @@
     <div class="page-title">
         <h3>
             <span class="fa fa-list-ol"></span> Batch Change
-            <span ng-if="batch.status =='Complete'" class="label label-success">{{batch.status}}</span>
-            <span ng-if="batch.status =='Pending' && batch.approvalStatus !='PendingApproval'" class="label label-info">{{batch.status}}</span>
-            <span ng-if="batch.status =='PartialFailure'" class="label label-warning">Partial Failure</span>
-            <span ng-if="batch.status =='Failed' && batch.approvalStatus != 'ManuallyRejected'" class="label label-danger">{{batch.status}}</span>
-            <span ng-if="batch.approvalStatus =='PendingApproval'" class="label label-warning">Pending Review</span>
+            <span ng-if="batch.status == 'Complete'" class="label label-success">{{batch.status}}</span>
+            <span ng-if="batch.status == 'Pending' && batch.approvalStatus !='PendingApproval'" class="label label-info">{{batch.status}}</span>
```
mitruly: `!='PendingApproval'"` should have some whitespace. Same with `batchChanges.scala.html`

pauljamescleary:
Feels like this kind of translation should be in the API / part of the core domain.  The interplay with approval status and batch status (and the forthcoming "scheduled") is tricky.  Imagine explaining this kind of matrix to API users, oi

modules/portal/app/views/batchChanges/batchChangeDetail.scala.html
```diff
@@ -45,6 +48,21 @@ <h3>
                         </span>
                     </p>
                 }
+                <p ng-if="batch.approvalStatus == 'ManuallyApproved'"><strong>Review Status:</strong> Approved</p>
+                <p ng-if="batch.reviewerUserName"><strong>Reviewer:</strong> {{batch.reviewerUserName}}</p>
```
pauljamescleary: would it make more sense to have all the fields show up if there is a reviewer?  For example, even if the reviewComment is empty, it should display as empty if there is no review?

I was thinking there could be an `ng-if batch.approvalStatus == 'ManuallyApproved'` and then just nest everything in side that.  If they are empty then the field name appears; however, the field would show no data.

modules/portal/app/views/batchChanges/batchChangeDetail.scala.html
```diff
@@ -16,9 +16,12 @@
...
             <span class="fa fa-list-ol"></span> Batch Change
-            <span ng-if="batch.status =='Complete'" class="label label-success">{{batch.status}}</span>
-            <span ng-if="batch.status =='PartialFailure'" class="label label-warning">Partial Failure</span>
-            <span ng-if="batch.status =='Failed'" class="label label-danger">{{batch.status}}</span>
+            <span ng-if="batch.status == 'Complete'" class="label label-success">{{batch.status}}</span>
+            <span ng-if="batch.status == 'Failed'" class="label label-danger">{{batch.status}}</span>
+            <span ng-if="batch.status == 'PartialFailure'" class="label label-warning">Partial Failure</span>
+            <span ng-if="batch.status == 'PendingProcessing'" class="label label-info">Pending Processing</span>
+            <span ng-if="batch.status == 'PendingReview'" class="label label-warning">Pending Review</span>
+            <span ng-if="batch.status == 'Rejected'" class="label label-danger">{{batchChange.status}}</span>
```
pauljamescleary: why is this `batchChange.status` as opposed to `batch.status`?

pauljamescleary: we should add an `ng-if` for the status of `Scheduled`.  That is what the API will return in the event that the batch change is scheduled but not yet processed (it is in pr #771 )

modules/portal/app/views/batchChanges/batchChangeNew.scala.html
```diff
@@ -50,6 +50,34 @@
...
+                            <div class="form-group row">
+                                <div class="col-md-6">
+                                    <label class="h5">Request Date/Time:</label>
+                                    <div class="radio no-top-margin">
+                                        <label>
+                                            <input type="radio" ng-model="scheduledTime" value="Now">
+                                            Now
+                                        </label><br/>
+                                        <label>
+                                            <input type="radio" ng-model="scheduledTime" value="Later">
```
pauljamescleary: It is confusing to use `scheduledTime` for the ng-model, I would use `scheduleOption` instead to avoid confusion

modules/portal/app/views/batchChanges/batchChangeDetail.scala.html
britneywright: the reviewer info and comments appear if the batch change is approved or rejected. 

We don't currently display the description field if it's blank, so following that same pattern of not displaying empty fields.

modules/portal/app/views/batchChanges/batchChangeDetail.scala.html
britneywright: fixed and added the scheduled status in batchChangeDetail.scala.html and batchChanges.scala.html

modules/portal/app/views/batchChanges/batchChangeDetail.scala.html
pauljamescleary: Is there a reason we decided to not show the description field if it is blank?  The absence of data is significant (i.e. it is evident there is no description because the field appears with no data next to it).

If someone rejects or approves a batch change, all the fields related to that action are in play.

My thoughts are:
1. absence of data is significant
2. it is jarring to a user to have some fields display sometimes, and not appear other times.

modules/portal/public/lib/batch-change/batch-change-detail.controller.js
```diff
@@ -26,6 +26,13 @@
             $scope.getBatchChange = function(batchChangeId) {
                 function success(response) {
                     $scope.batch = response.data;
+                    $scope.batch.createdTimestamp = utilityService.formatDateTime(response.data.createdTimestamp);
```
pauljamescleary: In the example above, we format the date time in the html, here we do it when we get the response in the controller.  Should we follow the same pattern in the html above?  I kinda like doing it in the controller and having less things happen in the html, but can't say I have a strong opinion on it

modules/portal/public/lib/batch-change/batch-changes.controller.js
```diff
@@ -101,6 +105,34 @@
                     });
             };
 
+            $scope.pendingFilter = function(){
```
mitruly: Can make this DRY-er:
```
$scope.togglePendingFilter = function(){
    $scope.onlyPendingReview = !$scope.onlyPendingReview
    if($scope.onlyPendingReview == true) {
        $scope.approvalStatus = undefined;
    } else {
        $scope.approvalStatus = 'PendingApproval';
    }
    $scope.refreshBatchChanges();
}
```

Also, should we be setting `approvalStatus` to `null` instead of `undefined`? I've never really seen variables explicitly _set_ to undefined.

modules/portal/public/lib/batch-change/batch-changes.controller.js
```diff
@@ -101,6 +105,34 @@
...
+                    $scope.approvalStatus = undefined;
+                    $scope.refreshBatchChanges();
+                } else {
+                    $scope.onlyPendingReview = true;
+                    $scope.approvalStatus = 'PendingApproval';
+                    $scope.refreshBatchChanges();
+                }
+            }
+
+            $scope.myRequests = function(){
```
mitruly: This and `allRequests` can be combined into something like:
```
$scope.getRequests = function(ignoreAccess) {
    $scope.ignoreAccess = ignoreAccess;
    $scope.refreshBatchChanges();
}
```

modules/portal/public/lib/batch-change/batch-change-detail.controller.js
britneywright: will look into that

modules/portal/app/views/batchChanges/batchChangeNew.scala.html
```diff
@@ -50,6 +50,34 @@
                                 </div>
                             </div>
                             }
+                            @if(meta.scheduledBatchChangesEnabled) {
+                            <div class="form-group row">
```
mitruly: Invalid dates (eg. 04/31/2019) should have a more obvious error message. Currently the field box just gets highlighted.

modules/portal/app/views/batchChanges/batchChangeNew.scala.html
```diff
@@ -50,6 +50,34 @@
...
+                                        </label><br/>
+                                        <label>
+                                            <input type="radio" ng-model="scheduledOption" ng-value="true">
+                                            Later
+                                        </label><br/>
+                                    </div>
+                                    <div ng-if="scheduledOption" class="form-inline">
+                                        <div class="form-group">
+                                            <input  class="form-control" type="datetime-local"
+                                                   ng-model="newBatch.scheduledTime"
```
mitruly: Would prefer these two be two separate boxes, especially since there are buttons to the right of the field for different sub-fields (eg. date selector, time selector).

Maybe also put the time locale next to the header (eg. `Request Date/Time (EDT)`).

modules/portal/public/lib/batch-change/batch-changes.controller.js
mitruly: `$scope.approvalStatus = 'PendingApproval';` needs to get updated to `$scope.approvalStatus = 'PendingReview';` (also corrected in proposed alternative above) with the rename.

modules/portal/app/views/batchChanges/batchChangeDetail.scala.html
```diff
@@ -16,9 +16,13 @@
     <div class="page-title">
         <h3>
             <span class="fa fa-list-ol"></span> Batch Change
-            <span ng-if="batch.status =='Complete'" class="label label-success">{{batch.status}}</span>
-            <span ng-if="batch.status =='PartialFailure'" class="label label-warning">Partial Failure</span>
-            <span ng-if="batch.status =='Failed'" class="label label-danger">{{batch.status}}</span>
+            <span ng-if="batch.status == 'Complete'" class="label label-success">{{batch.status}}</span>
```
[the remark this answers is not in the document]
britneywright: I'm basing the labels off the batch status so i would still have the conditionals

modules/portal/public/lib/batch-change/batch-change-detail.controller.js
britneywright: updated this in the other files

modules/portal/public/lib/batch-change/batch-changes.controller.js
britneywright: i removed this

modules/portal/app/views/batchChanges/batchChangeDetail.scala.html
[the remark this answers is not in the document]
mitruly: I would argue that both should be dynamically driven by variables rather than copy/paste.

mitruly:
Can we also fix the `Change Count` display for pending review changes? It currently shows as 0.

modules/portal/public/lib/batch-change/batch-changes.controller.js
```diff
@@ -44,13 +45,21 @@
             $scope.refreshBatchChanges = function() {
                 batchChangePaging = pagingService.resetPaging(batchChangePaging);
 
+                function formatBatchChanges(batchChanges) {
```
mitruly: This isn't being used anywhere.

modules/portal/app/views/batchChanges/batchChangeDetail.scala.html
```diff
@@ -45,6 +49,21 @@ <h3>
                         </span>
                     </p>
                 }
+                <p ng-if="batch.approvalStatus == 'ManuallyApproved'"><strong>Review Status:</strong> Approved</p>
```
pauljamescleary: If the approvalStatus is `Rejected`, should we display `Rejected` here as well?  The approval statuses are `AutoApproved, PendingReview, ManuallyApproved, ManuallyRejected`

modules/portal/app/views/batchChanges/batchChangeDetail.scala.html
pauljamescleary: Not seeing an update on this yet.  If the batch change is approved _or_ rejected, should all of the review fields including empty ones.  Update the `description` to always display even if absent.

modules/portal/app/views/batchChanges/batchChangeDetail.scala.html
britneywright: If the approval statuses mimic the approval statuses I opted not to show the approval statuses. Showing the manual approval because it could still result in one of the possible batch change statuses Complete, Partial Failure or Failure.

modules/portal/app/controllers/FrontendController.scala
```diff
@@ -100,7 +100,7 @@ class FrontendController @Inject()(
   }
 
   def index(): Action[AnyContent] = userAction.async { implicit request =>
-    Future(Ok(views.html.batchChanges.batchChanges(request.user.userName)))
+    Future(Ok(views.html.batchChanges.batchChanges(request.user.userName, request.user.isSuper)))
```
rebstar6: is this grouping super and support, or just super?

britneywright: this is updated to super and support users now

PR #759, opened 2019-07-22
fix a test that travis hates

rebstar6:
worth a shot.

pulling from timeout in https://typelevel.org/cats-effect/datatypes/io.html#race-conditions--race--racepair

seems functionally pretty much the same, but wondering if removing an IO will help us with this test thats failing on timeout

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/Interfaces.scala
+++ b/modules/api/src/main/scala/vinyldns/api/Interfaces.scala
@@ -60,10 +60,15 @@ object Interfaces {
         .handleError(e => Left(e))
     )
 
-  def withTimeout[A](theIo: => IO[A], duration: FiniteDuration, error: Throwable): Result[A] = {
-    val timeOut = IO.sleep(duration) *> IO(error)
-    EitherT(IO.race(timeOut, theIo).handleError(e => Left(e)))
-  }
+  def withTimeout[A](theIo: => IO[A], duration: FiniteDuration, error: Throwable): Result[A] =
+    EitherT(
+      IO.race(timer.sleep(duration), theIo)
+        .map {
+          case Left(_) => Left(error)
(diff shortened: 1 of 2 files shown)


modules/api/src/main/scala/vinyldns/api/Interfaces.scala
```diff
@@ -60,10 +60,15 @@ object Interfaces {
         .handleError(e => Left(e))
     )
 
-  def withTimeout[A](theIo: => IO[A], duration: FiniteDuration, error: Throwable): Result[A] = {
-    val timeOut = IO.sleep(duration) *> IO(error)
-    EitherT(IO.race(timeOut, theIo).handleError(e => Left(e)))
```
rebstar6:
```
    val timeOut = IO.sleep(duration).map(_ => error)
    EitherT(IO.race(timeOut, theIo).handleError(e => Left(e)))
```
may be sufficient too. or none of this and this test will just keep failing

rebstar6:
sad. failed.

modules/api/src/test/scala/vinyldns/api/domain/zone/ZoneConnectionValidatorSpec.scala
```diff
@@ -246,9 +247,10 @@ class ZoneConnectionValidatorSpec
         .when(mockDnsConnection)
         .resolve(badZone.name, badZone.name, RecordType.SOA)
 
-      val result = leftResultOf(underTest.validateZoneConnections(badZone).value)
```
rebstar6: leftResultOf is doing a timing at the same time that the underlying call has a timeout. thinking that structure is what's killing us

just doing an unsafeRunSync with the built in eithervalues/matchers feels way better. 🤞

PR #760, opened 2019-07-23
Add SNS Notifier

dscleaver:
Changes in this pull request:
- Notify SNS topic on batch completion

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/notifier/sns/SnsNotifier.scala
+++ b/modules/api/src/main/scala/vinyldns/api/notifier/sns/SnsNotifier.scala
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2018 Comcast Cable Communications Management, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package vinyldns.api.notifier.sns
+
+import vinyldns.core.notifier.{Notification, Notifier}
+import cats.effect.IO
+import cats.syntax.functor._
+import vinyldns.core.domain.batch.{BatchChange, BatchChangeInfo}
+import vinyldns.api.route.VinylDNSJsonProtocol
+import org.json4s.jackson.JsonMethods._
+import com.amazonaws.services.sns.AmazonSNS
+import com.amazonaws.services.sns.model.PublishRequest
+import com.amazonaws.services.sns.model.MessageAttributeValue
+import org.json4s.JsonAST.JNull
+
+class SnsNotifier(config: SnsNotifierConfig, sns: AmazonSNS)
+    extends Notifier
+    with VinylDNSJsonProtocol {
+
+  def notify(notification: Notification[_]): IO[Unit] =
+    notification.change match {
+      case bc: BatchChange => sendBatchChangeNotification(BatchChangeInfo(bc))
+      case _ => IO.unit
+    }
+
+  def sendBatchChangeNotification(bc: BatchChangeInfo): IO[Unit] =
+    IO {
+      val message =
+        compact(
+          render(BatchChangeInfoSerializer.toJson(bc).replace(List("changes"), JNull)).noNulls)
+      val request = new PublishRequest(config.topicArn, message)
+      request.addMessageAttributesEntry(
+        "userName",
+        new MessageAttributeValue().withDataType("String").withStringValue(bc.userName))
+      sns.publish(request)
+    }.void
+}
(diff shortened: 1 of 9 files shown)


modules/api/src/main/scala/vinyldns/api/notifier/sns/SnsNotifier.scala
```diff
@@ -0,0 +1,51 @@
...
+    notification.change match {
+      case bc: BatchChange => sendBatchChangeNotification(BatchChangeInfo(bc))
+      case _ => IO.unit
+    }
+
+  def sendBatchChangeNotification(bc: BatchChangeInfo): IO[Unit] =
+    IO {
+      val message =
+        compact(
+          render(BatchChangeInfoSerializer.toJson(bc).replace(List("changes"), JNull)).noNulls)
```
pauljamescleary: should we eliminate changes (perhaps changes makes the notification too big?)

project/Dependencies.scala
```diff
@@ -52,7 +52,9 @@ object Dependencies {
     "com.comcast"               %% "ip4s-cats"                      % ip4sV,
     "com.iheart"                %% "ficus"                          % "1.4.3",
     "com.sun.mail"              %  "javax.mail"                     % "1.6.2",
-    "javax.mail"                %  "javax.mail-api"                 % "1.6.2"
+    "javax.mail"                %  "javax.mail-api"                 % "1.6.2",
+    "com.amazonaws"             %  "aws-java-sdk-core"              % awsV withSources(),
+    "com.amazonaws"             %  "aws-java-sdk-sns"              % awsV withSources()
```
pauljamescleary: indentation / alignment

modules/api/src/main/scala/vinyldns/api/notifier/sns/SnsNotifier.scala
dscleaver: I was worried that a lot of changes could make the notification too large... I was also wary of posting out all the changes in a different security medium like AWS... Probably just paranoid on that one.

project/Dependencies.scala
dscleaver: I'll fix

project/Dependencies.scala
```diff
@@ -52,7 +52,9 @@ object Dependencies {
     "com.comcast"               %% "ip4s-cats"                      % ip4sV,
     "com.iheart"                %% "ficus"                          % "1.4.3",
     "com.sun.mail"              %  "javax.mail"                     % "1.6.2",
-    "javax.mail"                %  "javax.mail-api"                 % "1.6.2"
+    "javax.mail"                %  "javax.mail-api"                 % "1.6.2",
+    "com.amazonaws"             %  "aws-java-sdk-core"              % awsV withSources(),
```
mitruly: `aws-java-sdk-core` is already included higher up in `apiDependencies`.

modules/api/src/main/scala/vinyldns/api/notifier/sns/SnsNotifierProvider.scala
```diff
@@ -0,0 +1,54 @@
...
+
+  def load(config: NotifierConfig, userRepository: UserRepository): IO[Notifier] =
+    for {
+      snsConfig <- loadConfigF[IO, SnsNotifierConfig](config.settings)
+      client <- createClient(snsConfig)
+    } yield new SnsNotifier(snsConfig, client)
+
+  def createClient(config: SnsNotifierConfig): IO[AmazonSNS] = IO {
+    logger.error(
+      s"Setting up sns notifier client with settings: " +
```
mitruly: Technically don't need the `s` here since we're not interpolating.

modules/api/src/main/scala/vinyldns/api/notifier/sns/SnsNotifier.scala
```diff
@@ -0,0 +1,51 @@
...
+      val request = new PublishRequest(config.topicArn, message)
+      request.addMessageAttributesEntry(
+        "userName",
+        new MessageAttributeValue().withDataType("String").withStringValue(bc.userName))
+      sns.publish(request)
+    }.void
```
mitruly: `.void` doesn't appear to be needed here.

modules/api/src/test/scala/vinyldns/api/notifier/sns/SnsNotifierSpec.scala
```diff
@@ -0,0 +1,147 @@
...
+          "service-endpoint" -> "someValue",
+          "signing-region" -> "us-east-1",
+          "access-key" -> "access",
+          "secret-key" -> "secret"
+        ).asJava)
+      val notifier = new SnsNotifierProvider()
+        .load(NotifierConfig("", snsConfig), mockUserRepository)
+        .unsafeRunSync()
+
+      notifier.notify(new Notification("this won't be supported ever")) should be(IO.unit)
```
mitruly: `new` qualifier doesn't seem to be needed here.

project/Dependencies.scala
dscleaver: nice catch.

modules/api/src/main/scala/vinyldns/api/notifier/sns/SnsNotifierProvider.scala
dscleaver: I actually copied this from the SQS code. There's a fix needed there too, which I'll do in another PR.

modules/api/src/main/scala/vinyldns/api/notifier/sns/SnsNotifier.scala
dscleaver: `sns.publish` does not return Unit. `.void` converts the IO to IO[Unit]

mitruly: Gotcha; tests seemed to pass without it but explicit conversion to `IO[Unit]` is fine with me.

britneywright:
Seems we never got around to documentation about the email notifier. making a mental not that some info here may need to go into the docs.

PR #761, opened 2019-07-24
Add scheduled change feature flag and check

pauljamescleary:
Add a scheduled change feature flag.  If a user submits a batch change that is scheduled, and the feature flag is disabled, then raise an error.

* `BatchChangeValidations` - added a class member variable that holds the feature flag.  Added a function `validateScheduledChange` that has the business rule if the feature flag is disabled and scheduled time is set then raise an error
* `Boot` - pass the config value into the `BatchChangeValidations` constructor
* `BatchChangeRoute` - added error handler for `ScheduleChangeDisabled` error
* `VinylDNSConfig` - add a feature flag
* `api/reference.conf` - ensure that the scheduled batch change enabled flag is set to `false` by default (it is off)

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeErrors.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeErrors.scala
@@ -55,3 +55,8 @@ final case class BatchRequesterNotFound(userId: String, userName: String)
   def message: String =
     s"The requesting user with id $userId and name $userName cannot be found in VinylDNS"
 }
+
+case object ScheduledChangesDisabled extends BatchChangeErrorResponse {
+  val message: String =
+    s"""Cannot create a scheduled change, as it is currently disabled on this VinylDNS instance."""
+}
--- a/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeErrors.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeErrors.scala
@@ -58,5 +58,5 @@ final case class BatchRequesterNotFound(userId: String, userName: String)
 
 case object ScheduledChangesDisabled extends BatchChangeErrorResponse {
   val message: String =
-    s"""Cannot create a scheduled change, as it is currently disabled on this VinylDNS instance."""
+    "Cannot create a scheduled change, as it is currently disabled on this VinylDNS instance."
 }
--- a/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
@@ -72,17 +72,18 @@ class BatchChangeValidations(
       input: BatchChangeInput,
       existingGroup: Option[Group],
       authPrincipal: AuthPrincipal): BatchResult[Unit] = {
-    val validations = validateBatchChangeInputSize(input) |+| validateScheduledChange(
-      input,
-      scheduledChangesEnabled) |+| validateOwnerGroupId(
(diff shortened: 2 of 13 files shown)


modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
```diff
@@ -76,10 +77,13 @@ class BatchChangeValidations(
...
 
-    EitherT.fromEither(
-      validations
+    for {
+      _ <- validations
         .leftMap[BatchChangeErrorResponse](nel => InvalidBatchChangeInput(nel.toList))
-        .toEither)
+        .toEither
+        .toBatchResult
+      _ <- validateScheduledChange(input, scheduledChangesEnabled).toBatchResult
```
pauljamescleary: this cannot be in the `validations` above because it is not a `SingleChangeError`, it applies to the whole batch

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeErrors.scala
```diff
@@ -55,3 +55,8 @@ final case class BatchRequesterNotFound(userId: String, userName: String)
   def message: String =
     s"The requesting user with id $userId and name $userName cannot be found in VinylDNS"
 }
+
+case object ScheduledChangesDisabled extends BatchChangeErrorResponse {
+  val message: String =
+    s"""Cannot create a scheduled change, as it is currently disabled on this VinylDNS instance."""
```
mitruly: Technically don't need `s"""` since we're not interpolating or including `"` within the string.

britneywright:
What happens if `scheduled-changes-enabled` is true, but `manual-batch-review-enabled` is false.?

pauljamescleary:
@britneywright good question.  The "full" scheduled change solution will have them auto-process at the scheduled time.  We are implementing this first step (to go to manual review always) as an interim step toward the full solution.  Actually, we haven't implemented the auto-goto-manual code as of yet.

- if manual review enabled; then if a scheduled change fails then it gets kicked out for a human to intervene
- if manual review disabled; then if a scheduled change fails then it completes with a partial failure

There seem to be a few possible solutions:

1. Fail to start as the configuration is invalid (sched ON manual OFF).  This is the technically correct because the system is in an invalid state based on configuration and operate under the configuration.  The HUGE downside is that somehow linking these two feature flags together is very confusing, difficult to document, and feels wrong.

2. Reject batch requests (sched ON manual OFF).  This would avoid the confusion of solution 1, however, now the system is not behaving as configured, which is absolutely terrible.  Someone submits a batch change, everything checks out as valid, a scheduled time is set, but we reject it?  What is that message?  "Sorry user, invalid configuration is preventing me from accepting this request"?  This ends up also being confusing.

Given that this is an interim solution, I think failing the system startup (solution 1) is better as it will prevent the system from getting into a state where it is not behaving as desired.  Once the full solution is ready we can remove that check.  The downside is that we have not ever validated config like this, so the implementation on that is unclear.  I would prefer to do that in a separate PR and think it through.

rebstar6:
@pauljamescleary i would just fail to start if the configuration is invalid (sched ON manual OFF).

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
rebstar6: oh shoot, that was done wrong. validateBatchChangeInputSize and validateOwnerGroupId should have been BatchChangeErrorResponse

bummer.

rebstar6: oh hmm, on 2nd thought, i bet we did that for accumulation. lets you return multiple errors on the input in the case that there are multiple

PR #764, opened 2019-07-24
Add external-dns as an integration to Tools page

dgrizzanti:
Changes in this pull request:
- Add `external-dns` as a new "integration" to the tools page. I didn't mention the vinyl integration specifically in the external-dns line, let me know if you think that is needed/would read better

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -32,6 +32,7 @@ Thank you! If you have contributed in any way, but do not see your name here, pl
 - Luke Cori
 - Jearvon Dharrie
 - Andrew Dunn
+- David Grizzanti
 - Alejandro Guirao
 - Daniel Jin
 - Krista Khare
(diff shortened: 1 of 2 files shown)


rebstar6:
@dgrizzanti yeah the wording is tough. in theory a note about where vinyldns fits into external-dns would be good, but only if you can come up with something (I cant off the top of my head)

dgrizzanti:
@rebstar6 let me know if either of these sound any better. I forgot to include Cloud Foundry in the list so I add that too. I'll push once we decide on the verbiage 🙂

```
- [external-dns](https://github.com/kubernetes-incubator/external-dns) - synchronizes Cloud Foundry and Kubernetes resources with VinylDNS
- [external-dns](https://github.com/kubernetes-incubator/external-dns) - DNS provider-agnostic syncronization of Cloud Foundry and Kubernetes resources, including VinylDNS
```

Thanks!

rebstar6:
@dgrizzanti i like #2

PR #767, opened 2019-07-25
Zone discovery soft failure and tests

rebstar6:
this PR will be mostly func tests

### DIFF
--- a/docker/bind9/zones/not.loaded.hosts
+++ b/docker/bind9/zones/not.loaded.hosts
@@ -1,0 +1,9 @@
+$ttl 38400
+not.loaded.	IN	SOA	172.17.42.1. admin.test.com. (
+			1439234395
+			10800
+			3600
+			604800
+			38400 )
+not.loaded.      IN  NS  172.17.42.1.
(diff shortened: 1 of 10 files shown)


docker/bind9/zones/not.loaded.hosts
```diff
@@ -0,0 +1,9 @@
+$ttl 38400
+not.loaded.	IN	SOA	172.17.42.1. admin.test.com. (
```
pauljamescleary: in the rest of the func tests, for a zone that is loaded/un-loaded in a single func test we use "one-time" zone.  Is there something different about "not loaded"?

rebstar6: good call, forgot about that guy.

looking into it, i believe it may conflict with the create zone test, since that checks all the records (SOA/serial included).

I can change to that though and tweak the create test if you think its not worth creating a new zone for this

pauljamescleary: I am ok with leaving this, one-time is only used for zone tests (not recordsets).

PR #768, opened 2019-07-25
expand batch change statuses

britneywright:
Part of #659 

Changes in this pull request:
- Instead of using the existing Pending and Failed statuses use PendingReview for a batch change in the pending state and use Rejected when a batch change is ManuallyRejected.

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/engine/RecordSetChangeHandler.scala
+++ b/modules/api/src/main/scala/vinyldns/api/engine/RecordSetChangeHandler.scala
@@ -73,7 +73,7 @@ object RecordSetChangeHandler {
       batchChangeOption <- batchChangeRepository.updateSingleChanges(singleChangeStatusUpdates)
       _ <- batchChangeOption.map { bc =>
         bc.status match {
-          case BatchChangeStatus.Pending =>
+          case BatchChangeStatus.PendingProcessing | BatchChangeStatus.PendingReview =>
             IO.unit
           case _ =>
             notifiers.notify(Notification(bc))
--- a/modules/core/src/main/scala/vinyldns/core/domain/batch/BatchChange.scala
+++ b/modules/core/src/main/scala/vinyldns/core/domain/batch/BatchChange.scala
@@ -62,7 +62,8 @@ case class BatchChange(
  */
 object BatchChangeStatus extends Enumeration {
   type BatchChangeStatus = Value
-  val Complete, Failed, PartialFailure, PendingProcessing, PendingReview, Rejected, Scheduled = Value
+  val Complete, Failed, PartialFailure, PendingProcessing, PendingReview, Rejected, Scheduled =
+    Value
 
   def calculateBatchStatus(
       approvalStatus: BatchChangeApprovalStatus,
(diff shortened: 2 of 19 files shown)


modules/api/src/main/scala/vinyldns/api/engine/RecordSetChangeHandler.scala
```diff
@@ -73,7 +73,7 @@ object RecordSetChangeHandler {
       batchChangeOption <- batchChangeRepository.updateSingleChanges(singleChangeStatusUpdates)
       _ <- batchChangeOption.map { bc =>
         bc.status match {
-          case BatchChangeStatus.Pending =>
+          case BatchChangeStatus.PendingProcessing | BatchChangeStatus.PendingReview =>
```
pauljamescleary: wonder why we are keying of the status here, should we just say "if any of the remaining batch statuses are pending?"

modules/core/src/main/scala/vinyldns/core/domain/batch/BatchChange.scala
```diff
@@ -54,19 +54,19 @@ case class BatchChange(
  */
 object BatchChangeStatus extends Enumeration {
   type BatchChangeStatus = Value
-  val Pending, Complete, Failed, PartialFailure = Value
+  val PendingProcessing, PendingReview, Complete, Failed, PartialFailure, Rejected = Value
```
pauljamescleary: I like these statuses

mitruly:
On second glance, I think it's worthwhile to convert `SingleChangeStatus.Pending` to `SingleChangeStatus.PendingProcessing` as well for consistency.

modules/api/src/main/scala/vinyldns/api/engine/RecordSetChangeHandler.scala
britneywright: you mean the single changes?

britneywright:
> On second glance, I think it's worthwhile to convert `SingleChangeStatus.Pending` to `SingleChangeStatus.PendingProcessing` as well for consistency.

Unfortunately that’s one character longer than our limit for that field in the database. I’d save that for a future PR where we either extend or change that field to a different type.

modules/api/src/main/scala/vinyldns/api/engine/RecordSetChangeHandler.scala
pauljamescleary: @britneywright that is the idea, I guess it is one in the same though as the status is built off of the individual changes.  Going to resolve this

modules/core/src/main/scala/vinyldns/core/domain/batch/BatchChange.scala
```diff
@@ -54,19 +54,19 @@ case class BatchChange(
  */
 object BatchChangeStatus extends Enumeration {
   type BatchChangeStatus = Value
-  val Pending, Complete, Failed, PartialFailure = Value
```
rebstar6: there is a comment above here on the status definitions. can you update that comment?

PR #770, opened 2019-07-26
Add configuration for VinylDNS notifiers

mitruly:
Provide configuration information and example for notifiers.


britneywright:
Should also include something in the batch change api and portal docs that notes an email is sent if configured in the instance. And maybe a screenshot somewhere? Also maybe a feature to highlight on the homepage list?

modules/docs/src/main/tut/operator/config-api.md
```diff
@@ -458,6 +458,50 @@ has multiple records. Create batch will also fail if the user is attempting to m
...
+```yaml
+notifiers = ["email", "sns"]
+```
+
+#### E-mail notifier
+Configuration for the e-mail notifier appears like the following:
+
+```yaml
+email = {
+  class-name = "vinyldns.api.notifier.email.EmailNotifierProvider"
```
pauljamescleary: it is good form to put a comment line in the yaml itslef above anything that needs calling out.  I think here the from and smtp.host would be good for comments

PR #771, opened 2019-07-26
Advance scheduled change to PendingApproval

pauljamescleary:
When a scheduled change is submitted, if there are no hard errors advance to PendingApproval status.

* `BatchChange` - changed the calculation of the batch change status; if it is pending approval and the scheduled time is set it will be `Scheduled`
* `MySqlBatchChangeRepository` - updated the de-serialize to consider scheduled time so when "getting" a batch change, the status returned will appropriately be `Scheduled`
* `BatchChangeService` - updated the `buildResponse` method to consider scheduled time.  If no fatal errors, manual review enabled, and scheduled then goto PendingApproval

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
@@ -419,7 +419,7 @@ class BatchChangeService(
     } else {
       // we have soft errors
       // advance to manual review if this is scheduled OR manual review is ok
-      val gotoManualReview = isScheduled || (this.manualReviewEnabled && allowManualReview)
+      val gotoManualReview = this.manualReviewEnabled && (isScheduled || allowManualReview)
       if (gotoManualReview) {
         manualReviewResponse
       } else {
--- a/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
@@ -359,33 +359,12 @@ class BatchChangeService(
       auth: AuthPrincipal,
       allowManualReview: Boolean): Either[BatchChangeErrorResponse, BatchChange] = {
 
-    val allErrors = transformed
-      .collect {
-        case Invalid(e) => e
-      }
-      .flatMap(_.toList)
+    // Respond with a fatal error that kicks the change out to the user
+    def errorResponse =
+      InvalidBatchChangeResponses(batchChangeInput.changes, transformed).asLeft
 
-    val allNonFatal = allErrors.forall(!_.isFatal)
-    val isScheduled = batchChangeInput.scheduledTime.isDefined && this.scheduledChangesEnabled
-
-    // If someone says "don't manual review" but "schedule it", right now we ignore the former and
-    // just goto manual review.
-    val gotoManualReview = manualReviewEnabled && allNonFatal && (allowManualReview || isScheduled)
-
-    if (allErrors.isEmpty && !isScheduled) {
-      val changes = transformed.getValid.map(_.asStoredChange())
-      BatchChange(
-        auth.userId,
-        auth.signedInUser.userName,
-        batchChangeInput.comments,
-        DateTime.now,
-        changes,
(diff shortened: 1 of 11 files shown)


modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -362,8 +366,13 @@ class BatchChangeService(
       .flatMap(_.toList)
 
     val allNonFatal = allErrors.forall(!_.isFatal)
+    val isScheduled = batchChangeInput.scheduledTime.isDefined && this.scheduledChangesEnabled
 
-    if (allErrors.isEmpty) {
+    // If someone says "don't manual review" but "schedule it", right now we ignore the former and
+    // just goto manual review.
+    val gotoManualReview = manualReviewEnabled && allNonFatal && (allowManualReview || isScheduled)
+
```
pauljamescleary: this is an interesting logic decision:

Someone posts a batch change with `allowManualReview=false`, but it has a scheduled time set on it.  What do we do?

1. Chosen solution - here we ignore the `allowManualReview`

2. Alternative - we could add an if/else here that says `if allowManualReview == false && isScheduled then throw error`

Wonder what people's thoughts are.

modules/core/src/main/scala/vinyldns/core/domain/batch/BatchChange.scala
```diff
@@ -54,21 +60,23 @@ case class BatchChange(
...
+      case BatchChangeApprovalStatus.PendingApproval if isScheduled => BatchChangeStatus.Scheduled
       case BatchChangeApprovalStatus.PendingApproval => BatchChangeStatus.Pending
       case BatchChangeApprovalStatus.ManuallyRejected => BatchChangeStatus.Failed
       case _ =>
         (hasPending, hasFailed, hasComplete) match {
           case (true, _, _) => BatchChangeStatus.Pending
-          case (_, true, true) => BatchChangeStatus.PartialFailure
-          case (_, true, false) => BatchChangeStatus.Failed
+          case (false, true, true) => BatchChangeStatus.PartialFailure
+          case (false, true, false) => BatchChangeStatus.Failed
```
pauljamescleary: I explicitly called out `false` here because parsing these 3 booleans in the pattern match was hard to follow.  The first pattern that says `true` makes the next two a moot point, but that was _sooo_ easy to miss I had to stare at this longer than I needed to.

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
britneywright: I’d go with the ignore allowManualReview

rebstar6: i disagree, give them the errors if there are any.

the point of being able to set allowManualReview=false is so that people can see/addess their errors. if that is ignored, they dont have that power

pauljamescleary: @rebstar6 can add this as an upfront validation

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -355,47 +359,73 @@ class BatchChangeService(
...
+    val allErrors = transformed
+      .collect {
+        case Invalid(e) => e
+      }
+      .flatMap(_.toList)
+
+    // Tells us that we have soft errors, and no errors are hard errors
+    val hardErrorsPresent = allErrors.exists(_.isFatal)
+    val noErrors = allErrors.isEmpty
+    val isScheduled = batchChangeInput.scheduledTime.isDefined && this.scheduledChangesEnabled
```
mitruly: `this.` isn't needed

pauljamescleary: @mitruly it is a little of setting new precedent, but it is hard with a file this humongous to know where all these variables are coming from.  I liked using `this` to identify it as a class level variable.

PR #774, opened 2019-07-30
Sched approval

pauljamescleary:
Add check to make sure that a scheduled change is not past due when approving.

* `BatchChangeErrors` - added `ScheduledChangeNotDue` error type that will be raised in the event someone approves a batch change (i.e. processes) a scheduled change too soon.
* `BatchChangeValidations` - added a `validateScheduledApproval` function that raises the `ScheduledChangeNotDue` if the scheduled date is not past due
* `BatchChangeRouting` - added `ScheduledChangeNotDue` to `handleErrors` to return a `Forbidden`

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
@@ -117,7 +117,7 @@ class BatchChangeValidations(
       batchChange: BatchChange,
       authPrincipal: AuthPrincipal): Either[BatchChangeErrorResponse, Unit] =
     validateAuthorizedReviewer(authPrincipal, batchChange) |+| validateBatchChangePendingReview(
-      batchChange)
+      batchChange) |+| validateScheduledApproval(batchChange)
 
   def validateBatchChangePendingReview(
       batchChange: BatchChange): Either[BatchChangeErrorResponse, Unit] =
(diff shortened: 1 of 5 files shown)


modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
```diff
@@ -135,6 +135,12 @@ class BatchChangeValidations(
       UserNotAuthorizedError(batchChange.id).asLeft
     }
 
+  def validateScheduledApproval(batchChange: BatchChange): Either[BatchChangeErrorResponse, Unit] =
+    batchChange.scheduledTime match {
+      case Some(dt) if dt.isAfterNow => Left(ScheduledChangeNotDue(dt))
```
pauljamescleary: of note: `isAfterNow` compares the `DateTime` to `DateTime.currentTimeMillis` ... which returns "the current time in milliseconds from 1970-01-01T00:00:00Z" so also UTC / ZULU time

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
```diff
@@ -117,7 +117,7 @@ class BatchChangeValidations(
       batchChange: BatchChange,
       authPrincipal: AuthPrincipal): Either[BatchChangeErrorResponse, Unit] =
     validateAuthorizedReviewer(authPrincipal, batchChange) |+| validateBatchChangePendingApproval(
-      batchChange)
+      batchChange) |+| validateScheduledApproval(batchChange)
```
rebstar6: wow i just noticed we use |+| here to combine eithers rather than for comp. i didnt even realize you could do that!

PR #778, opened 2019-08-02
link to groups in the portal Manage Zone view

britneywright:
portal Manage Zones view with links to group pages

From a support perspective i find it annoying that I have to go to a zone to see an owner group or ACL group and then go to the Groups page to look up the group to see the members. Want to make the group names in the Manage Zones view link directly to the Group pages that list the members.

Changes in this pull request:
- Add a link under the admin group selection with a link to the current zone owner group
- Link the group name in ACL rules to the group page

### DIFF
--- a/modules/portal/app/views/zones/zoneTabs/manageZone.scala.html
+++ b/modules/portal/app/views/zones/zoneTabs/manageZone.scala.html
@@ -72,6 +72,7 @@
                                                 ng-selected="updateZoneInfo.adminGroupName == group.name">
                                             {{group.name}} ({{group.description}})</option>
                                     </select>
+                                    <a href="/groups/{{ updateZoneInfo.adminGroupId  }}">View group</a>
                                 </div>
                             </div>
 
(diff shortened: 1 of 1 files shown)


modules/portal/app/views/zones/zoneTabs/manageZone.scala.html
```diff
@@ -72,6 +72,7 @@ <h3 class="panel-title">Zone Info</h3>
                                                 ng-selected="updateZoneInfo.adminGroupName == group.name">
                                             {{group.name}} ({{group.description}})</option>
                                     </select>
+                                    <a href="/groups/{{ updateZoneInfo.adminGroupId  }}">View group</a>
```
mitruly: Nitpick: Extra whitespace at the end.

mitruly: May also be nice to add `target="_blank"` to both links so that it doesn't change the current window.

britneywright: Fixed the whitespace. We don't open internal pages in new tabs, if we want to do that it should be done across the portal to be consistent. Separate PR.

PR #779, opened 2019-08-05
Implement domains requiring manual review (via batch change interface)

mitruly:
Similar to existing high-value domains where we disallow certain records from being created via configuration, domains requiring manual review is a configured list that requires manual review (via batch interface) for the change to process. This PR takes care of only the batch interface process flow, with the record set restriction being handled separately.

Changes in this pull request:
- Add configuration (`manual-review-domains`) for changes that will automatically be flagged as non-fatal errors. If manual review is disabled, this will manifest as fatal errors. If manual review is enabled, these will be handled normally.
- Add tests

### DIFF
--- a/docker/api/docker.conf
+++ b/docker/api/docker.conf
@@ -210,7 +210,7 @@ vinyldns {
   # FQDNs / IPs that require manual review upon submission in batch change interface
   # domain-list used for all record types except PTR
   # ip-list used exclusively for PTR records
-  domains-requiring-manual-review = {
+  manual-review-domains = {
     domain-list = [
       "needs-review.*"
     ]
--- a/docker/api/docker.conf
+++ b/docker/api/docker.conf
@@ -207,6 +207,21 @@ vinyldns {
     ]
   }
 
+  # FQDNs / IPs that require manual review upon submission in batch change interface
+  # domain-list used for all record types except PTR
+  # ip-list used exclusively for PTR records
+  domains-requiring-manual-review = {
+    domain-list = [
+      "needs-review.*"
+    ]
+    ip-list = [
+      "192.0.2.254",
+      "192.0.2.255",
+      "fd69:27cc:fe91:0:0:0:ffff:1",
+      "fd69:27cc:fe91:0:0:0:ffff:2"
+    ]
+  }
+
   # types of unowned records that users can access in shared zones
   shared-approved-types = ["A", "AAAA", "CNAME", "PTR", "TXT"]
 
--- a/modules/api/functional_test/live_tests/batch/approve_batch_change_test.py
+++ b/modules/api/functional_test/live_tests/batch/approve_batch_change_test.py
@@ -11,7 +11,8 @@ def test_approve_pending_batch_change_success(shared_zone_test_context):
     approver = shared_zone_test_context.support_user_client
     batch_change_input = {
         "changes": [
-            get_change_A_AAAA_json("test-approve-success.not.loaded.", address="4.3.2.1")
+            get_change_A_AAAA_json("test-approve-success.not.loaded.", address="4.3.2.1"),
+            get_change_A_AAAA_json("needs-review.not.loaded.", address="4.3.2.1"),
         ],
         "ownerGroupId": shared_zone_test_context.ok_group['id']
     }
@@ -25,6 +26,8 @@ def test_approve_pending_batch_change_success(shared_zone_test_context):
         assert_that(get_batch['approvalStatus'], is_('PendingReview'))
         assert_that(get_batch['changes'][0]['status'], is_('NeedsReview'))
         assert_that(get_batch['changes'][0]['validationErrors'][0]['errorType'], is_('ZoneDiscoveryError'))
+        assert_that(get_batch['changes'][1]['status'], is_('NeedsReview'))
+        assert_that(get_batch['changes'][1]['validationErrors'][0]['errorType'], is_('RecordRequiresManualReview'))
 
         # need to create the zone so the change can succeed
         zone = {
@@ -43,12 +46,13 @@ def test_approve_pending_batch_change_success(shared_zone_test_context):
         to_delete = [(change['zoneId'], change['recordSetId']) for change in completed_batch['changes']]
 
         assert_that(completed_batch['status'], is_('Complete'))
-        assert_that(completed_batch['changes'][0]['status'], is_('Complete'))
+        for i in xrange(len(completed_batch['changes'])):
+            assert_that(completed_batch['changes'][i]['status'], is_('Complete'))
+            assert_that(len(completed_batch['changes'][i]['validationErrors']), is_(0))
         assert_that(completed_batch['approvalStatus'], is_('ManuallyApproved'))
         assert_that(completed_batch['reviewerId'], is_('support-user-id'))
         assert_that(completed_batch['reviewerUserName'], is_('support-user'))
         assert_that(completed_batch, has_key('reviewTimestamp'))
-        assert_that(len(completed_batch['changes'][0]['validationErrors']), is_(0))
     finally:
         clear_zoneid_rsid_tuple_list(to_delete, client)
         if to_disconnect is not None:
--- a/modules/api/functional_test/live_tests/batch/approve_batch_change_test.py
+++ b/modules/api/functional_test/live_tests/batch/approve_batch_change_test.py
@@ -46,9 +46,9 @@ def test_approve_pending_batch_change_success(shared_zone_test_context):
         to_delete = [(change['zoneId'], change['recordSetId']) for change in completed_batch['changes']]
 
         assert_that(completed_batch['status'], is_('Complete'))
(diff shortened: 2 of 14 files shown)


docker/api/docker.conf
```diff
@@ -207,6 +207,21 @@ vinyldns {
...
   }
 
+  # FQDNs / IPs that require manual review upon submission in batch change interface
+  # domain-list used for all record types except PTR
+  # ip-list used exclusively for PTR records
+  domains-requiring-manual-review = {
+    domain-list = [
+      "needs-review.*"
+    ]
+    ip-list = [
```
mitruly: Setup mirrors our high-value domains. Pre-emptively setting up IPs separately due to record set endpoint (not handled in this PR).

docker/api/docker.conf
```diff
@@ -207,6 +207,21 @@ vinyldns {
     ]
   }
 
+  # FQDNs / IPs that require manual review upon submission in batch change interface
+  # domain-list used for all record types except PTR
+  # ip-list used exclusively for PTR records
+  domains-requiring-manual-review = {
```
pauljamescleary: this is a rather long name, "manual-review-domains"

modules/api/functional_test/live_tests/batch/create_batch_change_test.py
```diff
@@ -860,6 +860,45 @@ def test_create_batch_change_with_high_value_domain_fails(shared_zone_test_conte
...
+            get_change_PTR_json("fd69:27cc:fe91:0:0:0:ffff:2"),
+            get_change_PTR_json("fd69:27cc:fe91:0:0:0:ffff:2", change_type="DeleteRecordSet"),
+
+            get_change_A_AAAA_json("i-can-be-touched.ok.", address="1.1.1.1")
+        ]
+    }
+    response = None
+
+    try:
+        response = client.create_batch_change(batch_change_input, status=202)
```
pauljamescleary: probably should add a check that it is in a PendingApproval state

pauljamescleary: probably should add a check that on the change, every one of the appropriate changes has an appropriate soft error

docker/api/docker.conf
rebstar6: this is on *domains (zones)* or *fqdns*?

I was thinking zones, in which case im not sure if the IP approach makes sense, should discuss

modules/api/functional_test/live_tests/batch/create_batch_change_test.py
rebstar6: yeah, i threw a lot of this logic in the approve_batch_change_test instead of create when doing zonediscovery (no real reason, just needed a home and both sides are tested at once).

should pull all the validations on the response from that

docker/api/docker.conf
rebstar6: looking through the code, i think im good with it being on the input name as it is, a few considerations though:
- does the IP list code CIDR or just single IPs?
- if we need to, is there a way to say "all ipv6 needs manual review"? not saying we will need that, but its not impossible...

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
```diff
@@ -181,7 +181,7 @@ class BatchChangeValidations(
       case other =>
         InvalidBatchRecordType(other.toString, SupportedBatchChangeRecordTypes.get).invalidNel[Unit]
     }
-    typedChecks |+| isNotHighValueDomain(change)
+    typedChecks |+| isNotHighValueDomain(change) |+| doesNotRequireManualReview(change)
```
britneywright: how does this work when a pending batch change is approved?

mitruly: Added a `isReviewing` flag to exclude this check for review mode.

modules/api/functional_test/live_tests/batch/approve_batch_change_test.py
```diff
@@ -43,12 +46,13 @@ def test_approve_pending_batch_change_success(shared_zone_test_context):
         to_delete = [(change['zoneId'], change['recordSetId']) for change in completed_batch['changes']]
 
         assert_that(completed_batch['status'], is_('Complete'))
-        assert_that(completed_batch['changes'][0]['status'], is_('Complete'))
+        for i in xrange(len(completed_batch['changes'])):
```
britneywright: typo?

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
britneywright: can you change that name to `isApproved` or `isAllowed` since rejections don't go through validation again. `isAllowed` might be better since we don't change the status to approve until after the batch change is validated again, and I think it's more understandable than `isReviewing`.

modules/api/functional_test/live_tests/batch/approve_batch_change_test.py
mitruly: `xrange` is an iterator for a collection while `range` actually constructs a new list.

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -107,11 +110,12 @@ class BatchChangeService(
 
   def applyBatchChangeValidationFlow(
       batchChangeInput: BatchChangeInput,
-      auth: AuthPrincipal): BatchResult[BatchValidationFlowOutput] =
+      auth: AuthPrincipal,
+      isReviewing: Boolean): BatchResult[BatchValidationFlowOutput] =
```
britneywright: made a comment here about renaming `isReviewing`: https://github.com/vinyldns/vinyldns/pull/779#discussion_r310741934

docker/api/docker.conf
pauljamescleary: @rebstar6 realistically since IP addresses work on the colon / nibble, you could say something like `[0-9a-f]{1-4}\:` and have that group repeat (sorry I didn't have the full solution) and would essentially match all IP v6

Ahh, here is one (long)

`(?:^|(?<=\s))(([0-9a-fA-F]{1,4}:){7,7}[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,7}:|([0-9a-fA-F]{1,4}:){1,6}:[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,5}(:[0-9a-fA-F]{1,4}){1,2}|([0-9a-fA-F]{1,4}:){1,4}(:[0-9a-fA-F]{1,4}){1,3}|([0-9a-fA-F]{1,4}:){1,3}(:[0-9a-fA-F]{1,4}){1,4}|([0-9a-fA-F]{1,4}:){1,2}(:[0-9a-fA-F]{1,4}){1,5}|[0-9a-fA-F]{1,4}:((:[0-9a-fA-F]{1,4}){1,6})|:((:[0-9a-fA-F]{1,4}){1,7}|:)|fe80:(:[0-9a-fA-F]{0,4}){0,4}%[0-9a-zA-Z]{1,}|::(ffff(:0{1,4}){0,1}:){0,1}((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])|([0-9a-fA-F]{1,4}:){1,4}:((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9]))(?=\s|$)
`

I do think we normalize IP6 addresses so ours could be much smaller

rebstar6: it we CIDR rule though...
we were discussing CIDR or regex for reverse, not both

modules/api/functional_test/live_tests/batch/approve_batch_change_test.py
rebstar6: fwiw, i think you could also do 
for change in completed_batch['changes']:
    assert_that(change['status'], is_('Complete'))
    assert_that(len(change['validationErrors']), is_(0))

britneywright:
This error should be dependent on `manual-batch-review-enabled` being enabled.

mitruly:
@britneywright Think there should be some discussion on this. If a user is configuring `manual-review-domains`, it is intentional. I opted to take the liberty of saying that a domain considered requiring approval despite manual review being disabled should be treated like our high-value domains.

I think that if we're upfront about how the implementation and behavior, then this approach is fine. Would like more input from others before changing this to completely disable if manual review is.

rebstar6:
@britneywright @mitruly I think as is is fine on config, but thats definitely a note that should be included in the operator guide when we update that.

Does bring up an interesting question of whether we should let support/super bypass that check...I'm thinking yes.

PR #780, opened 2019-08-05
correctly get batchChangeSummary totalChanges

britneywright:
cleanup for #659 

Changes in this pull request:
- Update the pending count to include single changes with a status of "Pending" or "NeedsReview"


modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlBatchChangeRepositoryIntegrationSpec.scala
```diff
@@ -183,6 +183,7 @@ class MySqlBatchChangeRepositoryIntegrationSpec
     actual.userName shouldBe expected.userName
     actual.createdTimestamp.getMillis shouldBe expected.createdTimestamp.getMillis +- 2000
     actual.ownerGroupId shouldBe expected.ownerGroupId
+    actual.totalChanges shouldBe expected.totalChanges
```
rebstar6: i dont see any tests here with NeedsReview changes anyway, so this would still have succeeded before the change.

can you just make sure theres one in here? id recommend updating `change_one` since thats already in PendingReview state

PR #782, opened 2019-08-05
allow underscores in batch change requests

britneywright:
domains like `_jabber._tcp.gmail.com` are valid but would fail the batch change validations because it contains an underscore.

Changes in this pull request:
- relax the `validFQDNRegex` by including underscores
- update the `InvalidDomainName` error message to include underscores as valid characters

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/DomainValidations.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/DomainValidations.scala
@@ -28,7 +28,7 @@ import scala.util.matching.Regex
  */
 object DomainValidations {
   val validFQDNRegex: Regex =
-    """^(?:([0-9a-zA-Z]{1,63}|[0-9a-zA-Z]{1}[0-9a-zA-Z\-\/]{0,61}[0-9a-zA-Z]{1})\.)*$""".r
+    """^(?:([0-9a-zA-Z_]{1,63}|[0-9a-zA-Z_]{1}[0-9a-zA-Z\-\/_]{0,61}[0-9a-zA-Z_]{1})\.)*$""".r
   val validIpv4Regex: Regex =
     """^(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$""".r
   val validIpv6Regex: Regex =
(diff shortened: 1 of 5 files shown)


modules/api/src/test/scala/vinyldns/api/domain/DomainValidationsSpec.scala
```diff
@@ -58,6 +58,7 @@ class DomainValidationsSpec
   property("Domain names beginning with invalid characters should fail with InvalidDomainName") {
     validateHostName("/slash.domain.name.").failWith[InvalidDomainName]
     validateHostName("-hyphen.domain.name.").failWith[InvalidDomainName]
+    validateHostName("_underscore.domain.name.").isValid
```
pauljamescleary: I wouldn't put this in this property.  The property is testing for InvalidDomainName, a difffer property that underscores should succeed would be better

britneywright:
> So fun facts, we can update bind9 to allow underscores (if we haven't already). I would add a func test to make sure that you can submit a batch change with a record that has an underscore.

I think bind9 already allows underscores. We can add records with underscores through Zones, just not through Batch Change.

modules/api/src/main/scala/vinyldns/api/domain/DomainValidations.scala
```diff
@@ -28,7 +28,7 @@ import scala.util.matching.Regex
  */
 object DomainValidations {
   val validFQDNRegex: Regex =
-    """^(?:([0-9a-zA-Z]{1,63}|[0-9a-zA-Z]{1}[0-9a-zA-Z\-\/]{0,61}[0-9a-zA-Z]{1})\.)*$""".r
+    """^(?:([0-9a-zA-Z_]{1,63}|[0-9a-zA-Z_]{1}[0-9a-zA-Z\-\/_]{0,61}[0-9a-zA-Z_]{1})\.)*$""".r
```
mitruly: This is a bit pedantic (especially since allowing leading underscores is different than the rule with hyphenation), but wondering if we need to support trailing underscores. Doesn't necessarily hurt, but wondering if we should try to limit what we allow through.

Easy route would be keeping these validations as-is.

modules/api/src/test/scala/vinyldns/api/domain/DomainValidationsSpec.scala
```diff
@@ -60,6 +60,12 @@ class DomainValidationsSpec
     validateHostName("-hyphen.domain.name.").failWith[InvalidDomainName]
   }
 
+  property("Domain names with underscores should pass property-based testing") {
+    validateHostName("_underscore.domain.name.").isValid
```
mitruly: If we're allowing trailing underscores, we should test that as well.

PR #783, opened 2019-08-06
Updating play to 2.7

pauljamescleary:

### DIFF
--- a/modules/portal/app/controllers/VinylDNS.scala
+++ b/modules/portal/app/controllers/VinylDNS.scala
@@ -17,22 +17,22 @@
 package controllers
 
 import java.util
+import java.util.HashMap
 
 import actions.{ApiAction, FrontendAction}
-import controllers.OidcAuthenticator.ErrorResponse
+import cats.data.EitherT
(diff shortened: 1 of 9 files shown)


modules/portal/app/controllers/VinylDNS.scala
```diff
@@ -102,9 +102,10 @@ class VinylDNS @Inject()(
     extends AbstractController(components)
     with CacheHeader {
 
-  import play.api.mvc._
   import VinylDNS._
+  import play.api.mvc._
 
+  private val logger = LoggerFactory.getLogger(classOf[VinylDNS])
```
rebstar6: we dont need this LoggerFactory stuff anymore, i think you can just do 
`val logger: Logger = Logger(...)`

pauljamescleary: I saw this, but also read here https://www.playframework.com/documentation/2.7.x/Migration27 where they mention you can use either or.

In most other places (including the portal) we use slf4j directly.  I figured it was better to be consistent across all modules.

PR #784, opened 2019-08-06
validate scheduled time is in the future

britneywright:
For scheduled changes we don't want users setting a time in the past.

Changes in this pull request:
- if scheduled changes are enabled check that the scheduled time is after the current time

I went the current time, but we could easily change it to an hour from the current time or some other set time from the current time.

### DIFF
--- a/modules/api/functional_test/live_tests/batch/create_batch_change_test.py
+++ b/modules/api/functional_test/live_tests/batch/create_batch_change_test.py
@@ -298,7 +298,7 @@ def test_create_scheduled_batch_change_with_zone_discovery_error_without_owner_g
     Test creating a scheduled batch without owner group ID fails if there is a zone discovery error
     """
     client = shared_zone_test_context.ok_vinyldns_client
-    dt = datetime.datetime.now().strftime('%Y-%m-%dT%H:%M:%SZ')
+    dt = (datetime.datetime.now() + datetime.timedelta(1)).strftime('%Y-%m-%dT%H:%M:%SZ')
 
     batch_change_input = {
         "comments": "this is optional",
@@ -311,6 +311,28 @@ def test_create_scheduled_batch_change_with_zone_discovery_error_without_owner_g
     errors = client.create_batch_change(batch_change_input, status=400)
 
     assert_that(errors, is_("Batch change requires owner group for manual review/scheduled changes."))
+
+
+@pytest.mark.manual_batch_review
+def test_create_scheduled_batch_change_with_scheduled_time_in_the_past_fails(shared_zone_test_context):
+    """
+    Test creating a scheduled batch with a scheduled time in the past
+    """
+    client = shared_zone_test_context.ok_vinyldns_client
+    yesterday = (datetime.datetime.now() - datetime.timedelta(1)).strftime('%Y-%m-%dT%H:%M:%SZ')
+
+    batch_change_input = {
+        "comments": "this is optional",
+        "changes": [
+            get_change_A_AAAA_json("no-owner-group.scheduled.parent.com.", address="4.5.6.7"),
+        ],
+        "ownerGroupId": shared_zone_test_context.ok_group['id'],
+        "scheduledTime": yesterday
+    }
+
+    errors = client.create_batch_change(batch_change_input, status=400)
+
+    assert_that(errors, is_("Scheduled time must be in the future."))
 
 
 def test_create_batch_change_without_scheduled_time_succeeds(shared_zone_test_context):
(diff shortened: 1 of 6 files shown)


modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
```diff
@@ -107,6 +107,12 @@ class BatchChangeValidations(
         else NotAMemberOfOwnerGroup(groupId, authPrincipal.signedInUser.userName).invalidNel
     }
 
+//  def validateScheduledTime(scheduledTime: Option[DateTime]): SingleValidation[Unit] =
```
mitruly: Commented code should be removed.

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
```diff
@@ -507,6 +513,11 @@ class BatchChangeValidations(
   def validateScheduledChange(
       input: BatchChangeInput,
       scheduledChangesEnabled: Boolean): Either[BatchChangeErrorResponse, Unit] =
-    if (scheduledChangesEnabled || input.scheduledTime.isEmpty) Right(())
-    else Left(ScheduledChangesDisabled)
+    (scheduledChangesEnabled, input.scheduledTime.isEmpty) match {
```
mitruly: This is more concise:
```
(scheduledChangesEnabled, input.scheduledTime) match {
      case (_, None) => Right(())
      case (true, Some(inputTime)) if inputTime.isAfterNow => Right(())
      case (true, _) => Left(ScheduledTimeMustBeInFuture)
      case (false, _) => Left(ScheduledChangesDisabled)
    }
```

modules/core/src/main/scala/vinyldns/core/domain/DomainValidationErrors.scala
```diff
@@ -35,6 +35,7 @@ final case class BatchChangeIsEmpty(limit: Int) extends DomainValidationError {
 final case class GroupDoesNotExist(id: String) extends DomainValidationError {
   def message: String = s"""Group with ID "$id" was not found"""
 }
+
```
mitruly: 💥

britneywright: that's valid, not sure when or why but the first few classes don't have spaces between them, the rest do. I'll update the file to have spaces between all of them.

mitruly:
There's one existing functional test that is failing (`test_create_batch_change_with_scheduled_time_and_owner_group_succeeds`) since it is using `datetime.datetime.now()`; it needs to be padded for a future time.

modules/api/functional_test/live_tests/batch/create_batch_change_test.py
```diff
@@ -276,7 +276,7 @@ def test_create_batch_change_with_scheduled_time_and_owner_group_succeeds(shared
     Test successfully creating a batch change with scheduled time and owner group set
     """
     client = shared_zone_test_context.ok_vinyldns_client
-    dt = datetime.datetime.now().strftime('%Y-%m-%dT%H:%M:%SZ')
+    dt = (datetime.datetime.now() + datetime.timedelta(1)).strftime('%Y-%m-%dT%H:%M:%SZ')
```
rebstar6: is this one ms? risky.

rebstar6: maybe not, can you just specify so its clear? `timedelta(hours=1)` or whatever

modules/api/functional_test/live_tests/batch/create_batch_change_test.py
```diff
@@ -313,6 +313,28 @@ def test_create_scheduled_batch_change_with_zone_discovery_error_without_owner_g
     assert_that(errors, is_("Batch change requires owner group for manual review/scheduled changes."))
 
 
+@pytest.mark.manual_batch_review
+def test_create_scheduled_batch_change_with_scheduled_time_in_the_past_fails(shared_zone_test_context):
+    """
+    Test creating a scheduled batch with a scheduled time in the past
+    """
+    client = shared_zone_test_context.ok_vinyldns_client
+    yesterday = (datetime.datetime.now() - datetime.timedelta(1)).strftime('%Y-%m-%dT%H:%M:%SZ')
```
rebstar6: same timedelta thing here on all of these

modules/api/functional_test/live_tests/batch/create_batch_change_test.py
britneywright: it's days, that's the default first parameter of timedelta

PR #785, opened 2019-08-06
show create recordset button based on ACL rules

britneywright:
Users with a Write  or Delete ACL rule aren't seeing the Create Recordset button in the portal.

Changes in this pull request:
- If a user has any Write or Delete ACL rule, display the Create Recordset button 
- This is in addition to the current rules, if a user has "Delete" level access to the zone overall (i.e. zone admins and super admins) or the zone is shared.


modules/portal/public/lib/controllers/controller.records.js
```diff
@@ -362,8 +363,9 @@ angular.module('controller.records', [])
     }
 
     function determineAdmin(){
-        $scope.isZoneAdmin = $scope.profile.isSuper || isInAdminGroup()
-        $scope.canReadZone = canReadZone()
+        $scope.isZoneAdmin = $scope.profile.isSuper || isInAdminGroup();
+        $scope.canReadZone = canReadZone();
+        $scope.canCreateRecords = $scope.zoneInfo.accessLevel == 'Delete' || $scope.zoneInfo.shared || canCreateRecords();
```
mitruly: Shouldn't the `$scope.zoneInfo.accessLevel` check also include `'Write'`?

modules/portal/public/lib/controllers/controller.records.js
```diff
@@ -384,6 +386,10 @@ angular.module('controller.records', [])
         return $scope.myGroupIds.indexOf(groupId) > -1;
     };
 
+    function canCreateRecords() {
```
mitruly: I would be inclined to have all of these checks occur in a single location (ie. inside `determineAdmin`) rather than a separate location, unless we rename this to something more specific like `canCreateRecordsViaAcl`.

modules/portal/public/lib/controllers/controller.records.js
britneywright: no, we only set the overall Zone access level to Delete, Read or NoAccess

https://github.com/vinyldns/vinyldns/blob/master/modules/api/src/main/scala/vinyldns/api/domain/AccessValidations.scala#L115-L120

mitruly: Gotcha - yeah, that's fairly interesting since we're translating zone access into record management access.

modules/portal/public/lib/controllers/controller.records.js
```diff
@@ -384,6 +391,7 @@ angular.module('controller.records', [])
         return $scope.myGroupIds.indexOf(groupId) > -1;
     };
 
+
```
mitruly: Can we remove this new line?

PR #786, opened 2019-08-07
rename SingleDeleteChange to SingleDeleteRRSetChange 

rebstar6:
in preparation for SingleDeleteRecordChange

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeConverter.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeConverter.scala
@@ -157,7 +157,7 @@ class BatchChangeConverter(batchChangeRepo: BatchChangeRepository, messageQueue:
     }
     val deletes = NonEmptyList.fromList {
       changes.collect {
-        case del: SingleDeleteChange if SupportedBatchChangeRecordTypes.get.contains(del.typ) => del
+        case del: SingleDeleteSetChange if SupportedBatchChangeRecordTypes.get.contains(del.typ) => del
       }
     }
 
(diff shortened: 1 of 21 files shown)


modules/api/src/main/scala/vinyldns/api/route/BatchChangeJsonProtocol.scala
```diff
@@ -132,8 +132,8 @@ trait BatchChangeJsonProtocol extends JsonValidation {
   }
 
   // recordName, zoneName, zoneId used to be required; getOrElse to maintain backwards compatability with clients
-  case object SingleDeleteChangeSerializer extends ValidationSerializer[SingleDeleteChange] {
-    override def toJson(sac: SingleDeleteChange): JValue =
+  case object SingleDeleteSetChangeSerializer extends ValidationSerializer[SingleDeleteSetChange] {
+    override def toJson(sac: SingleDeleteSetChange): JValue =
```
rebstar6: note: per this serializer, these names should only be in the code. we turn this into json with `changeType: Add` or `changeType: DeleteRecordSet ` when responding to users, so this change is just so the code isnt aggressively confusing

rebstar6:
may reopen, thinking through design

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeConverter.scala
```diff
@@ -157,7 +157,7 @@ class BatchChangeConverter(batchChangeRepo: BatchChangeRepository, messageQueue:
     }
     val deletes = NonEmptyList.fromList {
       changes.collect {
-        case del: SingleDeleteChange if SupportedBatchChangeRecordTypes.get.contains(del.typ) => del
+        case del: SingleDeleteSetChange if SupportedBatchChangeRecordTypes.get.contains(del.typ) => del
```
pauljamescleary: Can't say I love `SingleDeleteSetChange`, perhaps `SingleDeleteRecordSetChange` or if that is too long `SingleDeleteRRSetChange` <-- is pretty industry standard

PR #787, opened 2019-08-07
Fix docker releases

pauljamescleary:
There was an issue starting the docker containers due to how native packager
works where we were seeing issues with the container being able to start.

The issue was that we were assuming a "daemon" user to run the containers under.
At some point this changed to "1001:0".  As a result, there were not sufficient
privileges to start the containers because the "daemon" user was invalid or did
not have access to the scripts created by sbt native packager.

* `build.sbt` - update the user to "1001:0" for our custom install.  Cleaned up the hardcoded references in the script extras to `/opt/docker` to use the variable `app_home` instead.
* `plugins.sbt` - updated to the latest sbt native packager

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -166,7 +166,7 @@ lazy val portalDockerSettings = Seq(
   dockerCommands ++= Seq(
     Cmd("USER", "root"), // switch to root so we can install netcat
     ExecCmd("RUN", "apk", "add", "--update", "--no-cache", "netcat-openbsd", "bash"),
-    Cmd("USER", "1001:0") // switch back to the daemon user
+    Cmd("USER", "1001:0") // switch back to the user that runs the process
   ),
 
   credentials in Docker := Seq(Credentials(Path.userHome / ".ivy2" / ".dockerCredentials"))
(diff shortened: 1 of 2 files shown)


pauljamescleary:
To test:

1. Update your `docker/.env` file to change the VINYLDNS_VERSION to `0.9.3-SNAPSHOT`
2. Start up sbt (or reload)
3. `project api; docker:publishLocal`
4. `project portal; docker:publishLocal`
5. Then in a separate terminal, `./bin/docker-up-vinyldns.sh`

Make sure you can login to the portal and create a group, certifying everything works as expected.

build.sbt
```diff
@@ -134,7 +134,7 @@ lazy val apiDockerSettings = Seq(
   dockerCommands ++= Seq(
     Cmd("USER", "root"), // switch to root so we can install netcat
     ExecCmd("RUN", "apk", "add", "--update", "--no-cache", "netcat-openbsd", "bash"),
-    Cmd("USER", "daemon") // switch back to the daemon user
+    Cmd("USER", "1001:0") // switch back to the daemon user
```
mitruly: Nitpick: Might want to remove `daemon` from here.

pauljamescleary: done

PR #790, opened 2019-08-07
add review confirmation to DNS Change form

britneywright:
Trying this out. We may or may not decide to go with it.

Changes in this pull request:
- If manual review is enabled and the user submits a batch change with only soft errors we'll ask them to confirm they want to proceed to manual review. 
- change the alert message and change background colors to the warning yellow instead of danger red.
- Note: This first pass only includes the Zone Discovery Failed error message.

DNS-Change-Review-Confirmation

What else can we do in the portal to mitigate questions users might have about the review process?

### DIFF
--- a/modules/portal/public/lib/batch-change/batch-change-new.controller.js
+++ b/modules/portal/public/lib/batch-change/batch-change-new.controller.js
@@ -37,6 +37,9 @@
             $scope.ownerGroupError = false;
             $scope.formStatus = "pendingSubmit";
             $scope.scheduledOption = false;
+            $scope.allowManualReview = false;
+            $scope.confirmationPrompt = "Are you sure you want to submit this batch change request?";
+            $scope.manualReviewEnabled;
 
             $scope.addSingleChange = function() {
                 $scope.newBatch.changes.push({changeType: "Add", type: "A+PTR"});
@@ -93,7 +96,7 @@
 
                 formatData(payload);
 
-                return batchChangeService.createBatchChange(payload)
+                return batchChangeService.createBatchChange(payload, $scope.allowManualReview)
                     .then(success)
                     .catch(function (error){
                         if(error.data.errors || error.status !== 400 || typeof error.data == "string"){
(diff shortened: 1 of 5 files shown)


pauljamescleary:
I think this is a good idea, if there are issues, notifying the user that the thing will go to manual review (i.e. won't be implemented) is good.

I think the Confirm is consistent with other locations in the app, so good there.

Perhaps the messaging should be "cannot process immediately; change requires manual review" something along those lines.  We want to communicate to the user the _impact_ of what is happening (i.e. it is not going to be processed right now).

Also, how does this interplay with scheduled?  Messaging can be tricky there.

modules/portal/public/lib/batch-change/batch-change-new.controller.js
```diff
@@ -93,18 +96,26 @@
...
-                                .some(e => e.includes('owner group ID must be specified for record'));
-                            $scope.formStatus = "pendingSubmit";
-                            $scope.alerts.push({type: 'danger', content: 'Errors found. Please correct and submit again.'});
+                            $scope.listOfErrors = error.data.flatMap(d => d.errors)
+                            $scope.ownerGroupError = $scope.listOfErrors.some(e => e.includes('owner group ID must be specified for record'));
+                            var hardErrors = $scope.listOfErrors.every(e => ['Zone Discovery Failed'].includes(e));
+                            if ($scope.manualReviewEnabled && !hardErrors) {
+                                $scope.allowManualReview = true;
+                                $scope.formStatus = "pendingConfirm";
+                                $scope.alerts.push({type: 'warning', content: 'Issues found. Please correct or submit for review.'});
```
mitruly: Maybe something like `Issues found requiring manual review. Please correct or confirm submission for review.` is clearer.

modules/portal/public/lib/batch-change/batch-change-new.controller.js
```diff
@@ -93,18 +96,26 @@
...
                         } else {
                             $scope.newBatch.changes = error.data;
                             $scope.batchChangeErrors = true;
-                            $scope.ownerGroupError = error.data.flatMap(d => d.errors)
-                                .some(e => e.includes('owner group ID must be specified for record'));
-                            $scope.formStatus = "pendingSubmit";
-                            $scope.alerts.push({type: 'danger', content: 'Errors found. Please correct and submit again.'});
+                            $scope.listOfErrors = error.data.flatMap(d => d.errors)
+                            $scope.ownerGroupError = $scope.listOfErrors.some(e => e.includes('owner group ID must be specified for record'));
+                            var hardErrors = $scope.listOfErrors.every(e => ['Zone Discovery Failed'].includes(e));
```
mitruly: Also need to check for `RecordRequiresManualReview` now, since that flags an error for manual review.

britneywright:
> Also, how does this interplay with scheduled? Messaging can be tricky there.
If there is a scheduled date and no reviewable errors the batch change is accepted immediately on the initial submission.

If there are reviewable errors, regardless of a scheduled date, the user will get the prompt to confirm they want to go to manual review after their initial submission attempt.

britneywright:
> If there are reviewable errors, regardless of a scheduled date, the user will get the prompt to confirm they want to go to manual review after their initial submission attempt.
 
I got this part wrong. Based on the API if there's a scheduled change we immediately go to pending review. We look at scheduled or the allow manual review flag is true on the batch change. We should probably change this though so if it's scheduled and there are soft errors the user can still review those before sending to review.

modules/portal/public/lib/batch-change/batch-change-new.controller.js
britneywright: Updated to "Issues found that require manual review. Please correct or confirm submission for review."

modules/portal/public/lib/batch-change/batch-change-new.controller.js
britneywright: added

PR #791, opened 2019-08-07
support DeleteRecord in New DNS Change form

britneywright:
Screen Shot 2019-08-07 at 2 39 26 PM

Changes in this pull request:
- add DeleteRecord change type in New DNS Change form, requires record data
- update CSV import function to accept DeleteRecord change type

PR #792, opened 2019-08-07
add DeleteRecord info to the docs

britneywright:
Fixes #.

Changes in this pull request:
- add DeleteRecord info to the Batch Change model docs
- add DeleteRecord examples to the DNS Change CSV sample file

### DIFF
--- a/modules/docs/src/main/resources/microsite/static/dns-changes-csv-sample.csv
+++ b/modules/docs/src/main/resources/microsite/static/dns-changes-csv-sample.csv
@@ -12,3 +12,10 @@ DeleteRecordSet,AAAA+PTR,test7.example.com.,,fd69:27cc:fe91::60
 DeleteRecordSet,CNAME,test8.example.com.,,
 DeleteRecordSet,PTR,193.0.2.192,,
 DeleteRecordSet,TXT,test5.example.com,,
+DeleteRecord,A+PTR,test.example.,,1.1.1.1
+DeleteRecord,A,test1.example.,,1.2.3.4
+DeleteRecord,AAAA+PTR,test2.example.,,fd69:27cc:fe91::60
+DeleteRecord,AAAA,test3.example.com.,,fd69:27cc:fe91::60
+DeleteRecord,CNAME,test4.example.com.,,test0.example.com.
+DeleteRecord,PTR,192.0.2.193,,test.example.com.
+DeleteRecord,TXT,test5.example.com,,example text
--- a/modules/docs/src/main/resources/microsite/static/dns-changes-csv-sample.csv
+++ b/modules/docs/src/main/resources/microsite/static/dns-changes-csv-sample.csv
@@ -6,9 +6,7 @@ Add,AAAA,test3.example.com.,,fd69:27cc:fe91::60
 Add,CNAME,test4.example.com.,,test0.example.com.
 Add,PTR,192.0.2.193,200,test.example.com.
 Add,TXT,test5.example.com,7200,example text
-DeleteRecordSet,A+PTR,test5.example.com.,,1.1.1.1
 DeleteRecordSet,A,test6.exmaple.com.,,
-DeleteRecordSet,AAAA+PTR,test7.example.com.,,fd69:27cc:fe91::60
 DeleteRecordSet,CNAME,test8.example.com.,,
 DeleteRecordSet,PTR,193.0.2.192,,
 DeleteRecordSet,TXT,test5.example.com,,
--- a/modules/docs/src/main/tut/api/batchchange-model.md
+++ b/modules/docs/src/main/tut/api/batchchange-model.md
@@ -101,11 +101,11 @@ validationErrors | Array of BatchChangeError | Array containing any validation e
 id            | string        | The unique identifier for this change. |
 
 #### ChangeType Values <a id="changetype-values" />
-There are three valid changeTypes for a SingleChange, **Add**, **DeleteRecordSet** and **Delete**.
+There are three valid changeTypes for a SingleChange, **Add**, **DeleteRecordSet** and **DeleteRecord**.
 Used alone, or in conjunction with one another, they determine if you are creating, removing or modifying DNS records. Below are the change types to use based on intended outcome.
 
-- Create a new DNS record: **Add** changeType,
-- Delete an entire recordset: **DeleteRecordSet** changeType,
+- Create a new DNS record: **Add** changeType
+- Delete an entire recordset: **DeleteRecordSet** changeType
 - Delete a single record in a multi-record recordset: **DeleteRecord** changeType. Record types that only support one record data entry (i.e. CNAME records) will treat a DeleteRecord changeType as DeleteRecordSet.
 - Update an existing recordset: at least one entry with a **DeleteRecord** or **DeleteRecordSet** changeType and at least one entry with a **Add** changeType.
 
--- a/modules/docs/src/main/tut/api/batchchange-model.md
+++ b/modules/docs/src/main/tut/api/batchchange-model.md
@@ -68,7 +68,7 @@ A successful batch change response consists of a corresponding [SingleAddChange]
 
 name          | type          | description |
  ------------ | :------------ | :---------- |
-changeType    | ChangeInputType | Type of change input. Can be an **Add**, **DeleteRecordSet**, or **DeleteRecord**. |
+changeType    | ChangeInputType | Type of change input. Can be an **Add**, **DeleteRecordSet**, or **DeleteRecord**. [See full details](#changetype-values).|
 inputName     | string        | The fully-qualified domain name of the record which was provided in the create batch request. |
 type          | RecordType    | Type of DNS record, supported records for batch changes are currently: **A**, **AAAA**, **CNAME**, and **PTR**. |
 ttl           | long          | The time-to-live in seconds. |
@@ -87,7 +87,7 @@ id            | string        | The unique identifier for this change. |
 
 name          | type          | description |
  ------------ | :------------ | :---------- |
-changeType    | ChangeInputType | Type of change input. Can be an **Add**, **DeleteRecordSet**, or **DeleteRecord**. |
+changeType    | ChangeInputType | Type of change input. Can be an **Add**, **DeleteRecordSet**, or **DeleteRecord**. [See full details](#changetype-values).|
 inputName     | string        | The fully-qualified domain name of the record which was provided in the create batch request. |
 type          | RecordType    | Type of DNS record, supported records for batch changes are currently: **A**, **AAAA**, **CNAME**, and **PTR**. | 
 status        | SingleChangeStatus | Status for this change. Can be one of: **Pending**, **Complete**, **Failed**, **NeedsReview** or **Rejected**. |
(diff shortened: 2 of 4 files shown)


modules/docs/src/main/tut/api/batchchange-model.md
```diff
@@ -64,7 +64,7 @@ A successful batch change response consists of a corresponding [SingleAddChange]
 
 name          | type          | description |
  ------------ | :------------ | :---------- |
-changeType    | ChangeInputType | Type of change input. Can either be an **Add** or **DeleteRecordSet**. |
+changeType    | ChangeInputType | Type of change input. Can be an **Add**, **DeleteRecordSet**, or **DeleteRecord**. |
```
mitruly: Need to update `create-batchchange.md` where we should also provide more information on `DeleteRecord` vs `DeleteRecordSet`.

modules/docs/src/main/resources/microsite/static/dns-changes-csv-sample.csv
```diff
@@ -12,3 +12,10 @@ DeleteRecordSet,AAAA+PTR,test7.example.com.,,fd69:27cc:fe91::60
 DeleteRecordSet,CNAME,test8.example.com.,,
 DeleteRecordSet,PTR,193.0.2.192,,
 DeleteRecordSet,TXT,test5.example.com,,
+DeleteRecord,A+PTR,test.example.,,1.1.1.1
```
mitruly: For the existing `DeleteRecordSet`,`A+PTR`/`AAAA+PTR` entries, we need to split them out as separate `DeleteRecordSet` calls (individually for `A`, `AAAA` and `PTR`) per https://github.com/vinyldns/vinyldns/pull/791#pullrequestreview-293942390.

modules/docs/src/main/tut/api/batchchange-model.md
mitruly: We need to elaborate on the behavior between different permutations of `Add`, `DeleteRecord` and `DeleteRecordSet`.

Example:
Given existing an `A` record `foo.bar.` with addresses `1.1.1.1` and `2.2.2.2`

| Add(s)   | DeleteRecord(s)  | DeleteRecordSet | Permission level | Allowed | Notes |
| ------ | ---------------- | ----------------- | ---------------- | -------- | ------- |
| 3.3.3.3 |  |  No | Add | No | Must issue an update (with `DeleteRecord` or `DeleteRecordSet`) |
| 3.3.3.3 | 1.1.1.1 | No | Update | Yes | |
| 3.3.3.3 | 1.1.1.1, 2.2.2.2 | No | Update | Yes | |
| 3.3.3.3 | 1.1.1.1, 2.2.2.2 | Yes | Update | Yes |`DeleteRecord` can co-exist in a request with `DeleteRecordSet`; net result is a `DeleteRecordSet` (assuming valid record data for `DeleteRecord`) |
| 3.3.3.3, 4.4.4.4 | 1.1.1.1 | Yes | Update | Yes | |
| | 1.1.1.1 | No | Update | Yes | A single `DeleteRecord` for a DNS record containing multiple entries is an update |
| | 1.1.1.1, 2.2.2.2 | No | Delete | Yes | |
| | 1.1.1.1 | Yes | Delete | Yes | |
| | 3.3.3.3 | No | N/A | No | `DeleteRecord` data must be valid |

modules/docs/src/main/tut/api/create-batchchange.md
```diff
@@ -42,10 +42,12 @@ record        | [RecordData](../api/recordset-model#record-data) | yes         |
 
 name          | type          | required?   | description |
  ------------ | :------------ | ----------- | :---------- |
-changeType    | ChangeInputType | yes       | Type of change input. Must be **DeleteRecordSet** for *DeleteChangeInput*. |
+changeType    | ChangeInputType | yes       | Type of change input. Must be **DeleteRecord** or **DeleteRecordSet** for *DeleteChangeInput*. **DeleteRecord** deletes one record in a recordset. **DeleteRecordSet** deletes the entire recordset. |
 inputName     | string        | yes         | The fully qualified domain name of the record being deleted. |
 type          | RecordType    | yes         | Type of DNS record. Supported records for batch changes are currently: **A**, **AAAA**, **CNAME**, and **PTR**. |
+record        | [RecordData](../api/recordset-model#record-data) | sometimes         | One existing record data entry in the recordset. Required if the changeType is DeleteRecord. |
 
+Note: record types that only support one record data entry (i.e. CNAME records) will treat a DeleteRecord changeType as DeleteRecordSet
```
mitruly: For consistency, we should bolden occurences `DeleteRecord` and `DeleteRecordSet`.

modules/docs/src/main/resources/microsite/static/dns-changes-csv-sample.csv
britneywright: updated

modules/docs/src/main/tut/api/batchchange-model.md
britneywright: > Need to update `create-batchchange.md` where we should also provide more information on `DeleteRecord` vs `DeleteRecordSet`.

This is an interesting point. Somewhat related, It's been brought up before that we're probably over duplicating the batch change information. We should revisit that to discuss how to reduce duplication and also how to handle the attributes that require lengthier descriptions. Maybe we should make sections for attributes requiring more in depth descriptions rather than long table entries or footnotes. And other pages can link back to the main model page with all the details.

mitruly: @britneywright while I do agree about overhauling these docs overall, I do think that there are important aspects that we _need_ to mention _somewhere_ with these changes:
- We don't allow straight adds for existing record sets in any capacity
- `DeleteRecord` _must_ be for valid (existing) data
- `DeleteRecord` can be either update or delete permissions, based on the scenario

britneywright: Yes my point is what would be the most effective placement of the information because it is important.

mitruly: Either having an "information" page and explaining how batch change composition affects behavior or having sub-sections in the attributes section of the model page should suffice.

mitruly:
We also need to:
- Update the `SingleChange` attributes to include `DeleteRecord` in `changeType`
- Clean up the errors page: `ExistingMultiRecordError` and `NewMultiRecordError` are deprecated and `RecordAlreadyExists` needs to have an updated message.

britneywright:
> We also need to:
> 
> * Update the `SingleChange` attributes to include `DeleteRecord` in `changeType`

SingleChange is updated in the batchchange-model file. let me know if i missed it somewhere else.

modules/docs/src/main/tut/api/batchchange-model.md
```diff
@@ -100,6 +100,17 @@ recordSetId   | string        | Conditional: The unique identifier for the recor
...
 id            | string        | The unique identifier for this change. |
 
+#### ChangeType Values <a id="changetype-values" />
+There are three valid changeTypes for a SingleChange. These values determine whether you are adding, removing or updating DNS records. 
+
+- **Add**: To add a single DNS record.
+- **DeleteRecordSet**: To delete an entire recordset. This applies to record types that allow multi-record recordsets.
+- **DeleteRecord**: To delete one record in a recordset (i.e. A records). Record types that only support one record data entry (i.e. CNAME records) will treat a DeleteRecord changeType as DeleteRecordSet.
+
+To add a new record to an existing multi-record recordset you must first delete a single record in the recordset (DeleteRecord) and then add the new record data (Add) in the same batch change.
```
mitruly: It would be clearer to break these sentences down into "actions". Note that an update can be a `DeleteRecord` or `DeleteRecordSet`.

modules/docs/src/main/tut/api/batchchange-model.md
```diff
@@ -100,6 +100,17 @@ recordSetId   | string        | Conditional: The unique identifier for the recor
...
+
+To update an existing recordset that only supports one record, you must first delete the recordset (DeleteRecordSet) and then add the new record data (Add) in the same batch change.
```
mitruly: This can also be `DeleteRecord` with the record data. Not sure if this specific case is worth mentioning here, since it applies to `CNAME` support in general, and not specifically to batch. If we want to keep it, we should indicate the record types.

modules/docs/src/main/tut/api/batchchange-model.md
```diff
@@ -100,6 +100,17 @@ recordSetId   | string        | Conditional: The unique identifier for the recor
 validationErrors | Array of BatchChangeError | Array containing any validation errors associated with this SingleDeleteRRSetChange. *Note: These will only exist on `NeedsReview` or `Rejected` SingleChanges* |
 id            | string        | The unique identifier for this change. |
 
+#### ChangeType Values <a id="changetype-values" />
```
britneywright: broke the change type details out of the attribute table and into their own section. Vetted the text with a current user. They're excited.

modules/docs/src/main/tut/api/batchchange-model.md
britneywright: It's already mentioned earlier in the section that `Record types that only support one record data entry (i.e. CNAME records) will treat a DeleteRecord changeType as DeleteRecordSet.`

mitruly: Right, so breaking this down into actions and which actions aren't allowed makes more sense. Having information in one place talking about `DeleteRecord` and another place referring to `DeleteRecordSet` seems confusing to me.

Example:
- Add is only one or more `Add`s
- Update is one or more `Add`s with one or more `DeleteRecord`s/`DeleteRecordSet`
- Delete is `DeleteRecordSet` or `DeleteRecord` for all of the record data

I would get rid of record-specific stuff altogether since they are covered by the individual validations and just mention that adding to existing records with multiple entries is not supported. Record types supporting only one record data entry requiring an update is already implied.

modules/docs/src/main/tut/api/batchchange-model.md
```diff
@@ -100,6 +100,14 @@ recordSetId   | string        | Conditional: The unique identifier for the recor
...
```
britneywright: rejiggered this section.

modules/docs/src/main/tut/api/batchchange-model.md
```diff
@@ -101,16 +101,13 @@ validationErrors | Array of BatchChangeError | Array containing any validation e
 id            | string        | The unique identifier for this change. |
 
 #### ChangeType Values <a id="changetype-values" />
-There are three valid changeTypes for a SingleChange. These values determine whether you are adding, removing or updating DNS records. 
+There are three valid changeTypes for a SingleChange, **Add**, **DeleteRecordSet** and **Delete**.
```
mitruly: Last one should be `DeleteRecord`.

modules/docs/src/main/tut/api/batchchange-model.md
```diff
@@ -101,16 +101,13 @@ validationErrors | Array of BatchChangeError | Array containing any validation e
...
 
-- **Add**: To add a single DNS record.
-- **DeleteRecordSet**: To delete an entire recordset. This applies to record types that allow multi-record recordsets.
-- **DeleteRecord**: To delete one record in a recordset (i.e. A records). Record types that only support one record data entry (i.e. CNAME records) will treat a DeleteRecord changeType as DeleteRecordSet.
-
-To add a new record to an existing multi-record recordset you must first delete a single record in the recordset (DeleteRecord) and then add the new record data (Add) in the same batch change.
-
-To update an existing recordset that only supports one record, you must first delete the recordset (DeleteRecordSet) and then add the new record data (Add) in the same batch change.
-You can also use DeleteRecordSet to clear all the record data entries of a multi-record recordset followed by one or more Adds with new record data.
+- Create a new DNS record: **Add** changeType,
```
mitruly: Are these trailing commas for style?

modules/docs/src/main/tut/api/batchchange-model.md
```diff
@@ -101,16 +101,13 @@ validationErrors | Array of BatchChangeError | Array containing any validation e
...
+- Delete an entire recordset: **DeleteRecordSet** changeType,
```
mitruly: I think this suffices. A user can _technically_ issue multiple `DeleteRecord`s for valid data, but it's better to steer users to `DeleteRecordSet` if that's what they want to do.

britneywright:
updated

PR #794, opened 2019-08-08
API: cancel batch change that is pending review

britneywright:
Changes in this pull request:
- Give creators of a batch change the ability to cancel the change if it's approval status is Pending Review

### DIFF
--- a/modules/api/functional_test/live_tests/batch/cancel_batch_change_test.py
+++ b/modules/api/functional_test/live_tests/batch/cancel_batch_change_test.py
@@ -28,6 +28,10 @@ def test_cancel_batch_change_success(shared_zone_test_context):
     assert_that(get_batch['approvalStatus'], is_('Cancelled'))
     assert_that(get_batch['changes'][0]['status'], is_('Cancelled'))
     assert_that(get_batch, has_key('cancelledTimestamp'))
+    assert_that(get_batch, not(has_key('reviewTimestamp')))
+    assert_that(get_batch, not(has_key('reviewerId')))
+    assert_that(get_batch, not(has_key('reviewerUserName')))
+    assert_that(get_batch, not(has_key('reviewComment')))
 
 @pytest.mark.manual_batch_review
 def test_cancel_batch_change_fails_for_non_creator(shared_zone_test_context):
--- a/modules/api/functional_test/live_tests/batch/cancel_batch_change_test.py
+++ b/modules/api/functional_test/live_tests/batch/cancel_batch_change_test.py
@@ -27,6 +27,7 @@ def test_cancel_batch_change_success(shared_zone_test_context):
     assert_that(get_batch['status'], is_('Cancelled'))
     assert_that(get_batch['approvalStatus'], is_('Cancelled'))
     assert_that(get_batch['changes'][0]['status'], is_('Cancelled'))
+    assert_that(get_batch, has_key('cancelledTimestamp'))
 
 @pytest.mark.manual_batch_review
 def test_cancel_batch_change_fails_for_non_creator(shared_zone_test_context):
--- a/modules/api/functional_test/live_tests/batch/cancel_batch_change_test.py
+++ b/modules/api/functional_test/live_tests/batch/cancel_batch_change_test.py
@@ -1,0 +1,75 @@
+from hamcrest import *
+from utils import *
+
+
+@pytest.mark.manual_batch_review
+def test_cancel_batch_change_success(shared_zone_test_context):
+    """
+    Test cancelling a batch change succeeds for the user who made the batch change
+    """
+    client = shared_zone_test_context.ok_vinyldns_client
+    batch_change_input = {
+        "changes": [
+            get_change_A_AAAA_json("zone.discovery.failure.", address="4.3.2.1")
+        ],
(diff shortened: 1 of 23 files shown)


modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
```diff
@@ -143,6 +154,14 @@ class BatchChangeValidations(
       case _ => Right(())
     }
 
+  def validateCreatorCancellation(
+      batchChange: BatchChange,
+      auth: AuthPrincipal): Either[BatchChangeErrorResponse, Unit] =
+    if (batchChange.userId == auth.userId) {
```
pauljamescleary: Should we allow super/support users to cancel?  I know they can reject which is effectively the same, so thinking out loud a little.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlBatchChangeRepository.scala
```diff
@@ -90,7 +90,8 @@ class MySqlBatchChangeRepository
          |              bc.approval_status, bc.reviewer_id, bc.review_comment, bc.review_timestamp, bc.scheduled_time,
          |              SUM( case when sc.status like 'Failed' or sc.status like 'Rejected' then 1 else 0 end ) AS fail_count,
          |              SUM( case when sc.status like 'Pending' or sc.status like 'NeedsReview' then 1 else 0 end ) AS pending_count,
-         |              SUM( case sc.status when 'Complete' then 1 else 0 end ) AS complete_count
+         |              SUM( case sc.status when 'Complete' then 1 else 0 end ) AS complete_count,
+         |              SUM( case sc.status when 'Cancelled' then 1 else 0 end ) AS cancelled_count
```
mitruly: Are we opting for `cancelled` over `canceled`?

mitruly:
Pretty good with this PR.

@pauljamescleary I had thought about the same thing before; it's more of a design approach and I think it makes sense to only allow the submitting user to cancel the changes.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlBatchChangeRepository.scala
britneywright: cancelled 🙌🏽

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
britneywright: I think only a user can cancel their own request. Super and Support users would reject. No need to mix the two.

rebstar6: +1, they can reject

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -551,4 +561,16 @@ class BatchChangeService(
 
     batchChangeRepo.save(rejectedBatch).toBatchResult
   }
+
+  def cancelBatchChange(batchChange: BatchChange): BatchResult[BatchChange] = {
+    val cancelledSingleChanges = batchChange.changes.map(_.cancel)
+
+    // Update rejection attributes and single changes for batch change
+    val cancelledBatch = batchChange.copy(
```
rebstar6: do we care about any other metadata? must be this user, so dont need the user info, but what about cancelled time?

britneywright: Thought about cancelled time, I could go either way on it.

pauljamescleary: Good point, I think we should save a time when it was cancelled.

rebstar6: yeah, in general we cancel would only happen pretty immediately, but cant hurt to have. was wondering if we could leverage reviewTimestamp given that something cant be cancelled and any kind of other review, though the wording doesnt quite work

modules/api/functional_test/live_tests/batch/cancel_batch_change_test.py
```diff
@@ -0,0 +1,76 @@
...
+    assert_that(get_batch['changes'][0]['status'], is_('NeedsReview'))
+    assert_that(get_batch['changes'][0]['validationErrors'][0]['errorType'], is_('ZoneDiscoveryError'))
+
+    client.cancel_batch_change(result['id'], status=200)
+    get_batch = client.get_batch_change(result['id'])
+
+    assert_that(get_batch['status'], is_('Cancelled'))
+    assert_that(get_batch['approvalStatus'], is_('Cancelled'))
+    assert_that(get_batch['changes'][0]['status'], is_('Cancelled'))
+    assert_that(get_batch, has_key('cancelledTimestamp'))
```
mitruly: Can we add the following assertion:
```
assert_that(get_batch, not(has_key('reviewTimestamp')))
```

Can we also do something similar for the existing reject batch change functional test?

PR #795, opened 2019-08-08
JSON deserialization and protobuf changes for remove single DNS record entry

mitruly:
Prerequisite for remove single DNS record entry feature.

Changes in this pull request:
- Update protobuf for `SingleDeleteChange` to include an optional `recordData`
- Update unmarshalling to expect the correct data and output the correct types

Outstanding:
- Update unit and functional tests

### DIFF
--- a/modules/api/functional_test/live_tests/batch/create_batch_change_test.py
+++ b/modules/api/functional_test/live_tests/batch/create_batch_change_test.py
@@ -3197,6 +3197,7 @@ def test_create_batch_duplicates_update_check(shared_zone_test_context):
     finally:
         clear_recordset_list(to_delete, client)
 
+
 @pytest.mark.manual_batch_review
 def test_zone_name_requiring_manual_review(shared_zone_test_context):
     """
@@ -3229,3 +3230,34 @@ def test_zone_name_requiring_manual_review(shared_zone_test_context):
         # Clean up so data doesn't change
         if response:
             rejecter.reject_batch_change(response['id'], status=200)
+
+
+def test_create_batch_delete_record_fails(shared_zone_test_context):
+    """
+    Test creating batch change with DeleteRecord change input type is not recognized
+    """
+    client = shared_zone_test_context.ok_vinyldns_client
+    ok_zone = shared_zone_test_context.ok_zone
+    ok_group = shared_zone_test_context.ok_group
+
+    rs_to_create = get_recordset_json(ok_zone, "delete-record", "A", [{"address": "1.2.3.4"}], 200, ok_group['id'])
+
+    batch_change_input = {
+        "comments": "this is optional",
+        "changes": [
+            get_change_A_AAAA_json("delete-record.ok.", change_type="DeleteRecord")
+        ]
+    }
+
+    create_rs = None
+    try:
+        create_rs = client.create_recordset(rs_to_create, status=202)
+        client.wait_until_recordset_change_status(create_rs, 'Complete')
+
+        # TODO: Update this when DeleteRecord is supported
+        client.create_batch_change(batch_change_input, status=400)
+
+    finally:
+        if create_rs:
+            delete_rs = client.delete_recordset(ok_zone['id'], create_rs['recordSet']['id'], status=202)
+            client.wait_until_recordset_change_status(delete_rs, 'Complete')
--- a/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeProtocol.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeProtocol.scala
@@ -35,7 +35,8 @@ object BatchChangeInput {
   def apply(batchChange: BatchChange): BatchChangeInput = {
     val changes = batchChange.changes.map {
       case add: SingleAddChange => AddChangeInput(add)
-      case del: SingleDeleteRRSetChange => DeleteChangeInput(del)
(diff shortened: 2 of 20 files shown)


modules/api/functional_test/live_tests/batch/create_batch_change_test.py
```diff
@@ -3145,3 +3145,33 @@ def new_err(name, type):
...
+            get_change_A_AAAA_json("delete-record.ok.", change_type="DeleteRecord")
+        ]
+    }
+
+    create_rs = None
+    try:
+        create_rs = client.create_recordset(rs_to_create, status=202)
+        client.wait_until_recordset_change_status(create_rs, 'Complete')
+
+        # TODO: Update this when DeleteRecord is supported
```
mitruly: Added this test to setup `DeleteRecord`. This test should be reworked once `DeleteRecord` is fully supported in the flow.

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
```diff
@@ -169,7 +169,10 @@ class BatchChangeValidations(
       isApproved: Boolean): ValidatedBatch[ChangeInput] =
     input.map {
       case a: AddChangeInput => validateAddChangeInput(a, isApproved).map(_ => a)
-      case d: DeleteChangeInput => validateInputName(d, isApproved).map(_ => d)
+      case d: DeleteRRSetChangeInput => validateInputName(d, isApproved).map(_ => d)
+      // TODO: Add DeleteRecordChangeInput validations
```
mitruly: Adding this here pending validate `DeleteRecordChangeInput` changes.

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeProtocol.scala
```diff
@@ -89,6 +93,27 @@ final case class DeleteChangeInput(inputName: String, typ: RecordType) extends C
     )
 }
 
+// TODO: Remove coverage on/off
+// $COVERAGE-OFF$
```
mitruly: Blocking intermediate types that aren't being used from coverage (until `DeleteRecord` is included in the process flow).

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
```diff
@@ -169,7 +169,10 @@ class BatchChangeValidations(
...
+      case _: DeleteRecordChangeInput =>
+        throw new RuntimeException("DeleteRecordChangeInput validations not yet implemented.")
```
rebstar6: dont throw, return a `ValidatedNel[DomainValidationError, A]` which is always left ()

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchTransformations.scala
```diff
@@ -88,7 +88,10 @@ object BatchTransformations {
     def apply(zone: Zone, recordName: String, changeInput: ChangeInput): ChangeForValidation =
       changeInput match {
         case a: AddChangeInput => AddChangeForValidation(zone, recordName, a)
-        case d: DeleteChangeInput => DeleteChangeForValidation(zone, recordName, d)
+        case d: DeleteRRSetChangeInput =>
+          DeleteChangeForValidation(zone, recordName, d)
+        case d: DeleteRecordChangeInput =>
+          DeleteChangeForValidation(zone, recordName, d)
```
rebstar6: add a TODO here

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeProtocol.scala
```diff
@@ -73,7 +74,10 @@ final case class AddChangeInput(
   }
 }
 
-final case class DeleteChangeInput(inputName: String, typ: RecordType) extends ChangeInput {
+sealed trait DeleteChangeInput extends ChangeInput
```
rebstar6: i dont think we need this trait since were keeping these fully separate

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchTransformations.scala
```diff
@@ -130,19 +133,42 @@ object BatchTransformations {
       inputChange: DeleteChangeInput)
```
rebstar6: make this just always take to DeleteRecordChangeInput

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
mitruly: @rebstar6 I updated the PR to reflect this; the main downside to returning an `.invalidNel` here is that we would have to create a new one which would get permanently saved in `SingleChange` data.

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchTransformations.scala
```diff
@@ -124,25 +129,28 @@ object BatchTransformations {
...
-        inputChange.inputName,
-        inputChange.typ,
-        SingleChangeStatus.Pending,
-        None,
-        None,
-        None,
-        List.empty,
-        changeId.getOrElse(UUID.randomUUID().toString)
-      )
+      inputChange match {
```
britneywright: is this pattern matching needed?

PR #798, opened 2019-08-12
Allow configuration of ipv6 zone discovery search range

rebstar6:


modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeConverter.scala
```diff
@@ -157,7 +157,9 @@ class BatchChangeConverter(batchChangeRepo: BatchChangeRepository, messageQueue:
     }
     val deletes = NonEmptyList.fromList {
```
rebstar6: no changes in this file, formatter picked it up

modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
```diff
@@ -134,6 +135,24 @@ object VinylDNSConfig {
     .as[Option[Boolean]]("manual-batch-review-enabled")
     .getOrElse(false)
 
+  // defines nibble boundary for ipv6 zone discovery
+  // (min of 2, max of 3 means zones of form X.X.ip6-arpa. and X.X.X.ip6-arpa. will be discovered)
+  lazy val v6DiscoveryBoundries: IO[(Int, Int)] = IO {
```
rebstar6: can make this a type rather than tuple if anyone has strong feelings.

modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
```diff
@@ -134,6 +135,24 @@ object VinylDNSConfig {
...
+    val v6zoneNibbleMin: Int =
+      vinyldnsConfig.as[Option[Int]]("batch-v6-discovery-nibble-min").getOrElse(20)
```
rebstar6: err, comment to self, these are the wrong defaults. also need to ensure min > 0 and max < 32

modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
pauljamescleary: this really should be a type, ._1 and ._2 is difficult to follow

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -53,7 +53,9 @@ object BatchChangeService {
       manualReviewEnabled: Boolean,
       authProvider: AuthPrincipalProvider,
       notifiers: AllNotifiers,
-      scheduledChangesEnabled: Boolean): BatchChangeService =
+      scheduledChangesEnabled: Boolean,
+      v6zoneNibbleMin: Int,
```
pauljamescleary: rather than 2 separate values that are hard to pull around everywhere, `NibbleBoundary(min, max)` would be better

rebstar6: can do

modules/api/src/main/resources/application.conf
```diff
@@ -151,6 +151,5 @@ vinyldns {
 
   # feature flag for manual batch review
   manual-batch-review-enabled = true
-
-  scheduled-changes-enabled = false
+  scheduled-changes-enabled = true
```
britneywright: should this be true or we leaving the default as false?

rebstar6: in application.conf its ok, ref.conf is still false

modules/api/src/test/resources/application.conf
```diff
@@ -133,4 +133,5 @@ vinyldns {
   ]
 
   manual-batch-review-enabled = true
+  scheduled-changes-enabled = true
```
britneywright: same ? here

rebstar6: same A here :P

modules/docs/src/main/tut/operator/config-api.md
```diff
@@ -745,40 +797,4 @@ akka.http {
     illegal-header-warnings = on
   }
 }
-
-# types of unowned records that users can access in shared zones
-shared-approved-types = ["A", "AAAA", "CNAME", "PTR", "TXT"]
```
rebstar6: this all should have been nested in the vinyldns{}

modules/docs/src/main/tut/operator/config-api.md
```diff
@@ -530,6 +530,20 @@ manual review.
 manual-batch-review-enabled = true
 ```
 
+### IPv6 Zone Discovery Boundries
```
britneywright: 🤗

modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
```diff
@@ -134,6 +136,20 @@ object VinylDNSConfig {
     .as[Option[Boolean]]("manual-batch-review-enabled")
     .getOrElse(false)
 
+  // defines nibble boundary for ipv6 zone discovery
+  // (min of 2, max of 3 means zones of form X.X.ip6-arpa. and X.X.X.ip6-arpa. will be discovered)
+  lazy val v6DiscoveryBoundries: IO[V6DiscoveryNibbleBoundries] = IO {
+    val v6zoneNibbleMin: Int =
+      vinyldnsConfig.as[Option[Int]]("batch-v6-discovery-nibble-min").getOrElse(5)
```
pauljamescleary: are these in `reference.conf`?  if so, there is no need to load it as an option.

modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
```diff
@@ -134,6 +136,20 @@ object VinylDNSConfig {
...
+    val v6zoneNibbleMax: Int =
+      vinyldnsConfig.as[Option[Int]]("batch-v6-discovery-nibble-max").getOrElse(16)
+
+    V6DiscoveryNibbleBoundries(v6zoneNibbleMin, v6zoneNibbleMax)
+  }.flatMap {
+    case Right(x) => IO.pure(x)
```
pauljamescleary: What could fail here?

I would include those values in `reference.conf` and then just assume they are there.  An exception will be thrown if those aren't in there.  Also, why not use PureConfig for loading this?

modules/api/src/main/scala/vinyldns/api/domain/batch/V6DiscoveryNibbleBoundries.scala
```diff
@@ -0,0 +1,36 @@
...
+package vinyldns.api.domain.batch
+
+import vinyldns.api.VinylDNSConfig.VinylDNSConfigLoadError
+
+final case class V6DiscoveryNibbleBoundries(min: Int, max: Int)
+
+object V6DiscoveryNibbleBoundries {
+  def apply(min: Int, max: Int): Either[VinylDNSConfigLoadError, V6DiscoveryNibbleBoundries] =
+    if (min < 1) {
+      Left(VinylDNSConfigLoadError(s"v6zoneNibbleMin ($min) cannot be less than 1"))
```
pauljamescleary: Instead of this check here, you could use PureConfig to load, and add an `assert` to test the conditions.  This way, if the config values are missing, or not valid, PureConfig will wind up in an exception case.

modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
```diff
@@ -134,6 +136,20 @@ object VinylDNSConfig {
     .as[Option[Boolean]]("manual-batch-review-enabled")
     .getOrElse(false)
 
+  // defines nibble boundary for ipv6 zone discovery
+  // (min of 2, max of 3 means zones of form X.X.ip6-arpa. and X.X.X.ip6-arpa. will be discovered)
+  lazy val v6DiscoveryBoundries: IO[V6DiscoveryNibbleBoundries] = IO {
```
mitruly: Thinking this should be `v6DiscoveryBoundaries`.

modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
```diff
@@ -134,6 +136,20 @@ object VinylDNSConfig {
...
+    val v6zoneNibbleMin: Int =
+      vinyldnsConfig.as[Option[Int]]("batch-v6-discovery-nibble-min").getOrElse(5)
+    val v6zoneNibbleMax: Int =
+      vinyldnsConfig.as[Option[Int]]("batch-v6-discovery-nibble-max").getOrElse(16)
```
mitruly: Should this be `getOrElse(20)`?

modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
rebstar6: they are not. they can be though! i have no strong pref between the two

modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
rebstar6: V6DiscoveryNibbleBoundaries(v6zoneNibbleMin, v6zoneNibbleMax) has an apply() that returns an either ensuring the constraints on min/max

PR #800, opened 2019-08-12
add DNS Change review ability to portal

britneywright:
Changes in this pull request:
- For super and support users give them the ability to approve or reject pending DNS Changes through the portal

### DIFF
--- a/modules/portal/app/controllers/VinylDNS.scala
+++ b/modules/portal/app/controllers/VinylDNS.scala
@@ -628,8 +628,44 @@ class VinylDNS @Inject()(
         Status(response.status)(response.body)
           .withHeaders(cacheHeaders: _*)
       })
+    // $COVERAGE-ON$
+  }
+
+  def approveBatchChange(batchChangeId: String): Action[AnyContent] = userAction.async {
+    implicit request =>
+      // $COVERAGE-OFF$
+      val json = request.body.asJson
+      val payload = json.map(Json.stringify)
+      val vinyldnsRequest =
+        new VinylDNSRequest(
+          "POST",
+          s"$vinyldnsServiceBackend",
+          s"zones/batchrecordchanges/$batchChangeId/approve",
+          payload)
+      executeRequest(vinyldnsRequest, request.user).map(response => {
+        Status(response.status)(response.body)
+          .withHeaders(cacheHeaders: _*)
+      })
       // $COVERAGE-ON$
   }
+
+  def rejectBatchChange(batchChangeId: String): Action[AnyContent] =
+    userAction.async { implicit request =>
+      // $COVERAGE-OFF$
+      val json = request.body.asJson
+      val payload = json.map(Json.stringify)
+      val vinyldnsRequest =
+        new VinylDNSRequest(
+          "POST",
+          s"$vinyldnsServiceBackend",
+          s"zones/batchrecordchanges/$batchChangeId/reject",
+          payload)
+      executeRequest(vinyldnsRequest, request.user).map(response => {
+        Status(response.status)(response.body)
+          .withHeaders(cacheHeaders: _*)
+      })
+      // $COVERAGE-ON$
+    }
 
   def lockUser(userId: String): Action[AnyContent] = userAction.async { implicit request =>
     if (request.user.isSuper) {
--- a/modules/portal/app/views/batchChanges/batchChangeDetail.scala.html
+++ b/modules/portal/app/views/batchChanges/batchChangeDetail.scala.html
@@ -155,7 +155,7 @@
                                 </td>
                                 <td class="wrap-long-text">
                                     <p ng-repeat="error in change.validationErrors">
-                                        {{error.message}}
+                                        {{error.message ? error.message : error}}
                                     </p>
                                     {{change.systemMessage}}
                                 </td>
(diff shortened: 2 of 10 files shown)


modules/portal/app/views/batchChanges/batchChangeDetail.scala.html
```diff
@@ -156,6 +156,29 @@ <h3 class="panel-title">Changes</h3>
                         </table>
                     </div>
                 </div>
+                @if(rootAccountSuperAdmin || rootAccountSupportAdmin) {
+                    <div ng-if="batch.approvalStatus == 'PendingReview'" class="panel panel-default">
+                        <div class="panel-heading">
+                            <h3 class="panel-title">Review DNS Change</h3>
+                        </div>
+                        <div class="panel-body">
+                            <label>Review Comment (optional):</label> <input type="text" ng-model="reviewComment">
```
mitruly: This comment field is not properly getting set when approving/rejecting. Tested both approving and rejecting in the portal and this field did not display; tested approving and rejecting via API and this value was set.

modules/portal/public/lib/batch-change/batch-change-detail.controller.js
```diff
@@ -48,6 +51,58 @@
...
+                    .approveBatchChange($scope.batch.id, $scope.reviewComment)
+                    .then(success)
+                    .catch(function (error) {
+                        if (typeof error.data == "object") {
+                            for(var i = 0; i < error.data.length; i++) {
+                                if (error.data[i].errors) {
+                                    $scope.batch.changes[i].validationErrors = error.data[i].errors
+                                }
+                            }
+                            var errorAlert = {data: "Issues still remain, cannot approve DNS Change. Resolve all outstanding issues or reject the DNS Change.", status: error.status}
```
mitruly: Is it possible to indicate somehow what the remaining issues are? The only thing that appears right now is the alert box it can get confusing if there are multiple zone discovery errors.

Wonder if there would be a good approach to highlighting outstanding errors.

modules/portal/public/lib/batch-change/batch-change.spec.js
```diff
@@ -332,6 +389,24 @@ describe('BatchChange', function(){
                 });
             this.$httpBackend.flush();
         });
+
+        it('http backend gets called properly when approving a batch change', function () {
+            this.$httpBackend.expectPOST('/api/dnschanges/123/approve').respond('batch change created');
```
mitruly: Can we add an assertion for the comment here?

britneywright:
Screen Shot 2019-08-13 at 11 40 55 AM

added row highlighting for errors

modules/portal/app/controllers/VinylDNS.scala
```diff
@@ -628,9 +628,45 @@ class VinylDNS @Inject()(
         Status(response.status)(response.body)
           .withHeaders(cacheHeaders: _*)
       })
+    // $COVERAGE-ON$
+  }
+
+  def approveBatchChange(batchChangeId: String): Action[AnyContent] = userAction.async {
+    implicit request =>
+      // $COVERAGE-OFF$
```
pauljamescleary: no tests?

modules/portal/app/views/batchChanges/batchChangeDetail.scala.html
```diff
@@ -1,4 +1,4 @@
-@(rootAccountName: String)(implicit request: play.api.mvc.Request[Any], customLinks: models.CustomLinks, meta: models.Meta)
+@(rootAccountName: String, rootAccountSuperAdmin: Boolean, rootAccountSupportAdmin: Boolean)(implicit request: play.api.mvc.Request[Any], customLinks: models.CustomLinks, meta: models.Meta)
```
pauljamescleary: It would be better to have a boolean `canReview` here instead of passing in other booleans and determining the logic in page

modules/portal/public/lib/batch-change/batch-change-detail.controller.js
pauljamescleary: I would think that if errors "go away", that they should not be returned no?  It is confusing to "approve", have some new errors and some errors that are no longer there, and be presented with the old error set.  If this is in-fact the way things are then I am ok showing all the errors.  If for some reason we are not "updating" the error list we should discuss.

modules/portal/app/controllers/VinylDNS.scala
britneywright: added the base tests we have for the other routes.

modules/portal/public/lib/batch-change/batch-change-detail.controller.js
britneywright: If there are still errors that are returned after an admin tries to approve a DNS Change I've set it so the new errors replace the old errors. If a single change has no outstanding issues after re-validation I simply remove the original error message. 

Might be worth noting: new errors aren't saved on the API side so if the admin refreshes the page they would see the original errors again.

pauljamescleary: Ok.  Make a card to update that in the API.

modules/portal/app/controllers/VinylDNS.scala
pauljamescleary: remove `COVERAGE-OFF / COVERAGE-ON` if there are tests

britneywright: there aren't tests on the requests themselves, we currently only test whether a request can be made.

PR #802, opened 2019-08-14
lock down approve/reject for test support user

rebstar6:


modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -569,4 +570,22 @@ class BatchChangeService(
 
     batchChangeRepo.save(cancelledBatch).toBatchResult
   }
+
+  def getBypassTestCheckForReject(
+      rejecterAuth: AuthPrincipal,
+      batchChange: BatchChange): BatchResult[Boolean] =
+    if (!rejecterAuth.isTestUser) {
```
rebstar6: added this logic so as to not have the unnecessary user table call on every batch. will only get the requesting user in the case that the rejection call came from a test user

britneywright: why does this apply to rejection calls but not approval calls?

rebstar6: for approval we have to get the user that created the batch in order to rerun the validations, so the call has to be made regardless

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
```diff
@@ -141,8 +145,10 @@ class BatchChangeValidations(
 
   def validateAuthorizedReviewer(
       auth: AuthPrincipal,
-      batchChange: BatchChange): Either[BatchChangeErrorResponse, Unit] =
-    if (auth.isSystemAdmin) {
+      batchChange: BatchChange,
+      bypassTestValidation: Boolean): Either[BatchChangeErrorResponse, Unit] =
+    if (auth.isSystemAdmin && (bypassTestValidation || !auth.isTestUser)) {
```
mitruly: I think renaming `bypassTestValidation` to `isTestChange` is clearer. Worked out the logic and the flow makes sense.

rebstar6: yeah i started with isTestChange though its a lie a bit on the reject side. for reject, that will be true if its a test change OR just if the rejector user isnt test so we didnt bother checking

PR #805, opened 2019-08-15
enhance portal datetime form field

britneywright:
Screen Shot 2019-08-15 at 4 29 01 PM

Changes in this pull request:
- use [daterangepicker](http://www.daterangepicker.com/) and [moment.js](https://momentjs.com) to enhance the `Request Date and Time` field in the new DNS Change field. 
- change the display of the local time zone
- set a default date and time of the current date and top of the next hour
- reverse parsing of date time if there are errors in the form


modules/portal/.gitignore
```diff
@@ -2,7 +2,6 @@
 __pycache__
 *.pyc
 /func_test/.virtualenv
-bower_components/
```
mitruly: This doesn't feel like the right approach - we should be able to download and install the `moment` component instead of including it as part of the repo.

britneywright: Gentelella includes an older version of moment, that's why there's a newer one. I didn't include it initially, but travis failed without it.

mitruly: Running this locally, there is a similar message when loading for the first time:
```
custom.js:1584 Uncaught ReferenceError: moment is not defined
    at init_daterangepicker (custom.js:1584)
    at HTMLDocument.<anonymous> (custom.js:5021)
    at i (jquery.min.js:2)
    at Object.fireWith [as resolveWith] (jquery.min.js:2)
    at Function.ready (jquery.min.js:2)
    at HTMLDocument.J (jquery.min.js:2)
```

Feels like this is due to `moment` being called before the script is finished loading.

britneywright: you're seeing this error when you load the New DNS Change page?

britneywright: corrected the dependencies. not sure if that was related to the error you saw though, let me know if you're still seeing it.

mitruly: Not sure, but the error is gone now.

PR #807, opened 2019-08-16
Flag batch changes requiring manual review by zone name

mitruly:
Changes in this pull request:
- Allow flagging of batch changes that require manual review by zone name match
- Update tests
- Update docs


docker/api/docker.conf
```diff
@@ -220,6 +220,9 @@ vinyldns {
       "fd69:27cc:fe91:0:0:0:ffff:1",
       "fd69:27cc:fe91:0:0:0:ffff:2"
     ]
+    zone-name-list = [
+      "Non.test.shared"
```
britneywright: could you make a new zone for this so it doesn't impact existing tests.

mitruly: @britneywright Would like to get some input from @pauljamescleary on this, since in the past we've tried to avoid adding test zones when not necessary. Specifically chose the `non.test.shared.` zone since it is not used for any other tests.

britneywright:
can you add a test for this in the `approve_batch_change_test`

modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
```diff
@@ -36,6 +37,8 @@ import vinyldns.core.queue.MessageQueueConfig
 import vinyldns.core.repository.DataStoreConfig
 import vinyldns.core.notifier.NotifierConfig
 
+import scala.collection.immutable.HashSet
```
mitruly: `HashSet` is more optimized than using a sequenced collection since we can do a direct lookup.

modules/api/functional_test/live_tests/batch/create_batch_change_test.py
```diff
@@ -3145,3 +3145,36 @@ def new_err(name, type):
...
+        assert_that(get_batch['status'], is_('PendingReview'))
+        assert_that(get_batch['approvalStatus'], is_('PendingReview'))
+        for i in xrange(0, 3):
+            assert_that(get_batch['changes'][i]['status'], is_('NeedsReview'))
+            assert_that(get_batch['changes'][i]['validationErrors'][0]['errorType'], is_('RecordRequiresManualReview'))
+
+    finally:
+        # Clean up so data doesn't change
+        if response:
+            rejecter.reject_batch_change(response['id'], status=200)
```
rebstar6: new line :P

modules/api/functional_test/live_tests/shared_zone_test_context.py
```diff
@@ -281,6 +281,29 @@ def __init__(self):
...
+            # zone with name configured for manual review
+            requires_review_zone_change = self.ok_vinyldns_client.create_zone(
+                {
+                    'name': 'zone.requires.review.',
+                    'email': 'test@test.com',
+                    'shared': False,
+                    'adminGroupId': self.ok_group['id'],
+                    'isTest': True,
+                    'connection': {
+                        'name': 'review.',
```
rebstar6: can use the func-test-backend id fyi. just makes things shorter

modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
rebstar6: is the default `Set` sequenced?

mitruly: No, but `List` would be.

Scala's `Set` uses `HashSet` for implementations of size > 4, so I'll change this to `Set`.

PR #808, opened 2019-08-18
Refactor Portal Authentication

pauljamescleary:
This PR is an interim step to fixing OIDC.  The main issue with the existing authentication is that the controllers are tightly coupled to the authentication mechanisms themselves, namely the OidcAuthenticator and how we push / pull data from session state.  This has to be pulled out.

1. Introduce `SecuritySupport` which abstracts away the authentication and session management concerns
1. Introduce an `ApiActionBuilder` and `FrontendActionBuilder` which are actions that correspond to authentication and used heavily in the controllers.  These are now provided via `SecuritySupport` trait to the controllers via injection
1. Wrap the existing `ApiAction` and `FrontendAction` in `Legacy` implementations that mimic how everything works today

In addition, eliminated a lot of duplication in the `VinylDNSSpec` removing 400+ lines of duplicate code.

This change will allow us to tie in Pac4j updates for OIDC authentication in the `VinylDNSModule` at load time.

### DIFF
--- a/modules/portal/app/actions/ApiAction.scala
+++ b/modules/portal/app/actions/ApiAction.scala
@@ -21,7 +21,7 @@ import controllers.{CacheHeader, OidcAuthenticator}
 import javax.inject.Inject
 import org.slf4j.LoggerFactory
 import play.api.mvc.Results.{Forbidden, NotFound, Unauthorized}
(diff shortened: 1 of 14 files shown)


britneywright:
@pauljamescleary are there configuration changes that need to be made locally for this?

pauljamescleary:
@britneywright none, config changes will be introduced in a future PR

PR #809, opened 2019-08-19
Open list groups access

britneywright:
Changes in this pull request:
- API: Add parameter `ignoreAccess` to list groups route. Default value is false.
- API: If `ignoreAccess` is true return groups regardless of requester's membership in the group
- Portal: add "My Groups" and "All Groups" tabs in the Groups page of the portal. Similar to the Zones tabs. View, Edit, and Delete buttons are visible based on group membership and admin rights.

Doesn't change existing behavior, though we may want to in future PRs:
- support and super users have access to all groups, regardless of the value of `ignoreAccess`
 - non-admins cannot view individual groups they don't have access to, member lists or admin lists of those groups.

### DIFF
--- a/modules/api/functional_test/live_tests/membership/list_my_groups_test.py
+++ b/modules/api/functional_test/live_tests/membership/list_my_groups_test.py
@@ -30,6 +30,15 @@ class ListGroupsSearchContext(object):
                 pass
             raise
 
+    def verify_ignore_access(self, results):
+        assert_that(results, has_length(3))  # 3 fields
+
+        assert_that(len(results['groups']), greater_than(50))
+        assert_that(results, is_not(has_key('groupNameFilter')))
+        assert_that(results, is_not(has_key('startFrom')))
+        assert_that(results, is_not(has_key('nextId')))
+        assert_that(results['maxItems'], is_(100))
+
     def tear_down(self):
         clear_zones(self.client)
         clear_groups(self.client)
@@ -171,13 +180,7 @@ def test_list_my_groups_with_ignore_access_true(list_my_groups_context):
 
     results = list_my_groups_context.client.list_my_groups(ignore_access=True, status=200)
 
-    assert_that(results, has_length(3))  # 3 fields
-
-    assert_that(len(results['groups']), greater_than(50))
-    assert_that(results, is_not(has_key('groupNameFilter')))
-    assert_that(results, is_not(has_key('startFrom')))
-    assert_that(results, is_not(has_key('nextId')))
-    assert_that(results['maxItems'], is_(100))
+    list_my_groups_context.verify_ignore_access(results)
 
     my_results = list_my_groups_context.client.list_my_groups(status=200)
     my_results['groups'] = sorted(my_results['groups'], key=lambda x: x['name'])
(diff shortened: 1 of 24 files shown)


modules/api/functional_test/live_tests/membership/list_my_groups_test.py
```diff
@@ -163,3 +164,53 @@ def test_list_my_groups_no_deleted(list_my_groups_context):
...
+def test_list_my_groups_with_ignore_access_true(list_my_groups_context):
+    """
+    Test that we can get all the groups whether a user is a member or not
+    """
+
+    results = list_my_groups_context.client.list_my_groups(ignore_access=True, status=200)
+
+    assert_that(results, has_length(3))  # 3 fields
+
+    assert_that(len(results['groups']), greater_than(50))
```
mitruly: We can consolidate all of these shared assertions into a common helper.

modules/portal/app/controllers/VinylDNS.scala
```diff
@@ -67,6 +67,7 @@ object VinylDNS {
       lastName: Option[String],
       email: Option[String],
       isSuper: Boolean,
+      isSupport: Boolean,
```
mitruly: Doesn't seem like `isSupport` is used.

modules/portal/public/lib/controllers/controller.groups.js
```diff
@@ -101,19 +102,21 @@ angular.module('controller.groups', []).controller('GroupsController', function
             });
     };
 
-    $scope.refresh = function () {
-        //get users groups
+    $scope.refresh = function (ignoreAccess = null) {
```
mitruly: I'm not a fan of passing a parameter into `refresh` like this, especially when nothing inside the controller itself sets passes anything in and we're just using the existing `$scope.ignoreAccess` value.

Can we do something similar to `$scope.getAllRequests` in `batch-changes.controller.js` where we set `$scope.ignoreAccess` and then call `$scope.refresh()`?

modules/portal/app/controllers/VinylDNS.scala
britneywright: isSupport is used here https://github.com/vinyldns/vinyldns/pull/809/files#diff-9f95282be679fa1704072440c7c5e603R246

modules/portal/public/lib/controllers/controller.groups.js
```diff
@@ -236,6 +246,13 @@ angular.module('controller.groups', []).controller('GroupsController', function
         return (isAdmin || isSuper) ? true : false;
     }
 
+    $scope.canSeeGroup = function(group) {
+        var isMember = group.members.some(x => x.id === $scope.profile.id);
+        var isSupport = $scope.profile.isSupport;
+        var isSuper = $scope.profile.isSuper;
+        return (isMember || isSupport || isSuper) ? true : false;
```
mitruly: This is just `return isMember || isSupport || isSuper;`

britneywright: very true 😉

britneywright: updated

PR #811, opened 2019-08-21
update send to manual review logic

britneywright:
Changes in this pull request in the BatchChangeService `buildResponse` method:
- if a manual review is enabled, a batch change has no errors, and is scheduled it goes to manual review. since there are no errors on the changes we don't need to verify if an owner group ID is present. (note: if manual review is disabled the batch change will fail earlier in the `applyBatchChangeValidationFlow`)
- if manual review is enabled system-wide and allowManualReview attribute on a batch is true it will go  to manual review if an owner group ID is specified, otherwise returns a ManualReviewRequiresOwnerGroup error.

### DIFF
--- a/modules/api/functional_test/vinyldns_python.py
+++ b/modules/api/functional_test/vinyldns_python.py
@@ -553,13 +553,16 @@ class VinylDNSClient(object):
         response, data = self.make_request(url, u'GET', self.headers, **kwargs)
         return data
 
-    def create_batch_change(self, batch_change_input, **kwargs):
+    def create_batch_change(self, batch_change_input, allow_manual_review=True, **kwargs):
         """
         Creates a new batch change
         :param batch_change_input: the batchchange to be created
+        :param allow_manual_review: if true and manual review is enabled soft failure are treated as hard failures
         :return: the content of the response
         """
         url = urljoin(self.index_url, u'/zones/batchrecordchanges')
+        if allow_manual_review is not None:
+            url = url + (u'?' + u'allowManualReview={0}'.format(allow_manual_review))
         response, data = self.make_request(url, u'POST', self.headers, json.dumps(batch_change_input), **kwargs)
         return data
 
--- a/modules/api/functional_test/vinyldns_python.py
+++ b/modules/api/functional_test/vinyldns_python.py
@@ -557,7 +557,7 @@ class VinylDNSClient(object):
         """
         Creates a new batch change
         :param batch_change_input: the batchchange to be created
-        :param allow_manual_review: if true and manual review is enabled soft failure are treated as hard failures
+        :param allow_manual_review: if true and manual review is enabled soft failures are treated as hard failures
         :return: the content of the response
         """
         url = urljoin(self.index_url, u'/zones/batchrecordchanges')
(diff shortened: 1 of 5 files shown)


modules/api/functional_test/vinyldns_python.py
```diff
@@ -553,13 +553,16 @@ def list_recordsets(self, zone_id, start_from=None, max_items=None, record_name_
         response, data = self.make_request(url, u'GET', self.headers, **kwargs)
         return data
 
-    def create_batch_change(self, batch_change_input, **kwargs):
+    def create_batch_change(self, batch_change_input, allow_manual_review=True, **kwargs):
         """
         Creates a new batch change
         :param batch_change_input: the batchchange to be created
+        :param allow_manual_review: if true and manual review is enabled soft failure are treated as hard failures
```
mitruly: `soft failures`

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -424,20 +424,20 @@ class BatchChangeService(
     } else if (noErrors && !isScheduled) {
       // There are no errors and this is not scheduled, so process immediately
       processNowResponse
-    } else {
+    } else if (this.manualReviewEnabled && noErrors && isScheduled) {
```
mitruly: Feel like this logic is simplest to follow when we consolidate into four cases:
- Hard errors are present
- Process now
- Manual review
- Error fallback

```
    if (hardErrorsPresent) {
      // Always error out
      errorResponse
    } else if (noErrors && !isScheduled) {
      // There are no errors and this is not scheduled, so process immediately
      processNowResponse
    } else if (this.manualReviewEnabled && allowManualReview) {
      // Manual review is enabled
      if ((noErrors && isScheduled) || batchChangeInput.ownerGroupId.isDefined) {
        manualReviewResponse
      } else {
        ManualReviewRequiresOwnerGroup.asLeft
      }
    } else {
      // Cannot go to manual review, and we have soft errors, so just return a failure
      errorResponse
    }
```

mitruly:
What about the scenario where manual batch review is enabled, there are no errors, it is scheduled and the user explicitly passes in the `allowManualReview=false` parameter? IIRC the motivation for reworking this logic was because scheduled changes was not honoring the override parameter.

It feels inconsistent to conditionally honor this parameter override only when there are soft errors, though there may be some motivation from a portal perspective.

modules/api/functional_test/vinyldns_python.py
britneywright: fixed

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
britneywright: updated @mitruly

britneywright:
> What about the scenario where manual batch review is enabled, there are no errors, it is scheduled and the user explicitly passes in the `allowManualReview=false` parameter? IIRC the motivation for reworking this logic was because scheduled changes was not honoring the override parameter.
> 
> It feels inconsistent to conditionally honor this parameter override only when there are soft errors, though there may be some motivation from a portal perspective.

I completely missed this comment yesterday @mitruly. My initial commit did ignore the parameter override if a batch change was scheduled and had no errors. The refactor follows the parameter. In use it feels a little weird, but you're right about it being consistent.

PR #816, opened 2019-08-22
update Groups screenshots for new tabs

britneywright:
Changes in this pull request:
- update Groups screenshots in the documentation to show the new "My Groups" and "All Groups" tabs from #809


modules/docs/src/main/tut/api/list-groups.md
```diff
@@ -19,6 +19,7 @@ name          | type          | required?   | description |
 groupNameFilter    | string        | no          | One or more characters contained in the name of the group set to search for.  For example `TP`.  This is a contains search only, no wildcards or regular expressions are supported |
 startFrom     | *any*         | no          | In order to advance through pages of results, the startFrom is set to the `nextId` that is returned on the previous response.  It is up to the client to maintain previous pages if the client wishes to advance forward and backward.   If not specified, will return the first page of results |
 maxItems      | integer       | no          | The number of items to return in the page.  Valid values are 1 to 100. Defaults to 100 if not provided. |
+ignoreAccess  | boolean       | no          | If false, returns only groups the requesting user is a member of. If true, returns groups in the system, regardless of membership. Defaults to false if not provided. Super and support admin see all groups regardless of this value. |
```
mitruly: `If true, returns all groups ...`

mitruly: The request section should also be updated to include `ignoreAccess`:
```
GET /groups?startFrom={response.nextId}&maxItems={1 - 100}&groupNameFilter={filter}&ignoreAccess={true  &#124; false}
```

britneywright: we don't say "all" because there's still the maxItems restriction. There was confusion over that when we first implemented this param in Zones.

britneywright: updated

PR #817, opened 2019-08-22
update portal soft error check

britneywright:
Changes in this pull request:
- correctly aggregate the soft errors in the portal

### DIFF
--- a/modules/portal/public/lib/batch-change/batch-change-new.controller.js
+++ b/modules/portal/public/lib/batch-change/batch-change-new.controller.js
@@ -81,16 +81,11 @@
                               && ($scope.newBatch.changes[i].errors.every(e => e.includes('Zone Discovery Failed') || e.includes('requires manual review')))
                             ) {
                                 $scope.newBatch.changes[i].softErrors = true;
-                                $scope.newBatch.changes[i].hardErrors = false;
                             } else {
                                 $scope.anyHardErrors = true;
-                                $scope.newBatch.changes[i].softErrors = false;
                                 $scope.newBatch.changes[i].hardErrors = true;
                                 $scope.ownerGroupError = $scope.newBatch.changes[i].errors.filter(e => e.includes('owner group ID must be specified for record'))[0]
                             }
-                        } else {
-                            $scope.newBatch.changes[i].softErrors = false;
-                            $scope.newBatch.changes[i].hardErrors = false;
                         }
                     }
                 }
--- a/modules/portal/public/lib/batch-change/batch-change-new.controller.js
+++ b/modules/portal/public/lib/batch-change/batch-change-new.controller.js
@@ -103,7 +103,7 @@
 
                 formatData(payload);
 
-                return batchChangeService.createBatchChange(payload, true)
+                return batchChangeService.createBatchChange(payload, $scope.allowManualReview)
                     .then(success)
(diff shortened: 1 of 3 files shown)


modules/portal/public/lib/batch-change/batch-change-new.controller.js
```diff
@@ -112,9 +112,11 @@
                             $scope.newBatch.changes = error.data;
                             $scope.batchChangeErrors = true;
                             $scope.listOfErrors = error.data.flatMap(d => d.errors)
-                            $scope.ownerGroupError = $scope.listOfErrors.some(e => e.includes('owner group ID must be specified for record'));
-                            var hardErrors = $scope.listOfErrors.every(e => ['Zone Discovery Failed', 'requires manual review'].includes(e));
-                            if ($scope.manualReviewEnabled && !hardErrors) {
+                            $scope.ownerGroupError = $scope.listOfErrors.length > 0 && $scope.listOfErrors.some(e =>  e == undefined || e.includes('owner group ID must be specified for record'));
```
mitruly: This also needs to check for `Batch change requires owner group for manual review/scheduled changes` error.

modules/portal/public/lib/batch-change/batch-change-new.controller.js
```diff
@@ -89,33 +69,64 @@
...
+                function parseSingleChangeResponse(singleChanges) {
+                    $scope.allowManualReview = true;
+                    $scope.ownerGroupError = null;
+                    $scope.newBatch.changes = singleChanges;
+                    for(var i = 0; i < $scope.newBatch.changes.length; i++) {
+                        if ($scope.newBatch.changes[i].errors) {
+                            $scope.singleChangeErrors = true;
+                            if (
+                              $scope.manualReviewEnabled
+                              && ($scope.newBatch.changes[i].errors.every(e => e.includes('Zone Discovery Failed') || e.includes('requires manual review')))
```
britneywright: if we return the `isFatal` attribute on the errors from the API we could do something like `&& $scope.newBatch.changes[i].errors.every(e => !e.isFatal)` instead of needing to match on parts of the error messages.

modules/portal/public/lib/batch-change/batch-change-new.controller.js
```diff
@@ -89,33 +69,64 @@
...
-                            $scope.newBatch.changes = error.data;
-                            $scope.batchChangeErrors = true;
-                            $scope.listOfErrors = error.data.flatMap(d => d.errors)
-                            $scope.ownerGroupError = $scope.listOfErrors.some(e => e.includes('owner group ID must be specified for record'));
-                            $scope.softErrors = false;
-                            $scope.formStatus = "pendingSubmit";
-                            $scope.alerts.push({type: 'danger', content: 'Errors found. Please correct and submit again.'});
+                            parseSingleChangeResponse(error.data);
+                            if (!$scope.singleChangeErrors) {
+                                $scope.confirmationPrompt = "Would you like to submit this change to be processed at your requested time?"
```
pauljamescleary: why does the confirmation prompt say "at your requested time" here?

britneywright: Because it’s a scheduled change with no errors at that point.

PR #820, opened 2019-08-23
move fully to ip4s

rebstar6:
been meaning to do this, no need to have multiple ip/cidr libraries in here

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/zone/AclRuleOrdering.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/zone/AclRuleOrdering.scala
@@ -16,7 +16,7 @@
 
 package vinyldns.api.domain.zone
 
-import com.aaronbedra.orchard.CIDR
+import com.comcast.ip4s.Cidr
 import vinyldns.core.domain.record.RecordType
 import vinyldns.core.domain.zone.ACLRule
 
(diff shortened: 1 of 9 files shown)


modules/api/src/main/scala/vinyldns/api/domain/zone/AclRuleOrdering.scala
```diff
@@ -58,10 +58,11 @@ object ACLRuleOrdering extends ACLRuleOrdering {
...
   def sortableRecordMaskValue(rule: ACLRule): Int = {
-    val slash = rule.recordMask match {
-      case Some(cidrRule) => CIDR.valueOf(cidrRule).getMask
-      case None => 0
-    }
-    128 - slash
+    val prefix = for {
+      mask <- rule.recordMask
+      c <- Cidr.fromString(mask)
+    } yield c.prefixBits
```
rebstar6: pretty sure prefixBits is the right thing to be using, but need to double check

modules/api/src/test/scala/vinyldns/api/domain/AccessValidationsSpec.scala
```diff
@@ -812,7 +812,7 @@ class AccessValidationsSpec
         RecordSet("id", "255.255", RecordType.PTR, 200, RecordSetStatus.Active, DateTime.now)
 
       accessValidationTest.ruleAppliesToRecordName(rsTrue.name, rsTrue.typ, znTrue, aclRule) shouldBe true
-      accessValidationTest.ruleAppliesToRecordName(rsFalse.name, rsFalse.typ, znFalse, aclRule) shouldBe false
+      accessValidationTest.ruleAppliesToRecordName(rsFalse.name, rsFalse.typ, znFalse, aclRule) shouldBe true
```
rebstar6: this never made sense, /0 should match everything, no?

PR #823, opened 2019-08-30
save new batch change errors on approval attempt

britneywright:
For batch changes that are pending review, if an approval is attempted but there are new errors we should save the updated validation errors on the SingleChanges.


modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -492,6 +493,18 @@ class BatchChangeService(
       UnknownConversionError("Cannot convert a rejected batch change").toLeftBatchResult
   }
 
+  def getApprovalResult(batchChange: BatchChange, transformed: ValidatedBatch[ChangeForValidation])
+    : Either[BatchChangeErrorResponse, BatchChange] =
+    batchChange.approvalStatus match {
+      case ManuallyApproved => batchChange.asRight
+      case PendingReview => {
+        val batchChangeInput = BatchChangeInput(batchChange)
+        InvalidBatchChangeResponses(batchChangeInput.changes, transformed).asLeft
```
britneywright: Think I'm going to change this to a new error that can just use the updated batch change.

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -492,6 +493,17 @@ class BatchChangeService(
...
+      }
+      case _ => UnknownConversionError("Cannot convert to a batch change response.").asLeft
```
britneywright: we should never actually end up here with a batch change that isn't manually approved or pending review.

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -161,15 +161,15 @@ class BatchChangeService(
...
-        reviewInfo).toBatchResult
-      serviceCompleteBatch <- convertOrSave(
+        reviewInfo)
+      _ <- convertOrSave(
         changeForConversion,
         validationOutput.existingZones,
         validationOutput.existingRecordSets,
         batchChange.ownerGroupId)
-    } yield serviceCompleteBatch
+      result <- getApprovalResult(changeForConversion, validationOutput.validatedChanges).toBatchResult
```
mitruly: `getApprovalResult` should be renamed to `buildResponseForApprover` (since that's what it's doing). The existing `buildResponseForApprover` should be renamed to something like `buildBatchChange` or `updateBatchChangeWithErrors` and return a `BatchChange` type. We don't need to use `ValidatedBatch` if we're always returning `Right[BatchChange]`.

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -441,30 +441,31 @@ class BatchChangeService(
 
   def buildResponseForApprover(
       existingBatchChange: BatchChange,
-      batchChangeInput: BatchChangeInput,
       transformed: ValidatedBatch[ChangeForValidation],
-      reviewInfo: BatchChangeReviewInfo): Either[BatchChangeErrorResponse, BatchChange] =
```
mitruly: Based on comment about rename, this can just return a `BatchChange`.

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
mitruly: This could just follow what the previous logic did:
```
case ManuallyApproved => batchChange.asRight
case _ => InvalidBatchChangeResponses(batchChangeInput.changes, transformed).asLeft
```

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -493,6 +494,13 @@ class BatchChangeService(
       UnknownConversionError("Cannot convert a rejected batch change").toLeftBatchResult
   }
 
+  def buildResponseForApprover(
```
pauljamescleary: I know it isn't this PR, but `convertOrSave` has an issue.  The catch all `case _` assumes trying to convert or save on a rejected batch change approval status.  However, given the `if manualReviewEnabled` it is possible to fall into that catch all if `PendingReview and manualReviewEnabled == false`.

Need to clean up that `convertOrSave`, perhaps add a TODO

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -159,17 +159,17 @@ class BatchChangeService(
...
       validationOutput <- applyBatchChangeValidationFlow(asInput, requesterAuth, isApproved = true)
-      changeForConversion <- buildResponseForApprover(
+      changeForConversion <- rebuildBatchChangeForUpdate(
         batchChange,
-        asInput,
         validationOutput.validatedChanges,
-        reviewInfo).toBatchResult
-      serviceCompleteBatch <- convertOrSave(
+        reviewInfo)
+      _ <- convertOrSave(
```
pauljamescleary: we do not want to lose this result, as `convertOrSave` can add additional information to the batch change.

We still want to carry forward the `serviceCompleteBatch` result.

I am still teasing apart the flow here, but we should:

1. Revalidate; on failure we should simply do a save of the failures and return.  on success, we should continue with the existing flow as it is.

pauljamescleary: So, yea, carry this forward...

```scala
serviceCompleteBatch <- convertOrSave...
response <- buildResponseForApprover(serviceCompleteBatch).toBatchResult
} yield response
```

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
britneywright: added the todo here: https://github.com/vinyldns/vinyldns/pull/823/files#diff-3d42091f87b70df0dd77babb5a2a5a8fR489

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -493,6 +495,13 @@ class BatchChangeService(
       UnknownConversionError("Cannot convert a rejected batch change").toLeftBatchResult
   }
 
+  def buildResponseForApprover(
+      batchChange: BatchChange): Either[BatchChangeErrorResponse, BatchChange] =
+    batchChange.approvalStatus match {
+      case ManuallyApproved => batchChange.asRight
+      case _ => BatchChangeFailedApproval(batchChange.changes).asLeft
```
pauljamescleary: the existing code in `buildResponseForApprover` returns this if errors still exist - `InvalidBatchChangeResponses(batchChangeInput.changes, transformed).asLeft`

Shouldn't we return the same result?  This bubbles up into Routing and ultimately API clients as well

britneywright: I may need to double check this but I think I changed the return because at the point in the flow the types are different. In this PR the changes are SingleChanges at this point, whereas before we were failing while the changes were still ChangeInputs

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -440,32 +440,33 @@ class BatchChangeService(
     }
   }
 
-  def buildResponseForApprover(
+  def rebuildBatchChangeForUpdate(
```
mitruly: Can we rename this method for clarity? Something like `updateBatchChangeForApproval` is clearer to me.

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -440,32 +440,33 @@ class BatchChangeService(
...
       existingBatchChange: BatchChange,
-      batchChangeInput: BatchChangeInput,
       transformed: ValidatedBatch[ChangeForValidation],
-      reviewInfo: BatchChangeReviewInfo): Either[BatchChangeErrorResponse, BatchChange] =
-    if (transformed.forall(_.isValid)) {
-      val changes = transformed.getValid.zip(existingBatchChange.changes).map {
-        case (newValidation, existing) => newValidation.asStoredChange(Some(existing.id))
-      }
+      reviewInfo: BatchChangeReviewInfo): BatchResult[BatchChange] = {
```
mitruly: I still think that having a return signature of `BatchResult[BatchChange]` is misleading since I would think that there is a way for us to get an `Left` response, which isn't possible in this method.

This can simply return a `BatchChange`.

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
britneywright: `updateBatchChangeForApproval` isn't accurate. The batch change may or may not be approved, we're rebuilding the batch change with errors or with approval attributes to save further along in the process.

modules/api/src/test/scala/vinyldns/api/domain/batch/BatchChangeServiceSpec.scala
```diff
@@ -2264,4 +2260,55 @@ class BatchChangeServiceSpec
     }
   }
 
+  "buildResponseForApprover" should {
+    "return batch change has ManuallyApproved approval status" in {
```
mitruly: `return batch change with ManuallyApproved approval status`

PR #825, opened 2019-08-30
Revalidate scheduled batch changes

britneywright:
working out the idea of how to revalidate a batch change that is scheduled before its scheduled date and time. Currently on batch change approval we validate the scheduled time is in the past (and hard fail if it's not) before we validate the SingleChanges in the batch.

Changes in this pull request:
In the batch change approval flow:
  - move the scheduled time validation after the SingleChange validations
  - If there are still SingleChange errors on the scheduled batch change after the revalidation, return `BatchChangeFailedApproval`.
  - If there's no SingleChange errors but the scheduled time is still in the future return `ScheduledChangeNotDue`.
  - If there are no errors and the scheduled time is past, then the approval is accepted.
 - add a `revalidate` endpoint that only revalidates a scheduled batch change, does not attempt to approve the zone.
- add a Revalidate button to scheduled DNS Changes in the portal

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
@@ -452,7 +452,10 @@ class BatchChangeService(
         }
     }
 
-    if (transformed.forall(_.isValid)) {
+    val scheduledTimePastOrNonExistent =
+      existingBatchChange.scheduledTime.forall(t => t.isBeforeNow)
+
+    if (transformed.forall(_.isValid) && scheduledTimePastOrNonExistent) {
       existingBatchChange
         .copy(
           changes = changes,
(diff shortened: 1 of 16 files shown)


modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -498,6 +501,10 @@ class BatchChangeService(
       batchChange: BatchChange): Either[BatchChangeErrorResponse, BatchChange] =
     batchChange.approvalStatus match {
       case ManuallyApproved => batchChange.asRight
+      case _
```
mitruly: I am concerned about the way that this is being implemented, because the revalidating behavior solely happens if a pending review batch change is scheduled and approved prior to the scheduled time. I feel that revalidation should be more explicit than just approving before something goes through.

Would like some feedback from @pauljamescleary, but believe we should consider either having a 1. explicit endpoint at `zones/batchrecordchanges/:batchId/validate` or 2. at the very least have a parameter on the approval endpoint for `validateOnly` which _will not_ submit if it goes through.

Having time dependent-behavior (e.g. exact same approval request fails one second and passes one second later) isn't preferable imo.

pauljamescleary: We have to look at the workflow design spec - https://github.com/vinyldns/vinyldns/issues/839
I would use that as a guide for implementation

remerle:
This PR is a year old. Considering abandoned.

PR #819, opened 2019-08-23
[DeleteRecord] Refactor change validation map

mitruly:
Pre-requisite to updating `DeleteRecord` validations.

Refactors `ChangeForValidationMap` to consolidate information to use for `DeleteRecord` validation logic and batch change converter.

Changes in this pull request:
- Refactor `ChangeForValidationMap`. Mapping of `RecordKey` is no longer to applicable `ChangeForValidation`s, but instead to consolidated record data.
- New `LogicalChangeType` attribute added to tag the net type of DNS change resulting from all of the batch inputs and existing DNS records. Possible values are `Add`, `FullDelete` (i.e. record will no longer exist), `Update` and `NotEditedInBatch` (should never hit this case)
- Update unit tests

Changes in validation (following) PR:
- Update access validations for delete record
- Update validations to make sure that delete requests are valid

PR #826, opened 2019-09-03
Refactor cvm

rebstar6:
replaces #819, just handing minor comments

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
@@ -240,7 +240,7 @@ class BatchChangeValidations(
     // Updates are a combination of an add and delete for a record with the same name and type in a zone.
     changeGroups.changes.mapValid {
       case addUpdate: AddChangeForValidation
-          if changeGroups.getChangeForValidationDeletes(addUpdate.recordKey).nonEmpty =>
+          if changeGroups.containsValidDeleteChanges(addUpdate.recordKey) =>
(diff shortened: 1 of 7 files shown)


modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
```diff
@@ -437,27 +438,25 @@ class BatchChangeValidations(
...
-      recordTypeFilter: Option[RecordType]): SingleValidation[Unit] = {
-    val existingRecordTypesMatch = recordTypeFilter match {
-      case Some(recordType) =>
-        groupedChanges.existingRecordSets
-          .get(RecordKey(zoneId, recordName, recordType))
-          .map(_.typ)
-          .toList
-      case None =>
-        groupedChanges.existingRecordSets.getRecordSetMatch(zoneId, recordName).map(_.typ)
+      checkingCname: Boolean): SingleValidation[Unit] = {
```
britneywright: why not pass in the record type instead of a Boolean or an Option? The naming here is certainly more clear than previous iterations but this seems most clear to me: 

```
val conflictingExitingTypes = recordType match {
 case CNAME => groupedChanges.existingRecordSets.getRecordSetMatch(zoneId, recordName).map(_.typ)
 case _ => groupedChanges.getExistingRecordSet(RecordKey(zoneId, recordName, CNAME)).map(_.typ).toList
}
```

rebstar6: see update

PR #827, opened 2019-09-03
dont assign ownership to configured list

rebstar6:
untested, need to add.

this _should_ skip assigning ownership to records that are within a configured list. note that if the records already have an assignment, that wont be cleared.

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeConverter.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeConverter.scala
@@ -28,13 +28,20 @@ import vinyldns.api.domain.batch.BatchTransformations.{
   ExistingZones
 }
 import vinyldns.api.domain.record.RecordSetChangeGenerator
+import vinyldns.api.domain.zone.ZoneRecordValidations
 import vinyldns.core.domain.record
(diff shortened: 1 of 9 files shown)


modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeConverter.scala
```diff
@@ -256,7 +266,37 @@ class BatchChangeConverter(batchChangeRepo: BatchChangeRepository, messageQueue:
...
+    if (zoneIsShared && existingOwnerGroupId.isEmpty) {
+      inputOwnerGroupId.flatMap { og =>
+        typ match {
+          case RecordType.PTR =>
+            val shouldSkip = records
+              .collect {
+                case rd: PTRData => rd.ptrdname
+              }
+              .exists {
+                ZoneRecordValidations.isStringInRegexList(batchAccessSkippedFqdnList, _)
```
rebstar6: checking the PTR data on the reverse side. its a little meh, inconsistent with the way weve handled reverse in the past.

note that if any record matches, thats considered enough to skip the assignment (if its multirecord)

britneywright:
Closing in favor of the approach in #829 and #830

PR #829, opened 2019-09-04
global acl starter kit

rebstar6:
need to change tests for those to compile. also needs new tests

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/Boot.scala
+++ b/modules/api/src/main/scala/vinyldns/api/Boot.scala
@@ -87,6 +87,7 @@ object Boot extends App {
       restHost <- IO(VinylDNSConfig.restConfig.getString("host"))
       restPort <- IO(VinylDNSConfig.restConfig.getInt("port"))
       batchChangeLimit <- IO(VinylDNSConfig.vinyldnsConfig.getInt("batch-change-limit"))
+      globalAcls <- VinylDNSConfig.globalAcl
       syncDelay <- IO(VinylDNSConfig.vinyldnsConfig.getInt("sync-delay"))
       msgsPerPoll <- IO.fromEither(MessageCount(queueConfig.messagesPerPoll))
       healthCheckTimeout <- VinylDNSConfig.healthCheckTimeout
@@ -110,23 +111,25 @@ object Boot extends App {
         )
         .start
     } yield {
+      val batchAccessValidations = new AccessValidations(globalAcls)
+      val recordAccessValidations = new AccessValidations()
       val zoneValidations = new ZoneValidations(syncDelay)
       val batchChangeValidations = new BatchChangeValidations(
         batchChangeLimit,
-        AccessValidations,
+        batchAccessValidations,
         VinylDNSConfig.multiRecordBatchUpdateEnabled,
         VinylDNSConfig.scheduledChangesEnabled
       )
       val membershipService = MembershipService(repositories)
       val connectionValidator =
         new ZoneConnectionValidator(connections)
-      val recordSetService = RecordSetService(repositories, messageQueue, AccessValidations)
+      val recordSetService = RecordSetService(repositories, messageQueue, recordAccessValidations)
       val zoneService = ZoneService(
         repositories,
         connectionValidator,
         messageQueue,
         zoneValidations,
-        AccessValidations)
+        recordAccessValidations)
       val healthService = new HealthService(
         messageQueue.healthCheck :: connectionValidator.healthCheck(healthCheckTimeout) ::
           loaderResponse.healthChecks)
--- a/modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
+++ b/modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
@@ -25,6 +25,7 @@ import vinyldns.api.crypto.Crypto
 import com.comcast.ip4s._
 import net.ceedubs.ficus.Ficus._
 import net.ceedubs.ficus.readers.EnumerationReader._
+import vinyldns.api.domain.access.{GlobalAcl, SystemGlobalAcls}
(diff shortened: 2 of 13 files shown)


modules/api/src/main/scala/vinyldns/api/Boot.scala
```diff
@@ -110,23 +111,25 @@ object Boot extends App {
         )
         .start
     } yield {
+      val batchAccessValidations = new AccessValidations(globalAcls)
+      val recordAccessValidations = new AccessValidations()
```
rebstar6: this is probably not the move long term. if you want though, keeping this empty should ensure that you skip this new access in the recordset interface and just use in batch

modules/api/src/main/scala/vinyldns/api/domain/access/AccessValidations.scala
```diff
@@ -190,6 +191,8 @@ object AccessValidations extends AccessValidationAlgebra {
     case support if support.isSystemAdmin =>
       val aclAccess = getAccessFromAcl(auth, recordName, recordType, zone)
       if (aclAccess == AccessLevel.NoAccess) AccessLevel.Read else aclAccess
+    case globalAclUser if globalAcls.hasGlobalAcl(globalAclUser, recordName, zone) =>
```
rebstar6: with this, someone cant be a super/support user and also have global ACLs. the "my groups" part of auth for these folks is already messy, as i believe we just store all groups for them. in short, if there is any ACL, it will apply to the support users regardless of membership.

modules/api/src/main/scala/vinyldns/api/Boot.scala
pauljamescleary: So, @rebstar6 separated out validations that apply only in batch vs those that apply in normal record set update.

This makes sense for the time being, but ultimately we will need to consolidate all the validations.

modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
```diff
@@ -135,6 +136,9 @@ object VinylDNSConfig {
     .as[Option[Boolean]]("manual-batch-review-enabled")
     .getOrElse(false)
 
+  lazy val globalAcl: IO[SystemGlobalAcls] =
+    loadConfigF[IO, List[GlobalAcl]](vinyldnsConfig, "global-acl-rules").map(SystemGlobalAcls)
+
```
pauljamescleary: This is using pureconfig to load `SystemGlobalAcls` ... to be honest, I think we will rename it just `GlobalAcls`

modules/api/src/main/scala/vinyldns/api/domain/access/AccessValidations.scala
```diff
@@ -48,8 +49,9 @@ object AccessValidations extends AccessValidationAlgebra {
       auth: AuthPrincipal,
       recordName: String,
       recordType: RecordType,
-      zone: Zone): Either[Throwable, Unit] = {
-    val accessLevel = getAccessLevel(auth, recordName, recordType, zone)
+      zone: Zone,
+      recordData: List[RecordData] = List.empty): Either[Throwable, Unit] = {
+    val accessLevel = getAccessLevel(auth, recordName, recordType, zone, None, recordData)
```
pauljamescleary: seems like all the changes here simply pass new arguments into the `getAccessLevel` check

modules/api/src/main/scala/vinyldns/api/domain/access/AccessValidations.scala
```diff
@@ -180,7 +191,8 @@ object AccessValidations extends AccessValidationAlgebra {
       recordName: String,
       recordType: RecordType,
       zone: Zone,
-      recordOwnerGroupId: Option[String] = None): AccessLevel = auth match {
+      recordOwnerGroupId: Option[String] = None,
+      recordData: List[RecordData] = List.empty): AccessLevel = auth match {
```
pauljamescleary: this is where the magic happens.  We are passing in `recordData` so that we can also check the global access list for `PTR` and the `ptrdname`, which is an FQDN that may fall into the global acl list.

modules/api/src/main/scala/vinyldns/api/domain/access/AccessValidations.scala
pauljamescleary: I think I follow what @rebstar6 is saying here.  The pattern match hits the `support` user up above, so the global acl list will not be consulted for support users.

I am not sure why we cannot just have this pattern match above the support user check though.  Can't say I feel strongly either

modules/api/src/main/scala/vinyldns/api/domain/access/AccessValidationsAlgebra.scala
```diff
@@ -35,28 +36,32 @@ trait AccessValidationAlgebra {
       auth: AuthPrincipal,
       recordName: String,
       recordType: RecordType,
-      zone: Zone): Either[Throwable, Unit]
+      zone: Zone,
+      recordData: List[RecordData] = List.empty): Either[Throwable, Unit]
```
pauljamescleary: changes in this file are to add `recordData` to a bunch of methods

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
```diff
@@ -68,7 +68,7 @@ trait BatchChangeValidationsAlgebra {
 
 class BatchChangeValidations(
     changeLimit: Int,
-    accessValidation: AccessValidationAlgebra,
+    accessValidation: AccessValidationsAlgebra,
```
pauljamescleary: changes in this file are also to account for additional params, no logic changes

pauljamescleary:
Closing as taking this over in #830

PR #831, opened 2019-09-06
Set the maxItems to 1000 like list groups

pauljamescleary:
Following the same pattern we did for list groups, where we just set the default page size to 1000 when listing members.

- `service.groups.js` - modified the call to `getGroupMembers` to hard code the page size to 1000
- `service.groups.spec.js` - fixed unit tests
- `TestDataLoader` - added a sample group for dummy users that has all dummy users (200 of them) for simple testing.


pauljamescleary:
image

PR #832, opened 2019-09-06
Edit scheduled time for a batch change

britneywright:
Changes in this pull request:
- allow a creator ~~or system admin~~ to change the scheduledTime of a scheduled batch change
- also explores a different approach through the batch change service for re-validations

### DIFF
--- a/modules/api/functional_test/live_tests/batch/edit_batch_change_test.py
+++ b/modules/api/functional_test/live_tests/batch/edit_batch_change_test.py
@@ -3,9 +3,9 @@ from utils import *
 import datetime
 
 @pytest.mark.manual_batch_review
-def test_edit_batch_change_with_new_scheduled_time_succeeds(shared_zone_test_context):
+def test_edit_batch_change_with_new_scheduled_time_succeeds_for_creator(shared_zone_test_context):
     """
-    Test successfully changing the time of a scheduled batch change
+    Test successfully changing the time of a scheduled batch change if updated by the batch change creator
     """
     client = shared_zone_test_context.ok_vinyldns_client
     original_dt = (datetime.datetime.now() + datetime.timedelta(days=1)).strftime('%Y-%m-%dT%H:%M:%SZ')
@@ -42,3 +42,42 @@ def test_edit_batch_change_with_new_scheduled_time_succeeds(shared_zone_test_con
         if result:
             rejecter = shared_zone_test_context.support_user_client
             rejecter.reject_batch_change(result['id'], status=200)
+
+def test_edit_batch_change_with_new_scheduled_time_fails_for_non_creator(shared_zone_test_context):
+    """
+    Test changing the time of a scheduled batch change fails if not updated by the batch change creator
+    """
+    client = shared_zone_test_context.ok_vinyldns_client
+    support_client = shared_zone_test_context.support_user_client
+    original_dt = (datetime.datetime.now() + datetime.timedelta(days=1)).strftime('%Y-%m-%dT%H:%M:%SZ')
+    new_dt = (datetime.datetime.now() + datetime.timedelta(days=2)).strftime('%Y-%m-%dT%H:%M:%SZ')
+
+    original_batch_change_input = {
+        "comments": "this is optional",
+        "changes": [
+            get_change_A_AAAA_json("parent.com.", address="4.5.6.7"),
+        ],
+        "scheduledTime": original_dt,
+        "ownerGroupId": shared_zone_test_context.ok_group['id']
+    }
+    new_batch_change_input = {
+        "comments": "this is optional",
+        "changes": [
+            get_change_A_AAAA_json("parent.com.", address="4.5.6.7"),
+        ],
+        "scheduledTime": new_dt,
+        "ownerGroupId": shared_zone_test_context.ok_group['id']
+    }
+    result = None
+    try:
+        result = client.create_batch_change(original_batch_change_input, status=202)
+        assert_that(result['status'], 'Scheduled')
+        assert_that(result['scheduledTime'], original_dt)
+
+        error = support_client.edit_batch_change(result['id'], new_batch_change_input, status=403)
+
+        assert_that(error, is_("User does not have access to item " + result['id']))
+
+    finally:
+        if result:
+            support_client.reject_batch_change(result['id'], status=200)
--- a/modules/api/functional_test/live_tests/batch/edit_batch_change_test.py
+++ b/modules/api/functional_test/live_tests/batch/edit_batch_change_test.py
@@ -1,0 +1,44 @@
+from hamcrest import *
+from utils import *
+import datetime
+
+@pytest.mark.manual_batch_review
+def test_edit_batch_change_with_new_scheduled_time_succeeds(shared_zone_test_context):
+    """
+    Test successfully changing the time of a scheduled batch change
+    """
+    client = shared_zone_test_context.ok_vinyldns_client
+    original_dt = (datetime.datetime.now() + datetime.timedelta(days=1)).strftime('%Y-%m-%dT%H:%M:%SZ')
+    new_dt = (datetime.datetime.now() + datetime.timedelta(days=2)).strftime('%Y-%m-%dT%H:%M:%SZ')
+
+    original_batch_change_input = {
+        "comments": "this is optional",
+        "changes": [
+            get_change_A_AAAA_json("parent.com.", address="4.5.6.7"),
+        ],
+        "scheduledTime": original_dt,
+        "ownerGroupId": shared_zone_test_context.ok_group['id']
+    }
+    new_batch_change_input = {
+        "comments": "this is optional",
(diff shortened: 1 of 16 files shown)


modules/api/src/main/scala/vinyldns/api/route/BatchChangeRouting.scala
```diff
@@ -137,6 +137,17 @@ class BatchChangeRoute(
               complete(StatusCodes.OK, chg)
             }
           }
+        } ~
+        path("zones" / "batchrecordchanges" / Segment / "edit") { id =>
```
mitruly: Shouldn't this just be `PUT` to `zones/batchrecordchanges/segment`? This endpoint doesn't feel very RESTful..

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -181,6 +181,36 @@ class BatchChangeService(
...
+        BatchRequesterNotFound(batchChange.userId, batchChange.userName)
+      )
+      _ <- validateScheduledBatchChangeEdit(
+        batchChange,
+        batchChangeInput.scheduledTime,
+        authPrincipal,
+        requesterAuth.isTestUser).toBatchResult
+      rescheduledBatchChange <- updateScheduledTime(batchChange, batchChangeInput.scheduledTime)
+      asInput = BatchChangeInput(rescheduledBatchChange)
+      thing <- applyBatchChange(asInput, authPrincipal, true)
```
mitruly: `thing` is too descriptive here 🙃

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
```diff
@@ -120,27 +128,40 @@ class BatchChangeValidations(
...
       batchChange,
       authPrincipal)
 
+  def validateScheduledBatchChangeEdit(
+      batchChange: BatchChange,
+      scheduledTime: Option[DateTime],
+      authPrincipal: AuthPrincipal,
+      isTestChange: Boolean): Either[BatchChangeErrorResponse, Unit] =
+    validateBatchChangePendingReview(batchChange, "edited") |+|
+      validateScheduledTimeExists(batchChange, scheduledChangesEnabled) |+|
```
mitruly: Does a scheduled time need to exist? A possible case is someone submitted a batch change that is pending review and want to switch it to scheduled. Maybe the converse is true.

Is this a stop gap for users being able to edit a batch? My concern is that this functionality seems to be focused on strictly editing an existing scheduled time that may be expanded in the future. Since we're not reapplying the batch change validation flow, we are not checking for `manualReviewEnabled` flag, either.

modules/api/src/test/scala/vinyldns/api/route/BatchChangeJsonProtocolSpec.scala
```diff
@@ -285,9 +315,25 @@ class BatchChangeJsonProtocolSpec
...
+      val resultChangesIds = result.changes.map(_.id)
+
+      val batchChangeResult = BatchChangeInput(
+        None,
+        List(
+          deleteAChangeInput.copy(id = resultChangesIds.head),
+          addAChangeInput.copy(id = resultChangesIds(1)),
+          addAAAAChangeInput.copy(id = resultChangesIds(2)),
+          addCNAMEChangeInput.copy(id = resultChangesIds(3)),
+          addPTRChangeInput.copy(id = resultChangesIds(4)),
```
mitruly: @britneywright this trailing comma is what is causing the scalastyle error.

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
```diff
@@ -64,6 +65,12 @@ trait BatchChangeValidationsAlgebra {
   def validateBatchChangeCancellation(
       batchChange: BatchChange,
       authPrincipal: AuthPrincipal): Either[BatchChangeErrorResponse, Unit]
+
+  def validateScheduledBatchChangeEdit(
```
mitruly: I'm not a fan of "edit" being used over "update" as it is inconsistent with our other endpoints.

modules/api/functional_test/live_tests/batch/edit_batch_change_test.py
```diff
@@ -0,0 +1,83 @@
...
+
+        assert_that(retrieved_batch_change['status'], 'Scheduled')
+        assert_that(retrieved_batch_change['scheduledTime'], new_dt)
+
+    finally:
+        if result:
+            rejecter = shared_zone_test_context.support_user_client
+            rejecter.reject_batch_change(result['id'], status=200)
+
+def test_edit_batch_change_with_new_scheduled_time_fails_for_non_creator(shared_zone_test_context):
```
pauljamescleary: be sure to have 2 lines between test functions.  We don't have an auto formatter yet for our python tests (we really should).  All functions should have 2 blank lines between them.  In general, if you use IntelliJ, it will complain.  Can also auto-format the file

modules/api/functional_test/vinyldns_python.py
```diff
@@ -611,6 +611,17 @@ def cancel_batch_change(self, batch_change_id, **kwargs):
         _, data = self.make_request(url, u'POST', self.headers, **kwargs)
         return data
 
+    def edit_batch_change(self, batch_change_id, batch_change_input, **kwargs):
```
pauljamescleary: Note: classes (not standalone functions) leave 1 blank line between methods

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
pauljamescleary: I agree that we should not concern ourselves with the disposition of the scheduled time.  A `PUT` carries semantic that indicate "just make the old values these new values".  Certain validations are correct (like making sure the person can update the change), but it is not needed in this instance.

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
```diff
@@ -570,15 +591,23 @@ class BatchChangeValidations(
...
     }
 
+  def validateScheduledTimeExists(
+      batchChange: BatchChange,
+      scheduledChangesEnabled: Boolean): Either[BatchChangeErrorResponse, Unit] =
+    (scheduledChangesEnabled, batchChange.scheduledTime) match {
+      case (true, Some(_)) => Right(())
+      case (_, None) => Left(BatchChangeNotScheduled(batchChange.id))
+      case (false, _) => Left(ScheduledChangesDisabled)
+    }
```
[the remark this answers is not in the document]
pauljamescleary: If we regularly miss new lines and other formatting details, let's update our build / scalafmt to auto correct those things

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
britneywright: I thought we discussed previously that we only want to update existing scheduled batch changes, part of the rationale was uncertainty around timing of editing batch changes that are only pending review. And we were kicking that can down the road and putting this as an initial step towards more editing abilities for batch changes.

mitruly: In light of @pauljamescleary's comment, think it makes more sense for the service to go through `applyBatchChangeValidationFlow`.

For the time being (until we can actually use batch change input), we can create an `UpdateBatchChangeInput` that only accepts values that we are allowing to update (see `UpdateZoneInput` in `Zone.scala`).

britneywright: It already goes through `applyBatchChangeValidationFlow` when it goes through `applyBatchChange`

mitruly: I mistyped: `validateScheduledBatchChangeEdit` is calling `validateScheduledChange`, which is also being invoked in the `applyBatchChangeValidationFlow` validation flow so that specifically should not be here. Similarly, since we are saying that it makes logical sense to allow a user to remove a scheduled time from their batch, then we can also get rid of `validateScheduledTimeExists`; if we decide to keep it, this check should be moved and consolidated with the existing scheduled change check.

We should not be duplicating the same validation at different points in the service flow.

britneywright: `validateScheduledTimeExists` is to validate that they are attempting to edit a batch change that is scheduled. The edit can include a time or not include a time, that's a separate validation.

remerle:
Closing as out-of-date, original author MIA

PR #838, opened 2019-09-14
Par func tests

pauljamescleary:
Major overhaul of func tests to allow them to run in parallel.  Major changes include:

1. Consolidate all separate test fixtures into a single test fixture in the `shared_zone_test_context`
1. Add `xdist` to allow running tests in parallel
1. Add hooks in main `conftest.py` to setup the test fixture before workers run, and tear it down when workers are finished
1. After fixture is setup, save state in a local `tmp.out` so the workers will use that state instead of trying to recreate the fixture.
1. Add a `utils.generate_record_name` which generates a unique record name in order to avoid conflicts when running tests in parallel
1. Add a `pytest.mark.serial` for func tests that just cannot be run in serial
1. Tests are now run in two phases, first we run in parallel, and if that is successful, we run the serial tests
1. Add a `--teardown` flag, this allows us to reuse the test fixture between the two phases parallel and serial


bin/func-test-api-travis.sh
```diff
@@ -27,13 +27,13 @@ echo "Copy the vinyldns.jar to the api docker folder so it is in context..."
...
     cd $DIR/../
-    sbt api/clean api/assembly
+    sbt build-api
     cd $DIR
 fi
 cp -f $DIR/../modules/api/target/scala-2.12/vinyldns.jar $WORK_DIR/docker/api
 
 echo "Starting docker environment and running func tests..."
-docker-compose -f $WORK_DIR/docker/docker-compose-func-test.yml --project-directory $WORK_DIR/docker up --exit-code-from functest
+docker-compose -f $WORK_DIR/docker/docker-compose-func-test.yml --project-directory $WORK_DIR/docker up --build --exit-code-from functest
```
pauljamescleary: this is necessary when trying to test locally, otherwise the func tests will pull an old docker image from your local workstation

build.sbt
```diff
@@ -530,6 +530,6 @@ addCommandAlias("verify", "; project root; killDocker; " +
 )
 
 // Build the artifacts for release
-addCommandAlias("build-api", ";project api;clean;assembly")
-addCommandAlias("build-portal", ";project portal;clean;preparePortal;dist")
+addCommandAlias("build-api", ";project api;clean;coverageOff;assembly")
+addCommandAlias("build-portal", ";project portal;clean;coverageOff;preparePortal;dist")
```
pauljamescleary: not sure why my version of sbt hit this, but it is a good idea to make sure when we build to disable scoverage otherwise we could hit runtime issues.

docker/bind9/etc/named.conf.local
```diff
@@ -187,3 +187,9 @@ zone "zone.requires.review" {
 	file "/var/bind/zone.requires.review.hosts";
 	allow-update { key "vinyldns."; };
 	};
+
+zone "list-records" {
```
pauljamescleary: we used to actually create records for the list record tests.  It is much better to have them already exist in a zone, this will cut down on execution times

docker/docker-compose-func-test.yml
```diff
@@ -7,6 +7,8 @@ services:
     container_name: "vinyldns-mysql"
     ports:
       - "19002:3306"
+    logging:
```
pauljamescleary: disable logging on other containers we don't care about, makes parsing the func test runs easier.

docker/functest/run.sh
```diff
@@ -43,12 +43,29 @@ echo "Running live tests against ${VINYLDNS_URL} and DNS server ${DNS_IP}"
 
 cd /app
 
+# Cleanup any errant cached file copies
+find . -name "*.pyc" -delete
+find . -name "__pycache__" -delete
+
```
pauljamescleary: this was a change needed to run the func-test-api.sh script locally, gotta remove the cached files

docker/functest/run.sh
```diff
@@ -43,12 +43,29 @@ echo "Running live tests against ${VINYLDNS_URL} and DNS server ${DNS_IP}"
...
 # If PROD_ENV is not true, we are in a local docker environment so do not skip anything
 if [ "${PROD_ENV}" = "true" ]; then
     # -m plays havoc with -k, using variables is a headache, so doing this by hand
-    echo "./run-tests.py live_tests -m \"not skip_production\" -v --url=${VINYLDNS_URL} --dns-ip=${DNS_IP} ${TEST_PATTERN}"
-    ./run-tests.py live_tests -v -m "not skip_production" --url=${VINYLDNS_URL} --dns-ip=${DNS_IP} ${TEST_PATTERN}
+    # run parallel tests first (not serial)
```
pauljamescleary: below, we have to run in two phases.  The first phase runs in parallel (-n2), and the second runs tests marked with `pytest.mark.serial`

modules/api/functional_test/conftest.py
```diff
@@ -33,8 +29,8 @@ def pytest_addoption(parser):
                      help="OAuth credentials in consumer:secret format")
     parser.addoption("--environment", dest="cim_env", action="store", default="test",
                      help="CIM_ENV that we are testing against.")
-    parser.addoption("--log-level", dest="logging_level",
-                     help="logging level should be CRITICAL, ERROR, WARNING, INFO or DEBUG")
+    parser.addoption("--teardown", dest="teardown", action="store", default="True",
```
pauljamescleary: allows us to share the test fixture file across parallel and serial (as opposed to creating the test fixture twice, once for parallel and once for serial)

modules/api/functional_test/conftest.py
```diff
@@ -61,10 +57,39 @@ def pytest_configure(config):
...
-                               config.getoption("dns_key"),
-                               config.getoption("url"))
+                                  config.getoption("dns_zone"),
+                                  config.getoption("dns_key_name"),
+                                  config.getoption("dns_key"),
+                                  config.getoption("url"),
+                                  config.getoption("teardown"))
+
+    from shared_zone_test_context import SharedZoneTestContext
+    if not hasattr(config, 'workerinput'):
```
pauljamescleary: important!  we check if we are master by checking the `workerinput`, if we are master we setup the test fixture as it can only be setup ONCE!!

modules/api/functional_test/conftest.py
```diff
@@ -61,10 +57,39 @@ def pytest_configure(config):
...
+            SharedZoneTestContext('tmp.out')
+        else:
+            print 'No fixture file found, loading a new test fixture'
+            ctx = SharedZoneTestContext()
+            ctx.out_fixture_file("tmp.out")
+    else:
+        print 'This is a worker'
+
+
+def pytest_unconfigure(config):
```
pauljamescleary: Here we teardown, if we are master AND `--teardown=true`, then we will teardown the fixture

modules/api/functional_test/live_tests/batch/list_batch_change_summaries_test.py
```diff
@@ -1,112 +1,16 @@
+import pytest
 from hamcrest import *
 from utils import *
-import time
-import pytest
-from vinyldns_python import VinylDNSClient
 from vinyldns_context import VinylDNSTestContext
+from vinyldns_python import VinylDNSClient
```
pauljamescleary: Moved the test fixture out of the file ... this is the new pattern

modules/api/functional_test/live_tests/conftest.py
```diff
@@ -1,28 +1,7 @@
...
-
-    context = ZoneHistoryContext()
-
-    def fin():
-        context.tear_down()
-
-    request.addfinalizer(fin)
-
-    return context
+    return SharedZoneTestContext("tmp.out")
```
pauljamescleary: the fixture in the nested module ALWAYS loads from the fixture file `tmp.out`; the `tmp.out` is created in the higher directory's `conftest.py`

modules/api/functional_test/live_tests/internal/status_test.py
```diff
@@ -20,6 +20,8 @@ def test_get_status_success(shared_zone_test_context):
     assert_that(result['keyName'], not_none())
     assert_that(result['version'], not_none())
 
+
+@pytest.mark.serial
```
pauljamescleary: geez, don't accidentally flip processing while other tests are running in parallel!

modules/api/functional_test/live_tests/list_batch_summaries_test_context.py
```diff
@@ -0,0 +1,106 @@
+import time
+from hamcrest import *
+from utils import *
+from vinyldns_context import VinylDNSTestContext
+from vinyldns_python import VinylDNSClient
+
+
+class ListBatchChangeSummariesTestContext():
```
pauljamescleary: this is really just moving the test fixture out of the file and into it's own file at the top level (new pattern)

modules/api/functional_test/live_tests/membership/get_group_changes_test.py
```diff
@@ -1,84 +1,53 @@
...
-    def fin():
-        if created_group:
-            client.delete_group(created_group['id'], status=(200,404))
-
-    request.addfinalizer(fin)
-
     return {
-        'created_group': created_group,
-        'updated_groups': updated_groups
+        'created_group': shared_zone_test_context.group_activity_created,
```
pauljamescleary: please don't look at this file, it is not pleasant to follow

modules/api/functional_test/requirements.txt
```diff
@@ -6,9 +6,10 @@
...
 pyhamcrest==1.8.0
 pytz>=2014
-pytest==2.6.4
+pytest==4.4.1
 mock==1.0.1
 dnspython==1.14.0
 boto==2.48.0
 future==0.17.0
 requests==2.20.0
+pytest-xdist==1.29.0
```
pauljamescleary: dependencies needed to be updated

modules/api/src/main/scala/vinyldns/api/repository/TestDataLoader.scala
```diff
@@ -130,6 +130,18 @@ object TestDataLoader {
     isTest = true
   )
 
+  final val listRecordsUser = User(
```
pauljamescleary: created a new user to segregate list records from all other record operations, isolation enabled parallelism!

modules/api/functional_test/vinyldns_python.py
```diff
@@ -26,7 +26,7 @@
 
 __all__ = [u'VinylDNSClient', u'MAX_RETRIES', u'RETRY_WAIT']
 
-MAX_RETRIES = 30
+MAX_RETRIES = 40
```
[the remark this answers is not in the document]
pauljamescleary: Yep, we are dialed way down in DEV and now that we can run in parallel I have seen occasional timeouts waiting for things.

modules/api/functional_test/live_tests/batch/list_batch_change_summaries_test.py
```diff
@@ -1,112 +1,16 @@
...
-            assert_that(summary["totalChanges"], equal_to(self.completed_changes[i + start_from]["totalChanges"]))
-            assert_that(summary["status"], equal_to(self.completed_changes[i + start_from]["status"]))
-            assert_that(summary["id"], equal_to(self.completed_changes[i + start_from]["id"]))
-            assert_that(summary["approvalStatus"], equal_to("AutoApproved"))
-            assert_that(summary, is_not(has_key("reviewerId")))
-
-
-@pytest.fixture(scope = "module")
+@pytest.fixture(scope="module")
 def list_fixture(request, shared_zone_test_context):
```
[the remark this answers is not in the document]
pauljamescleary: Let me make sure it still works if removed

PR #856, opened 2019-09-30
remove access validation from get zone by name service

britneywright:
The terraform provider doesn't support batch change so users need to make records in shared zones through the recordset API endpoint. The recordset endpoint requires a zone ID, so terraform clients need to be able to get a zone by name to find out the zone id. 

This is a read-only endpoint so there's not a need for access validations based on zone ownership or system administration. There's also already initial validations at the routing endpoint to ensure locked users and users without access to the system cannot make the request.
 
Changes in this pull request:
- remove access validation from getZoneByName method in the Zone Service

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneService.scala
@@ -126,7 +126,6 @@ class ZoneService(
   def getZoneByName(zoneName: String, auth: AuthPrincipal): Result[ZoneInfo] =
     for {
       zone <- getZoneByNameOrFail(ensureTrailingDot(zoneName))
-      _ <- canSeeZone(auth, zone).toResult
       aclInfo <- getZoneAclDisplay(zone.acl)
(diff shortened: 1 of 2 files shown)


modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneService.scala
```diff
@@ -126,7 +126,6 @@ class ZoneService(
   def getZoneByName(zoneName: String, auth: AuthPrincipal): Result[ZoneInfo] =
     for {
       zone <- getZoneByNameOrFail(ensureTrailingDot(zoneName))
-      _ <- canSeeZone(auth, zone).toResult
```
pauljamescleary: we had zero unit tests?

britneywright: we have unit tests for get zone by name, but we did not have one for the access validation failure case.

mdb:
For record keeping purposes, this is the corresponnding `terraform-provider-vinyldns` issue:
https://github.com/vinyldns/terraform-provider-vinyldns/issues/60

PR #859, opened 2019-10-03
Support non ActiveDirectory LDAP

pauljamescleary:
Support non ActiveDirectory LDAP

This PR has an optional local portal setup against this docker container - https://github.com/rroemhild/docker-test-openldap

The base modifications for LDAP was to change the actual authentication flow.  Before, we only attempted to bind (setting up a DirContext and relying on an exception).  We would test all of the search bases until we exhausted the list.

The new approach works differently:

1. First, login using the main service account
2. Second, do a lookup of the user
3. Finally, attempt to bind to that user's context directly using the password provided.

This works fine with both AD LDAP as well as the example docker container which uses OpenLDAP

Besides these changes, need to make configurable the userNameField, which is the ldap attribute that is used to search for the username sent in the login screen.  In AD, this is `sAMAccountName`, but in the example it is `uid`, the logon field is up to the way LDAP is setup

- `docker-up-vinyldns.sh` - fixed a quote issue with the startup script to properly send in the version of vinyldns
- `docker-compose-build.yml` - added the `ldap` container so the portal can connect as `vinyldns-ldap`
- `docker/portal/application.conf` - new config file so that we can connect to the new ldap container
- `docker-compose.yml` - added the `ldap` container here as well so we can play with it using `reStart` in sbt instead of `docker-up-vinyldns.sh` - simplifies local testing
- `LdapAuthenticator.scala` - this is where the main changes happen

### TO TEST DOCKER

**publish docker locally**
`> sbt docker:publish`

_This should publish 0.9.4-SNAPSHOT localy_

**start up the docker vinyldns including portal**
`> bin/docker-up-vinyldns.sh --version 0.9.4-SNAPSHOT`

**login to the portal once started up**
User is **professor** **professor**

### TO TEST LOCAL
To test this PR, check it out and do the following...

**Update your local.conf**

```
LDAP {
  user = "cn=admin,dc=planetexpress,dc=com"
  password = "GoodNewsEveryone"
  domain = ""
  userNameAttribute = "uid"
  searchBase = [
    {organization = "", domainName = "ou=people,dc=planetexpress,dc=com"},
  ]
  context {
    initialContextFactory = "com.sun.jndi.ldap.LdapCtxFactory"
    securityAuthentication = "simple"
    providerUrl = "ldap://localhost:19008"
  }

  user-sync {
    enabled = false
    hours-polling-interval = 1
  }
}
```

**Start up the API**

1. start `sbt`
2. go to api project `sbt> project api`
3. start up docker `sbt> dockerComposeUp`
4. start up api `sbt> reStart`
5. go to portal project `sbt> project portal`
6. run the portal `sbt> run`
7. Wait until ready, then go to http://localhost:9001
8. Login with **professor** **professor**

### DIFF
--- a/docker/portal/application.conf
+++ b/docker/portal/application.conf
@@ -9,7 +9,7 @@ LDAP {
   domain = ""
 
   # This will be the name of the LDAP field that carries the user's login id (what they enter in the username in login form)
-  userNameField = "uid"
+  userNameAttribute = "uid"
 
   # For ogranization, leave empty for this demo, the domainName is what matters, and that is the LDAP structure
   # to search for users that require login
--- a/docker/portal/application.conf
+++ b/docker/portal/application.conf
@@ -1,0 +1,59 @@
+LDAP {
+  # For OpenLDAP, this would be a full DN to the admin for LDAP / user that can see all users
+  user = "cn=admin,dc=planetexpress,dc=com"
+
+  # Password for the admin account
+  password = "GoodNewsEveryone"
+
+  # Keep this as an empty string for OpenLDAP
+  domain = ""
+
+  # This will be the name of the LDAP field that carries the user's login id (what they enter in the username in login form)
+  userNameField = "uid"
+
+  # For ogranization, leave empty for this demo, the domainName is what matters, and that is the LDAP structure
+  # to search for users that require login
+  searchBase = [
+    {organization = "", domainName = "ou=people,dc=planetexpress,dc=com"},
+  ]
+  context {
+    initialContextFactory = "com.sun.jndi.ldap.LdapCtxFactory"
+    securityAuthentication = "simple"
+
+    # Note: The following assumes a purely docker setup, using container_name = vinyldns-ldap
+    providerUrl = "ldap://vinyldns-ldap:389"
+  }
+
+  # This is only needed if keeping vinyldns user store in sync with ldap (to auto lock out users who left your
+  # company for example)
+  user-sync {
+    enabled = false
+    hours-polling-interval = 1
+  }
+}
+
+# Note: This MUST match the API or strange errors will ensure, NoCrypto should not be used for production
+crypto {
+  type = "vinyldns.core.crypto.NoOpCrypto"
+}
+
+http.port = 9001
+
+data-stores = ["mysql"]
+
+# Note: The default mysql settings assume a local docker compose setup with mysql named vinyldns-mysql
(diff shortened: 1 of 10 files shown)


docker/portal/application.conf
```diff
@@ -0,0 +1,59 @@
+LDAP {
+  # For OpenLDAP, this would be a full DN to the admin for LDAP / user that can see all users
+  user = "cn=admin,dc=planetexpress,dc=com"
+
+  # Password for the admin account
+  password = "GoodNewsEveryone"
+
+  # Keep this as an empty string for OpenLDAP
+  domain = ""
```
pauljamescleary: note: if you omit this it will inherit the default from `reference.conf`, and we EXPLICITLY want this to be empty, it is only used for ActiveDirectory

docker/portal/application.conf
```diff
@@ -0,0 +1,59 @@
...
+
+  # This will be the name of the LDAP field that carries the user's login id (what they enter in the username in login form)
+  userNameField = "uid"
+
+  # For ogranization, leave empty for this demo, the domainName is what matters, and that is the LDAP structure
+  # to search for users that require login
+  searchBase = [
+    {organization = "", domainName = "ou=people,dc=planetexpress,dc=com"},
```
pauljamescleary: For non ActiveDirectory, this is empty as well

modules/portal/app/controllers/LdapAuthenticator.scala
```diff
@@ -117,36 +118,63 @@ object LdapAuthenticator {
...
         searchDomain: LdapSearchDomain,
         username: String,
-        password: String): Either[LdapException, LdapUserDetails] =
-      createContext(s"${searchDomain.organization}\\$username", password).flatMap { context =>
-        searchContext(context, searchDomain.organization, username)
-      }
+        password: String): Either[LdapException, LdapUserDetails] = {
+
+      // Login as the service account
+      val qualifiedName =
```
pauljamescleary: you may wonder about this check, whether the domain is set.  In Active directory, you prepend the domain to the user like so `domain\userName`; however, in OpenLDAP, you do not do that, but instead pass in the DN of the user like `cn=foo,dc=example,dc=com` and omit the prefix.

The reason to do this is to make sure that configuration is backward compatible.  In theory, there should be an AD specific config section but that would break backward compatibility

pauljamescleary:
Verified this in a lower environment as well

PR #862, opened 2019-10-03
Add logging around sent notifications in the API

### LINKED ISSUE #852: Add Logging around sent notifications in the API
**Motivation**
There are no log statements when we emit notifications from the API. 

**Scope of change**

- `AllNotifiers` - wrap the individual `notify` `IO` with a `monitor` function

```scala
final case class AllNotifiers(...) extends Monitored {
...
  def notify(notifier: Notifier, notification: Notification[_]): IO[Unit] =
    monitor(notifier.getClass.getSimpleName) {
      notifier.notify(notification).handleErrorWith { e =>
        IO { logger.error(s"Notifier ${notifier.getClass.getSimpleName} failed.", e) }
      }
  }
...
```

ericcheatham:
Fixes #852 

Changes in this pull request:
- Adds logging to notification events by wrapping `notify.io` with a `monitor` class.

### DIFF
--- a/modules/core/src/main/scala/vinyldns/core/notifier/AllNotifiers.scala
+++ b/modules/core/src/main/scala/vinyldns/core/notifier/AllNotifiers.scala
@@ -35,7 +35,7 @@ final case class AllNotifiers(notifiers: List[Notifier])(implicit val cs: Contex
     monitor(notifier.getClass.getSimpleName) {
       notifier.notify(notification).handleErrorWith { e =>
         IO {
-          logger.error(s"Notifi er ${notifier.getClass.getSimpleName} failed.", e)
+          logger.error(s"Notifier ${notifier.getClass.getSimpleName} failed.", e)
         }
       }
     }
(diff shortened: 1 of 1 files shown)


modules/core/src/main/scala/vinyldns/core/notifier/AllNotifiers.scala
```diff
@@ -30,7 +32,11 @@ final case class AllNotifiers(notifiers: List[Notifier])(implicit val cs: Contex
     } yield ()
 
   def notify(notifier: Notifier, notification: Notification[_]): IO[Unit] =
-    notifier.notify(notification).handleErrorWith { e =>
-      IO { logger.error(s"Notifier ${notifier.getClass.getSimpleName} failed.", e) }
+    monitor(notifier.getClass.getSimpleName) {
+      notifier.notify(notification).handleErrorWith { e =>
+        IO {
+          logger.error(s"Notifi er ${notifier.getClass.getSimpleName} failed.", e)
```
pauljamescleary: Looks like a space in `Notifi er` in the `logger.error`

ericcheatham: Good catch.

ericcheatham: Would you prefer that I make an additional commit on this branch or amend my current commit?

pauljamescleary: We do squash merge here, and it is easier just to add another commit :)

pauljamescleary:
Thanks @ericcheatham !!!

PR #867, opened 2019-10-08
Feature flag multi record delete

### LINKED ISSUE #860: Feature Flag Multi-record support in Batch
We have a feature ready to support multi-records via batch change; however, it has some issues with clients that need to be updated.

As such, we need to add a feature flag to both the Portal and the API so we can merge all of the multi-record support.

britneywright:
fixes #860

Changes in this pull request:
- reimplement reverted commits related to multi-record batch change support (except docs commit)
- add feature flags to the API and portal `enable-multi-record-batch-update`, default to false.
- add markers to the func tests that depend on the `enable-multi-record-batch-update` config value. One marker is `multi_record_enabled`, the other is `multi_record_disabled`. Right now I've set the tests to not run `multi_record_enabled` tests for all environments.

### DIFF
--- a/modules/api/functional_test/live_tests/batch/create_batch_change_test.py
+++ b/modules/api/functional_test/live_tests/batch/create_batch_change_test.py
@@ -55,15 +55,15 @@ def assert_change_success_response_values(changes_json, zone, index, record_name
     assert_that(changes_json[index]['id'], is_not(none()))
     assert_that(changes_json[index]['changeType'], is_(change_type))
     assert_that(record_type, is_in(['A', 'AAAA', 'CNAME', 'PTR', 'TXT', 'MX']))
-    if record_type in ["A", "AAAA"] and change_type == "Add":
+    if record_type in ["A", "AAAA"] and change_type != "DeleteRecordSet":
         assert_that(changes_json[index]['record']['address'], is_(record_data))
-    elif record_type == "CNAME" and change_type == "Add":
+    elif record_type == "CNAME" and change_type != "DeleteRecordSet":
         assert_that(changes_json[index]['record']['cname'], is_(record_data))
-    elif record_type == "PTR" and change_type == "Add":
+    elif record_type == "PTR" and change_type != "DeleteRecordSet":
         assert_that(changes_json[index]['record']['ptrdname'], is_(record_data))
-    elif record_type == "TXT" and change_type == "Add":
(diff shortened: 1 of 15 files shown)


modules/api/functional_test/live_tests/batch/create_batch_change_test.py
```diff
@@ -3755,11 +3755,29 @@ def test_create_batch_with_irrelevant_global_acl_rule_applied_fails(shared_zone_
 
 @pytest.mark.serial
 @pytest.mark.skip_production
+@pytest.mark.multi_record_enabled
 def test_create_batch_duplicates_add_check(shared_zone_test_context):
     """
-    Test recordsets with multiple records cannot be added in batch (relies on config, skip-prod)
```
mitruly: We need to keep the original test as-is but flagged with a marker (e.g. `@pytest.mark.multi_record_disabled`). We need to also test behavior when multi-record is disabled to ensure that it is correct.

modules/api/functional_test/live_tests/batch/create_batch_change_test.py
```diff
@@ -3773,98 +3791,256 @@ def test_create_batch_duplicates_add_check(shared_zone_test_context):
...
+            get_change_TXT_json(txt_fqdn, text="some-multi-text"),
+            get_change_TXT_json(txt_fqdn, text="more-multi-text"),
+            get_change_MX_json(mx_fqdn, preference=0),
+            get_change_MX_json(mx_fqdn, preference=1000, exchange="bar.foo.")
         ]
     }
 
-    response = client.create_batch_change(batch_change_input, status=400)
+    def err(name):
+        return 'Record "{}" Already Exists: cannot add an existing record; to update it, issue a DeleteRecord or DeleteRecordSet then an Add.'.format(name)
```
mitruly: With the proposal to have a feature flag, we actually need to retain the old message for code when multi-record is disabled (since `DeleteRecord` would not be valid).

modules/api/src/main/resources/reference.conf
```diff
@@ -132,7 +132,6 @@ vinyldns {
     }
   }
 
-  scheduled-changes-enabled = false
```
mitruly: Do we need to remove this?

britneywright:
closing this, we're going to take a fresh approach to this.

PR #871, opened 2019-10-10
feature flag portal multi-record batch change support

britneywright:
portal flag for #867

Changes in this pull request:
- feature flag for the New DNS Change form that determines if the Record Data field shows for a change if the is a DeleteRecordSet type. 
- `multi-record-batch-change-enabled` config value defaults to false if not set. We use the term "DNS Change" in the portal but the config values for the API and portal should match hence the use of "batch-change".
- Include help text when the feature is enabled to let the user know when Record Data is not a required field.

Screenshot of portal with `multi-record-batch-change-enabled` set to true:
Screen Shot 2019-10-10 at 3 14 58 AM

### DIFF
--- a/modules/portal/app/models/Meta.scala
+++ b/modules/portal/app/models/Meta.scala
@@ -23,7 +23,8 @@ case class Meta(
     batchChangeLimit: Int,
     defaultTtl: Long,
     manualBatchChangeReviewEnabled: Boolean,
-    scheduledBatchChangesEnabled: Boolean)
+    scheduledBatchChangesEnabled: Boolean,
+    multiRecordBatchChangeEnabled: Boolean)
 object Meta {
   def apply(config: Configuration): Meta =
     Meta(
@@ -32,6 +33,7 @@ object Meta {
       config.getOptional[Int]("batch-change-limit").getOrElse(1000),
       config.getOptional[Long]("default-ttl").getOrElse(7200L),
       config.getOptional[Boolean]("manual-batch-review-enabled").getOrElse(false),
-      config.getOptional[Boolean]("scheduled-changes-enabled").getOrElse(false)
+      config.getOptional[Boolean]("scheduled-changes-enabled").getOrElse(false),
+      config.getOptional[Boolean]("multi-record-batch-change-enabled").getOrElse(false)
     )
 }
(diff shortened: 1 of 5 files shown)


modules/portal/app/views/batchChanges/batchChangeNew.scala.html
```diff
@@ -151,42 +148,46 @@ <h4>Changes</h4>
...
+                                    <td ng-if="change.type=='AAAA'">
+                                        <input name="record_address_{{$index}}" type="text" ng-model="change.record.address" ng-required="change.changeType=='Add'" class="form-control" placeholder="e.g. fd69:27cc:fe91::60" ng-disabled="change.changeType=='DeleteRecordSet' && !@meta.multiRecordBatchChangeEnabled" ipv6>
                                         <p ng-show="createBatchChangeForm.$submitted">
                                             <span ng-show="createBatchChangeForm.record_address_{{$index}}.$error.required" class="batch-change-error-help">Record data is required!</span>
                                             <span ng-show="createBatchChangeForm.record_address_{{$index}}.$error.ipv6" class="batch-change-error-help">must be a valid IPv6 Address!</span>
                                         </p>
+                                        <p class="help-block" ng-if="change.changeType=='DeleteRecordSet' && @meta.multiRecordBatchChangeEnabled">Record Data is optional.</p>
```
britneywright: I'm concerned users will think DeleteRecordSet changes require record data if the feature flag is enabled, the help text is intended to let them know when it's optional. This is a starting point. One possible iteration is that we link that text to a longer explanation in our docs. Would like to test the form out with some users and get feedback.

modules/portal/public/lib/batch-change/batch-change.directive.js
```diff
@@ -39,50 +38,58 @@
...
+    app.directive('fqdn', function () {
         return {
             require: 'ngModel',
-            link: function(scope, elm, attrs, ctrl) {
-                ctrl.$validators.fqdn = function(modelValue, viewValue) {
-                    return FQDN_REGEX.test(viewValue);
-                }
+            link: function (scope, elm, attrs, ctrl) {
+                ctrl.$validators.fqdn = function (modelValue, viewValue) {
+                    if (attrs.required || (viewValue !== undefined && viewValue.length > 0)) {
```
britneywright: had to update these record data validations so they only run when the record data field is required or it has content in it.

PR #872, opened 2019-10-10
Enable multi-record functionality in batch change

mitruly:
Prototype for multi-record enabled.

Changes in this PR:
- Allow optional `record` data with `DeleteRecordSet` requests (controlled by configuration value of `multi-record-batch-change-enabled`)
- Update unit tests

Outstanding issues:
- Update functional tests (with testing behavior dependent on multi-record enabled)

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
@@ -175,7 +175,7 @@ class BatchChangeValidations(
       isApproved: Boolean): ValidatedBatch[ChangeInput] =
     input.map {
       case a: AddChangeInput => validateAddChangeInput(a, isApproved).map(_ => a)
-      case d: DeleteRRSetChangeInput => validateInputName(d, isApproved).map(_ => d)
+      case d: DeleteRRSetChangeInput => validateDeleteRRSetChangeInput(d, isApproved).map(_ => d)
     }
 
   def validateAddChangeInput(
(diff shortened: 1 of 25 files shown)


modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
```diff
@@ -266,13 +265,13 @@ class BatchChangeValidations(
       groupedChanges: ChangeForValidationMap): SingleValidation[Unit] =
     change match {
       // For DeleteRecord inputs, need to verify that the record data actually exists
-      case deleteRecord: DeleteRecordChangeForValidation
-          if !groupedChanges
+      case deleteRecord: DeleteRRSetChangeForValidation
+          if deleteRecord.inputChange.record.isDefined && !groupedChanges
             .getExistingRecordSet(change.recordKey)
-            .exists(_.records.contains(deleteRecord.inputChange.record)) =>
+            .exists(_.records.contains(deleteRecord.inputChange.record.get)) =>
```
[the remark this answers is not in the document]
mitruly: @pauljamescleary yeah, there is code clean up that I intended to have; this was just to get the functionality we were looking for during prototype phase :)

modules/api/src/main/scala/vinyldns/api/route/VinylDNSJsonProtocol.scala
```diff
@@ -24,6 +24,8 @@ trait VinylDNSJsonProtocol
     with BatchChangeJsonProtocol
     with ACLJsonProtocol {
 
+  implicit val multiRecordEnabled: Boolean
```
pauljamescleary: Need to pull this out.

We should not do any kind of business logic in the JSON / Routes, those should be purely for in/out of HTTP things.

The only place we need to worry about `multiRecordEnabled` should be where we currently do the check during batch change processing.

I have to hunt it down, but somewhere in batch change processing we should have logic that says `if this record set has multiple records then raise an error`, the config flag should bypass that check if multi record is enabled.

There is a question of "do we serialize to json the record data if it is present".  I am unclear on that but we should proceed with everything else and just test that really well against older clients.

modules/api/src/test/scala/vinyldns/api/notifier/email/EmailNotifierSpec.scala
```diff
@@ -222,9 +223,6 @@ class EmailNotifierSpec
             row.contains(ac.ttl.toString) should be(true)
           case _: SingleDeleteRRSetChange =>
             row.contains("Delete") should be(true)
-          case dc: SingleDeleteRecordChange =>
-            row.contains("Delete") should be(true)
-            row.contains(dc.recordData) should be(true)
```
pauljamescleary: Should we verify still that the record data is there in the row?

mitruly:
@pauljamescleary yes, I left some feedback on the func tests; will merge those in once they have been incorporated. Definitely agree that we should wait for tests before merging in source.

PR #836, opened 2019-09-11
[DeleteRecord] Remove multi-record config

mitruly:
Remove multi-record config and support DeleteRecord inputs.

Changes in this pull request:
- Remove multi-record configs and associated errors
- Enable `DeleteRecord` input
- Update existing tests

Outstanding:
- Add more exhaustive (functional and unit) testing for `DeleteRecord`

Note: This PR is dependent on both https://github.com/vinyldns/vinyldns/pull/833 and https://github.com/vinyldns/vinyldns/pull/834.

PR #877, opened 2019-10-11
Func tests for multi record flag

britneywright:
rebased off #872

### DIFF
--- a/docker/functest/run.sh
+++ b/docker/functest/run.sh
@@ -55,8 +55,8 @@ if [ "${PROD_ENV}" = "true" ]; then
     ./run-tests.py live_tests -n2 -v -m "not skip_production and not serial and not multi_record_enabled" --url=${VINYLDNS_URL} --dns-ip=${DNS_IP} ${TEST_PATTERN} --teardown=False
     if [ $? -eq 0 ]; then
       # run serial tests second (serial marker)
-      echo "./run-tests.py live_tests -n0 -v -m \"not skip_production and serial and not multi_record_disabled\" -v --url=${VINYLDNS_URL} --dns-ip=${DNS_IP} ${TEST_PATTERN} --teardown=True"
-      ./run-tests.py live_tests -n0 -v -m "not skip_production and serial and not multi_record_disabled" --url=${VINYLDNS_URL} --dns-ip=${DNS_IP} ${TEST_PATTERN} --teardown=True
+      echo "./run-tests.py live_tests -n0 -v -m \"not skip_production and serial and not multi_record_enabled\" -v --url=${VINYLDNS_URL} --dns-ip=${DNS_IP} ${TEST_PATTERN} --teardown=True"
+      ./run-tests.py live_tests -n0 -v -m "not skip_production and serial and not multi_record_enabled" --url=${VINYLDNS_URL} --dns-ip=${DNS_IP} ${TEST_PATTERN} --teardown=True
     fi
 else
     # run parallel tests first (not serial)
--- a/docker/functest/run.sh
+++ b/docker/functest/run.sh
@@ -51,21 +51,21 @@ find . -name "__pycache__" -delete
 if [ "${PROD_ENV}" = "true" ]; then
     # -m plays havoc with -k, using variables is a headache, so doing this by hand
     # run parallel tests first (not serial)
-    echo "./run-tests.py live_tests -n2 -v -m \"not skip_production and not serial\" -v --url=${VINYLDNS_URL} --dns-ip=${DNS_IP} ${TEST_PATTERN} --teardown=False"
-    ./run-tests.py live_tests -n2 -v -m "not skip_production and not serial" --url=${VINYLDNS_URL} --dns-ip=${DNS_IP} ${TEST_PATTERN} --teardown=False
+    echo "./run-tests.py live_tests -n2 -v -m \"not skip_production and not serial and not multi_record_enabled\" -v --url=${VINYLDNS_URL} --dns-ip=${DNS_IP} ${TEST_PATTERN} --teardown=False"
+    ./run-tests.py live_tests -n2 -v -m "not skip_production and not serial and not multi_record_enabled" --url=${VINYLDNS_URL} --dns-ip=${DNS_IP} ${TEST_PATTERN} --teardown=False
     if [ $? -eq 0 ]; then
       # run serial tests second (serial marker)
-      echo "./run-tests.py live_tests -n0 -v -m \"not skip_production and serial\" -v --url=${VINYLDNS_URL} --dns-ip=${DNS_IP} ${TEST_PATTERN} --teardown=True"
-      ./run-tests.py live_tests -n0 -v -m "not skip_production and serial" --url=${VINYLDNS_URL} --dns-ip=${DNS_IP} ${TEST_PATTERN} --teardown=True
+      echo "./run-tests.py live_tests -n0 -v -m \"not skip_production and serial and not multi_record_disabled\" -v --url=${VINYLDNS_URL} --dns-ip=${DNS_IP} ${TEST_PATTERN} --teardown=True"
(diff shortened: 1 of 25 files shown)


docker/functest/run.sh
```diff
@@ -51,21 +51,21 @@ find . -name "__pycache__" -delete
...
+    echo "./run-tests.py live_tests -n2 -v -m \"not skip_production and not serial and not multi_record_enabled\" -v --url=${VINYLDNS_URL} --dns-ip=${DNS_IP} ${TEST_PATTERN} --teardown=False"
+    ./run-tests.py live_tests -n2 -v -m "not skip_production and not serial and not multi_record_enabled" --url=${VINYLDNS_URL} --dns-ip=${DNS_IP} ${TEST_PATTERN} --teardown=False
     if [ $? -eq 0 ]; then
       # run serial tests second (serial marker)
-      echo "./run-tests.py live_tests -n0 -v -m \"not skip_production and serial\" -v --url=${VINYLDNS_URL} --dns-ip=${DNS_IP} ${TEST_PATTERN} --teardown=True"
-      ./run-tests.py live_tests -n0 -v -m "not skip_production and serial" --url=${VINYLDNS_URL} --dns-ip=${DNS_IP} ${TEST_PATTERN} --teardown=True
+      echo "./run-tests.py live_tests -n0 -v -m \"not skip_production and serial and not multi_record_disabled\" -v --url=${VINYLDNS_URL} --dns-ip=${DNS_IP} ${TEST_PATTERN} --teardown=True"
+      ./run-tests.py live_tests -n0 -v -m "not skip_production and serial and not multi_record_disabled" --url=${VINYLDNS_URL} --dns-ip=${DNS_IP} ${TEST_PATTERN} --teardown=True
```
mitruly: These should have `and not multi_record_enabled`.

modules/api/functional_test/utils.py
```diff
@@ -413,45 +413,39 @@ def clear_groups(client, exclude=[]):
...
-                "record": {
-                    "address": address
-                }
-            }
-        else:
-            json = {
-                "changeType": "DeleteRecordSet",
-                "inputName": input_name,
-                "type": record_type
+        json = {
```
mitruly: I like the functionality changes; minor feedback (applies to all similar helpers) on cleaning up a bit better for clarity:
```
    json = {
        "changeType": change_type,
        "inputName": input_name,
        "type": record_type,
    }

    if change_type == "Add":
        json["ttl"] = ttl
        json["record"] = {
            "address": address or "1.1.1.1"
        }
    else:
        if address is not None:
            json["record"] = {
                "address": address
            }

    return json
```

modules/api/functional_test/live_tests/batch/create_batch_change_test.py
```diff
@@ -3794,8 +3795,106 @@ def err(name, type):
...
+
+        assert_failed_change_in_error_response(errors[0], input_name="delete-non-existent-record.ok.", record_data="1.1.1.1", change_type="DeleteRecordSet",
+                                               error_messages=['Record "delete-non-existent-record.ok." Does Not Exist: cannot delete a record that does not exist.'])
+        assert_failed_change_in_error_response(errors[1], input_name=a_delete_fqdn, record_data="4.5.6.7", change_type="DeleteRecordSet",
+                                               error_messages=['Record data AData(4.5.6.7) does not exist for "' + a_delete_fqdn + '".'])
+
+    finally:
+        clear_recordset_list(to_delete, client)
+
+# def test_create_batch_delete_recordset_fails(shared_zone_test_context):
```
mitruly: This test can be removed since multi-record enabled and disabled will now recognize `record` if provided. Validation is also confirmed by the `test_create_batch_delete_record_for_invalid_record_data_fails` test.

modules/api/functional_test/live_tests/batch/create_batch_change_test.py
```diff
@@ -3869,68 +3968,388 @@ def new_err(name, type):
         clear_recordset_list(to_delete, client)
 
 
-@pytest.mark.manual_batch_review
-def test_create_batch_with_zone_name_requiring_manual_review(shared_zone_test_context):
+@pytest.mark.multi_record_disabled
+def test_create_batch_deletes_fails(shared_zone_test_context):
```
mitruly: Can we rename this for clarity to something like `test_create_batch_with_multi_record_deletes_fails`?

modules/api/functional_test/live_tests/batch/create_batch_change_test.py
```diff
@@ -3753,6 +3753,7 @@ def test_create_batch_with_irrelevant_global_acl_rule_applied_fails(shared_zone_
             shared_client.wait_until_recordset_change_status(delete_a_rs, 'Complete')
 
 
+@pytest.mark.multi_record_disabled
```
mitruly: We need a version of these tests where new adds succeed for when `multi_record_enabled` (though multi-record adds for existing records will _still_ fail). See: https://github.com/vinyldns/vinyldns/pull/836/files#diff-090b2d00e8c30c6696c34e10ed759ac5.

modules/api/functional_test/live_tests/batch/create_batch_change_test.py
```diff
@@ -3869,68 +3968,388 @@ def new_err(name, type):
...
+        "comments": "this is optional",
         "changes": [
-            get_change_A_AAAA_json("add-test-batch.zone.requires.review.", address="1.1.1.1"),
-            get_change_A_AAAA_json("update-test-batch.zone.requires.review.", change_type="DeleteRecordSet"),
-            get_change_A_AAAA_json("update-test-batch.zone.requires.review.", address="1.1.1.1"),
-            get_change_A_AAAA_json("delete-test-batch.zone.requires.review.", change_type="DeleteRecordSet")
-        ],
-        "ownerGroupId": shared_zone_test_context.ok_group['id']
+            get_change_A_AAAA_json(rs_fqdn, address="1.2.3.4", change_type="DeleteRecordSet"),
+            get_change_A_AAAA_json(rs_fqdn_2, address="1.2.3.4", change_type="DeleteRecordSet"),
```
mitruly: One of these tests shouldn't be passing in `address` (so that we are testing `DeleteRecordSet` with no data against the single record).

docker/functest/run.sh
britneywright: think you looked before i updated that

mitruly:
Can we also have a functional test added for checking for invalid record data for `Add` and `DeleteRecordSet`? Just want to ensure that we get an expected error rather than a HTTP 500 response from the API.

modules/api/functional_test/live_tests/batch/create_batch_change_test.py
britneywright: This isn't only testing multi record, this is testing all forms of delete: single record without record data, single record with record data multi-record without record data and multi-record with record data.

britneywright:
> Can we also have a functional test added for checking for invalid record data for `Add` and `DeleteRecordSet`? Just want to ensure that we get an expected error rather than a HTTP 500 response from the API.

There's one for invalid record data for DeleteRecordSet `test_create_batch_delete_record_for_invalid_record_data_fails` and there's already tests for invalid record data for Adds for each of the record types.

mitruly:
@britneywright Am good with these changes once we add a JSON validation.

The `test_create_batch_delete_record_for_invalid_record_data_fails` test has properly formatted record. And we should have just one test similar to the adds for deletes - the reason being that we caught an issue where we weren't actually validating the record data specifically for `DeleteRecordSet`: https://github.com/vinyldns/vinyldns/pull/872/commits/e1c144c063b283280e75a23bb434e2ad0b2bd8d8. The add test follows a separate flow in the validation process so we should check the flow for `DeleteRecordSet`.

mitruly:
Closing as I merged this PR into https://github.com/vinyldns/vinyldns/pull/872.

PR #878, opened 2019-10-14
Update documentation

mitruly:
Update API, portal and configuration documentation for handling multi-record changes.


modules/docs/src/main/tut/api/batchchange-model.md
```diff
@@ -101,6 +102,16 @@ validationErrors | Array of BatchChangeError | Array containing any validation e
...
 
+#### ChangeType Values <a id="changetype-values" />
+There are two valid `changeType`s for a `SingleChange`: **Add** and **DeleteRecordSet**
+
+`changeType`s can be used independently or combined to achieve the desired behavior described below.
+
+- Create a new DNS record: **Add** with record data
+- Delete an entire record set: 1. **DeleteRecordSet** without specifying existing record data *or* 2. **DeleteRecordSet** for each entry of the DNS record
+- Delete a single entry from DNS record with multiple entries: **DeleteRecordSet** specifying existing record data
+- Update an existing record set (i.e. value reassignment): 1. **DeleteRecordSet** specifying existing record data (single entry delete) or not specifying record data (full delete) *and* 2. **Add** with record data
```
britneywright: why the `(i.e. value reassignment)`?

modules/docs/src/main/tut/api/create-batchchange.md
```diff
@@ -45,6 +45,7 @@ name          | type          | required?   | description |
 changeType    | ChangeInputType | yes       | Type of change input. Must be **DeleteRecordSet** for *DeleteChangeInput*. |
 inputName     | string        | yes         | The fully qualified domain name of the record being deleted. |
 type          | RecordType    | yes         | Type of DNS record. Supported records for batch changes are currently: **A**, **AAAA**, **CNAME**, and **PTR**. |
+record        | [RecordData](../api/recordset-model#record-data) | no         | The data for the record. If specified, only this DNS entry for the existing DNS recordset will be deleted; if unspecified, all DNS entries associated to the DNS recordset will be deleted. |
```
britneywright:
```suggestion
record        | [RecordData](../api/recordset-model#record-data) | no         | The data for the record. If specified, only this DNS entry for the existing DNS recordset will be deleted; if unspecified, the entire DNS recordset will be deleted. |
```

modules/docs/src/main/tut/portal/dns-changes.md
```diff
@@ -13,7 +13,7 @@ DNS Changes is an alternative to submitting individual RecordSet changes and pro
 **Note**: DNS Change is portal-only terminology. The API equivalent is [batch change](../api/batchchange-model).
 
 #### Access
-* Access permissions will follow existing rules (admin group or ACL access). Note that an update (delete and add of the same record name, zone and record type combination) requires **Write** or **Delete** access.
+* Access permissions will follow existing rules (admin group or ACL access). Note that an update (delete and add of the same record name *or* delete of single entry of multi-record DNS record set, zone and record type combination) requires **Write** or **Delete** access.
```
britneywright: this added text is confusing and unnecessary. The intended point is that because you need to do a delete (of any kind) and an add to achieve a record update you must have write and delete access.

modules/docs/src/main/tut/api/batchchange-model.md
mitruly: @britneywright this is in reference to "re-IP" or IP reassignment, though I can remove this if this is confusing.

britneywright: yeah i don't think it's needed

modules/docs/src/main/tut/portal/dns-changes.md
mitruly: @britneywright I don't quite agree that this is unnecessary, though I'm open to suggestions to clarify up any confusing verbiage.

VinylDNS treats updates with `Write` permissions. If we delete a single entry from a multi-record DNS record set, only `Write` permission is required (and not `Delete`).

britneywright: This note is aiming to explain when both Write and Delete ACL access is required, 
so we could clarify that it's a full recordset delete and an add that require Write and Delete Access.

mitruly: @britneywright can you elaborate on your comment? _Any_ type of delete (whether it is a "full" record set delete or single DNS record) with an add is an update, which is *Write* permissions. Access levels can be considered bucketed/tiered, so *Delete* access would also work but isn't required.

Here is the full breakdown of add/deletes given:
`foo.bar.   A   1.1.1.1`
`foo.bar.   A   2.2.2.2`

- Add `foo2.bar.   A   3.3.3.3`: write access
- Delete `foo.bar.   A   2.2.2.2`: write access
- Delete `foo.bar.   A   1.1.1.1` and `foo.bar.   A  2.2.2.2`: delete access
- Delete `foo.bar.   A   1.1.1.1` and `foo.bar.   A  2.2.2.2`; add `foo.bar.   A   3.3.3.3`: write access
- Delete `foo.bar.   A   1.1.1.1` and add `foo.bar.   A   3.3.3.3`: write access

britneywright: If delete access isn't required than this note doesn't seem needed at all and we could just delete the last sentence.

mitruly: This pre-existing note is mentioning that you would need either `Write` or `Delete` access for an "update" operation and this PR is adding on a new scenario for an update.

As we don't yet have a comprehensive breakdown on access levels required for which operations in batch change, I'm inclined to say that we keep this for now.

britneywright: 👍🏽

PR #882, opened 2019-10-18
remove empty record data fields from json

britneywright:
Screen Shot 2019-10-18 at 1 48 52 PM

Changes in this pull request:
- In the DNS Change form in the portal if there's an entry with a changeType of "DeleteRecordSet" and the record data value is undefined or an empty string we will delete the record data attributes before sending the JSON to the API.

To test this:
- Choose "DeleteRecordSet" change type, enter data in the Input Name and Record Data fields.
- Delete the data in the Record Data field
- Submit the DNS Change

The DNS Change should not return an error related invalid record data


britneywright:
going to change this in a bit.

modules/portal/public/lib/dns-change/dns-change-new.controller.js
```diff
@@ -86,6 +86,17 @@
                             var newEntry = {changeType: entry.changeType, type: "PTR", ttl: entry.ttl, inputName: entry.record.address, record: {ptrdname: entry.inputName}}
                             payload.changes.splice(i+1, 0, newEntry)
                         }
+                        if(entry.changeType == 'DeleteRecordSet' && entry.record) {
+                            var recordDataEmpty = true;
+                            for (var attr in entry.record) {
+                                if (entry.record[attr] !== undefined && entry.record[attr].length > 0) {
```
mitruly: We need to do a different check here:
```
if (entry.record[attr] === undefined || entry.record[attr].length == 0) {
  delete entry.record[attr];
} else {
    recordDataEmpty = false
}
```

If I create an `MX` record using a whitespace character, the request hits the API with `"exchange": ""` since the attribute is not `undefined`.

modules/portal/app/views/dnsChanges/dnsChangeNew.scala.html
```diff
@@ -187,7 +187,7 @@ <h4>Changes</h4>
                                     </td>
                                     <td ng-if="change.type=='MX'">
                                         <label class="batch-label">Preference</label>
-                                        <input name="record_mx_preference_{{$index}}" type="number" ng-model="change.record.preference" ng-required="change.changeType=='Add'" class="form-control" placeholder="e.g. 1" ng-min="0" ng-max="65535" ng-disabled="change.changeType=='DeleteRecordSet' && !@meta.multiRecordBatchChangeEnabled">
+                                        <input name="record_mx_preference_{{$index}}" type="number" ng-model="change.record.preference" ng-required="change.changeType=='Add' || change.record.exchange" class="form-control" placeholder="e.g. 1" ng-min="0" ng-max="65535" ng-disabled="change.changeType=='DeleteRecordSet' && !@meta.multiRecordBatchChangeEnabled">
```
britneywright: updated the form for MX so preference and exchange are dependent on one another. If one field has data in it the other will be required.

modules/portal/public/lib/dns-change/dns-change-new.controller.js
```diff
@@ -86,6 +86,18 @@
                             var newEntry = {changeType: entry.changeType, type: "PTR", ttl: entry.ttl, inputName: entry.record.address, record: {ptrdname: entry.inputName}}
                             payload.changes.splice(i+1, 0, newEntry)
                         }
+                        if(entry.changeType == 'DeleteRecordSet' && entry.record) {
+                            var recordDataEmpty = true;
+                            for (var attr in entry.record) {
+                                console.log(entry.record[attr])
+                                if (entry.record[attr] !== undefined && entry.record[attr].toString().length > 0) {
```
britneywright: the MX preference field was error-ing on `entry.record[attr].length` because it's a number not a string.

modules/portal/public/lib/dns-change/dns-change-new.controller.js
```diff
@@ -86,6 +90,18 @@
...
+                            for (var attr in entry.record) {
+                                console.log(entry.record[attr])
+                                if (entry.record[attr] != undefined && entry.record[attr].toString().length > 0) {
+                                    recordDataEmpty = false
+                                }
+                            }
+                            if (recordDataEmpty) {
+                                delete entry.record
+                            }
+                        }
```
mitruly: The behavior is still not working as expected since `0` preference with whitespace will still still send `"exchange": ""` to the API:
Screen Shot 2019-10-22 at 8 36 26 AM
Screen Shot 2019-10-22 at 8 36 39 AM

We need to have an:
```
else {
    delete entry.record[attr];
}
```

britneywright:
@mitruly are you sure you're in the correct branch with the latest commits when testing this locally. It shouldn't be possible to submit partially completed MX record data now. If preference is filled in then exchange is required and vice versa. Exchange field with spaces would fail the validation that requires it to be an absolute value.

Screen Shot 2019-10-22 at 10 17 27 AM

mitruly:
@britneywright yes, I am sure and I just re-tested this again. Note the screen captures I provided where the preference is `0`.

Note: Realistically, VinylDNS will be supporting more record types via batch change interface in the future and the proposed solution should make sure that the request JSON is properly formatted, regardless of the type. It would be desirable to prevent sending empty strings in our request JSON (through validation ahead of time).

mitruly:
Also noticed that upon error for `DeleteRecordSet`, the record data is being cleared which is inconsistent with `Add` behavior.

britneywright:
> Also noticed that upon error for `DeleteRecordSet`, the record data is being cleared which is inconsistent with `Add` behavior.

That's what my other PR is for to add record data to the JSON protocol so the data is returned. 😄

modules/portal/app/views/dnsChanges/dnsChangeNew.scala.html
```diff
@@ -198,7 +198,7 @@ <h4>Changes</h4>
                                         <br />
 
                                         <label class="batch-label">Exchange</label>
-                                        <input name="record_mx_exchange_{{$index}}" type="text" ng-model="change.record.exchange" ng-required="change.changeType=='Add'" class="form-control" placeholder="e.g. test.example.com." ng-disabled="change.changeType=='DeleteRecordSet' && !@meta.multiRecordBatchChangeEnabled" fqdn>
+                                        <input name="record_mx_exchange_{{$index}}" type="text" ng-model="change.record.exchange" ng-required="change.changeType=='Add' || change.record.preference==0 || change.record.preference>0" class="form-control" placeholder="e.g. test.example.com." ng-disabled="change.changeType=='DeleteRecordSet' && !@meta.multiRecordBatchChangeEnabled" fqdn>
```
mitruly:
```suggestion
                                        <input name="record_mx_exchange_{{$index}}" type="text" ng-model="change.record.exchange" ng-required="change.changeType=='Add' || change.record.preference>=0" class="form-control" placeholder="e.g. test.example.com." ng-disabled="change.changeType=='DeleteRecordSet' && !@meta.multiRecordBatchChangeEnabled" fqdn>
```

britneywright: `>=` doesn't work because `null >= 0` is considered true. I ran into issues with that.

britneywright: Did `Number.isInteger(change.record.preference)` work for you? I tried it before and just tried it again but it did not trigger the exchange field to be required.

mitruly: @britneywright I tried this locally and it seemed to work for me; the scenario is that if a number was entered into the `preference` field, we would be required to enter in an `exchange`.

mitruly: @britneywright Can you test out `Number(change.record.preference) !== 'NaN'`?

mitruly: Discussed offline with @britneywright and think that the current implementation works for now.

PR #883, opened 2019-10-18
Build optimizations

pauljamescleary:
A few specific build optimizations:

1. Consolidated `dockerComposeUp` to only use a single `root/docker/docker-compose.yml` instead of each module having its own docker-compose files.  This eliminates additional waits for docker containers to startup and stop, as well as reduces memory consumption during the build
2. Cleaned up `VinylDNSSpec` - I noticed that this spec was taking 3 minutes to run!  I discovered that the way we were mocking the `WSClient` was largely to blame.  Was able to get tests to run in **16 SECONDS** using a library called `mock-ws`.  This is where we see most savings.
3. Added back `dynamodb-local` instead of running it in `localstack`.  Integration tests for dynamodb were very slow in localstack.  This added an additional 20-30 second improvement.

After doing several tests locally running the following command...

```
> SECONDS=0; sbt verify; echo "DURATION = $SECONDS SECONDS"
```

Current master took 535 seconds to run; with these optimizations it took **211 SECONDS** - that is a 60% improvement.

The initial Travis builds reported a run time of 13 minutes as opposed to 19 minutes; this would save some 6 minutes off of Travis build times (or 30% improvement).


nimaeskandary:
🔥

docker/functest/run.sh
```diff
@@ -47,25 +47,31 @@ cd /app
...
+    result=$?
+    if [ $result -eq 0 ]; then
       # run serial tests second (serial marker)
       echo "./run-tests.py live_tests -n0 -v -m \"serial and not multi_record_disabled\" --url=${VINYLDNS_URL} --dns-ip=${DNS_IP} ${TEST_PATTERN} --teardown=True"
       ./run-tests.py live_tests -n0 -v -m "serial and not multi_record_disabled" --url=${VINYLDNS_URL} --dns-ip=${DNS_IP} ${TEST_PATTERN} --teardown=True
+      result=$?
     fi
 fi
+
+exit $result
```
mitruly: Need a new line here.

PR #886, opened 2019-10-22
Allow overriding parallelism for func tests

pauljamescleary:
Allow overriding of the number of CPUs to use for running func tests.  Leaving this as an override as `auto` is not always optimal and will depend on the machine performing the func tests.

### DIFF
--- a/bin/func-test-api-travis.sh
+++ b/bin/func-test-api-travis.sh
@@ -35,8 +35,9 @@ cp -f $DIR/../modules/api/target/scala-2.12/vinyldns.jar $WORK_DIR/docker/api
 echo "Starting docker environment and running func tests..."
 
 if [ -z "${PAR_CPU}" ]; then
-  PAR_CPU=2
+  export PAR_CPU=2
 fi
+
 docker-compose -f $WORK_DIR/docker/docker-compose-func-test.yml --project-directory $WORK_DIR/docker up --build --exit-code-from functest
 test_result=$?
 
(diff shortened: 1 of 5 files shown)


docker/functest/Dockerfile
```diff
@@ -16,5 +16,8 @@ ADD functional_test /app
 # Install our func test requirements
 RUN pip install --index-url https://pypi.python.org/simple/ -r /app/requirements.txt
 
+# Specifies how many CPUs to use for func tests; the more the better or specifiy "auto" for optimal results
+ENV PAR_CPU=2
```
mitruly: Do we technically need to set this if it is the default value?

pauljamescleary: Yes, afaik we technically need to give it a default value

bin/func-test-api-travis.sh
```diff
@@ -33,6 +33,10 @@ fi
 cp -f $DIR/../modules/api/target/scala-2.12/vinyldns.jar $WORK_DIR/docker/api
 
 echo "Starting docker environment and running func tests..."
+
```
pauljamescleary: default in travis is 2

docker/functest/Dockerfile
pauljamescleary: Confirmed...
`Error response from daemon: failed to parse Dockerfile: ENV must have two arguments`

mitruly: @pauljamescleary I also confirmed locally; resetting back to commit https://github.com/vinyldns/vinyldns/pull/886/commits/82352153a8e5b4b70a9ceb9473d589f17673b93f should do the trick.

pauljamescleary: I think we are good here now?  I tested this locally in our CI and it works.

mitruly: @pauljamescleary I think so, too; just need to roll back the last commit.

pauljamescleary: Bah, I put the default back in, as it can be unclear as to why it failed which is worse.

PR #891, opened 2019-10-28
make User Is Not Authorized error more informative

britneywright:
Add record or zone contact information to User is Not Authorized error message in batch change/DNS Change.

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
@@ -154,8 +154,8 @@ class BatchChangeService(
             val updatedError = UserIsNotAuthorizedError(
               e.userName,
               e.ownerGroupId,
-              group.map(_.email).getOrElse(e.contactEmail),
               e.ownerType,
+              group.map(_.email),
               group.map(_.name))
             logger.error(updatedError.message)
             updatedError
--- a/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
@@ -32,7 +32,12 @@ import vinyldns.core.domain.auth.AuthPrincipal
 import vinyldns.core.domain.batch.BatchChangeApprovalStatus.BatchChangeApprovalStatus
 import vinyldns.core.domain.batch._
 import vinyldns.core.domain.batch.BatchChangeApprovalStatus._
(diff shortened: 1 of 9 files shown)


modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
```diff
@@ -486,7 +494,19 @@ class BatchChangeValidations(
...
+      .leftMap(_ =>
+        ownerGroupId match {
+          case Some(id) =>
+            UserIsNotAuthorizedError(authPrincipal.signedInUser.userName, id, "", "record")
+          case None =>
+            UserIsNotAuthorizedError(
+              authPrincipal.signedInUser.userName,
+              input.zone.adminGroupId,
+              input.zone.email,
+              "zone")
```
pauljamescleary: As opposed to a hardcoded string, instead to use a Sum type (enum) for the value

britneywright:
> @britneywright can we also have tests added for user should not have access to modify a record that they are the owner group of in a private zone and is returning the correct error message? Our access validations don't allow this, but just want to make sure that the error messaging reflects this.

This sounds like it should be a separate issue and PR.

britneywright:
> Not sure I follow, @britneywright - that should be this PR since before it was the same error message before and we're now changing the error message and how the fields are populated.

Added a record to test in the existing `test_a_recordtype_update_delete_checks` func test.

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -126,7 +131,40 @@ class BatchChangeService(
...
+            val group = groups.find(_.id == e.ownerGroupId)
+            val updatedError = UserIsNotAuthorizedError(
+              e.userName,
+              e.ownerGroupId,
+              e.ownerType,
+              group.map(_.email),
+              group.map(_.name))
+            logger.error(updatedError.message)
+            updatedError
+          case a =>
```
mitruly: `e` is preferable here for errors.

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -126,7 +131,40 @@ class BatchChangeService(
...
+            logger.error(a.message)
+            a
+        }.invalid
+      case Valid(a) => a.validNel
```
mitruly: This can just be `case valid => valid`.

PR #892, opened 2019-10-29
set help text for DNS Change form input name and record data

### LINKED ISSUE #879: Improve help text on DNS Change form
Per user feedback, we should try to improve the help text on the DNS Change form so users better understand what information goes in what fields.

One suggestion is to have the heading "Input Name" include the help text "existing name" and the heading "Record Data" have the help text "alias"

britneywright:
Fixes #879.

Changes in this pull request:

- Update the Input Name help text and add Record Data help text in the DNS Change form. 
  - Input Name help text: The name of the record set. Fully qualified domain name or IP address, depending on the record type.
  - Record data help text: Where the record should point.

Screen Shot 2019-10-29 at 11 05 58 AM

Screen Shot 2019-10-29 at 11 14 10 AM

### DIFF
--- a/modules/portal/app/views/dnsChanges/dnsChangeNew.scala.html
+++ b/modules/portal/app/views/dnsChanges/dnsChangeNew.scala.html
@@ -85,12 +85,16 @@
                                     <th class="col-md-2">Change Type</th>
                                     <th>Record Type</th>
                                     <th>Input Name
-                                        <span data-toggle="tooltip" data-placement="top" title="Fully qualified domain name or IP address, depending on the record type.">
+                                        <span data-toggle="tooltip" data-placement="top" title="The name of the record set. Fully qualified domain name or IP address, depending on the record type.">
                                             <span class="fa fa-question-circle"></span>
                                         </span>
                                     </th>
                                     <th class="col-md-1">TTL (optional)</th>
-                                    <th>Record Data</th>
+                                    <th>Record Data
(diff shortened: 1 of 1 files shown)


britneywright:
Why don't you think it's helpful in the context of an MX record? The record points to a mail servers. I think the text meets the goal of helping users distinguish between the input name field and the record data field.

mitruly:
It might make sense to have text for `A`/`AAAA+PTR` where input name would list **FQDN** and record data would list **IP Address** for those rows.
CNAME could have **FQDN** for record name and **Alias** for record data.

If I'm not mistaken, I think this is what @pauljamescleary is referring to, including a tooltip to elaborate where necessary. Regarding `MX` records, it's not super obvious (to me at least) that preference denotes priority order and exchange is the FQDN of the mail exchange server.

pauljamescleary:
@britneywright `MX` has exchange + preference, `SRV` has 5 fields.  For `TXT` records it doesn't make sense.  "Points to" doesn't apply in all cases.  Having help on the actual text fields in the row is preferable.

britneywright:
The scope of this PR is specifically general help tooltips on the input name and record data field names. Regardless of what we end up doing with the specific data fields I would like to implement this.

pauljamescleary:
I'm good with the feature, just need something more descriptive than "where the record should point".

Something like "record data linked to the FQDN or IP address in the Input Name" would be preferable as the record data can be many different things.

PR #901, opened 2019-11-04
DNS Change detail notices

britneywright:
Changes in this pull request:
- Add configurable notices to the DNS Change detail page based on the DNS Change status. The notices are intended to give users additional context about the state of their DNS Change.
- Config key is "dnsChangeNotices" and requires a list. The list can be empty or contain objects with the attributes: "status", "text", "alertType", "href" (optional), "hrefText" (optional).
- status is matched against our current list of valid batch change statuses: Cancelled, Complete, Failed, PartialFailure, PendingProcessing, PendingReview, Rejected, Scheduled. If the given status doesn't match we set the status to "Unknown". This determines which notice appears on which DNS Change.
- alertType is matched against a list of valid Bootstrap 3 alert types: info, success, danger, warning. If the given alertType doesn't match we set the alertType to "info". This sets the styling of the notice.
- If included, the "href" and "hrefText" will appear after the given "text" value in the notice message.

**Documentation Note**: will include documentation updates in this PR, but will wait for feedback to confirm the configuration.

Screenshots of the different style notices based on "alertType":

info alertType:
Screen Shot of alert-info style notice

success alertType:
Screen Shot of alert-info style notice

warning alertType:
Screen Shot of alert-info style notice

danger alertType:
Screen Shot of alert-info style notice

### DIFF
--- a/modules/portal/app/models/DnsChangeNotices.scala
+++ b/modules/portal/app/models/DnsChangeNotices.scala
@@ -55,12 +55,12 @@ case class DnsChangeNotice(
     alertType: DnsChangeNoticeType,
     text: String,
     hrefText: String,
-    href: String)
+    href: String
+)
 
 object DnsChangeStatus extends Enumeration {
   type DnsChangeStatus = Value
-  val Cancelled, Complete, Failed, PartialFailure, PendingProcessing, PendingReview, Rejected,
-  Scheduled = Value
+  val Cancelled, Complete, Failed, PartialFailure, PendingProcessing, PendingReview, Rejected, Scheduled = Value
 }
 
 object DnsChangeNoticeType extends Enumeration {
--- a/modules/portal/app/models/DnsChangeNotices.scala
+++ b/modules/portal/app/models/DnsChangeNotices.scala
@@ -1,0 +1,93 @@
+/*
+ * Copyright 2018 Comcast Cable Communications Management, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package models
+
+import com.typesafe.config.Config
+import models.DnsChangeNoticeType.DnsChangeNoticeType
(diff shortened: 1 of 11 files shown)


modules/portal/app/models/DnsChangeNotices.scala
```diff
@@ -0,0 +1,95 @@
...
+          "alertType" -> n.alertType.toString,
+          "text" -> n.text,
+          "href" -> n.href.getOrElse(""),
+          "hrefText" -> n.hrefText.getOrElse("")
+        )
+      }
+    )
+
+  def formatDnsChangeNotice(config: Config): DnsChangeNotice = {
+    val status = try {
```
pauljamescleary: in general, the rule is that if there is bad config to fail on load.   Fail fast instead of try/catching here

britneywright: I had to do a try/catch for implementation of ficus. It doesn't allow a fallback if the value given doesn't match the enumeration. For status and alertType i want to accept values outside of the enumeration in the event the supported values change I don't want to break someone's application or if they want to set up their config in anticipation of new values. 

I could go back to the mapping in the earlier commit if we don't like the try and catch handling needed with ficus.

modules/portal/app/models/DnsChangeNotices.scala
```diff
@@ -0,0 +1,93 @@
...
+ * limitations under the License.
+ */
+
+package models
+
+import com.typesafe.config.Config
+import models.DnsChangeNoticeType.DnsChangeNoticeType
+import models.DnsChangeStatus.DnsChangeStatus
+import play.api.libs.json.{JsValue, Json}
+import play.api.{ConfigLoader}
```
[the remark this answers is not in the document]
pauljamescleary: @mitruly these should be handled by scalafmt.  @britneywright you can always run scalafmt by hand in sbt, or I believe it is tied into compile and you can run from sbt>root`;compile;test:compile;it:compile`

modules/portal/app/models/DnsChangeNotices.scala
pauljamescleary: @britneywright you mean ficus doesn't throw if it fails?  I.e. if you do `config.as[DnsChangeStatus]("foo")` is just says "ok"?

britneywright:
updated this PR description to include screenshots of the different notice styles based on the given alertTtype in the config.

modules/portal/public/templates/notice.html
```diff
@@ -0,0 +1,4 @@
+<div class="alert" ng-class="'alert-'+ngModel.alertType" role="alert">
```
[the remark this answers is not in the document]
britneywright: missed this comment, but i did update all of the styling. The text is all white or off white now and the linked text is bold and underlined for distinction. Screenshots of each alert are in the description.

PR #903, opened 2019-11-07
Convert batch change notification to command

mitruly:
Currently the notifications for batch changes are fired during the record set change processing stage, but there is an edge case where the notification doesn't trigger despite processing completion. Queueing a new batch change command allows us the following:
- Retries on the processing notification
- Middle ground before batch processing is supported by the `CommandHandler`.

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/engine/BatchChangeHandler.scala
+++ b/modules/api/src/main/scala/vinyldns/api/engine/BatchChangeHandler.scala
@@ -1,0 +1,100 @@
+/*
+ * Copyright 2018 Comcast Cable Communications Management, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package vinyldns.api.engine
+
+import cats.effect._
+import cats.implicits._
+import org.slf4j.LoggerFactory
+import vinyldns.core.domain.batch.{
+  BatchChange,
+  BatchChangeCommand,
+  BatchChangeRepository,
+  BatchChangeStatus
+}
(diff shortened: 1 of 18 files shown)


modules/api/src/main/scala/vinyldns/api/engine/BatchChangeHandler.scala
```diff
@@ -0,0 +1,100 @@
...
+  private final case class Pending(
+      change: Option[BatchChange],
+      batchChangeCommand: BatchChangeCommand)
+      extends BatchChangeProcessorState
+
+  private final case class PendingBatchNotificationError(change: BatchChange) extends Throwable
+
+  private final case class Completed(change: Option[BatchChange]) extends BatchChangeProcessorState
+
+  private def fsm(
```
pauljamescleary: my .02 without knowing a whole lot here, I don't think you need something like an `fsm` here.  It is more-or-less "if the batch change is not complete then let the message timeout and retry again later, otherwise send the notification (on error retry later)" let me know if I am oversimplifying, but feels like a lot of code for what I mentioned

modules/api/src/test/scala/vinyldns/api/backend/CommandHandlerSpec.scala
```diff
@@ -64,6 +64,7 @@ class CommandHandlerSpec
   private val mockRecordChangeProcessor =
     mock[(DnsConnection, RecordSetChange) => IO[RecordSetChange]]
   private val mockZoneSyncProcessor = mock[ZoneChange => IO[ZoneChange]]
+  private val mockBatchChangeProcessor = mock[BatchChangeCommand => IO[Option[BatchChange]]]
```
pauljamescleary: probably should at least add one unit test no?

modules/api/src/main/scala/vinyldns/api/engine/BatchChangeHandler.scala
dscleaver: I was trying to figure out if the fsm could be useful later, but my first take was that it's probably overkill here as well.

dscleaver:
Looks good from the "it does what we wanted to do" perspective

modules/api/src/test/scala/vinyldns/api/backend/CommandHandlerSpec.scala
mitruly: Added a couple of tests here.

modules/api/src/main/scala/vinyldns/api/engine/BatchChangeHandler.scala
mitruly: Makes sense; stripped out the FSM stuff.

PR #905, opened 2019-11-14
validate CNAME record data is not an IP address in DNS Change form

britneywright:
Screen Shot 2019-11-14 at 11 19 47 AM

Changes in this pull request:
- add an invalid IP check to the CNAME record data field in the DNS Change form. If the record data is an IPV4 address (with or without a trailing dot) return an error that states "CNAME data must be a domain name, not an IP address!" Opted for this new message as opposed to tying it into the existing FQDN validation in order to provide users with more context about why the data they provided is invalid, and help them understand that CNAMES are for FQDNS not IP addresses. 

Intentionally kept the scope of this PR small, but we should make additional changes to the FQDN regex in the portal and the API to not allow IP addresses to be considered valid FQDNs. It's a little tricky right now between the regex and the way we handle adding trailing dots. I believe we add or ensure there's a trailing dot before we do the FQDN regex validation in the API. So `1.1.1.1` would normally fail but by the time it gets to the validation it's `1.1.1.1.` which passes.

### DIFF
--- a/modules/portal/app/views/dnsChanges/dnsChangeNew.scala.html
+++ b/modules/portal/app/views/dnsChanges/dnsChangeNew.scala.html
@@ -168,8 +168,8 @@
                                         <input name="record_cname_{{$index}}" type="text" ng-model="change.record.cname" ng-required="change.changeType=='Add'" class="form-control" placeholder="e.g. test.example.com." ng-disabled="change.changeType=='DeleteRecordSet'" fqdn invalidip>
                                         <p ng-show="createBatchChangeForm.$submitted">
                                             <span ng-show="createBatchChangeForm.record_cname_{{$index}}.$error.required" class="batch-change-error-help">Record data is required!</span>
-                                            <span ng-show="createBatchChangeForm.record_cname_{{$index}}.$error.fqdn" class="batch-change-error-help">CNAME data must be a full domain name!</span>
-                                            <span ng-show="createBatchChangeForm.record_cname_{{$index}}.$error.invalidip" class="batch-change-error-help">CNAME data must be a domain name, not an IP address!</span>
+                                            <span ng-show="createBatchChangeForm.record_cname_{{$index}}.$error.fqdn" class="batch-change-error-help">CNAME data must be a fully qualified domain name!</span>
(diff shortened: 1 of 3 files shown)


britneywright:
Users seem to mostly run into this issue because they mean to create an A record rather than a CNAME. we could also include in the error message something like "If you're record data is an IP address you might be trying to make an A record rather than a CNAME", but in a more concise way.

modules/portal/app/views/dnsChanges/dnsChangeNew.scala.html
```diff
@@ -165,10 +165,11 @@ <h4>Changes</h4>
...
+                                        <input name="record_cname_{{$index}}" type="text" ng-model="change.record.cname" ng-required="change.changeType=='Add'" class="form-control" placeholder="e.g. test.example.com." ng-disabled="change.changeType=='DeleteRecordSet'" fqdn invalidip>
                                         <p ng-show="createBatchChangeForm.$submitted">
                                             <span ng-show="createBatchChangeForm.record_cname_{{$index}}.$error.required" class="batch-change-error-help">Record data is required!</span>
-                                            <span ng-show="createBatchChangeForm.record_cname_{{$index}}.$error.fqdn" class="batch-change-error-help">CNAME data must be absolute!</span>
+                                            <span ng-show="createBatchChangeForm.record_cname_{{$index}}.$error.fqdn" class="batch-change-error-help">CNAME data must be a full domain name!</span>
```
mitruly:
```suggestion
                                            <span ng-show="createBatchChangeForm.record_cname_{{$index}}.$error.fqdn" class="batch-change-error-help">CNAME data must be a fully-qualified domain name!</span>
```

modules/portal/public/lib/dns-change/dns-change.directive.js
```diff
@@ -17,6 +17,7 @@
 (function () {
     var app = angular.module('dns-change');
     var FQDN_REGEX = /\./;
+    var INVALID_FQDN_REGEX = /^(?:[0-9]{1,3}\.){3}[0-9]{1,3}\.?$/;
```
mitruly: This regex isn't entirely accurate, as `999.999.999.999` would get flagged (despite not being a valid IP). Is there a way that we can have a "base" regex for the IPv4 matching and then compose the `IPV4_REGEX` and `INVALID_FQDN_REGEX` from it?

PR #907, opened 2019-11-21
retry some failed record changes

### LINKED ISSUE #842: Implement backoff+retry for each phase of record change handling
### Background
The `RecordSetChangeHandler` is the state machine "engine" that processes validated requests that are queued for processing. There are two targets for improvement, but the design will focus on implementing retry with backoff:
1. Retry change when receiving "recoverable" DNS error responses from DNS backend: when applying the requested change, we interpret any type of error including `SocketTimeout`, `TryAgain`, `Refused`, `ServerFailure` and `DnsFailure` as immediate stops and return a failure response back to the user. We should relax our constraints and retry given certain DNS responses.
~2. Our retry policy when verifying that a change has been made: We currently verify whether a change has successfully completed, retrying the query every 500ms up to a total of 10 times (for a total of 5s) before responding to the user. Taking traffic into account, we should realistically be re-queuing the request with an increased retry timeout to free up resources for a new change.~

### Current process flow
Current Record Set Change Handler

Currently, the `RecordSetChangeHandler` follows the current process flow to process successfully validated requests:
1. Fetch a change request from the message queue.

1. `Validate` that the change has not yet been implemented by querying the DNS backend
  - If the change has _already_ been implemented, indicate `AlreadyApplied` and return with a successful response.
  - If the state of the record is in good shape (i.e. no conflict with the proposed request), indicate `ReadyToApply` and move on to `Apply` step.
  - If there are conflicting issues with the request and state of the DNS backend (e.g. create request for a record that already exists in the DNS backend, delete request for a record that does not exist in the DNS backend, update request has existing records as exact match against DNS backend), return a `Failure` response back to the user.

1. `Apply` the requested change to the DNS backend
  - If there are any errors, indicate that the request completed with a failure.
  - If there are no errors, indicate that the request has been `Applied` and proceed to the `Verify` step.

1. `Verify` that the requested change was applied to the DNS backend
  - If the change has been applied, return a successful response to the user.
  - If the change has not yet been applied, retry query every 500ms up to a total of 10 times. If change still has not been applied after retries have been exhausted, return a failure response.
  - If there is a failure response, bubble it back to the user.

### Proposed process flow
Improved Record Set Change Handler
1. Fetch a change request from the message queue. If the change has exceeded the maximum retry count, remove the change from the queue and return a failure response back to the user. Otherwise, proceed to the `Validate` step.

1. `Validate` that the change has not yet been implemented by querying the DNS backend
  - If the change has _already_ been implemented, indicate `AlreadyApplied` and return with a successful response
  - If the state of the record is in good shape (i.e. no conflict with the proposed request), indicate `ReadyToApply` and move on to `Apply` step.
  - If there are conflicting issues with the request and state of the DNS backend (e.g. create request for a record that already exists in the DNS backend, delete request for a record that does not exist in the DNS backend, update request has existing records as exact match against DNS backend), return a `Failure` response back to the user. _Also_, if the request is out or retries, return an out-of-retries failure response back to the user.
- If there are recoverable errors or any IO exceptions, increase retry counter for message and re-queue.

1. `Apply` the requested change to the DNS backend
  - If there are any unrecoverable errors (eg. SRVFAIL, etc), indicate that the request completed with a failure.
  - If there are recoverable errors or any IO exceptions, increase retry counter for message and re-queue.
  - If there are no errors, indicate that the request has been `Applied` and proceed to the `Verify` step.

1. `Verify` that the requested change was applied to the DNS backend
  - If the change has been applied, return a successful response to the user and delete the message from the queue.
  - If the change has not yet been applied, increase the retry counter and re-queue.
  - If there is an unrecoverable DNS error response, bubble it back to the user and delete the message from the queue.

Note: The `Validate`-`Apply`-`Verify` steps are collectively a higher-order function that will be retried in its entirety.

#### Add retry to apply step with exponential backoff
We can have a retry loop similar to the existing `Verify` step where we are retrying DNS changes when receiving recoverable DNS error responses. Updating the retry policies to use exponential backoff would achieve the same desired behavior. However, this adds more potential for resources to get tied up.

#### Delegate retry policy to message queue
We can simplify the retry counts by relying on the queue retry policy. However, 1. the queue implementation being used is not guaranteed to support automatic retry/dead-letter, and 2. we would have to poll the dead-letter queue to send back failure responses to the user. The trade-off is not worthwhile.

### Summary
With the new design considerations, we are increasing the success rate of VinylDNS processing against the DNS backend due to 1. scenarios in which we are reattempting DNS changes and 2. the span of time across the DNS reattempts are being retried. We are also freeing up resources to process queued messages (rather than suspending threads waiting for retry timeouts).

The main compelling reason the proposed design, is that there shouldn't be an issue attempting to reapply a change.

#### Update VinylDNS responses
Currently there are two `DnsConnection` operations that interact with the DNS backend:
- `applyChange`: attempts to apply the change to DNS backend. Possible values include: `NOERROR`, `FORMERR`, `SERVFAIL`, `NXDOMAIN`, `NOTIMP`, `REFUSED`, `YXDOMAIN`, `YXRRSET`, `NXRRSET`, `NOTAUTH`, `NOTZONE`, `BADVERS`, `BADSIG`, `BADKEY`, `BADTIME`, `BADMODE`

- `resolve`: queries the backend for a DNS record response. Possible values include: `SUCCESSFUL`, `UNRECOVERABLE`, `TRY_AGAIN`, `HOST_NOT_FOUND`, `TYPE_NOT_FOUND`

We need to introduce new wrapper error classes rather than just using a "catch-all" for `Throwable`:
```
sealed abstract class ApplyChangeError(cause: Throwable) extends Throwable(cause)
final case class UnrecoverableApplyChangeError(cause: Throwable) extends ApplyChangeError(cause)
final case class RecoverableApplyChangeError(cause: Throwable) extends ApplyChangeError(cause)

sealed abstract class ResolutionError(cause: Throwable) extends Throwable(cause)
final case class UnrecoverableResolutionError(cause: Throwable) extends ResolutionError(cause)
final case class RecoverableResolutionError(cause: Throwable) extends ResolutionError(cause)
```

#### Update `MySqlMessageQueue`
We currently hard-code `100` value for the maximum retry attempts. This needs to be configurable.

#### Update the RecordSetChangeHandler finite-state machine (fsm)
`RecoverableApplyChangeError`s: `REFUSED`
`RecoverableResolutionError`s: `TRY_AGAIN`

In `RecordSetChangeHandler`, update the fsm to:
- Re-queue the message if a recoverable error is encountered _or_ there is a failure with the promise (i.e. `IO`). If there is no retry count attribute on the message, we should assume that this is the first `attempt` and set the new `attempts` value to 2.
- Remove message from the queue during fetch stage if `attempts` count > maximum retries.

New outcomes need to be added for the FSM result:
- `Succeeded`: Remove message from queue and return good response to the user
- `Failed`: Remove message from queue and return bad response to the user
- `Retry`: Do nothing

britneywright:
close #842 

Retry failed record changes if the record change error is Refused during the apply step of the finite state machine, or if the record change error is Try Again during the verify step of the finiite state machine in the RecordSetChangeHandler. Raising an error in the RecordSetChangeHandler causes the change to go back on the MySqlMessageQueue.

Tested against a commonly failing scenario, changing an A record to a CNAME through a batch change. Currently the A record deletes successfully but the CNAME fails in the backend. With the re-queue the CNAME add appears to succeed.

record statuses based on current flow:
Screen Shot 2019-11-20 at 9 12 29 PM

record statuses based on updated flow:
Screen Shot 2019-11-20 at 9 12 46 PM

Need to add/update more tests.

### DIFF
--- a/modules/api/functional_test/vinyldns_python.py
+++ b/modules/api/functional_test/vinyldns_python.py
@@ -26,7 +26,7 @@ logger = logging.getLogger(__name__)
 
 __all__ = [u'VinylDNSClient', u'MAX_RETRIES', u'RETRY_WAIT']
 
-MAX_RETRIES = 60
+MAX_RETRIES = 75
 RETRY_WAIT = 0.05
 
 class VinylDNSClient(object):
--- a/modules/api/functional_test/vinyldns_python.py
+++ b/modules/api/functional_test/vinyldns_python.py
@@ -26,7 +26,7 @@ logger = logging.getLogger(__name__)
 
 __all__ = [u'VinylDNSClient', u'MAX_RETRIES', u'RETRY_WAIT']
 
-MAX_RETRIES = 40
+MAX_RETRIES = 100
 RETRY_WAIT = 0.05
 
 class VinylDNSClient(object):
--- a/modules/api/functional_test/vinyldns_python.py
+++ b/modules/api/functional_test/vinyldns_python.py
@@ -26,7 +26,7 @@ logger = logging.getLogger(__name__)
 
 __all__ = [u'VinylDNSClient', u'MAX_RETRIES', u'RETRY_WAIT']
 
-MAX_RETRIES = 100
+MAX_RETRIES = 40
 RETRY_WAIT = 0.05
 
 class VinylDNSClient(object):
--- a/modules/api/functional_test/vinyldns_python.py
+++ b/modules/api/functional_test/vinyldns_python.py
@@ -26,7 +26,7 @@ logger = logging.getLogger(__name__)
 
 __all__ = [u'VinylDNSClient', u'MAX_RETRIES', u'RETRY_WAIT']
 
-MAX_RETRIES = 40
+MAX_RETRIES = 60
 RETRY_WAIT = 0.05
 
 class VinylDNSClient(object):
--- a/modules/api/src/main/scala/vinyldns/api/engine/RecordSetChangeHandler.scala
+++ b/modules/api/src/main/scala/vinyldns/api/engine/RecordSetChangeHandler.scala
@@ -20,20 +20,18 @@ import cats.effect.{ContextShift, IO, Timer}
 import cats.implicits._
 import org.slf4j.LoggerFactory
 import vinyldns.api.domain.dns.DnsConnection
-import vinyldns.api.domain.dns.DnsProtocol.NoError
+import vinyldns.api.domain.dns.DnsProtocol.{NoError, Refused, TryAgain}
 import vinyldns.api.domain.record.RecordSetHelpers._
 import vinyldns.core.domain.batch.{BatchChangeRepository, SingleChange}
 import vinyldns.core.domain.record._
 
-import scala.concurrent.ExecutionContext
-import scala.concurrent.duration._
-
 object RecordSetChangeHandler {
 
   private val logger = LoggerFactory.getLogger("vinyldns.api.engine.RecordSetChangeHandler")
-  private implicit val timer: Timer[IO] = IO.timer(ExecutionContext.global)
   private implicit val cs: ContextShift[IO] =
     IO.contextShift(scala.concurrent.ExecutionContext.global)
+
+  private final case class Requeue(change: RecordSetChange) extends Throwable
 
   def apply(
       recordSetRepository: RecordSetRepository,
@@ -96,6 +94,8 @@ object RecordSetChangeHandler {
 
   private final case class Completed(change: RecordSetChange) extends ProcessorState
 
+  private final case class Retrying(change: RecordSetChange) extends ProcessorState
+
   private case class ProcessingError(message: String)
 
   sealed trait ProcessingStatus
@@ -110,6 +110,9 @@ object RecordSetChangeHandler {
   // ReadyToApply will attempt to retry until max retries limit is reached,
   // at which point the response will be returned.
   final case class ReadyToApply(change: RecordSetChange) extends ProcessingStatus
+
+  // Failure to process change. Permitted to retry.
+  final case class Retry(change: RecordSetChange) extends ProcessingStatus
 
   def getProcessingStatus(change: RecordSetChange, dnsConn: DnsConnection): IO[ProcessingStatus] = {
     def isDnsMatch(dnsResult: List[RecordSet], recordSet: RecordSet, zoneName: String): Boolean =
@@ -145,6 +148,7 @@ object RecordSetChangeHandler {
             if (existingRecords.nonEmpty) ReadyToApply(change) // we have a record set, move forward
             else AlreadyApplied(change) // we did not find the record set, so already applied
         }
+      case Left(_: TryAgain) => Retry(change)
       case Left(error) => Failure(change, error.getMessage)
     }
   }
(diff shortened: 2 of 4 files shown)


modules/api/src/main/scala/vinyldns/api/engine/RecordSetChangeHandler.scala
```diff
@@ -225,32 +236,27 @@ object RecordSetChangeHandler {
...
-          IO.sleep(500.milliseconds) *> loop(retries - 1)
-        case Failure(_, message) =>
-          IO.pure(
-            Completed(
-              change.failed(
-                s"Failed verifying update to DNS for change ${change.id}:${change.recordSet.name}: $message"
-              )
+  /* Step 3: Verify the record was created. */
+  private def verify(change: RecordSetChange, dnsConn: DnsConnection): IO[ProcessorState] =
+    getProcessingStatus(change, dnsConn).flatMap {
```
mitruly: We don't need to `flatMap` with the removal of nested `loop`; this can be consolidated to:
```
    getProcessingStatus(change, dnsConn).map {
      case AlreadyApplied(_) => Completed(change.successful)
      case ReadyToApply(_) | Retry(_) => Retrying(change)
      case Failure(_, message) =>
        Completed(
          change.failed(
            s"Failed verifying update to DNS for change ${change.id}:${change.recordSet.name}: $message"
          )
        )
    }
```

britneywright:
I added to the existing batch change func test that tests updating records. Made an A record then tested deleting that A record and adding a CNAME with the same name. Before the RecordSetChangeHandler updates the result was a PartialFailure, after the updates it was Complete. I did have to increase the MAX_RETRIES in the func tests. I changed from 40 to 100 arbitrarily, but i'll test to see how low it can be and pass consistently.

modules/api/functional_test/vinyldns_python.py
```diff
@@ -26,7 +26,7 @@
 
 __all__ = [u'VinylDNSClient', u'MAX_RETRIES', u'RETRY_WAIT']
 
-MAX_RETRIES = 40
+MAX_RETRIES = 100
```
pauljamescleary: oof, I am going to go out on a limb here and say this is something we do not want to e2e test, as we just spent a lot of time speeding up our builds.  Perhaps just leave it to unit tests down below where we have more control and tests are fast

britneywright: For what it's worth the run time seems to be about the same with this change.

mitruly: Think having a test should be fine @pauljamescleary since in our SQS requeue, we set to 10 seconds (instead of defaulting to default visibility timeout). That said, having this as a separate test for changing a record type in our functional tests would probably be better (1. so we don't have to increase `MAX_RETRIES` and 2. so the test can run in parallel with others).

britneywright:
I'll move the update record type to a separate func test and see if that allows me to bring down the retries.

britneywright:
The func tests passed once with the MAX_RETRIES back at 40, but then i rebased and they continually failed. Also failed at 60. Passed at 75 once. will rerun them couple times to confirm it's consistently passing.

PR #908, opened 2019-11-25
reject recordset names with spaces

britneywright:
When making a new recordset (not through batch change/DNS Change) if a user includes spaces the record name is accepted and the space is transformed to `\32` in the record name. This is unintended, we should not accept spaces in any recordset names.

Changes in this pull request:
- add a recordset validation to check if the given record name contains spaces

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetValidations.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetValidations.scala
@@ -42,15 +42,6 @@ object RecordSetValidations {
         )
     }
 
-  def validRecordName(recordSet: RecordSet): Either[Throwable, Unit] =
-    ensuring(
-      InvalidRequest(
-        s"RecordSet name is not valid."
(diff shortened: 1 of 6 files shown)


modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetValidations.scala
```diff
@@ -42,6 +42,15 @@ object RecordSetValidations {
         )
     }
 
+  def validRecordName(recordSet: RecordSet): Either[Throwable, Unit] =
```
pauljamescleary: could this more simply be in the `regex` in the `DnsProtocol` up front?

I know that feels weird, I am just unclear on where most of these kinds of validations should live.

I imagine regex ones should be in the json validations, and more complex ones live back here

britneywright: good point, moved it to the `DnsJsonProtocol`

PR #909, opened 2019-11-29
make Batch Change AAAA record data flexible

britneywright:
For Batch Changes we now accept record data for DeleteRecordSet to delete only specific records in a recordset rather than the entire recordset. We do an exact match on the existing records in the recordset which is fine for most record data types, except IPv6 address for AAAA records. They can be written in many forms and we should match our existing records across any given format.

Changes in this pull request:
- try to match the given IPv6 address with existing record data in `BatchChangeValidations`
- choose the existing record data that is the equivalent of the given record data in `BatchTransformations`

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
@@ -297,9 +297,9 @@ class BatchChangeValidations(
     else
       ().validNel
 
-  def compareIpv6Addresses(existingRecordSetData: List[AAAAData], recordData: AAAAData): Boolean =
-    existingRecordSetData.exists { rd =>
-      InetAddress.getByName(recordData.address).getHostName == InetAddress
+  def matchIpv6Address(existingRecordSetData: List[RecordData], address: String): Boolean =
+    existingRecordSetData.asInstanceOf[List[AAAAData]].exists { rd =>
+      InetAddress.getByName(address).getHostName == InetAddress
         .getByName(rd.address)
         .getHostName
     }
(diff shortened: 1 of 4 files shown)


modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
```diff
@@ -295,12 +297,29 @@ class BatchChangeValidations(
     else
       ().validNel
 
+  def matchIpv6Address(existingRecordSetData: List[RecordData], address: String): Boolean =
```
mitruly: It's better to use pattern matching than casting (using `asInstanceOf`):
```
  def matchIpv6Address(existingRecordSetData: List[RecordData], address: String): Boolean =
    existingRecordSetData
      .exists {
        case AAAAData(existingAddress) =>
          InetAddress.getByName(address).getHostName == InetAddress
            .getByName(existingAddress)
            .getHostName
        case _ => false
      }
```

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
```diff
@@ -295,12 +297,29 @@ class BatchChangeValidations(
...
     change match {
       // For DeleteRecord inputs, need to verify that the record data actually exists
+      case DeleteRRSetChangeForValidation(
+          _,
+          _,
+          DeleteRRSetChangeInput(inputName, AAAA, Some(AAAAData(address)))
+          ) =>
+        if (groupedChanges
+            .getExistingRecordSet(change.recordKey)
+            .exists(rs => matchIpv6Address(rs.records, address))) {
```
mitruly: We should be able to consolidate these two cases here (using something similar to `convertedRecordData`).

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchTransformations.scala
```diff
@@ -197,6 +198,16 @@ object BatchTransformations {
       // existing DNS entries in the event of DeleteRecordSet
       val deleteChangeSet = changes
         .collect {
+          case DeleteRRSetChangeForValidation(
```
mitruly: We don't need to add this case here; can use the existing one below with the addition of a helper `convertedRecordData`:
```
  def convertedRecordData(
      existingRecords: Set[RecordData],
      recordData: RecordData
  ): Set[RecordData] =
    recordData match {
      case AAAAData(address) =>
        existingRecords.collect {
          case AAAAData(existingAddress)
              if InetAddress.getByName(existingAddress).getHostName ==
                InetAddress.getByName(address).getHostName =>
            AAAAData(existingAddress)
          case _ =>
            recordData
        }
      case _ => Set(recordData)
    }
```

This case would turn into something like:
```
          case DeleteRRSetChangeForValidation(
              _,
              _,
              DeleteRRSetChangeInput(_, _, Some(recordData))
              ) =>
            convertedRecordData(existingRecords, recordData)
```

britneywright: i don't like the nested case statements here, but I did refactor what I had so I'm not using `asInstanceOf`.

mitruly: @britneywright think I'm good with these changes.

PR #910, opened 2019-11-29
Add max retries to MySQL queue configuration

mitruly:
Changes in this PR:
- Add new optional `maxRetries` config value for `MySqlMessageQueue` with default value of 100
- Update example config in documentation

----
Note: If encountering a config error during API boot, try doing a clean (`sbt ";project api ;clean ;run"`).

### DIFF
--- a/modules/mysql/src/main/scala/vinyldns/mysql/queue/MySqlMessageQueue.scala
+++ b/modules/mysql/src/main/scala/vinyldns/mysql/queue/MySqlMessageQueue.scala
@@ -44,12 +44,18 @@ object MySqlMessageQueue {
   final case class InvalidMessageTimeout(timeout: Int)
       extends Throwable(s"Invalid message timeout $timeout")
   final val QUEUE_CONNECTION_NAME = 'queue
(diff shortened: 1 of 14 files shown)


modules/mysql/src/main/scala/vinyldns/mysql/queue/MySqlMessageQueue.scala
```diff
@@ -44,13 +44,19 @@ object MySqlMessageQueue {
   final case class InvalidMessageTimeout(timeout: Int)
       extends Throwable(s"Invalid message timeout $timeout")
   final val QUEUE_CONNECTION_NAME = 'queue
+  final val DEFAULT_MAX_RETRIES = 100
 }
 
-class MySqlMessageQueue extends MessageQueue with Monitored with ProtobufConversions {
+class MySqlMessageQueue(maxRetries: Option[Int])
```
britneywright: Should we put the default value in https://github.com/vinyldns/vinyldns/blob/master/modules/api/src/main/resources/reference.conf

britneywright:
Looks like `MessageQueueLoaderSpec` needs updates

PR #911, opened 2019-12-02
Coursier plugin update 

raghu999:
Fixes #.
We are running into few errors while running `sbt:verify` through our CI/CD. This PR contains a change to upgrade Coursier sbt build plugin  which will fix the `OverlappingFileLockException`. Copying the stacktrace below for your reference

Changes in this pull request:
- Update `sbt-coursier` to latest  `2.0.0-RC4-1`
(Coursier Docs)[https://get-coursier.io/docs/sbt-coursier]

```
Downloaded https://repo1.maven.org/maven2/com/ironcorelabs/cats-scalatest_2.12/2.3.1/
[error] coursier.ResolutionException: Exception during resolution
[error] Caused by: java.nio.channels.OverlappingFileLockException
[error] 	at sun.nio.ch.SharedFileLockTable.checkList(FileLockTable.java:255)
``

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -38,4 +38,4 @@ addSbtPlugin("org.xerial.sbt" % "sbt-sonatype" % "2.3")
 
 addSbtPlugin("io.crashbox" % "sbt-gpg" % "0.2.0")
 
-addSbtPlugin("io.get-coursier" % "sbt-coursier" % "1.0.3")
+addSbtPlugin("io.get-coursier" % "sbt-coursier" % "2.0.0-RC4-1")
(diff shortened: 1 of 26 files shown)

### CODE REVIEW

File: project/plugins.sbt [hunk 1]
```diff
@@ -38,4 +38,4 @@ addSbtPlugin("org.xerial.sbt" % "sbt-sonatype" % "2.3")
 
 addSbtPlugin("io.crashbox" % "sbt-gpg" % "0.2.0")
 
-addSbtPlugin("io.get-coursier" % "sbt-coursier" % "1.0.3")
+addSbtPlugin("io.get-coursier" % "sbt-coursier" % "2.0.0-RC4-1")
```
mitruly: I think I'm alright if this passes, though the current version (`1.0.3`) is the most recent _full_ release for this plugin; the interesting thing is that I haven't encountered this error when running locally or within certain containers so I wonder if there is some other issue at play with the error.

Note: As this is a release candidate, we should promptly update this to `2.0.0` when that is finally released.

PR #912, opened 2019-12-02
Zone recordset search enhancements

### LINKED ISSUE #896: Record set search by zone enhancements
From #861 discussion.

## Existing API Endpoint:

```
GET /zones/:zoneId/recordsets
```

## Current Request Parameters

 | Name | Type | Required | Description |
 |------------|------|---|------|
 | recordNameFilter  | string  | No | Characters that are part of the record name to search for.  |
  startFrom  | any  | No | In order to advance through pages of results, the startFrom is set to the nextId that is returned on the previous response. |
 | maxItems  | integer  | No | The number of items to return in the page. Valid values are 1 to 100. Default is 100. |

## Proposed Request Parameters

 | Name | Type |  Required | Description |
 |------------|------|---|------|
 | recordTypeFilter | string | No | A list of comma separated record types. Won't filter results by record type if none are provided. |
| sort | string | No | Direction of sort by record name. Either `asc` or `desc`. Default is `asc`|

## API Changes
RecordSetRouting:
- update the `getRecordSets` Endpoint to include recordTypes, and sort as optional parameters.
- set the sort default value to "asc"
- validate that given sort is either "desc" or "asc" (case-insensitive) and convert to SortDirection enumerable (need to create)
- validate that given recordTypes are valid and convert to an optional list of RecordTypes
- rename `getRecordSets` to `getRecordSetsByZone`
- update `ListRecordSetsResponse`

RecordSetService:
- Update `listRecordSets` to include new parameters

RecordSetServiceAlgebra:
- Update `listRecordSets` to include new parameters

RecordSetRepository:
- Update `listRecordSets` to include new parameters

MySqlRecordSetRepository:
- Update `listRecordSets` to include new parameters

DynamoDBRecordSetRepository:
- Update `listRecordSets` to include new parameters

## Portal Changes
VinylDNS.scala:
- Rename `getRecordSets` to `getRecordSetsByZone` include new parameters
- Change parameter from `id` to `zId`

routes:
- Change `@controllers.VinylDNS.getRecordSets(id: String)` to  `@controllers.VinylDNS.getRecordSetsByZone(zid: String)`

service.records.js:
- Change `getRecordSets` to `getRecordSetsById`

controller.records.js:
- Update recordService calls from `getRecordSets` to `getRecordSetsById`
- Include new query parameters

manageRecords.scala.html:
- Add sort on Name column. Default to ascending.
- Add filter by record type. List of checkboxes. Default is All. Consider only showing record types that are valid in the given zone. For example don't show PTR records as an option in forward zones.

## Documentation Changes
API:
- Add new parameters to List / Search RecordSets

Portal:
- Update screenshots on Manage Records page

## Tooling Changes:
The tools can be updated at a later point as the current query parameters aren't changing in terms of requirements or type and the new query parameters are optional. These are initial recommendations, may not include all files and functions that require updates.

- go-vinyldns:
  update `RecordSetsListAll` in [recordsets.go](https://github.com/vinyldns/go-vinyldns/blob/master/vinyldns/recordsets.go)
 update `RecordSetCollector` in [recordsets.go](https://github.com/vinyldns/go-vinyldns/blob/master/vinyldns/recordsets.go)
update `RecordSetsResponse` in [recordsets.go](https://github.com/vinyldns/go-vinyldns/blob/master/vinyldns/recordsets_resources.go) 
may need to to create recordset specific ListFilter see [resources.go](https://github.com/vinyldns/go-vinyldns/blob/master/vinyldns/recordsets_resources.go) 
consider `RecordSetsListAll` naming as it will relate to future recordset search outside of zones

- vinyldns-cli:
  don't think changes are needed as it calls out to go-vinyldns
  consider `recordSets` naming as it will relate to future recordset search outside of zones

- vinyldns-js:
  add new query parameters for `getRecordSets` in [vinyldns.js](https://github.com/vinyldns/vinyldns-js/blob/master/src/vinyldns.js#L104-L114)
 consider `getRecordSets` naming as it will relate to future recordset search outside of zones

- vinyldns-java:
   update [ListRecordSetsRequest](https://github.com/vinyldns/vinyldns-java/blob/master/src/main/java/io/vinyldns/java/model/record/set/ListRecordSetsRequest.java)
  update [ListRecordSetsResponse](https://github.com/vinyldns/vinyldns-java/blob/master/src/main/java/io/vinyldns/java/model/record/set/ListRecordSetChangesResponse.java)
  consider `ListRecordSetsRequest` and `ListRecordSetsResponse` naming as it will relate to future recordset search outside of zones

- terraform-provider-vinyldns:
  don't think changes are needed
 
- vinyldns-ruby:
  update RecordSet search in [zone.rb](https://github.com/vinyldns/vinyldns-ruby/blob/master/lib/vinyldns/api/zone/zone.rb#L107-L112)

- vinyldns-python:
  update `ListRecordSetsResponse` in [record.py](https://github.com/vinyldns/vinyldns-python/blob/master/src/vinyldns/record.py#L211-L227)
  update `list_record_sets` in [client.py](https://github.com/vinyldns/vinyldns-python/blob/master/src/vinyldns/client.py#L545-L566)
  consider `list_record_sets` naming as it will relate to future recordset search outside of zones

britneywright:
Fixes #896

Changes in this pull request:
- Add filter by record type and a sort order (by recordset name) to list recordsets by zone endpoint.
- Updated endpoint and integrated in the portal

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/route/RecordSetRouting.scala
+++ b/modules/api/src/main/scala/vinyldns/api/route/RecordSetRouting.scala
@@ -104,10 +104,6 @@ class RecordSetRoute(
               validate(
                 0 < maxItems && maxItems <= DEFAULT_MAX_ITEMS,
                 s"maxItems was $maxItems, maxItems must be between 0 and $DEFAULT_MAX_ITEMS"
-              )
-              validate(
-                "ASC" == sort.toUpperCase || "DESC" == sort.toUpperCase,
-                s"""sort was $sort, sort must be "ASC" or "DESC"."""
               ) {
                 authenticateAndExecute(
                   recordSetService
--- a/modules/api/src/main/scala/vinyldns/api/route/RecordSetRouting.scala
+++ b/modules/api/src/main/scala/vinyldns/api/route/RecordSetRouting.scala
@@ -22,7 +22,8 @@ import akka.util.Timeout
 import vinyldns.api.Interfaces._
 import vinyldns.api.domain.record.RecordSetServiceAlgebra
 import vinyldns.api.domain.zone._
-import vinyldns.core.domain.record.RecordSet
+import vinyldns.core.domain.record.RecordType.RecordType
+import vinyldns.core.domain.record.{RecordSet, RecordType}
 import vinyldns.core.domain.zone.ZoneCommandResult
 
 import scala.concurrent.duration._
@@ -34,7 +35,7 @@ case class ListRecordSetsResponse(
     nextId: Option[String] = None,
     maxItems: Option[Int] = None,
     recordNameFilter: Option[String] = None,
-    recordTypeFilter: Option[String] = None,
+    recordTypeFilter: Option[Set[RecordType]] = None,
     sort: String
 )
 
@@ -78,7 +79,7 @@ class RecordSetRoute(
           "maxItems".as[Int].?(DEFAULT_MAX_ITEMS),
           "recordNameFilter".?,
           "recordTypeFilter".?,
-          "sort".as[String].?("asc")
+          "sort".as[String].?("ASC")
         ) {
           (
               startFrom: Option[String],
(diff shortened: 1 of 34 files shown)


modules/api/src/main/scala/vinyldns/api/route/RecordSetRouting.scala
```diff
@@ -71,16 +75,51 @@ class RecordSetRoute(
...
+          "sort".as[String].?("ASC")
+        ) {
+          (
+              startFrom: Option[String],
+              maxItems: Int,
+              recordNameFilter: Option[String],
+              recordTypeFilter: Option[String],
+              sort: String
+          ) =>
+            val convertedRecordTypeFilter = recordTypeFilter match {
```
mitruly: This can be consolidated to:
`val convertedRecordTypeFilter = recordTypeFilter.map(_.split(",").flatMap(RecordType.find).toSet)`

britneywright: this doesn't work if only invalid record types are given.

britneywright: it can be consolidated differently but will still require a case statement or if statement at some point.

PR #914, opened 2019-12-04
fix user lock status in group response

britneywright:
Noticed our get group endpoint was returning all user lock statuses with the value of Unlocked even if the user is Locked. Looked into this more closely and realized we're using the UserInfo class as part of the response. In the UserInfo class all fields are optional except ID and LockStatus, but when retrieving a group we only have the member IDs so I set a default LockStatus of Unlocked on initial implementation. 

In reality if we're only handling user IDs in certain instances we should just be using a type that only includes the user ID as an attribute. I added UserID that is used in the get, create and update group endpoints. And updated UserInfo to not have a default LockStatus. This also removes LockStatus from the get group endpoint which matches what our documentation currently states the response is.

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipProtocol.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipProtocol.scala
@@ -43,8 +43,8 @@ object GroupInfo {
     description = group.description,
     created = group.created,
     status = group.status,
-    members = group.memberIds.map(UserId(_)),
-    admins = group.adminUserIds.map(UserId(_))
+    members = group.memberIds.map(UserId),
+    admins = group.adminUserIds.map(UserId)
   )
 }
 
@@ -68,15 +68,7 @@ object GroupChangeInfo {
   )
 }
 
-case class UserId(
-    id: String
-)
-object UserId {
-  def apply(user: User): UserId =
-    UserId(
-      id = user.id
-    )
-}
+case class UserId(id: String)
 
 case class UserInfo(
     id: String,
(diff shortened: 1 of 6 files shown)


modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipProtocol.scala
```diff
@@ -43,8 +43,8 @@ object GroupInfo {
     description = group.description,
     created = group.created,
     status = group.status,
-    members = group.memberIds.map(UserInfo(_)),
-    admins = group.adminUserIds.map(UserInfo(_))
+    members = group.memberIds.map(UserId(_)),
```
mitruly: `(_)` can be removed here (i.e. `members = groups.memberIds.map(UserID)`).

modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipProtocol.scala
```diff
@@ -68,14 +68,24 @@ object GroupChangeInfo {
   )
 }
 
+case class UserId(
+    id: String
+)
+object UserId {
```
mitruly: Don't think we need this since it's not used in main source code.

modules/api/src/test/scala/vinyldns/api/route/MembershipRoutingSpec.scala
```diff
@@ -64,6 +64,7 @@ class MembershipRoutingSpec
   // this is avoided since were working with 0s there anyway
   val baseTime = new DateTime(0)
   val okUserInfo: UserInfo = UserInfo(okUser).copy(created = Some(baseTime))
+  val okUserId: UserId = UserId(okUser)
```
mitruly: Can change this to `val okUserId: UserId = UserId(okUser.id)`.

PR #915, opened 2019-12-05
[ci skip] Documentation for zone record set sort

mitruly:
Documentation for record set sort feature work (https://github.com/vinyldns/vinyldns/pull/912).

Changes in this pull request:
- Add information for record sort/filtering
- Update manage record portal images


modules/docs/src/main/tut/portal/manage-records.md
```diff
@@ -35,3 +35,10 @@ The *Additional Info* column will contain details of the change failure.
 If you make changes to the records in a zone outside of VinylDNS you must manually sync the zone so VinylDNS is aware of those records. Use the *Sync Zone* button in the *Records* pane.
 
 [![Sync zones screenshot](../img/portal/sync-zones-annotated.png){: .screenshot}](../img/portal/sync-zones-annotated.png)
+
+### Record sorting and filtering
+It is now possible to sort records in VinylDNS by record name ascending/descending order in addition to filtering the record types that will be displayed, providing more ease and control when navigating your DNS records.
+To use this functionality, make sure that you are on the *Manage Records* tab. To change the sort order for record name, click on the *Name* header in the *Records* table to toggle between ascending and descending order.
+To filter record types, click on the *Filter By Record Type* link to reveal the record type options and adjust your selection.
+  
+[![Sort records screenshot](../img/portal/manage-records-sort.png){: .screenshot}](../img/portal/manage-records-sort.png)
```
britneywright: this screenshot link is broken. Not sure if this section is really necessary, seems fairly intuitive. But if we want to include it I think it should be moved above the Sync Zone material. Also you don't need to say "make sure that you are on the *Manage Records* tab." since that what this doc is about and the user is told to navigate to that tab in the initial information in this page. Also can you update the first screenshot?

britneywright: We should also consider adding information about record set search either in this PR or a separate one, presumably that information would be in the same section. Exact match vs partial match searching using `*`s

PR #916, opened 2019-12-10
update CNAME placeholders in DNS Change form

britneywright:
There's confusion over which domains go in which DNS Change form fields for CNAME records. Both of the placeholder values are currently "test.example.com". We can change the placeholder values to be more helpful.

Changes in this pull request:
- change the Input Name placeholder to "alias.example.com"
- change the Record Data placeholder to "answer.example.com"

### DIFF
--- a/modules/portal/app/views/dnsChanges/dnsChangeNew.scala.html
+++ b/modules/portal/app/views/dnsChanges/dnsChangeNew.scala.html
@@ -116,7 +116,8 @@
                                         </select>
                                     </td>
                                     <td>
-                                        <input ng-if="change.type!=='PTR'" name="inputName_{{$index}}" type="text" ng-model="change.inputName" required="string" class="form-control" placeholder="e.g. test.example.com.">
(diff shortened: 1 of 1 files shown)


modules/portal/app/views/dnsChanges/dnsChangeNew.scala.html
```diff
@@ -165,7 +166,7 @@ <h4>Changes</h4>
                                         <p class="help-block" ng-if="change.changeType=='DeleteRecordSet'">Record Data is optional.</p>
                                     </td>
                                     <td ng-if="change.type=='CNAME'">
-                                        <input name="record_cname_{{$index}}" type="text" ng-model="change.record.cname" ng-required="change.changeType=='Add'" class="form-control" placeholder="e.g. test.example.com." ng-disabled="change.changeType=='DeleteRecordSet'" fqdn invalidip>
+                                        <input name="record_cname_{{$index}}" type="text" ng-model="change.record.cname" ng-required="change.changeType=='Add'" class="form-control" placeholder="e.g. answer.example.com." ng-disabled="change.changeType=='DeleteRecordSet'" fqdn invalidip>
```
mitruly: Think `target.example.com.` is more intuitive than `answer.example.com.`. "answer" is associated with the DNS response that we get back from a DNS query, which doesn't necessarily have to be a CNAME; CNAME record data is typically "target", "alias target" or some variation.

Issue #853, opened 2019-09-30
Improve out-of-sync record handling


mitruly:
### Background
VinylDNS does not currently handle DNS requests well when the database differs from the source of truth (i.e. DNS backend). The current process is that users attempt to change a record that VinylDNS thinks exists or doesn't exist when it does. This results in the user having to sync in changes from the zone and then resubmit the change.

The proposal is to:
1. Check against the DNS backend when making DNS changes (i.e. adds, deletes or updates). If everything is deemed correct, send to the backend for processing.
2. When processing record set changes, we will determine during the `getProcessingStatus` stage whether or not the record set in the database is out-of-sync with the DNS backend. If so, generate both a record set and record set change with a new `RecordSetChangeStatus.Synced`.

#### Current flow
Below are the current (simplified) sad path process flows for `RecordSetService` and `BatchChangeService` (with the batch interface collecting the record state for multiple records at the same time):

Records in the DNS backend:
```
delete.foo.  A  1.1.1.1
update.foo.   A   2.2.2.2
```

Records in VinylDNS:
```
add.foo.    A  1.1.1.1
update.foo.   A  3.3.3.3
```

*Adds* (sad path, record exists in VinylDNS but not in DNS backend):
1. User tries to add `add.foo.  A   1.1.1.1` which exists in VinylDNS but not the DNS backend
1. VinylDNS attempts to retrieve `add.foo.  A` from the database
1. A `RecordSetAlreadyExists` error is returned to the user
1. User syncs in `foo.` zone and the `add.foo.  A` record is deleted from VinylDNS
1. User resubmits change which is successfully validated and queued for processing

*Deletes* (sad path, record doesn't exist in VinylDNS but exists in DNS backend):
1. User tries to delete `delete.foo.  A` which does not exist in VinylDNS but does in the DNS backend
1. VinylDNS attempts to retrieve `delete.foo.  A` from the database
1. A `RecordDoesNotExist` error is returned to the user
1. User syncs in `foo.` and the `delete.foo.  A` record appears in VinylDNS
1. User resubmits change which is successfully validated and queued for processing

#### Proposed flow
*Adds* (sad path, record exists in VinylDNS but not in DNS backend):
1. User tries to add `add.foo.  A  1.1.1.1` which exists in VinylDNS but not the DNS backend
1. VinylDNS attempts to retrieve `add.foo.  A` from the database
1. VinylDNS attempts to retrieve `add.foo.  A` from the DNS backend
[...]

PR #921, opened 2019-12-20
Implement record set sync against DNS backend in RecordSetChangeHandler process flow

mitruly:
Tied to https://github.com/vinyldns/vinyldns/issues/853.

Changes in this pull request:
- Introduce new `RecordSetChangeType.Sync` and `RecordSetChangeStatus.Synced` enumerations
- Implement a sync against DNS backend check that happens during the `validate` stage of record set change processing
- Update unit tests

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetChangeGenerator.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetChangeGenerator.scala
@@ -123,7 +123,11 @@ object RecordSetChangeGenerator extends DnsConversions {
   def forDelete(recordSet: RecordSet, zone: Zone, auth: AuthPrincipal): RecordSetChange =
     forDelete(recordSet, zone, auth.userId, List())
 
-  def forSyncAdd(recordSet: RecordSet, zone: Zone): RecordSetChange =
+  private def forSyncAdd(
+      recordSet: RecordSet,
+      zone: Zone,
+      systemMessage: Option[String]
+  ): RecordSetChange =
     RecordSetChange(
       zone = zone,
       recordSet = recordSet
@@ -131,10 +135,15 @@ object RecordSetChangeGenerator extends DnsConversions {
       userId = "system",
       changeType = RecordSetChangeType.Create,
       status = RecordSetChangeStatus.Complete,
-      systemMessage = Some("Change applied via zone sync")
+      systemMessage = systemMessage
     )
 
-  def forSyncUpdate(replacing: RecordSet, newRecordSet: RecordSet, zone: Zone): RecordSetChange =
+  private def forSyncUpdate(
+      replacing: RecordSet,
+      newRecordSet: RecordSet,
+      zone: Zone,
+      systemMessage: Option[String]
+  ): RecordSetChange =
     RecordSetChange(
       zone = zone,
       recordSet = newRecordSet.copy(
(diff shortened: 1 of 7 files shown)


modules/api/src/main/scala/vinyldns/api/engine/RecordSetChangeHandler.scala
```diff
@@ -207,9 +253,74 @@ object RecordSetChangeHandler {
...
+     * - Delete record from database for ADD request if exists in database but does not exist in DNS backend
+     */
+    def syncDnsBackendRRSet(
+        storedRRSet: Option[RecordSet],
+        dnsBackendRRSet: Option[RecordSet],
+        zone: Zone,
+        recordSetRepository: RecordSetRepository,
+        recordChangeRepository: RecordChangeRepository
+    ): IO[Unit] = {
+      val recordSetToSync = (storedRRSet, dnsBackendRRSet) match {
```
mitruly: During sync check, determine if there is scenario where a record should be synced in. If the scenario does not apply, don't do anything; if the scenario applies, generate a record set change and record set to save in the database and process the request as normal.

Note: We only want to check for these specific scenarios because we don't want to save a record set change/record set for `Delete` if the record has already been deleted from the DNS backend (which is consistent with existing behavior).

modules/api/src/main/scala/vinyldns/api/engine/RecordSetChangeHandler.scala
```diff
@@ -207,9 +253,74 @@ object RecordSetChangeHandler {
...
+          for {
+            _ <- recordChangeRepository.save(changeSet)
+            _ <- recordSetRepository.apply(changeSet)
+          } yield ()
+        }
+        .getOrElse(IO.unit)
+    }
+
+    for {
+      storedRRSet <- recordSetRepository
```
mitruly: This is subtle, but we need to use `getRecordSetsByName` since fetching a record that exists in the DNS backend may already exist in the database with an unknown record set identifier.

modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetChangeGenerator.scala
```diff
@@ -148,17 +157,47 @@ object RecordSetChangeGenerator extends DnsConversions {
...
+      newRecordSet: RecordSet,
+      zone: Zone
+  ): RecordSetChange =
+    forSyncUpdate(replacing, newRecordSet, zone, Some("Change applied via zone sync"))
+
+  def forZoneSyncDelete(recordSet: RecordSet, zone: Zone): RecordSetChange =
+    forSyncDelete(recordSet, zone, Some("Change applied via zone sync"))
+
+  def forRecordSyncAdd(recordSet: RecordSet, zone: Zone): RecordSetChange =
+    forSyncAdd(recordSet, zone, Some("Change applied via record sync"))
```
britneywright:
```suggestion
    forSyncAdd(recordSet, zone, Some("Change applied via recordset sync"))
```

modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetChangeGenerator.scala
```diff
@@ -148,17 +157,47 @@ object RecordSetChangeGenerator extends DnsConversions {
...
+
+  def forRecordSyncUpdate(
+      replacing: RecordSet,
+      newRecordSet: RecordSet,
+      zone: Zone
+  ): RecordSetChange =
+    forSyncUpdate(replacing, newRecordSet, zone, Some("Change applied via record sync"))
+
+  def forRecordSyncDelete(recordSet: RecordSet, zone: Zone): RecordSetChange =
+    forSyncAdd(recordSet, zone, Some("Change applied via record sync"))
```
britneywright:
```suggestion
    forSyncAdd(recordSet, zone, Some("Change applied via recordset sync"))
```

modules/api/src/main/scala/vinyldns/api/engine/RecordSetChangeHandler.scala
```diff
@@ -207,9 +253,74 @@ object RecordSetChangeHandler {
...
+        .getRecordSetsByName(change.zoneId, change.recordSet.name)
+      _ <- syncDnsBackendRRSet(
+        storedRRSet.find(_.typ == change.changeType),
```
mitruly: This is also subtle, but we are allowed to grab the first zone, record name and record type match since there should only be one record set.

PR #922, opened 2019-12-23
Implement global recordset search: API

### LINKED ISSUE #861: Global recordset search
## Current state of recordset search and access in VinylDNS

Recordsets are only accessible through zones. The requesting user must have access to the zone as a member of the zone owner group, with a read, write or delete ACL rule, or as system admin. 

With shared zones in particular users own recordsets but do not have the ability to view their current state if they don't have access to the zones the recordsets is in. Additionally, we've seen that groups own individual recordsets across many zones, so there's a need to list recordsets beyond the zone scope. Lastly we want to give users autonomy to look up recordsets and find owner groups if DNS conflicts arise.   

## API Endpoints

Proposing three routes, two would be new, one already exists but will have additional query parameters.

- List all recordsets in VinylDNS
```
GET /recordsets
```

- List all records for a given zone (this endpoint already exists)
```
GET /zones/:zoneId/recordsets
```

- List all records owned by a given group. Direct record ownership only. Not including indirect ownership through zone ownership or ACL rules.
```
GET /groups/:groupId/recordsets
```

## Parameters
 | Name | Type | Description |
 |------------|------|------|
 | recordNameFilter  | string  | Text to filter against the recordset name attribute. case-insensitive. exact match unless wildcards (`*`) are used. |
  fqdnFilter  | string  | Text to filter against FQDN attribute. case-insensitive. exact match unless wildcards (`*`) are used. |
 | recordTypes  | string  | A list of comma separated record types. Default is all record types if not provided. |
 | direction  |  string | The direction of the sort. Either `asc` or `desc`. Default: `desc` |
 | sort  |  string | What to sort results by. `fqdn`, `created`, `updated`. Default: `fqdn`  |
 | since  | date-time | Only recordsets updated at or after this time. This is a timestamp in ISO 8601 format: YYYY-MM-DDTHH:MM:SSZ. | 
 | startFrom  | string | Recordset ID. Should be set to the nextId that is returned on the previous response to return the next page of results. | 
 | maxItems  | integer | The number of items to return in the page. Valid values are 1 to 100. Defaults to 100 if not provided. | 

## Response Attributes
| Name | Type | Description |
|------------|------|------|
| recordsets  | array  | Array of recordsets. See recordset response for attributes. |
| startFrom  | string  | startFrom sent in request. Will not be returned if not provided |
| maxItems  | string  | maxItems sent in request. Default is 100. |
| sort  |  string | Given sort string `fqdn`, `name`, `created`, `updated`. Default to `fqdn` if not provided.  |
| nextId  | string  | startFrom sent in request. Will not be returned if not provided |
| direction  |  string | The given direction of the sort `asc` `desc`. Default to `asc` if not provided. |
| recordNameFilter  | string  | Requested record name to filter against. will not be returned if not provided. **Note:** only for zone recordset endpoint. |
| fqdnNameFilter  | string  | Requested fqdn to filter against. will not be returned if not provided. **Note**: only for groups and global recordset endpoints. |
| recordTypes  | string  | A list of comma separated record types requested. Will not be returned if not provided. |
| since  | string | Given date time to look for updated recordsets. Will not be returned if not provided. | 
| ipAddressFilter | string | Filter PTR records by their IP address formatted name |

  
### Recordset Response Attributes
| Name | Type | Description|
|------------|------|------|
| fqdn  | string  | Full qualified domain name |
| type  | string  | DNS record type |
| zoneId  |  string | The ID of the zone the record is in |
| name  | string  | Record name, Does not include the zone name |
| zoneName  |  string | The name of the zone the record is in |
| ttl  |  long | the TTL in seconds for the recordset  |
| status  | string | The current state of the Recordset. `Active`, `Inactive`, `Pending`, `PendingUpdate`, `PendingDelete` | 
| created  | date-time | The timestamp (UTC) when the recordset was created. ISO 8601 format: YYYY-MM-DDTHH:MM:SSZ. | 
| updated  | date-time | The timestamp (UTC) when the recordset was most recently updated. ISO 8601 format: YYYY-MM-DDTHH:MM:SSZ. | 
| records  | array of RecordData | Array of record data, a single RecordSet can have multiple DNS records as long as they are all the same type and the type supports multiple records | 
| accessLevel  | string | Access the requesting user has to the recordset. `None`, `Read`, `Delete` | 
| ownerGroupId  | string | The ID of the group that owns the record | 
| ownerGroupName  | string | The name of the group that owns the record | 
    
## Access
All authenticated, unlocked users should be able to make requests to these endpoints.
This is a change for the existing recordsets by zone endpoint. It is currently restricted based on zone access.

### Records
```
/records
```

New page in the portal at the root level.

Screen Shot 2019-10-03 at 1 43 54 PM
Table headings:

| FQDN | Type |	TTL |	Record Data | Zone Type | Owner Group Name | Actions |
|------|------|-----|-------------|---------|---------|---------|

Table options:
- Sort by FQDN in ascending or descending order. Default to ascending.

Search options:
- FQDN, wildcard support
- Owner group name. See about making this predictive/autocomplete, otherwise drop down of group names.
- Filter by record type. Default to all.

### Group Records
```
/groups/:groupId/records
```
Similar to current Zone area of the site, make tabbed pages in the Group. One for the existing group data, "Membership" and another for the new data, "Records". Explicit routing is not required, but would be an improvement over our existing tab behavior.

Screen Shot 2019-10-03 at 12 14 58 AM

Table headings:

| FQDN | Type |	TTL |	Record Data |	Actions |
|------|------|-----|-------------|---------|

Table options:
- Sort by FQDN in ascending or descending order. Default to ascending.

Search options:
- FQDN, wildcard support.
- Filter by record type. Checkboxes. Defaulted to all.

### Zone Records
```
/zones/:zoneId
```

Existing page with proposed modifications.

Table updates:
- Sort by Name in ascending or descending order. Default to ascending.

Search updates:
- Filter by record type. Checkboxes. Defaulted to all.

  
Table headings:

| FQDN | Type |	TTL |	Record Data |		Owner Group Name  |Actions |
|------|------|-----|-------------|---------|---------|

Table options:
- Sort by FQDN in ascending or descending order. Default to descending.

Search options:
- FQDN, wildcard support.
- Filter by record type. Checkboxes. Defaulted to all.

## Considerations
API & Portal:
- Linking to groups in the portal will require access validation changes in the API.

API:
- Should the default direction change from `asc` to `desc` if the sort is `created` or `updated`
- Need to think about IP address search, ho we'd do it and how to address clashes with other filters a user might pass in.

Portal:
- If we allow updating records in the Groups and Records areas we should only allow TTL, Record Data and Owner Group changes. Should not allow changing the FQDN which could change the record name and/or the zone.

britneywright:
Fixes #861

Changes in this pull request:
- Add GET `/recordsets` endpoint
  - Requires `recordNameFilter` query parameter with minimum of two alphanumeric characters. `recordNameFilter` is run on the `fqdn` column in the recordset table in the database. Can use `*` for fuzz matching.
  - If given an IP address (IPv4 or IPv6) we will convert it to an fqdn that we can then pass in for the database query.
  - Optional query parameters are nameSort (default is "ASC" for ascending, other option is "DESC" for descending), recordTypeFilter (an array of record types), maxItems, startFrom, and nextId.
  - Doesn't require any access validations


modules/api/src/test/scala/vinyldns/api/domain/record/RecordSetServiceSpec.scala
```diff
@@ -781,6 +819,64 @@ class RecordSetServiceSpec
       doReturn(
         IO.pure(
           ListRecordSetResults(
+            List(sharedZoneRecord),
```
mitruly: For some reason, the mock repos are not returning the mocked data properly. Was able to use `any[]` to get this test to pass:
```
      doReturn(IO.pure(Set(okGroup)))
        .when(mockGroupRepo)
        .getGroups(any[Set[String]])

      doReturn(
        IO.pure(
          ListRecordSetResults(
            List(sharedZoneRecord),
            recordNameFilter = "aaaa*",
            nameSort = NameSort.ASC
          )
        )
      ).when(mockRecordRepo)
        .listRecordSets(
          startFrom = any[Option[String]],
          maxItems = any[Option[Int]],
          recordNameFilter = any[String],
          recordTypeFilter = any[Option[Set[RecordType.RecordType]]],
          nameSort = any[NameSort.NameSort]
        )
```

britneywright: modified the test to this, i'll revisit though.

britneywright:
The portal code is due for a refactoring, there's a lot of duplicate code I'll remove but the initial UI implementation can be reviewed locally by going to `http://localhost:9001/recordsets` and an individual recordset can be viewed by going to `http://localhost:9001/recordsets/:rid`

The api can be tested locally through the portal or through Postman with `http://127.0.0.1:9000/recordsets?recordNameFilter=*dummy` updating the record name value to any two-character minimum.

modules/api/src/main/scala/vinyldns/api/domain/record/ListRecordSetChangesResponse.scala
```diff
@@ -41,3 +41,26 @@ object ListRecordSetChangesResponse {
       listResults.maxItems
     )
 }
+
+case class ListRecordSetRecordSetChangesResponse(
```
britneywright: I have to come up with a better name here, or I may call this `ListRecordSetChangesResponse` and change the existing `ListRecordSetChangesResponse` to `ListZoneRecordSetChangesResponse`

modules/api/src/main/scala/vinyldns/api/domain/record/ListRecordSetChangesResponse.scala
```diff
@@ -41,3 +41,26 @@ object ListRecordSetChangesResponse {
...
+    recordSetId: String,
+    recordSetChanges: List[RecordSetChangeInfo] = Nil,
+    nextId: Option[String],
+    startFrom: Option[String],
+    maxItems: Int
+)
+
+object ListRecordSetRecordSetChangesResponse {
```
britneywright: same as comment above

modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetService.scala
```diff
@@ -123,13 +126,22 @@ class RecordSetService(
     } yield change
 
   def getRecordSet(
+      recordSetId: String,
+      authPrincipal: AuthPrincipal
```
britneywright: Think I can take authPrincipal out of here. Anyone would be able to view a recordset, this wouldn't give them ability to modify it.

modules/mysql/src/main/resources/db/migration/V3.21__RecordChangeChangeRecordSetId.sql
```diff
@@ -0,0 +1,6 @@
+CREATE SCHEMA IF NOT EXISTS ${dbName};
```
britneywright: will remove this migration, after discussion we no longer need this. Will put modifications to RecordSetChange in a future PR.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
```diff
@@ -354,7 +431,7 @@ object MySqlRecordSetRepository extends ProtobufConversions {
   val unknownRecordType: Int = 100
 
   def toRecordSet(rs: WrappedResultSet): RecordSet =
-    fromPB(VinylDNSProto.RecordSet.parseFrom(rs.bytes(1)))
+    fromPB(VinylDNSProto.RecordSet.parseFrom(rs.bytes(1)), rs.stringOpt(2))
```
britneywright: We don't currently store the FQDN in the protobuf so include it separately here to pass to the `fromPB` method that converts the data to a RecordSet type. If we add FQDNs to the protobuf we won't need to do this, however we would need to update the data blobs for all existing records. 

I prefer this approach because it doesn't require changes to the database so there won't be compatibility issues, i need test the queries to determine if there is a significant time impact though. Alternative suggestions are welcome.

modules/core/src/main/scala/vinyldns/core/protobuf/ProtobufConversions.scala
```diff
@@ -63,19 +63,19 @@ trait ProtobufConversions {
...
       systemMessage = if (chg.hasSystemMessage) Option(chg.getSystemMessage) else None,
-      updates = if (chg.hasUpdates) Option(fromPB(chg.getUpdates)) else None,
+      updates = if (chg.hasUpdates) Option(fromPB(chg.getUpdates, None)) else None,
       id = chg.getId,
       singleBatchChangeIds = chg.getSingleBatchChangeIdsList.asScala.toList
     )
   }
 
-  def fromPB(rs: VinylDNSProto.RecordSet): RecordSet =
+  def fromPB(rs: VinylDNSProto.RecordSet, fqdn: Option[String]): RecordSet =
```
mitruly: `fqdn` should be removed as a parameter since this should just be the data blob deserialized (which defaults `fqdn` to `None` which is fine).

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
mitruly: By removing `fqdn` as a parameter from `fromPB`, this can become: `fromPB(VinylDNSProto.RecordSet.parseFrom(rs.bytes(1))).copy(fqdn = rs.stringOpt(2))`

modules/core/src/main/scala/vinyldns/core/domain/record/ListRecordSetResults.scala
```diff
@@ -30,6 +30,16 @@ object NameSort extends Enumeration {
 }
 
 case class ListRecordSetResults(
+    recordSets: List[RecordSet] = List[RecordSet](),
+    nextId: Option[String] = None,
+    startFrom: Option[String] = None,
+    maxItems: Option[Int] = None,
+    recordNameFilter: String,
```
mitruly: Is `recordNameFilter` required? Think this would also be `Option[String]` meaning we can just use a single `ListRecordSetResults` for both `listRecordSets` and `listRecordSetsByZone`.

modules/core/src/main/protobuf/VinylDNSProto.proto
```diff
@@ -129,6 +129,7 @@ message RecordSet {
   repeated RecordData record = 9;
   required string account = 10;
   optional string ownerGroupId = 11;
+  optional string fqdn = 12;
```
mitruly: This should be removed since we're not storing this in the blob.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
```diff
@@ -170,6 +170,74 @@ class MySqlRecordSetRepository extends RecordSetRepository with Monitored {
       }.as(changeSet)
     }
 
+  def listRecordSets(
+      startFrom: Option[String],
+      maxItems: Option[Int],
+      recordNameFilter: String,
+      recordTypeFilter: Option[Set[RecordType]],
+      nameSort: NameSort
```
mitruly: Most of this logic is shared with `listRecordSetsByZone`; we could consolidate it and make `zoneId` an `Option[String]` which would determine whether we have `zone_id = {zoneId}` added to the `WHERE` clause of the query.

modules/core/src/main/scala/vinyldns/core/domain/record/ListRecordSetResults.scala
britneywright: yes it's required, two character limit.

mitruly: @britneywright Although `recordNameFilter` is required, keeping it as an option makes things more compact and modular. We did something similar a while back for `SingleChange` (https://github.com/vinyldns/vinyldns/commit/e4b2d7e499b4393664a5ad8991d6c7639af0a53f#diff-0e527c5212dcba3c3f158debef65564f).

britneywright:
Consolidated the`listRecordSetsByZone` and `listRecordSets` repository queries.

modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetValidations.scala
```diff
@@ -302,4 +303,12 @@ object RecordSetValidations {
       (),
       InvalidRequest("Cannot update RecordSet's zone ID.")
     )
+
+  def validRecordNameFilterLength(recordNameFilter: String): Either[Throwable, Unit] =
+    ensuring(
+      InvalidRequest("recordNameFilter must contain at least two letters or numbers.")
+    ) {
+      val searchRegex: Regex = """[a-zA-Z0-9].*[a-zA-Z0-9]+""".r
```
britneywright: this is to ensure at least two alphanumeric characters are in the name search. we want this to be active searches for particular records, not a blanket search to try and see what's in the system.

modules/core/src/main/scala/vinyldns/core/domain/record/RecordSet.scala
```diff
@@ -47,7 +47,8 @@ case class RecordSet(
     records: List[RecordData] = List.empty,
     id: String = UUID.randomUUID().toString,
     account: String = "system",
-    ownerGroupId: Option[String] = None
+    ownerGroupId: Option[String] = None,
+    fqdn: Option[String] = None
```
mitruly: At first glance, `fqdn` should probably be required here since it's required in our database. I think we're coupling two different data structures here: 1. the record set input that is submitted in a request, and 2. the actual record set as it is reflected in our database.

modules/api/src/main/scala/vinyldns/api/route/DnsJsonProtocol.scala
```diff
@@ -183,7 +183,8 @@ trait DnsJsonProtocol extends JsonValidation {
           .andThen(extractRecords(_, js \ "records")),
         (js \ "id").default[String](UUID.randomUUID().toString),
         (js \ "account").default[String]("system"),
-        (js \ "ownerGroupId").optional[String]
+        (js \ "ownerGroupId").optional[String],
+        (js \ "fqdn").optional[String]
```
mitruly: We should make this required and decouple the `CreateRecordSetInput`/`UpdateRecordSetInput`/`DeleteRecordSetInput` similar to how we did for zones. It doesn't make sense that a user would pass in an `fqdn` with a create request. Testing locally, it seems we don't actually use this to generate the `fqdn` that gets saved so this can suffice as a stop gap.

modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetValidations.scala
mitruly: Need to double-check since historically the regex match in Scala appeared to require a full string match, meaning that we may need to include an optional terminating dot (`\.?`) and may need to allow any character preceding the first part (`.*`).

modules/dynamodb/src/main/scala/vinyldns/dynamodb/repository/DynamoDBRecordSetRepository.scala
```diff
@@ -136,15 +142,30 @@ class DynamoDBRecordSetRepository private[repository] (
...
+      startFrom: Option[String],
+      maxItems: Option[Int],
+      recordNameFilter: String,
+      recordTypeFilter: Option[Set[RecordType]],
+      nameSort: NameSort
+  ): IO[ListRecordSetResults] =
+    monitor("repo.RecordSet.listRecordSets") {
+      IO.raiseError(
+        UnsupportedDynamoDBRepoFunction(
+          s"listRecordSets is not supported by VinylDNS DynamoDB RecordSetRepository"
```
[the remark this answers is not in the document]
mitruly: Also think we should only deprecate this if `zoneId` is `None` for the newer functionality; I'm fine with not supporting newer functionality but don't think we should necessarily remove existing ones.

PR #925, opened 2020-01-07
disallow recordset name and type updates

britneywright:
We want to prevent users from changing recordset names and types during recordset updates. If a user is trying to change a recordset name or type they should really be creating a new record and deleting the existing record. This is going to help us move toward being able to display recordset change history based on the unique combination of a recordset's name and type.

Changes in this pull request:
- Add RecordSet validation to confirm the recordset name, type and zone ID in the given data for recordset update matches the name of the existing recordset. we already checked for zone ID changes, but this moves that into the same validation.

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetValidations.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetValidations.scala
@@ -272,6 +272,16 @@ object RecordSetValidations {
         else InvalidRequest(s"""User not in record owner group with id "$groupId"""").asLeft
     }
 
+  def onlyUpdatesModifiableAttributes(
+      existing: RecordSet,
+      updates: RecordSet
+  ): Either[Throwable, Unit] =
+    Either.cond(
+      updates.name == existing.name && updates.zoneId == existing.zoneId && updates.typ == existing.typ,
+      (),
+      InvalidRequest("Can only update RecordSet's record data, TTL, or owner group.")
+    )
+
   def recordSetIsInZone(recordSet: RecordSet, zone: Zone): Either[Throwable, Unit] =
     Either.cond(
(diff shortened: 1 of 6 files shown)


modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetValidations.scala
```diff
@@ -272,10 +272,13 @@ object RecordSetValidations {
...
     }
 
-  def recordSetIsInZone(recordSet: RecordSet, zone: Zone): Either[Throwable, Unit] =
+  def onlyUpdatesModifiableAttributes(
+      existing: RecordSet,
+      updates: RecordSet
+  ): Either[Throwable, Unit] =
     Either.cond(
-      recordSet.zoneId == zone.id,
+      updates.name == existing.name && updates.zoneId == existing.zoneId && updates.typ == existing.typ,
```
mitruly: Need to also account for trailing dot/`@` with apex records.

modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetValidations.scala
```diff
@@ -272,10 +258,34 @@ object RecordSetValidations {
...
     }
 
-  def recordSetIsInZone(recordSet: RecordSet, zone: Zone): Either[Throwable, Unit] =
+  def unchangedRecordName(
+      existing: RecordSet,
+      updates: RecordSet,
+      zone: Zone
+  ): Either[Throwable, Unit] = Either.cond(
+    updates.name.toLowerCase == existing.name.toLowerCase
+      || (updates.name == "@" && existing.name.toLowerCase == zone.name),
```
mitruly: `zone.name` should also be `.toLowerCase` here as we use zone name casing as-is.

PR #926, opened 2020-01-07
add unique FQDN type index on recordset

britneywright:
Changes in this pull request:
- Add unique fqdn_type_index on recordset to avoid adding duplicate record entries during long zone syncs where a job might restart.


britneywright:
closing this, will make a new PR that uses the existing zone_id_name_index but make it unique

PR #932, opened 2020-02-04
add stats mysql table

britneywright:
Changes in this pull request:
- add a table named "stats" to the mysql database to support tracking and storing various database related information. An example would be tracking the number of recordsets in the database at a particular time.

### DIFF
--- a/modules/mysql/src/main/resources/db/migration/V3.21__Stats.sql
+++ b/modules/mysql/src/main/resources/db/migration/V3.21__Stats.sql
@@ -1,0 +1,12 @@
+CREATE SCHEMA IF NOT EXISTS ${dbName};
+
+USE ${dbName};
+
+CREATE TABLE stats (
+  id INT NOT NULL AUTO_INCREMENT,
+  name VARCHAR(255) NOT NULL,
+  count BIGINT NOT NULL,
+  created DATETIME NOT NULL,
+  PRIMARY KEY (id),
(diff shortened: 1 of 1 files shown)


modules/mysql/src/main/resources/db/migration/V3.21__Stats.sql
```diff
@@ -0,0 +1,12 @@
+CREATE SCHEMA IF NOT EXISTS ${dbName};
+
+USE ${dbName};
+
+CREATE TABLE stats (
+  id INT NOT NULL AUTO_INCREMENT,
```
mitruly: Is `id` necessary for `stats`? `name` could arguably be forced to be unique and used as the primary key.

mitruly: Oh, I think I see - so `id` here is used to track a particular "log" a stat where the total being tracked is linked by `name`.

If this is the case, I think updating the index to `name`, `created` makes more sense. We may also want to use `BIGINT` or `CHAR(36)` (to remain consistent with trend of UUIDs) for `id` rather than `INT` here.

mitruly: Note: I think using a UUID would be marginally better due to lack of lock contention on the table when doing many stat inserts, but this table is probably not going to have as many write operations compared to record set/batch change tables.

PR #933, opened 2020-03-05
Support DNS backend checks for batch changes

mitruly:
Support DNS backend verification for `RecordDoesNotExist` and `RecordAlreadyExists` conflicts that surface when requested DNS changes conflict with the VinylDNS database disposition despite being valid via the DNS backend.

Changes in this pull request:
- Support checking against DNS backend for record existence conflicts with the VinylDNS database. **Note**: The DNS backend would only get polled for DNS records provided in the batch change. This means that the VinylDNS database will still be used as the source of truth for conflicting records that exist since updating records outside of the scope of submitted batch changes should be handled via zone syncs.

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/Boot.scala
+++ b/modules/api/src/main/scala/vinyldns/api/Boot.scala
@@ -119,7 +119,11 @@ object Boot extends App {
       val batchChangeValidations = new BatchChangeValidations(
         batchChangeLimit,
         batchAccessValidations,
-        VinylDNSConfig.scheduledChangesEnabled
+        (zone, connections) =>
+          DnsConnection(ZoneConnectionValidator.getZoneConnection(zone, connections)),
+        connections,
+        VinylDNSConfig.scheduledChangesEnabled,
+        VinylDNSConfig.validateRecordLookupAgainstDnsBackend
       )
       val membershipService = MembershipService(repositories)
       val connectionValidator =
--- a/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeInterfaces.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeInterfaces.scala
@@ -26,6 +26,7 @@ object BatchChangeInterfaces {
 
   type SingleValidation[A] = ValidatedNel[DomainValidationError, A]
   type ValidatedBatch[A] = List[ValidatedNel[DomainValidationError, A]]
+  type ValidatedBatchMonad[F[_], A] = List[F[ValidatedNel[DomainValidationError, A]]]
   type BatchResult[A] = EitherT[IO, BatchChangeErrorResponse, A]
 
   implicit class IOBatchResultImprovements[A](theIo: IO[A]) {
@@ -61,6 +62,16 @@ object BatchChangeInterfaces {
       batch.map {
         case Valid(item) => fn(item)
         case Invalid(errList) => errList.invalid
+      }
+
+    def mapValidMonad[F[_], B](
+        validFn: A => F[SingleValidation[B]]
+    )(
+        invalidFn: NonEmptyList[DomainValidationError] => F[SingleValidation[B]]
+    ): ValidatedBatchMonad[F, B] =
+      batch.map {
+        case Valid(item) => validFn(item)
+        case Invalid(errList) => invalidFn(errList)
       }
 
     def getValid: List[A] = batch.collect {
--- a/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
@@ -377,17 +377,16 @@ class BatchChangeValidations(
               zoneDoesNotRequireManualReview(change, isApproved) |+|
               ensureRecordExists(change, groupedChanges)
           )
-        case None =>
-          if (validateRecordLookupAgainstDnsBackend) {
-            dnsConnection(change.zone, configuredDnsConnections)
-              .resolve(change.recordName, change.zone.name, change.inputChange.typ)
-              .value
-              .map {
-                case Right(dnsBackendRecords) if dnsBackendRecords.nonEmpty =>
-                  ensureRecordExists(change, dnsBackendRecords)
-                case _ => RecordDoesNotExist(change.inputChange.inputName).invalidNel
-              }
-          } else IO(RecordDoesNotExist(change.inputChange.inputName).invalidNel)
+        case None if validateRecordLookupAgainstDnsBackend =>
+          dnsConnection(change.zone, configuredDnsConnections)
+            .resolve(change.recordName, change.zone.name, change.inputChange.typ)
+            .value
+            .map {
+              case Right(dnsBackendRecords) if dnsBackendRecords.nonEmpty =>
+                ensureRecordExists(change, dnsBackendRecords)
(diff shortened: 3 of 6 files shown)


modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
```diff
@@ -327,22 +344,52 @@ class BatchChangeValidations(
...
-          userCanDeleteRecordSet(change, auth, rs.ownerGroupId, rs.records) |+|
-            zoneDoesNotRequireManualReview(change, isApproved) |+|
-            ensureRecordExists(change, groupedChanges)
-        case None => RecordDoesNotExist(change.inputChange.inputName).invalidNel
+          IO(
+            userCanDeleteRecordSet(change, auth, rs.ownerGroupId, rs.records) |+|
+              zoneDoesNotRequireManualReview(change, isApproved) |+|
+              ensureRecordExists(change, groupedChanges)
+          )
+        case None =>
```
clinedome, line 380: Just minor style, but could probably make this `case None if validateRecordLookupAgainstDnsBackend =>`, and the `else` clause could just be the `case _ =>` or `case None =>`. Feels a little awkward to have a single case with the feature flag `if` inside of it. Same with the other instances of this pattern below.

modules/api/src/main/scala/vinyldns/api/Boot.scala
```diff
@@ -119,7 +119,11 @@ object Boot extends App {
       val batchChangeValidations = new BatchChangeValidations(
         batchChangeLimit,
         batchAccessValidations,
-        VinylDNSConfig.scheduledChangesEnabled
+        (zone, connections) =>
+          DnsConnection(ZoneConnectionValidator.getZoneConnection(zone, connections)),
```
clinedome, line 123: just making sure I understand: we need to do this as a function here because the batch is comprised of lots of _changes_, and each change has an associated zone. this gives us a way to grab a connection to poll the backend about a particular zone _inside_ the validation.

question: are we cool doing a whole boatload of backend DNS polls if, say, a batch change is submitted against an entirely unsynced zone?

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeInterfaces.scala
```diff
@@ -26,6 +26,7 @@ object BatchChangeInterfaces {
 
   type SingleValidation[A] = ValidatedNel[DomainValidationError, A]
   type ValidatedBatch[A] = List[ValidatedNel[DomainValidationError, A]]
+  type ValidatedBatchMonad[F[_], A] = List[F[ValidatedNel[DomainValidationError, A]]]
```
clinedome, line 29: this is just a new type of validation that expects a function as its input, right?

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
```diff
@@ -327,22 +344,52 @@ class BatchChangeValidations(
...
+    change match {
+      // For DeleteRecord inputs, need to verify that the record data actually exists
+      case DeleteRRSetChangeForValidation(
+          _,
+          _,
+          DeleteRRSetChangeInput(inputName, _, Some(recordData))
+          )
+          if !dnsBackendRecords
+            .exists(rs => matchRecordData(rs.records, recordData)) =>
+        DeleteRecordDataDoesNotExist(inputName, recordData).invalidNel
```
clinedome, line 360: so this is rejecting a DeleteRecordSet where the record neither exists in vinyl or the backend dns, right?

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
```diff
@@ -351,53 +398,74 @@ class BatchChangeValidations(
...
+          IO(
+            userCanUpdateRecordSet(change, auth, rs.ownerGroupId, List(change.inputChange.record)) |+|
+              ownerGroupProvidedIfNeeded(
+                change,
+                groupedChanges.existingRecordSets
+                  .get(change.zone.id, change.recordName, change.inputChange.typ),
+                batchOwnerGroupId
+              ) |+|
+              zoneDoesNotRequireManualReview(change, isApproved)
+          )
```
clinedome, line 421: Are we wrapping this in an IO now because there's a possibility of some async IO happening? I'm not familiar with the `IO` structure, can you explain briefly how it's used here?

modules/api/src/main/scala/vinyldns/api/Boot.scala
mitruly: @clinedome this is a good point. Ideally all changes would get made to a DNS zone _within_ VinylDNS post-connecting to a zone. When a zone is connected, a sync is automatically performed so this circumstance would only surface if many changes are being made external to VinylDNS .

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeInterfaces.scala
mitruly: This is a type that handles a monad (e.g. `Future`, `IO`) so that there's less typing too carry down to the function logic.

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
mitruly: Yes, this is correct.

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
mitruly: Right, we are adding an `IO` call in the event that we get no results but are configured to validate the lookup against the DNS backend.

Since `dnsConnection.resolve(...)` returns us a `Result[List[RecordSet]]`, we need to convert the method to return an `IO` for the cases that are simply returning a validation so that the typing matches.

pauljamescleary: @clinedome `IO` is a pure wrapper around side effects / i.e. things that can terminate unreasonably (like throw an SocketException if the network is down right now, but work fine in a minute).  It also is used to "describe" a computation that is evaluated "at the end of the world", i.e. you are describing a computation and when servicing an HTTP request that computation is then evaluated.

In the old days, we used scala's `Future`, however that has a bunch of issues.  We converted to IO which makes concurrency explicit, and evaluation is lazy as opposed to eager with `Future`.

modules/api/src/main/scala/vinyldns/api/Boot.scala
pauljamescleary: @clinedome I will add that "entirely unsynced zone" is an edge case.  When we connect to the zone in VinylDNS, we automatically load in the records.  It is theoretically possible that someone loads the zone into VinylDNS, then makes changes for a while outside of VinylDNS, only coming back to it some point in the future.  But either way, some portion of the records in every zone will live in VinylDNS

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeInterfaces.scala
pauljamescleary: Might be overkill since we are standardized on IO tho right?

pauljamescleary:
Closing this due to age, it is a good idea that we can refactor, incorporate in the future.  Marking as postponed.

PR #939, opened 2020-04-26
Update dependencies

pauljamescleary:
Updated almost all dependencies to current.  There were some issues with akka-http 10.1.11 so I stayed with 10.1.10 for the time being.
    
Func tests passed locally and manual review of the UI looks to be good
    
Significant changes are:
    
- `pureconfig` - this update had breaking syntax, so I had to update everywhere
we use pureconfig.  Functionally it is the same, just different syntax
- `scalatest` - this was a big change, as scalatest has refactored out things
like Mockito and scalacheck.  Many imports changed.
- `Java11` - formally moved everything to java 11.  This required some new
dependencies like `javax.activation` and `java.xml.bind`


pauljamescleary:
You can disregard that first codecov check.  We were previously not excluding `Boot` the right way, I fixed that so it looks like stuff went down but it didn't.

build.sbt
```diff
@@ -12,46 +12,13 @@ resolvers ++= additionalResolvers
 
 lazy val IntegrationTest = config("it") extend Test
 
-// Needed because we want scalastyle for integration tests which is not first class
```
pauljamescleary, line 15: I removed scalastyle.  It has not been updated since 2017 and will inevitably generate heartburn as we continue to update sbt.  We have a lot of coverage between scalafmt and wart remover

build.sbt
```diff
@@ -113,7 +80,7 @@ lazy val apiAssemblySettings = Seq(
 )
 
 lazy val apiDockerSettings = Seq(
-  dockerBaseImage := "openjdk:8u191-jdk-alpine3.9",
+  dockerBaseImage := "adoptopenjdk/openjdk11:jdk-11.0.7_10-alpine",
```
pauljamescleary, line 83: JDK11 here

modules/api/src/it/scala/vinyldns/api/domain/batch/BatchChangeRepositoryIntegrationSpec.scala
```diff
@@ -17,7 +17,8 @@
 package vinyldns.api.domain.batch
 
 import org.joda.time.DateTime
-import org.scalatest.{Matchers, WordSpecLike}
+import org.scalatest.matchers.should.Matchers
```
pauljamescleary, line 20: this is everywhere.  The `minor` scalatest update had breaking changes.  I plan on creating a common test trait to extend that has all of this in there.  It was tedious to make all the changes.

modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
```diff
@@ -51,7 +53,9 @@ object VinylDNSConfig {
       .asScala
       .toList
       .map { configKey =>
-        loadConfigF[IO, DataStoreConfig](vinyldnsConfig, configKey)
+        Blocker[IO].use(
+          ConfigSource.fromConfig(vinyldnsConfig).at(configKey).loadF[IO, DataStoreConfig](_)
+        )
```
pauljamescleary, line 58: This is the `pureconfig` update changes.  Since updating everything to current (cats, cats-effect) we need to keep all the typelevel libs up to date.  Pureconfig also had breaking changes, so you will see this pattern basically everywhere as well.  Between scalatest and pureconfig, those are 99% of the changes in this PR

modules/api/src/main/scala/vinyldns/api/VinylDNSConfig.scala
```diff
@@ -122,7 +130,14 @@ object VinylDNSConfig {
...
 
   lazy val healthCheckTimeout: IO[Int] =
-    loadConfigF[IO, Option[Int]](vinyldnsConfig, "health-check-timeout").map(_.getOrElse(10000))
+    Blocker[IO].use(
+      ConfigSource
+        .fromConfig(vinyldnsConfig)
+        .at("health-check-timeout")
+        .loadF[IO, Option[Int]](_)
+        .map(_.getOrElse(10000))
+        .handleError(_ => 10000)
```
pauljamescleary, line 139: sadly, in the latest pureconfig, it raises an error if the config entry is not present :(

So, need to handle when it is explicitly set to `null`, as well as handle the resulting error and providing a default.

modules/api/src/test/scala/vinyldns/api/domain/zone/ZoneConnectionValidatorSpec.scala
```diff
@@ -241,25 +243,6 @@ class ZoneConnectionValidatorSpec
       result.getMessage should include("transfer connection failure!")
     }
 
-    "respond with a failure if loadDns times out" in {
```
pauljamescleary, line 244: I deleted this test as it was flaky.  We are literally trying to invoke a timeout in a test which is not a good idea.

modules/api/src/test/scala/vinyldns/api/route/VinylDNSRouteTestHelper.scala
```diff
@@ -18,9 +18,15 @@ package vinyldns.api.route
...
 import akka.http.scaladsl.server.{MalformedRequestContentRejection, RejectionHandler}
+import akka.http.scaladsl.testkit.ScalatestRouteTest
 import org.json4s.MappingException
 
-trait VinylDNSRouteTestHelper {
+trait VinylDNSRouteTestHelper { this: ScalatestRouteTest =>
+  import scala.concurrent.duration._
+  import akka.http.scaladsl.testkit.RouteTestTimeout
+
+  implicit val testTimeout = RouteTestTimeout(10.seconds)
```
pauljamescleary, line 28: the route tests after updating sometimes timed out.  I believe the increased speed / parallelism in SBT 1.3.9 maybe the issue.  I saw this locally, so it may have just been my PC (I have sbt issues locally).  Upped the timeout to 10 seconds to avoid issues again.

modules/portal/app/controllers/Settings.scala
```diff
@@ -55,8 +57,18 @@ class Settings(private val config: Configuration) {
...
+        .fromConfig(config.underlying)
+        .at("data-stores")
+        .loadF[IO, List[String]](blocker)
+        .flatMap { lst =>
+          lst
+            .map(
+              ConfigSource.fromConfig(config.underlying).at(_).loadF[IO, DataStoreConfig](blocker)
+            )
+            .parSequence
+        }
```
pauljamescleary, line 71: can I just mention how great the new scalafmt defaults are, dig it

project/Dependencies.scala
```diff
@@ -1,22 +1,22 @@
 import sbt._
 object Dependencies {
 
-  lazy val akkaHttpV = "10.1.9"
+  lazy val akkaHttpV = "10.1.10"
```
pauljamescleary, line 4: the latest here is 10.1.11, however, someone decided to put in some breaking changes for our route tests that was just too much to work through.  I have aspirations to move to http4s and maybe replace the frontend with typescript-react, so I will take the small update for the time being.

project/Dependencies.scala
```diff
@@ -71,7 +71,9 @@ object Dependencies {
     "com.sun.xml.bind"          %  "jaxb-impl"                      % jaxbV,
     "ch.qos.logback"            %  "logback-classic"                % "1.0.7",
     "io.dropwizard.metrics"     %  "metrics-jvm"                    % "3.2.2",
-    "co.fs2"                    %% "fs2-core"                       % "1.0.0"
+    "co.fs2"                    %% "fs2-core"                       % "2.3.0",
+    "javax.xml.bind"            %  "jaxb-api"                       % "2.3.0",
+    "javax.activation"          %  "activation"                     % "1.1.1"
```
pauljamescleary, line 76: So, the `ee` stuff was left out of jdk 11, so we have to pull it in by hand.

project/metals.sbt
```diff
@@ -0,0 +1,4 @@
+// DO NOT EDIT! This file is auto-generated.
+// This file enables sbt-bloop to create bloop config files.
+
+addSbtPlugin("ch.epfl.scala" % "sbt-bloop" % "1.4.0-RC1-190-ef7d8dba")
```
pauljamescleary, line 4: I will .gitignore this

PR #940, opened 2020-04-28
MySQL Membership updates

mitruly:
Update VinylDNS to use MySQL implementation for `membership` table.

Changes in this pull request:
- Add `is_admin` column to track admin status of a user within a group. **Note**: the admin status is still grabbed via the `Group` model, but the handling will make the column data consistent for switching data later on.
- Update configs to use MySQL instead of DynamoDB for membership table.


modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
```diff
@@ -50,13 +50,21 @@ class MembershipService(
...
       _ <- groupChangeRepo.save(GroupChange.forAdd(newGroup, authPrincipal)).toResult[GroupChange]
       _ <- groupRepo.save(newGroup).toResult[Group]
-      _ <- membershipRepo.addMembers(newGroup.id, newGroup.memberIds).toResult[Set[String]]
+      // save admin and non-admin members separately
+      _ <- membershipRepo
+        .saveMembers(newGroup.id, adminMembers, isAdmin = true)
+        .toResult[Set[String]]
+      _ <- membershipRepo
+        .saveMembers(newGroup.id, nonAdminMembers, isAdmin = false)
+        .toResult[Set[String]]
```
clinedome, line 67: Feels a little awkward that the `isAdmin` is a field on the `saveMembers` method, rather than a property of the member itself. It probably makes sense to do it this way, since that second parameter is just a `Set[String]`, but in some ways it'd be nicer to have a concept of a user that more closely mapped to the data (where user has an ID and an admin property). It'd probably mean that we could save these in one batch, and finding all the admins in a set is a `filter` rather than `Set` diffs. Not a dealbreaker by any means though.

modules/dynamodb/src/main/scala/vinyldns/dynamodb/repository/DynamoDBMembershipRepository.scala
```diff
@@ -95,7 +95,7 @@ class DynamoDBMembershipRepository private[repository] (
       dynamoDBHelper.query(queryRequest).map(result => result.getItems.asScala.map(fromItem).toSet)
     }
 
-  def addMembers(groupId: String, memberUserIds: Set[String]): IO[Set[String]] =
+  def saveMembers(groupId: String, memberUserIds: Set[String], isAdmin: Boolean): IO[Set[String]] =
```
clinedome, line 98: No-op for setting a member to `isAdmin = true` in dynamo? Should we probably either set it, log the fact that we aren't, or reject an `isAdmin = true`?

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlMembershipRepository.scala
```diff
@@ -37,22 +45,28 @@ class MySqlMembershipRepository extends MembershipRepository with Monitored {
...
             DB.localTx { implicit s =>
-              val valueClause = " VALUES " + nonEmpty.as("(?, ?)").mkString(",")
-              val query = BASE_ADD_MEMBERS + valueClause
-              val valueParams: List[String] = nonEmpty.flatMap(Seq(_, groupId))
-              SQL(query)
-                .bind(valueParams: _*)
-                .update
-                .apply()
-
+              SAVE_MEMBERS.batchByName(saveParams(nonEmpty, groupId, isAdmin): _*).apply()
```
clinedome, line 69: This is a totally decent pattern for this kind of thing :+1:

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlMembershipRepository.scala
```diff
@@ -94,4 +108,32 @@ class MySqlMembershipRepository extends MembershipRepository with Monitored {
...
+        query.append(BASE_GET_USERS_FOR_GROUP)
+        val baseConditions = Seq('groupId -> groupId)
+
+        // extra conditions based on whether isAdmin is set
+        val extraConditions = isAdmin match {
+          case None => Seq()
+          case Some(adminFlag) =>
+            query.append(" AND is_admin = {isAdmin}")
+            Seq('isAdmin -> adminFlag)
+        }
```
clinedome, line 127: instead of side-effecting the `query.append`, we could just have this be `val (extraQuery, extraConditions) = { ... }` and return a tuple, then update both after this block. I suspect it would read a little bit more clearly.

mitruly: Sounds good, @clinedome - updated.

modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
mitruly: Yeah, I agree that this makes sense as the path forward. Gonna table this with https://github.com/vinyldns/vinyldns/issues/941.

modules/dynamodb/src/main/scala/vinyldns/dynamodb/repository/DynamoDBMembershipRepository.scala
mitruly: Opted for logging here. `isAdmin` is unwrapping the `memberIds` and `adminUserIds` from `Group` blobs, but we are still using `Group` blobs as the source of truth for admin determination.

PR #951, opened 2020-06-09
Slightly reduce long txt record test size

clinedome:
The BLOB type is juuuust too small for the previous value. Reduce the
number of characters by two -- the purpose of the test is still being
fulfilled. A migration to MEDIUMBLOB or LARGEBLOB will be required if we
need this to be bigger.


clinedome:
Unsure why travis didn't report back, but it passed: https://travis-ci.org/github/vinyldns/vinyldns/builds/696490604

Issue #955, opened 2020-06-30
Create script for managing super users


remerle:
There is no administrative interface, or config-level control over the super user.  Super users are controlled with a flag in the binary blob in the backing store, which is opaque from a query and update point of view.

**Describe the solution you'd like**
As a short term solution, as part of #953, VinylDNS should provide a script that can manage the super user flag.  This should be a simple solution with minimal dependencies as to limit documentation overhead.  It should assume that the data is stored in a MySQL database using the [VinylDNS protobuf definition](modules/core/src/main/protobuf/VinylDNSProto.proto).  It should provide the ability to set a given user as a super user, as well as remove the super user flag from a given user.

The expectation is that this is a management task that would not occur often, but this is still as stopgap until a more formal approach to role-based security and/or group management is provided.

_Suggest this be placed in a top-level `/scripts` directory to prevent it from being confused with the scripts in the `/bin` directory, which are more deployment related._

**Describe alternatives you've considered**
See #953 for context

**Additional context**
See #953 for context

JanuszU:
If there you provide a data structure description then I could write a script.
YES it is definitely needed.

pauljamescleary:
I created a PR for this - https://github.com/vinyldns/vinyldns/pull/956

Recommend python, just need to compile the proto.  I put it all in a docker file for right now, but the script can run anywhere as long as the proto is compiled and available

PR #956, opened 2020-07-02
Adding scripts to update support users

pauljamescleary:
In order to update support users, we need to be able to unpack/pack the VinylDNS protobuf.

This requires us to 1) compile the proto file and 2) call out to the vinyldns Database

Added `update-support-user.py` that provides a simple way to pass in a vinyldns username and `Boolean` to update a user to be a support user.

Built a dockerfile that compiles the Vinyldns.proto file to Python.  This is to avoid a user from having to have `protoc` installed locally and having to know the correct commands.  An alternative would be to run `protoc` and output the proto file to `target` or something.  This approach requires `docker` which felt preferable, but I could go either way.

To execute, you must setup the environment variables or pass them in on the command line.

```
> ./bin/update-support-user.sh -u root -p pass -h host.docker.internal -name --port 19002 vinyldns dummy False
```

The above is useful when running against the quick start setup.

### DIFF
--- a/bin/update-support-user.sh
+++ b/bin/update-support-user.sh
@@ -1,0 +1,84 @@
+#!/usr/bin/env bash
+
+usage () {
+    echo -e "Description: Updates a user in VinylDNS to a support user, or removes the user as a support user.\n"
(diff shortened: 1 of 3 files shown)


remerle:
Addresses #955

PR #965, opened 2020-08-17
Update update-support-user.py script

remerle:
Changes in this pull request:

- Update usage to more clearly define params
- Fix error in help text re mysql port default
- Fix arg parsing in shell script
- Fix pulling latest proto regardless of local version
- Add clean up to docker run in shell script
- Remove unnecessary actions from Dockerfile

### DIFF
--- a/bin/update-support-user.sh
+++ b/bin/update-support-user.sh
@@ -2,27 +2,27 @@
 
 usage () {
     echo -e "Description: Updates a user in VinylDNS to a support user, or removes the user as a support user.\n"
-    echo -e "Usage: update-support-user.sh [OPTIONS] [VINYLDNS USERNAME] [SUPPORT - True | False]\n"
+    echo -e "Usage: update-support-user.sh [OPTIONS] <username> <enableSupport>\n"
+    echo -e "Required Parameters:"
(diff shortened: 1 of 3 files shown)


bin/update-support-user.sh
```diff
@@ -2,27 +2,27 @@
 
 usage () {
     echo -e "Description: Updates a user in VinylDNS to a support user, or removes the user as a support user.\n"
-    echo -e "Usage: update-support-user.sh [OPTIONS] [VINYLDNS USERNAME] [SUPPORT - True | False]\n"
+    echo -e "Usage: update-support-user.sh [OPTIONS] <username> <enableSupport>\n"
+    echo -e "Required Parameters:"
+    echo -e "username\tThe VinylDNS user for which to change the support flag"
+    echo -e "enableSupport\t'true' to set the user as a support user; 'false' to remove support privileges\n"
```
Soloflex95, line 8: Is the 'True' case sensitive here?  If not, then we should be good.

remerle: No, it's not.  We do a case insensitive check:
https://github.com/vinyldns/vinyldns/blob/211c995fc140d564374062e1c45ce3998c1ac090/docker/admin/update-support-user.py#L9

Issue #968, opened 2020-08-28
Replace the portal


pauljamescleary:
The portal is built in the Play Framework, and is little more than a thin wrapper around the API.  There are two main issues with the portal:

1. It mandates a separate artifact to be deployed and managed.  This adds undue maintenance and overhead on what is an otherwise simple system
2. The code is in places in need of updates and simplification

Some objectives for the new portal include:

1. Simple - it really should be very easy to develop and maintain
2. Single artifact - the portal should be deployed along with the API to simplify maintenance

Issue #929, opened 2020-01-24
group membership check in case of LDAP authentication


JanuszU:
Please consider adding a group membership check in case of LDAP authentication, it is much more flexible than just narrowing users to OU

JanuszU:
a simple solution:
1. in ../vinyldns/docker/portal/application.conf  in section "LDAP", add a new config parameter
    groupDN = "CN=vinyl_admins,OU=ourDept,DC=company,DC=com,DC=pl"
2. in ../vinyldns/modules/portal/app/controllers/Settings.scala add a line:
    val ldapGroupDN: String = config.get[String]("LDAP.groupDN")
3. in ../vinyldns/modules/portal/app/controllers/LdapAuthenticator.scala  replace a line
old line (line# 129):
      s"(${settings.ldapUserNameAttribute}=$lookupUserName)",
new line:
      s"(&(objectClass=user)(${settings.ldapUserNameAttribute}=$lookupUserName)(memberOf=${settings.ldapGroupDN}))",

Enjoy ;)
PS. sorry , no group recursion

pauljamescleary:
@JanuszU yes this should be possible.  If the `groupDN` is missing, works like it does today, if present it will be used.  Sounds pretty straightforward

Issue #603, opened 2019-04-26
Ability to search/filter Groups by keyword


daveback:
As an administrator, I am looking for the ability to keyword search/filter the list of groups in the Groups view.  (Similar to how the search in the Zones view functions.)

An additional suggestion would be having the keyword highlighted in the filtered results. I would consider that part optional and not truly in scope for this request.

nimaeskandary:
Noting we can close if https://github.com/vinyldns/vinyldns/pull/621 is merged

britneywright:
Also noting that a current workaround is using the in-browser search functionality since we currently return all groups on the groups page

remerle:
We should address this as part of #968

Issue #415, opened 2019-01-02
Better groups control


rebstar6:
We are using a groups dropdown in the following locations:

- Manage Zone - zone admin group dropdown
- Manage Zone - add ACL rule modal (ACL for group)
- [Coming Soon] New Batch Change (see https://github.com/vinyldns/vinyldns/issues/343)

This is not a great interface for super users (all groups in the system), or the ACL dropdown, which should be expanded to allow any group to be entered. 

We should have a new groups form fill option that will autocomplete to any group in the system. We should use this field for super users in all 3 locations, and other users in the ACL modal.

Note: the field is optional, so may be left empty, on the new batch change screen. Other 2 places it's required.

remerle:
We should address this as part of #968

nspadaccino:
Addressed in portal overhaul https://github.com/vinyldns/vinyldns/issues/968

PR #973, opened 2020-09-02
Increase max groups list result size

remerle:
Groups ultimately need pagination and searching (#603, #415) as well as support for LDAP (#929) groups. As such this fix is to extend the life of the existing implementation until we have time to revisit this (likely with #968).


pauljamescleary (approved):

clinedome (commented):
:+1:

PR #977, opened 2020-09-11
Add support for simple request tracing

### LINKED ISSUE #976: Add basic log event correlation between portal and API
Correlating requests between the portal and the API has notoriously been difficult. VinylDNS isn't a highly distributed application, so a full-fledged tracing implementation seems like overkill. Of primary concern is the ability to correlate events in the system to the originating request.

Initially we want to correlate the Portal->API.  Future issues will deal with correlating API requests with queued messages as well as capturing more pertinent context information for every logged event.

- Define tracing id for inbound requests
- Pass tracing id from Portal to API
- Add a `trace.id` value to logged requests
- Correlate authentication with the request `trace.id`

### Example Output
Correlated on GUID `77b3b671-9d00-42be-b0f7-f4a64ba8fc9f `

**Portal**
```
INFO - filters.AccessLoggingFilter - Request: method=GET, path=/api/dnschanges?maxItems=100, remote_address=127.0.0.1, user_agent=Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/85.0.4183.83 Safari/537.36 | trace.id=77b3b671-9d00-42be-b0f7-f4a64ba8fc9f |Response: status_code=200
```

**API**
```
INFO  | [scala-execution-context-global-35] (v.a.r.BatchChangeRoute:55) | AKKA_SOURCE= | Request started by user testuser | trace.id=77b3b671-9d00-42be-b0f7-f4a64ba8fc9f
INFO  | [VinylDNS-akka.actor.default-dispatcher-4] (a.a.ActorSystemImpl:92) | AKKA_SOURCE=akka.actor.ActorSystemImpl(VinylDNS) | Request: method=GET, path=http://localhost:9000/zones/batchrecordchanges?maxItems=100 | trace.id=77b3b671-9d00-42be-b0f7-f4a64ba8fc9f | Headers: [Timeout-Access='<function1>', Host='localhost:9000', X-Amz-Date='20200910T161926Z', X-VinylDNS-TraceId='77b3b671-9d00-42be-b0f7-f4a64ba8fc9f', Accept='*/*', User-Agent='AHC/2.1'] | Response: status_code=200 | Duration: request_duration=892
```

remerle:
- Add request tracing header for calls from `portal->API`
- Update request logging to include `trace.id`
- Tame down the monitor logging to prevent log pollution
- Add support for JVM param to `build.sbt` to prevent aggressive `scalafmt`

Closes #976

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,11 +1,12 @@
-import CompilerOptions._
-import Dependencies._
 import Resolvers._
+import Dependencies._
+import CompilerOptions._
 import com.typesafe.sbt.packager.docker._
-import microsites._
-import org.scalafmt.sbt.ScalafmtPlugin._
-import sbtrelease.ReleasePlugin.autoImport.ReleaseTransformations._
 import scoverage.ScoverageKeys.{coverageFailOnMinimum, coverageMinimum}
+import org.scalafmt.sbt.ScalafmtPlugin._
+import microsites._
+import ReleaseTransformations._
+import sbtrelease.Version
 
 import scala.util.Try
 
@@ -33,8 +34,8 @@ lazy val sharedSettings = Seq(
   ),
 
   // scala format
-  scalafmtOnCompile := getEnvFlagOrDefault("build.scalafmtOnCompile", true),
-  scalafmtOnCompile in IntegrationTest := true,
+  scalafmtOnCompile := getPropertyFlagOrDefault("build.scalafmtOnCompile", true),
+  scalafmtOnCompile in IntegrationTest := getPropertyFlagOrDefault("build.scalafmtOnCompile", true),
 
   // coverage options
   coverageMinimum := 85,
@@ -217,6 +218,8 @@ lazy val coreBuildSettings = Seq(
   // to write a crypto plugin so that we fall back to a noarg constructor
   scalacOptions ++= scalacOptionsByV(scalaVersion.value).filterNot(_ == "-Ywarn-unused:params")
 ) ++ pbSettings
+
+import xerial.sbt.Sonatype._
 lazy val corePublishSettings = Seq(
   publishMavenStyle := true,
   publishArtifact in Test := false,
@@ -422,23 +425,20 @@ lazy val initReleaseStage = Seq[ReleaseStep](
   setSonatypeReleaseSettings
 )
 
-lazy val finalReleaseStage = Seq[ReleaseStep](
+lazy val finalReleaseStage = Seq[ReleaseStep] (
   releaseStepCommand("project root"), // use version.sbt file from root
   commitReleaseVersion,
   setNextVersion,
   commitNextVersion
 )
 
-def getEnvOrDefault(name: String, value: String) =
-  Option(System.getProperty(name)).getOrElse(value)
-
-def getEnvFlagOrDefault(name: String, value: Boolean): Boolean =
-  Option(System.getProperty(name)).flatMap(x => Try(x.toBoolean).toOption).getOrElse(value)
+def getPropertyFlagOrDefault(name: String, value: Boolean): Boolean =
+  sys.props.get(name).flatMap(propValue => Try(propValue.toBoolean).toOption).getOrElse(value)
 
 releaseProcess :=
   initReleaseStage ++
-    sonatypePublishStage ++
-    finalReleaseStage
+  sonatypePublishStage ++
+  finalReleaseStage
 
 // Let's do things in parallel!
 addCommandAlias("validate", "; root/clean; " +
--- a/build.sbt
+++ b/build.sbt
@@ -1,12 +1,13 @@
-import Resolvers._
-import Dependencies._
 import CompilerOptions._
+import Dependencies._
+import Resolvers._
 import com.typesafe.sbt.packager.docker._
-import scoverage.ScoverageKeys.{coverageFailOnMinimum, coverageMinimum}
-import org.scalafmt.sbt.ScalafmtPlugin._
 import microsites._
-import ReleaseTransformations._
-import sbtrelease.Version
+import org.scalafmt.sbt.ScalafmtPlugin._
+import sbtrelease.ReleasePlugin.autoImport.ReleaseTransformations._
+import scoverage.ScoverageKeys.{coverageFailOnMinimum, coverageMinimum}
+
+import scala.util.Try
 
 resolvers ++= additionalResolvers
 
@@ -32,7 +33,7 @@ lazy val sharedSettings = Seq(
   ),
 
   // scala format
-  scalafmtOnCompile := true,
+  scalafmtOnCompile := getEnvFlagOrDefault("build.scalafmtOnCompile", true),
   scalafmtOnCompile in IntegrationTest := true,
 
   // coverage options
@@ -216,8 +217,6 @@ lazy val coreBuildSettings = Seq(
   // to write a crypto plugin so that we fall back to a noarg constructor
   scalacOptions ++= scalacOptionsByV(scalaVersion.value).filterNot(_ == "-Ywarn-unused:params")
 ) ++ pbSettings
-
-import xerial.sbt.Sonatype._
 lazy val corePublishSettings = Seq(
   publishMavenStyle := true,
   publishArtifact in Test := false,
@@ -314,7 +313,7 @@ lazy val portal = (project in file("modules/portal")).enablePlugins(PlayScala, A
     routesGenerator := InjectedRoutesGenerator,
     coverageExcludedPackages := "<empty>;views.html.*;router.*;controllers\\.javascript.*;.*Reverse.*",
     javaOptions in Test += "-Dconfig.file=conf/application-test.conf",
-    
+
     // ads the version when working locally with sbt run
     PlayKeys.devSettings += "vinyldns.base-version" -> (version in ThisBuild).value,
 
@@ -423,17 +422,23 @@ lazy val initReleaseStage = Seq[ReleaseStep](
   setSonatypeReleaseSettings
 )
 
-lazy val finalReleaseStage = Seq[ReleaseStep] (
+lazy val finalReleaseStage = Seq[ReleaseStep](
   releaseStepCommand("project root"), // use version.sbt file from root
   commitReleaseVersion,
   setNextVersion,
   commitNextVersion
 )
 
+def getEnvOrDefault(name: String, value: String) =
+  Option(System.getProperty(name)).getOrElse(value)
+
+def getEnvFlagOrDefault(name: String, value: Boolean): Boolean =
+  Option(System.getProperty(name)).flatMap(x => Try(x.toBoolean).toOption).getOrElse(value)
+
 releaseProcess :=
   initReleaseStage ++
-  sonatypePublishStage ++
-  finalReleaseStage
+    sonatypePublishStage ++
+    finalReleaseStage
 
 // Let's do things in parallel!
 addCommandAlias("validate", "; root/clean; " +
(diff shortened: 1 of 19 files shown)


pauljamescleary:
Woot!  Nice, I will take a look at this over the weekend.

modules/api/src/main/scala/vinyldns/api/route/RequestLogging.scala
```diff
@@ -0,0 +1,138 @@
...
+
+    val traceId = RequestTracing.extractTraceId(request.headerMap).getOrElse("(unknown)")
+    s"Request: method=${request.method.value}, path=${request.uri} | " +
+      s"trace.id=$traceId | " +
+      s"Headers: [$requestHeadersNoAuth] | " +
+      s"$response | " +
+      s"Duration: request_duration=$duration"
+  }
+
+  private[route] final class VinylDnsRequestIdHeader(requestId: IO[String])
```
pauljamescleary, line 125: why make `requestId` an `IO`?

modules/core/src/main/scala/vinyldns/core/logging/RequestTracing.scala
```diff
@@ -0,0 +1,43 @@
...
+  val requestIdHeaderName = "X-VinylDNS-TraceId"
+
+  /** Extracts the trace identifier from a map of headers */
+  def extractTraceId(headerMap: Map[String, String]): Option[String] =
+    headerMap.get(requestIdHeaderName)
+
+  /** Generates a trace identifier */
+  def generateTraceId: IO[String] = IO {
+    UUID.randomUUID().toString
+  }
```
pauljamescleary, line 34: Curious as to why this is in `IO`?

modules/portal/app/models/VinylRequest.scala
```diff
@@ -33,7 +35,7 @@ case class VinylDNSRequest(
     payload: Option[String] = None,
     parameters: util.HashMap[String, java.util.List[String]] =
       new util.HashMap[String, java.util.List[String]]()
-)
+)(implicit val userRequest: UserRequest[AnyContent])
```
pauljamescleary, line 38: what is this implicit for?  

I have never seen an implicit on a case class itself anywhere.  Typically, you will see the implicit there because it is used elsewhere within the class / instance itself.  Here it seems to be attached as another value?

modules/api/src/main/scala/vinyldns/api/route/RequestLogging.scala
```diff
@@ -0,0 +1,138 @@
...
+
+    /** Returns the request headers as a map. See [[HttpRequestExtension]] */
+    def headerMap: Map[String, String] = req.headers.map(x => x.name() -> x.value()).toMap
+  }
+
+  /** Transforms an incoming request by adding an HTTP header with a tracking ID (if one is not already present).
+    * Should be called before the request is handled by the routes.
+    */
+  def injectTrackingId: Directive0 = mapRequest { request: HttpRequest =>
+    if (request.getHeader(RequestTracing.requestIdHeaderName).isPresent) request
```
pauljamescleary, line 47: this is better as `request.headerMap.contains(RequestTracing.requestIdHeaderName)`

pauljamescleary, line 47: Not a fan that akka is co-mingling java bits and scala bits...

Scala bits ...
`request.headers.exists(_.name == RequestTracing.requestIdHeaderName)`

modules/core/src/main/scala/vinyldns/core/logging/RequestTracing.scala
remerle: `randomUUID` is not referentially transparent and is impure. It's wrapped in the IO monad to convey that.

modules/api/src/main/scala/vinyldns/api/route/RequestLogging.scala
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  /** A custom [[ExceptionHandler]] which logs the exception with the request context. */
+  def loggingExceptionHandler: ExceptionHandler = ExceptionHandler {
+    case NonFatal(e) =>
+      ctx => {
+        val message = Option(e.getMessage).getOrElse("(No error message supplied)")
+        val traceId = RequestTracing.extractTraceId(ctx.request.headerMap).getOrElse("(unknown)")
+
+        val context = s"Exception during request - $message | trace.id=$traceId"
```
pauljamescleary, line 58: to avoid all the codecov gobbly cook, if there isn't anything super important to test, can disable coverage by doing a 

```
// $COVERAGE-OFF$

.. un tested ..
// $COVERAGE-ON$
```

modules/api/src/main/scala/vinyldns/api/route/RequestLogging.scala
remerle: Yeah, tried to be vigilant in avoiding Java. really wish it wasn't possible to do the less-idomatic thing. I'll drop in the scalarific version.

remerle: Yeah, tried to be vigilant in avoiding Java. really wish it wasn't possible to do the less-idomatic thing. I'll drop in the scalarific version.

modules/portal/app/models/VinylRequest.scala
remerle: Yeah, this shouldn't be here. It's effectively adding an additional field (but one that doesn't participate in matching unless you `unapply`). There's a simpler, idiomatic, solution to what I was trying to achieve.

remerle: Yeah, this isn't the best approach. It's effectively adding an additional field with a curried constructor, but I can curry the `executeRequest` method, instead.

modules/api/src/main/scala/vinyldns/api/route/RequestLogging.scala
remerle: I was trying to defer the side effects of generating a request id, but was pretty quickly dead-ended since akka-http doesn't play well with cats.

modules/core/src/main/scala/vinyldns/core/logging/RequestTracing.scala
remerle: My hope was that this would be cleaner with the introduction of http4s and its support for cats IO. I can remove it for now if it adds more confusion than value.

modules/api/src/main/scala/vinyldns/api/route/RequestLogging.scala
remerle: See notes on [related comment](https://github.com/vinyldns/vinyldns/pull/977/files#r487443880)

modules/core/src/main/scala/vinyldns/core/logging/RequestTracing.scala
clinedome: I don't have strong feelings here, but it does seem technically correct. I don't have a problem with wrapping this up in IO, even though it doesn't functionally offer us much right now.

build.sbt
```diff
@@ -423,17 +422,23 @@ lazy val initReleaseStage = Seq[ReleaseStep](
...
-lazy val finalReleaseStage = Seq[ReleaseStep] (
+lazy val finalReleaseStage = Seq[ReleaseStep](
   releaseStepCommand("project root"), // use version.sbt file from root
   commitReleaseVersion,
   setNextVersion,
   commitNextVersion
 )
 
+def getEnvOrDefault(name: String, value: String) =
+  Option(System.getProperty(name)).getOrElse(value)
```
clinedome, line 433: FWIW the scala-native way is to use `scala.sys.env` here, which just returns a map of the env vars. So `sys.env.getOrElse(name, value)` will do what you need.

modules/portal/test/models/VinylRequestSpec.scala
```diff
@@ -20,15 +20,16 @@ import java.io.{BufferedInputStream, BufferedReader, ByteArrayInputStream, Input
...
 import com.amazonaws.HttpMethod
 import com.amazonaws.http.HttpMethodName
 import org.specs2.mock.Mockito
 import org.specs2.mutable._
 import org.specs2.runner._
 import org.junit.runner._
+import play.api.mvc.AnyContent
 
 import scala.collection.JavaConversions._
-
```
clinedome, line 31: You can probably revert these last two files' import changes

clinedome (commented):
Looks pretty workable to me, will need to look again with fresh eyes. Can you drop an example of the updated request logging output into the description? I'm sure it'll want to change to ECS or some such soon, but it's worth noting what this generates.

remerle:
> Looks pretty workable to me, will need to look again with fresh eyes. Can you drop an example of the updated request logging output into the description? I'm sure it'll want to change to ECS or some such soon, but it's worth noting what this generates.

I added details in #976 which show the correlation.  Basically we can see the request on the portal side, the request on the API side, and a correlation to the `userId` (which was `testuser` in my test - it'd be a GUID for a real user)

modules/api/src/main/scala/vinyldns/api/route/RequestLogging.scala
pauljamescleary: Sounds good, this is part of the motivation to http4s, will simplify the API code

modules/core/src/main/scala/vinyldns/core/logging/RequestTracing.scala
pauljamescleary: I agree that keeping impure stuff out is good, but at the same time sprinkling `unsafeRunSync` in the code imo is worse, it's called "unsafe" because you really shouldn't use it in the production code, it should be used at the end of the world when your program is run.  We push all `unsafe` ops into the akka http API via `unsafeToFuture`.

modules/core/src/main/scala/vinyldns/core/logging/RequestTracing.scala
```diff
@@ -0,0 +1,43 @@
...
+
+  /** Creates a trace header name/value pair */
+  def createTraceHeader: (String, IO[String]) = (requestIdHeaderName, generateTraceId)
+
+  /** Retrieves the trace header name/value pair from the given headerMap */
+  def extractTraceHeader(headerMap: Map[String, String]): (String, String) =
+    requestIdHeaderName -> headerMap.getOrElse(requestIdHeaderName, generateTraceId.unsafeRunSync())
```
pauljamescleary, line 41: for example here, `unsafeRunSync` has the ability to throw an exception, but nothing on the return type here indicates an exception maybe thrown.

remerle: Yeah, agreed. I'm going to pull out the IO monad since it doesn't offer the needed guarantees when you can't defer execution back to the origin. Trying to achieve purity in an already impure language is like herding cats.

modules/core/src/main/scala/vinyldns/core/logging/RequestTracing.scala
remerle: Agreed. More precisely because we are indicating intent at the definition, but lose that intent at the caller once we unwrap/`*run*` the monad. 

Pragmatically, conveying intent matters more than trying to work around an IO monad shoehorned into an impure language.

remerle: Agreed. More precisely because we are indicating intent at the definition, but lose that intent at the caller once we unwrap/`*run*` the monad. 

Pragmatically, conveying intent matters more than trying to work around an IO monad shoehorned into an impure language.

clinedome (approved):
I'm satisfied with where this is. Thanks for putting this together! Looking forward to the next steps.

PR #980, opened 2020-09-16
Add backend provider

pauljamescleary:
Introduces the concept of a `Backend` into VinylDNS.  This will allow support for any DNS backend in the future, including AwS Route53 for example.  This is consistent with other "provider" things for dynamic loading of classes (Notifier, Repository, Queue, etc.)

The initial implementation builds on what we have already, that is when creating a zone one can choose a `backendId` that is configured in the `application.conf`.  If no `backendId` is specified, we attempt to map like we do today, so the exact same functionality.

We expand that by allowing one to map a `backendId` to a different provider (like aws). 

After this PR:
1. If someone specifies a zone connection on a zone, it will work exactly like it does today, namely go through the `DnsBackend` to connect.
2. If someone specifies a `backendId` when setting up a zone, the naive mapping will take place to map that zone to the `Backend` implementation that is configured with that `backendId`.   For example, if you have configured a backend id `aws` that connects to Route53, and you specify `aws` when connecting the zone, it will connect to it in Route 53 **Note: we still do not support zone create, but that is much closer to reality with this PR, much much**
3. If someone specifies NEITHER, the `defaultBackendId` will be used, which could be on any one of the backend providers configured.

To start, there is a new `vinyldns.core.domain.backend` package that contains the main classes for the system.  In there you will find the following:

- `BackendProvider` - this is to be implemented by each provider.  Adds a means of pre-loading zones, and providing connections to zones. 
- `Backend` - provides connectivity to a particular backend instance.  For example, a particular DNS Authoritative server.  This is where the real work happens of interacting with whatever backend.  For example, `DnsConnection` implements this to send DDNS messages to the DNS system.  Consider this the "main" thing to implement, where the rubber meets the road, the meat and potatoes
- `BackendProviderLoader` - to be implemented by each provider, knows how to load it's single instance `BackendProvider`, as well as possibly pre-loading configured `Backends` or anything else it needs to do to get ready.  It provides a dynamic hook via the `def load` method that is called by the `BackendLoader` to load a specific `Backend`
- `BackendResolver` - the main, default, BackendResolver.  It holds all `BackendProvider` instances loaded via the `BackendLoader` and provides right now a naive lookup mechanism to find `Backend`s.  Really, this is more of a `Router` or `Resolver`, as in the future it could use more advanced techniques to finding connections than right now
- `BackendConfigs` - used by the `BackendRegistry` as the entrypoint into configuration for all backends
- `BackendProviderConfig` - a single backend provider configuration, specifies a `className` that should be the `BackendProviderLoader` implementation to be loaded, and a `settings` that is passed into the `BackendProvider` to load itself.  This is consistent with other providers.
- `BackendResponse` - uniform responses across all providers to the rest of the VinylDNS System

**Workflow**
During initialization of the system:

1. The `BackendResolver` loads the `BackendConfigs` from the application configuration.  This contains configuration for ALL backends
2. The `BackendResolver` utilizes the `BackendLoader` to dynamically load each backend individually.  If any backend cannot be loaded, it will fail.
3. The `BackendLoader` creates a new instance of each `className` for each `BackendConfig`, this points to the `BackendProviderLoader` implementation which takes care of loading the specific `BackendProvider` provided the configuration
4. The `BackendProviderLoader` does any initialization necessary to ensure it is ready.  In the case of `Route53`, it will pre-load and cache all hosted zones that are available for the AWS account that is configured.  For Route53, a single `Route53Backend` is setup right now.  For `DnsBackend`, a connection (server, port, tsig key) is setup for each DNS Authoritative system to integrate with.

During runtime of the system:

1. When anything is needed, the `BackendResolver` is consulted that will determine how to lookup the `Backend` that is needed.  This is done right now by naively scanning all `BackendProvider` instances it has to say "can anyone connect to this zone".  More intelligent discovery rules can be added in the future
2. Once a `Backend` is obtained, any operation can be performed:
    1. `ZoneConnectionValidator` uses `zoneExists` and `loadZone` to validate a zone is usable by VinylDNS
    2. `RecordSetChangeHandler` uses `resolve` and `applyChange` to apply changes to the DNS backend
    3. `ZoneSyncHandler` and `DnsZoneViewLoader` use `loadZone` in order to load records into VinylDNS

**What else is here**

- Provided an implementation of a backend provider for DNS via `Backend`
- Updated all of VinylDNS to use `Backends` instead of hard coded to DNS
- Provided an implementation of a backend provider for AWS Route 53 as an example to follow for other providers

**Example configuration**

```
vinyldns {
  backend {
    default-backend-id = "r53"

    backend-providers = [
      {
        class-name = "vinyldns.route53.backend.Route53BackendProviderLoader"
        settings = {
          backends = [
            {
              id = "test"
              access-key = "vinyldnsTest"
              secret-key = "notNeededForSnsLocal"
              service-endpoint = "http://127.0.0.1:19009"
              signing-region = "us-east-1"
            }
          ]
        }
      }
    ]
  }
}
```


modules/core/src/main/scala/vinyldns/core/domain/backend/Backend.scala
```diff
@@ -0,0 +1,64 @@
...
+import vinyldns.core.domain.record.RecordType.RecordType
+import vinyldns.core.domain.zone.Zone
+
+/**
+  * Provides the backend interface to work with any kind of DNS backend
+  *
+  * Implement this interface for your own backend.  The default backend is the DnsBackend that talks DDNS.
+  *
+  */
+trait Backend {
```
pauljamescleary, line 30: 👀  This is what matters, all the other stuff is just fluff to get tests to pass and the like

modules/dns/src/main/scala/vinyldns.dns/DnsBackend.scala
```diff
@@ -0,0 +1,98 @@
...
+import org.xbill.DNS.{TSIG, ZoneTransferIn}
+import vinyldns.core.domain.backend.{Backend, BackendResponse}
+import vinyldns.core.domain.record.{RecordSet, RecordSetChange, RecordType}
+import vinyldns.core.domain.record.RecordType.RecordType
+import vinyldns.core.domain.zone.{Zone, ZoneConnection}
+
+import scala.collection.JavaConverters._
+import org.xbill._
+
+class DnsBackend(zoneConnection: ZoneConnection, conn: DnsConnection)
```
pauljamescleary, line 29: Default DNS backend implementation, right now wraps existing calls for the most part.  The load zone was copy 🍝  from the `DnsZoneViewLoader`

This may evolve a little, trying to limit changes on these methods as those will have a larger impact on integration, but we shall see.

The idea with DnsBackend is that you should have one DnsBackend configured per DNS server you are integrating with.

pauljamescleary:
I am actually going to make some changes to this tonight.

Seems like a better approach is to have a `BackendResolver` that has a method `Zone => BackendConnection` 

Also, right now we kind of discard the `backendId` when it comes in.  This should be saved on the `Zone` itself.

This will replace `ZoneConnection` as the mechanism for "connecting" to the dns backend.

The `BackendResolver` will be built in `Boot` using the loader / provider that is already in this PR.

modules/api/src/main/scala/vinyldns/api/backend/dns/DnsConnection.scala
```diff
@@ -84,25 +87,55 @@ class DnsQuery(val lookup: DNS.Lookup, val zoneName: DNS.Name) {
   def error: String = lookup.getErrorString
 }
 
-class DnsConnection(val resolver: DNS.SimpleResolver) extends DnsConversions {
+final case class TransferInfo(address: SocketAddress, tsig: DNS.TSIG)
+
```
pauljamescleary, line 91: changed this up a few times.  As part of moving the ability to do a zone transfer / zone load into `DnsConnection`, I needed certain info (`SocketAddress`, `TSIG`)

modules/api/src/main/scala/vinyldns/api/backend/dns/DnsConnection.scala
```diff
@@ -84,25 +87,55 @@ class DnsQuery(val lookup: DNS.Lookup, val zoneName: DNS.Name) {
...
-    IO {
+  def resolve(name: String, zoneName: String, typ: RecordType): IO[List[RecordSet]] =
+    IO.fromEither {
       for {
         query <- toQuery(name, zoneName, typ)
         records <- runQuery(query)
       } yield records
-    }.toResult
+    }
+
```
pauljamescleary, line 113: This is 98% copy/pasta from `DnsZoneViewLoader`

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneService.scala
```diff
@@ -233,7 +237,7 @@ class ZoneService(
   }
 
   def getBackendIds(): Result[List[String]] =
-    VinylDNSConfig.configuredDnsConnections.dnsBackends.map(_.id).toResult
+    backendRegistry.ids.toList.toResult
```
pauljamescleary, line 240: the super cool thing about this one line change is that any backend configured in the system will be available in the UI!

modules/api/src/main/scala/vinyldns/api/backend/CommandHandler.scala
```diff
@@ -51,14 +50,14 @@ object CommandHandler {
...
-      recordChangeHandler: (DnsConnection, RecordSetChange) => IO[RecordSetChange],
+      recordChangeHandler: (BackendConnection, RecordSetChange) => IO[RecordSetChange],
       zoneSyncHandler: ZoneChange => IO[ZoneChange],
       batchChangeHandler: BatchChangeCommand => IO[Option[BatchChange]],
       mq: MessageQueue,
       count: MessageCount,
       pollingInterval: FiniteDuration,
       pauseSignal: SignallingRef[IO, Boolean],
-      connections: ConfiguredDnsConnections,
+      backendRegistry: BackendRegistry,
```
pauljamescleary, line 60: You will see this change in a lot of places.  Instead of using "DNS" things, now we use "Backend" things, the refactoring while a little pervasive wasn't too bad (I didn't update tests yet so grain of salt on that)

modules/api/src/main/scala/vinyldns/api/backend/dns/DefaultDnsBackend.scala
```diff
@@ -0,0 +1,58 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package vinyldns.api.backend.dns
+
+import vinyldns.core.crypto.CryptoAlgebra
+import vinyldns.core.domain.backend.{Backend, BackendConnection}
+import vinyldns.core.domain.zone.Zone
+
```
pauljamescleary, line 22: Provides a `DnsBackend` implementation that wraps how we currently do things

modules/api/src/main/scala/vinyldns/api/backend/dns/DnsBackendProvider.scala
```diff
@@ -0,0 +1,63 @@
...
+  private implicit val cs: ContextShift[IO] =
+    IO.contextShift(scala.concurrent.ExecutionContext.global)
+
+  /**
+    * Loads a backend based on the provided config so that it is ready to use
+    * This is internally used typically during startup
+    *
+    * @param config The BackendConfig, has settings that are specific to this backend
+    * @return A ready-to-use Backend instance, or does an IO.raiseError if something bad occurred.
+    */
```
pauljamescleary, line 35: Loads DNS backend ids.  The important thing in here is `legacy` support.  The big win there is that this will work without any configuration changes for existing users.

I named it "legacy" because the intent is to move away from it eventually.

This subtle change makes adopting a release with this backend easy (well, existing users do not need to do anything so that's cool)

modules/api/src/main/scala/vinyldns/api/backend/dns/DnsConnection.scala
```diff
@@ -200,19 +251,38 @@ class DnsConnection(val resolver: DNS.SimpleResolver) extends DnsConversions {
 
 object DnsConnection {
 
-  def apply(conn: ZoneConnection): DnsConnection = new DnsConnection(createResolver(conn))
+  def apply(
```
pauljamescleary, line 254: can ignore this, nothing special right here

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneConnectionValidator.scala
```diff
@@ -117,22 +78,22 @@ class ZoneConnectionValidator(connections: ConfiguredDnsConnections)
       ConnectionFailed(zone, "Unable to connect to zone: Transfer connection invalid")
     )
 
-  def hasSOA(records: List[RecordSet], zone: Zone): Result[Unit] = {
```
pauljamescleary, line 120: this used to be the check for "does this zone exist", instead we push that behavior to the backend because it will be different for each provider

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneViewLoader.scala
```diff
@@ -18,49 +18,24 @@ package vinyldns.api.domain.zone
...
+import vinyldns.core.route.Monitored
 
 trait ZoneViewLoader {
   def load: () => IO[ZoneView]
 }
 
 object DnsZoneViewLoader extends DnsConversions {
-
   val logger = LoggerFactory.getLogger("DnsZoneViewLoader")
-
```
pauljamescleary, line 39: moved to `DnsConnection` as it is part of `loadZone` now

modules/api/src/main/scala/vinyldns/api/engine/RecordSetChangeHandler.scala
```diff
@@ -333,12 +333,12 @@ object RecordSetChangeHandler {
...
-    dnsConn.applyChange(change).value.map {
-      case Right(_: NoError) =>
-        Applied(change)
-      case Left(_: Refused) =>
+  private def apply(change: RecordSetChange, conn: BackendConnection): IO[ProcessorState] =
+    conn.applyChange(change).attempt.map {
+      case Right(BackendResponse.Retry(_)) =>
         Retrying(change)
+      case Right(BackendResponse.NoError(_)) =>
+        Applied(change)
```
pauljamescleary, line 341: subtle but important changes.  Before the `RecordSetChangeHandler` used to rely on `DnsResponse` to figure out if it should retry.

Now, with this change, it relies on `BackendResponse` to figure out what to do.  

This same logic of retry was moved to `DnsConnection.applyChange`

modules/core/src/main/scala/vinyldns/core/domain/Fqdn.scala
```diff
@@ -19,6 +19,13 @@ package vinyldns.core.domain
 import DomainHelpers.{ensureTrailingDot, removeWhitespace}
 
 case class Fqdn(fqdn: String) {
+
```
pauljamescleary, line 22: added some nice helpers into Fqdn.  We have code that tries to sort through combining and splitting record and zone names.  I put it all here as that is where it makes sense.  Route53 turns out relies more on FQDN things than DNS

modules/core/src/main/scala/vinyldns/core/domain/backend/BackendConnection.scala
```diff
@@ -0,0 +1,77 @@
...
+import cats.effect.IO
+import vinyldns.core.domain.record.{RecordSet, RecordSetChange}
+import vinyldns.core.domain.record.RecordType.RecordType
+import vinyldns.core.domain.zone.Zone
+
+/**
+  * Provides the backend interface to work with any kind of DNS backend
+  *
+  * Implement this interface for your own backend.  The default backend is the DnsBackend that talks DDNS.
+  */
```
pauljamescleary, line 28: The workhorse of the backend providers

modules/core/src/main/scala/vinyldns/core/domain/record/RecordData.scala
```diff
@@ -19,38 +19,106 @@ package vinyldns.core.domain.record
 import scodec.bits.ByteVector
 import vinyldns.core.domain.Fqdn
 
-sealed trait RecordData
+import scala.util.Try
```
pauljamescleary, line 22: added a lot of convenience helpers and such here as well, in general these are universally useful and we can start taking advantage of them elsewhere.  Right now, used mostly in Route 53

modules/r53/src/it/scala/vinyldns/route53/backend/Route53IntegrationSpec.scala
```diff
@@ -0,0 +1,115 @@
...
+import vinyldns.core.domain.zone.Zone
+
+import scala.collection.JavaConverters._
+import org.scalatest.OptionValues._
+import org.scalatest.EitherValues._
+import vinyldns.core.domain.backend.BackendResponse
+import vinyldns.core.domain.{Fqdn, record}
+import vinyldns.core.domain.record.{RecordSet, RecordType}
+
+class Route53IntegrationSpec
```
pauljamescleary, line 16: Here is the only test I did.  tbh, could be more exhaustive but it tests a lot of scenarios for integration with Route 53 so I am pretty happy with the coverage it will provide.

modules/r53/src/main/scala/vinyldns/route53/backend/Route53BackendConnection.scala
```diff
@@ -0,0 +1,278 @@
...
+/**
+  * Connects to the backend for a single AWS account
+  *
+  * @param id VinylDNS backend identifier used to connect to route 53
+  * @param hostedZones A list of hosted zones, loaded when the application is started.  Necessary
+  *                    as most interactions with Route53 go through the zone id, not the zone name.
+  *                    This will be used as a cache, and on cache miss will lookup the zone in real time
+  * @param client A route 53 client with credentials that can talk to this route 53 aws account
+  */
+class Route53BackendConnection(
```
pauljamescleary, line 47: The meat and potatoes for the Route 53 provider

modules/r53/src/main/scala/vinyldns/route53/backend/Route53BackendConnection.scala
```diff
@@ -0,0 +1,278 @@
...
+    * The record name + zone name should form the FQDN
+    *
+    * @param name     The name of the record (without the zone - e.g. www)
+    * @param zoneName The full domain name (e.g. example.com)
+    * @param typ      The type of record (e.g. AAAA)
+    * @return A list of record sets matching the name, empty if not found
+    */
+  def resolve(name: String, zoneName: String, typ: RecordType): IO[List[RecordSet]] = {
+    for {
+      hostedZoneId <- lookupHostedZone(zoneName)
```
pauljamescleary, line 103: Important: this will return an empty list if the zone is not found.  This is consistent with how things work with DNS in VinylDNS today

modules/r53/src/main/scala/vinyldns/route53/backend/Route53BackendConnection.scala
```diff
@@ -0,0 +1,278 @@
...
+    * Loads all record sets in a zone.  Used typically for zone syncs.
+    *
+    * Note, this will cause memory issues for large zones (100,000s of records).  Need to make
+    * zone sync memory safe before changing this
+    *
+    * @param zone        The zone to load
+    * @param maxZoneSize The maximum number of records that we allow loading, typically configured
+    * @return All record sets in the zone
+    */
+  def loadZone(zone: Zone, maxZoneSize: Int): IO[List[RecordSet]] = {
```
pauljamescleary, line 182: The intent here is to load ALL record sets in the zone.

This is not memory safe, but neither is the rest of Vinyl right now.

Route 53 provides up to 100 records in a single ListResourceRecordSetsRequest, and provides a means to know if there are more records.  Here we recurse until `isTruncated` is false

modules/r53/src/main/scala/vinyldns/route53/backend/Route53BackendProvider.scala
```diff
@@ -0,0 +1,41 @@
...
+ * limitations under the License.
+ */
+
+package vinyldns.route53.backend
+
+import cats.implicits._
+import cats.effect.{ContextShift, IO}
+import vinyldns.core.domain.backend.{Backend, BackendConfig, BackendProvider}
+
+class Route53BackendProvider extends BackendProvider {
```
pauljamescleary, line 23: pretty simple, just loads any pre-configured connections.

modules/r53/src/main/scala/vinyldns/route53/backend/Route53Conversions.scala
```diff
@@ -0,0 +1,93 @@
...
+import org.joda.time.DateTime
+import vinyldns.core.domain.Fqdn
+import vinyldns.core.domain.record.{RecordData, RecordSet, RecordSetStatus}
+import vinyldns.core.domain.record.RecordType.RecordType
+import vinyldns.core.domain.record.RecordType._
+import vinyldns.core.domain.zone.Zone
+
+import scala.collection.JavaConverters._
+
+trait Route53Conversions {
```
pauljamescleary, line 29: buncho helpers for converting from/to vinyldns/route53 things.  I crated a `Conversions` as it follows precedent laid down by `DnsConversions`

pauljamescleary:
Considering this good to go now.  

1. Did some final "renaming" that I am happier with
2. Updated _all_ tests and integration tests (everything passes)
3. All functional tests pass

remerle:
I haven't had a chance to go through all of the code, but your description for the PR describes the implementation exactly as I would imagine it would be implemented. It seems like a pretty standard plugin pattern - _"define a contract and any number of implementations can be provided dynamically through config"_.

Hopefully @clinedome can get some 👀 on this.

pauljamescleary:
@remerle - that is exactly the idea.  The concepts are significant, even if some of the initial implementation is naive:

1. `BackendResolver` - allows VinylDNS to have mechanisms of selecting a backend to send a request to
2. `BackendProvider` - allows a specific backend to decide how or if it will service a request
3. `Backend` - the meat and potatoes once a backend is chosen

Everything else is config or loading.  The dynamic loading using `Class.forName` obviates the need for something like spring, plus it is consistent with the way we do dynamic loading via config in the rest of the system.

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneConnectionValidator.scala
```diff
@@ -117,22 +72,22 @@ class ZoneConnectionValidator(connections: ConfiguredDnsConnections)
...
-    } else {
-      ().asRight[Throwable]
-    }
-  }.toResult
+  def zoneExists(zone: Zone, connection: Backend): Result[Unit] =
+    connection
+      .zoneExists(zone)
+      .ifM(
+        IO(Right(())),
+        IO(Left(ConnectionFailed(zone, "SOA Record for zone not found")))
```
nimaeskandary, line 80: Are there reasons for a backend to fail finding a zone other than `SOA Record for zone not found`? Cant think of any but maybe this error message could come from the provider

modules/r53/src/main/scala/vinyldns/route53/backend/Route53Backend.scala
```diff
@@ -0,0 +1,284 @@
...
+    hostedZones.map(z => z.getName -> z.getId): _*
+  )
+
+  /* Lookup in the local cache, if a new zone is added since start, we have to retrieve it in real time */
+  private def lookupHostedZone(zoneName: String): OptionT[IO, String] = {
+    // pain but we must parse to use the hosted zone ids from the cache
+    def parseHostedZoneId(hzid: String): String = {
+      val lastSlash = hzid.lastIndexOf('/')
+      if (lastSlash > 0) {
+        hzid.substring(lastSlash + 1)
```
nimaeskandary, line 68: would a `substring(lastslash).dropLeft` be safer here?

nimaeskandary (changes requested):
Had a couple minor comments, this is super cool, being able to use vinyl with other providers is an awesome upgrade 🥇  🥳

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneConnectionValidator.scala
pauljamescleary: @nimaeskandary good catch, this is me being lazy before I updated the unit and func tests.  I can change the description.

modules/r53/src/main/scala/vinyldns/route53/backend/Route53Backend.scala
pauljamescleary: Hmm, maybe, let me test that out

pauljamescleary: I don't see a `dropLeft`, the format of the hosted zone id is `/hostedzone/Z111111QQQQQQQ`

nimaeskandary: Ah my bad thought they had a helper like that. If thats the format then 👍

nimaeskandary (approved):

remerle (approved):
LGTM

PR #986, opened 2020-09-30
Fix #985

### LINKED ISSUE #985: Content-Type header not passed from portal to API
**Describe the bug**
Content-Type header not passed from portal to API. When executing a request from the portal which requires a request to the API with a JSON body, we fail to pass the `Content-Type` header resulting in:
```
The request's Content-Type [Some(Some(text/plain; charset=UTF-8))] is not supported. Expected:
application/json
```

This will apply to any portal actions that receive user input (e.g., Batch Changes, Groups, RecordSets, Zones)

**VinylDNS Version**
0.9.5

**To Reproduce**
1. Groups
1. New Group
1. Complete form
1. Click `Create`

**Expected behavior**
Expect a group to be created

**Screenshots**
image

remerle:
- Add the `Content-Type` header implicitly to string request bodies (the default is `text/plain`)

Fixes #985


modules/portal/app/controllers/VinylDNS.scala
```diff
@@ -592,6 +593,14 @@ class VinylDNS @Inject() (
     val credentials = new BasicAWSCredentials(user.accessKey, crypto.decrypt(user.secretKey))
     signer.sign(signableRequest, credentials)
     logger.info(s"Request to send: [${signableRequest.getResourcePath}]")
+
+    // We stringify JSON before it gets here, so we need to specify the content type through this
+    // ugly implicit. If the body came as a JsonValue, we'd get this for free. Also, WSClient does
+    // some weird stuff inside such as preserve content type when specifically overriding the headers.
+    implicit val stringToJsonBody: BodyWritable[String] = {
```
remerle, line 600: I don't like this, but it was the least-bad option. 

[Previously ](https://github.com/vinyldns/vinyldns/blame/c179bcb3a44f00f59e653f2f16921d7730cc99c5/modules/portal/app/controllers/VinylDNS.scala#L594) we had code that set the `Content-Type` header with `withHttpHeaders` as we do below, but this is supposed to replace the headers. Given that we call it twice it was assumed that the `Content-Type` header being set would be lost with the subsequent `withHttpHeaders`.  There's some fun logic in `WSClient` that preserves the `Content-Type` even when the documentation states that `withHttpHeaders` will replace existing headers.

From the docs:
>  Use addHttpHeaders to append additional headers, and withHttpHeaders to overwrite all headers.

nimaeskandary:
tested posts on these portal views from my local and can verify they work

* groups
* zones
* batch 
* recordset search

nimaeskandary (approved):

modules/portal/app/controllers/VinylDNS.scala
pauljamescleary: Why not put back the removed line?  I just did that and it worked...

```
.withHttpHeaders("Content-Type" -> signableRequest.contentType)
```

Since that was proven for the last few years.

pauljamescleary: Deleted my other comment as I read what you typed here.  I'll confirm as well now and then will merge this straight away if all is good.

I will move the 0.9.6 tag to this revision and then do the release.

Thanks!

pauljamescleary (approved):

PR #1001, opened 2020-10-07
review comment replaced with text area

### LINKED ISSUE #995: Increase size of "Review comment" on the Manual Approval screen
Right now the text box is really small.  Given the real estate, this can be a bigger Text Area instead.

P3trur0:
Fixes #995.

Changes in this pull request:

- `dnsChangeDetail.scala.html` modified with the introduction of a `textArea` instead of the `input` tag.

### DIFF
--- a/modules/portal/app/views/dnsChanges/dnsChangeDetail.scala.html
+++ b/modules/portal/app/views/dnsChanges/dnsChangeDetail.scala.html
@@ -174,7 +174,7 @@
                             <h3 class="panel-title">Review DNS Change</h3>
                         </div>
                         <div class="panel-body">
-                            <label>Review Comment (optional):</label> <input type="text" ng-model="$parent.reviewComment">
+                            <label>Review Comment (optional):</label> <textarea ng-model="$parent.reviewComment" class="form-control" rows="3"></textarea>
                         </div>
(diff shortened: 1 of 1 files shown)


pauljamescleary:
The styling on the text area is not great.  The screenshot is below.  Will need to tinker with the CSS a little before merging.

To test this out, need to:

1. Modify `modules/portal/conf/reference.conf` to include the following config entries
2. In sbt type `project root`, then type `dockerComposeUp`
3. In the same console, type `project api`, then `reStart` (case is important)
4. Then hit enter in the same sbt console, and type `project portal` and then `run`
5. Login locally at http://localhost:9001 as `testuser` + `testpassword`
6. Create a DNS Change, make sure the input field is set to `needs-review.shared.` and add an arbitrary A record
7. Click Submit, it goes to the pending review screen where I got this screen shot 

image

pauljamescleary:
Seems like adding ...

```
class="form-control" rows="3"
```

To the `textarea` works pretty well.

P3trur0:
@pauljamescleary, done. Please let me know if it now fits your needs. 😸

pauljamescleary:
Looks great @P3trur0 ! 🎉 

image

pauljamescleary (approved):

PR #1011, opened 2020-10-14
Added key algorithm support

### LINKED ISSUE #156: Make TSIG key algorithm configurable
TSIG keys are currently created and applied in `named.conf` (or `named.conf.local`). As of BIND version 9.4.0 (we are running 9.11.3 in vinyldns-bind9), key algorithms greater than HMAC-MD5 128-bit strength are supported.

While the BIND server can accept different TSIG key algorithms, we need to pass that information to:
- The `ZoneViewLoader` and `DnsConnection` in the main source code
- The Update call in our DNS Python `utils.py`, specifically `dns_do_command`

Make TSIG key algorithm configurable in such a way that different algorithm types can be used with a single running instance of our BIND server.

Alternatively, we can force users to only use HMAC-SHA512.

**Additional context**
We currently generate our HMAC-SHA512 keys with the following command:
```
dnssec-keygen -a HMAC-SHA512 -b 512 -n user <key_name>
```
`dnssec-keygen` requires that bind-utils are installed:
- Download a release tagged with `P` (eg. `v9_10_2_P4`): https://github.com/tigeli/bind-utils/releases
- Unarchive file (eg. `cd ~/Downloads && tar -xzf <file_name>`)
- Run `configure` in new directory (eg. `cd <directory_name> && configure). May need to link `openssl` for installation to work.
- Run `make`.
- Run `make install`. May need to run with `sudo`. Utils will appear in `bin` folder.

### LINKED ISSUE #964: make possible other key algorithms than HMAC-MD5, maybe HMAC-SHA256
only HMAC-MD5 is possible for now

possibility of using stronger keys would be a nice enhancement

@pauljamescleary suggested that maybe it is possible to change a key name to "algorithm:keyName"

pauljamescleary:
Fixes #964 

- Updated the `ZoneConnection` model to allow specifying the key algorithm.
- Added an `Algorithm` to the protobuf file, defaults to HMAC-MD5
- Updated JSON serialization to serdes the algorithm
- Updated the Portal to allow the user to specify the algorithm when connecting to a zone or managing a zone

Supported algorithms are:

```
  case object HMAC_MD5 extends Algorithm("HMAC-MD5.SIG-ALG.REG.INT")
  case object HMAC_SHA1 extends Algorithm("hmac-sha1.")
  case object HMAC_SHA224 extends Algorithm("hmac-sha224.")
  case object HMAC_SHA256 extends Algorithm("hmac-sha256")
  case object HMAC_SHA384 extends Algorithm("hmac-sha384.")
  case object HMAC_SHA512 extends Algorithm("hmac-sha512.")
```

**Note: needs some tests**


pauljamescleary:
@JanuszU let me know your thoughts if you happen across this

pauljamescleary:
Example "Manage Zone" tab...

image

pauljamescleary:
Example "Connect to zone" dialog...

image

modules/core/src/main/scala/vinyldns/core/domain/zone/Zone.scala
```diff
@@ -139,7 +140,33 @@ final case class ZoneACL(rules: Set[ACLRule] = Set.empty) {
   def deleteRule(rule: ACLRule): ZoneACL = copy(rules = rules - rule)
 }
 
-case class ZoneConnection(name: String, keyName: String, key: String, primaryServer: String) {
+sealed abstract class Algorithm(val name: String) {
+  override def toString: String = name
+}
+object Algorithm {
+  case object HMAC_MD5 extends Algorithm("HMAC-MD5.SIG-ALG.REG.INT")
```
pauljamescleary, line 147: I could change these to more "humanized" names, as the important conversion happens in the `DnsBackend` anyways.

modules/core/src/main/protobuf/VinylDNSProto.proto
```diff
@@ -3,11 +3,21 @@ syntax = "proto2";
...
 option optimize_for = SPEED;
 
+enum Algorithm {
+  HMAC_MD5 = 0;
+  HMAC_SHA1 = 1;
+  HMAC_SHA224 = 2;
+  HMAC_SHA256 = 3;
+  HMAC_SHA384 = 4;
+  HMAC_SHA512 = 5;
+}
```
pauljamescleary, line 13: protobuf has to be a positive int (or 0)

modules/core/src/main/protobuf/VinylDNSProto.proto
```diff
@@ -3,11 +3,21 @@ syntax = "proto2";
...
+
 message ZoneConnection {
   required string name = 1;
   required string keyName = 2;
   required string key = 3;
   required string primaryServer = 4;
+  optional Algorithm algorithm = 5 [default = HMAC_MD5];
```
pauljamescleary, line 20: have to include a default (which is MD5) since we have lots of zone connections out there (all of which are MD5)

modules/core/src/main/scala/vinyldns/core/domain/zone/Zone.scala
JanuszU: Looks great ! Exactly what I meant
Heapy coding

pauljamescleary:
Adding end to end tests now.  I created additional keys for all algorithms and added them to the "one-time" zone, this PR will be updated shortly to include those changes and then I think we are good.

modules/core/src/main/scala/vinyldns/core/domain/zone/Zone.scala
pauljamescleary: @JanuszU I changed the names to be "humanized", i.e. "HMAC-MD5" instead of "HMAC-MD5.SIG-ALG.REG.INT".

Also, fwiw, the configuration file will use those same values when configuring default connections / backends

slandry90:
Another note on this, and this may be better suited for a different MR, but it would also be nice to have the entire tsig be optional. We attempted to test with CoreDNS and were unable to make a connection to the zone because coredns does not currently support using tsig for zone transfers.

pauljamescleary:
> Another note on this, and this may be better suited for a different MR, but it would also be nice to have the entire tsig be optional. We attempted to test with CoreDNS and were unable to make a connection to the zone because coredns does not currently support using tsig for zone transfers.

Interesting, I will give this some thought.  Since it is already required, making it optional would be difficult from a protobuf / domain standpoint.  

Since it is just a string, perhaps could make it an `Option[String]` if it is an empty string `""` assume it is not there and skip the TSIG key.  Could do that for both zone connection and transfer connection

pauljamescleary:
@slandry90 we can fix that issue in a separate PR (making tsig key optional for a transfer connection).  I think it is best if it is implemented in a different way.   Meanwhile, this is good to go I think now that I have a func test in.

PR #1015, opened 2020-10-14
Route 53 fixes for connect

### LINKED ISSUE #1012: Route53 Zone doesn't connect and load records properly
**Describe the bug**
When trying to connect a route53 zone, we encounter an error "Missing apex NS record". This is happening because name of the map on the records does not match the name of the maps for the recordSets and is therefore just returning None.

We did some digging in the code and were able to isolate the issue down to Route53Conversions `toVinylRecordSet` function.. `Fqdn.merge(r53RecordSet.getName, zoneName).firstLabel` should be `r53RecordSet.getName.toString` so the name of the map is correct consistently.

Once we fixed the above and allowed the zone to be added we ran into additional issues where no records show up in the zone on vinyl (even though they exist in r53, no errors) and records cannot be added or removed (error out with "Failed validating update to DNS for change 784b4a3d-b3a8-4b82-8596-0c684fa34ef3:jenkins: Incompatible record already exists in DNS." for example)

Records appear to be missing because the zoneid is "unknown" in Route53Conversions, we updated `toVinylRecordSets` in the `loadZone` function which added the zoneid to the recordSets and they now appear in the zone in vinyl but we are still unable to CRUD records through vinyl for the zone

sidebar, SOA records are also broken but we'll open a new issue for that and we'll have an MR for that as well.

**VinylDNS Version**
0.9.7

**To Reproduce**
1. Add route53 zone to vinyldns through portal returns "Missing apex NS record"

pauljamescleary:
Fixes #1012 

Changes in this pull request:
- When available, ensure that the `zoneId` is passed when loading vinyldns record sets
- Use the `DelegationSet` on the hosted zone to _simulate_ NS records.  NS records do not exist by default on private hosted zones

### DIFF
--- a/modules/api/src/it/scala/vinyldns/api/route53/Route53ApiIntegrationSpec.scala
+++ b/modules/api/src/it/scala/vinyldns/api/route53/Route53ApiIntegrationSpec.scala
@@ -18,12 +18,14 @@ package vinyldns.api.route53
 
 import cats.data.NonEmptyList
 import cats.effect.IO
+import com.amazonaws.services.route53.model.DeleteHostedZoneRequest
 import org.scalatest.concurrent.ScalaFutures
 import org.scalatest.matchers.should.Matchers
-import org.scalatest.wordspec.AnyWordSpecLike
+import org.scalatest.wordspec.AnyWordSpec
 import org.scalatest.{BeforeAndAfterAll, BeforeAndAfterEach}
 import org.scalatestplus.mockito.MockitoSugar
 import scalikejdbc.DB
+import vinyldns.api.domain.zone.ZoneConnectionValidator
 import vinyldns.api.engine.ZoneSyncHandler
 import vinyldns.api.{MySqlApiIntegrationSpec, ResultHelpers}
 import vinyldns.core.TestRecordSetData._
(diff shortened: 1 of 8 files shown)


modules/api/src/it/scala/vinyldns/api/route53/Route53ApiIntegrationSpec.scala
```diff
@@ -0,0 +1,105 @@
...
+import scalikejdbc.DB
+import vinyldns.api.engine.ZoneSyncHandler
+import vinyldns.api.{MySqlApiIntegrationSpec, ResultHelpers}
+import vinyldns.core.TestRecordSetData._
+import vinyldns.core.domain.backend.{Backend, BackendResolver}
+import vinyldns.core.domain.record.{NameSort, RecordType}
+import vinyldns.core.domain.zone.{Zone, ZoneChange, ZoneChangeType}
+import vinyldns.core.health.HealthCheck.HealthCheck
+import vinyldns.route53.backend.{Route53Backend, Route53BackendConfig}
+
```
pauljamescleary, line 35: This will be a good place to put future integration tests that test more than just the backend

modules/r53/src/main/scala/vinyldns/route53/backend/Route53Conversions.scala
```diff
@@ -76,9 +85,31 @@ trait Route53Conversions {
 
   def toVinylRecordSets(
       r53RecordSets: java.util.List[ResourceRecordSet],
-      zoneName: String
+      zoneName: String,
+      zoneId: String = "unknown"
   ): List[RecordSet] =
-    r53RecordSets.asScala.toList.map(toVinylRecordSet(zoneName, _))
+    r53RecordSets.asScala.toList.map(toVinylRecordSet(zoneName, zoneId, _))
+
```
pauljamescleary, line 92: NS record sets from a delegation set have a record name that equals the zone name, and a single record set with each name server from the delegation set on the hosted zone

PR #1017, opened 2020-10-16
Fix route53 issue with merge fqdn

pauljamescleary:
Addresses #1012

Was able to get a "real" public hosted zone connected.  The `Fqdn.merge` was not working properly when merging record name and zone names when they matched.

Added a bunch of unit tests along with the fix.

This was tested against a "real" public hosted zone.  Verified connect, load zone, add record, delete record.


pauljamescleary:
Addresses #1012

Was able to get a "real" public hosted zone connected.  The `Fqdn.merge` was not working properly when merging record name and zone names when they matched.

Added a bunch of unit tests along with the fix.

This was tested against a "real" public hosted zone.  Verified connect, load zone, add record, delete record.

PR #1019, opened 2020-10-16
R53 soa fix

### LINKED ISSUE #1018: SOA Records not read into Route53
**Describe the bug**
SOA Records from route53 are not correctly populated into vinyldns

**VinylDNS Version**
0.9.7 updated master branch

**To Reproduce**
import a r53 zone, see unknown soa record

slandry90:
Fixes #1018 .

Also fixes a configuration merge bug now that access-key and secret-key are optional fields, this is non breaking since no one has x/x for their key/secret

### DIFF
--- a/modules/api/src/main/resources/reference.conf
+++ b/modules/api/src/main/resources/reference.conf
@@ -72,8 +72,8 @@ vinyldns {
 
     settings {
       # AWS access key and secret.
-      access-key = "x"
-      secret-key = "x"
+      # access-key = "x"
+      # secret-key = "x"
 
       # Regional endpoint to make your requests (eg. 'us-west-2', 'us-east-1', etc.). This is the region where your queue is housed.
       signing-region = "x"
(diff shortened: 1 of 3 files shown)


modules/api/src/main/resources/reference.conf
```diff
@@ -72,8 +72,8 @@ vinyldns {
 
     settings {
       # AWS access key and secret.
-      access-key = "x"
-      secret-key = "x"
+      # access-key = "x"
```
abhide, line 75: https://github.com/vinyldns/vinyldns/issues/1020

modules/core/src/main/scala/vinyldns/core/domain/record/RecordData.scala
```diff
@@ -46,6 +46,7 @@ object RecordData {
     case SRV => SRVData.fromString(value)
     case SSHFP => SSHFPData.fromString(value)
     case TXT => TXTData.fromString(value)
+    case SOA => SOAData.fromString(value)
```
pauljamescleary, line 49: Nice catch

pauljamescleary (approved):

PR #1022, opened 2020-10-22
Overhaul config

### LINKED ISSUE #1010: Fix build issue with config
Used to see this frequently in travis, seeing it in GHA as well.  Unclear why this is happening but really need to fix this so our builds are less flaky...

```
com.typesafe.config.ConfigException$Missing: merge of system properties,reference.conf @ jar:file:/home/runner/.sbt/boot/scala-2.12.10/org.scala-sbt/sbt/1.3.9/ssl-config-core_2.12-0.4.0.jar!/reference.conf: 1: No configuration setting found for key 'vinyldns'
```

pauljamescleary:
This is a rather big change.  There are a few significant issues with the way that the API config is presently loaded:

1. We use effectively global variables throughout the system, which is a bad practice in general
2. We have inconsistent loading of configuration values, some used at boot up, some used elsewhere

In addition, we get sporadic build failures due to how these "global config" values are loaded, based on timing and parallelism that are impossible to reproduce.

This PR addresses these issues:

1. Create a `VinylDNSConfig` that loads all configuration in one place
2. Create custom `ConfigReader` implementations that read config values (ideally we would have used pureconfig from the start to automatically read sane config values but here we are)
3. Segment config into different case classes.  The groupings are not totally arbitrary, but I did my best at logical groupings of settings
4. Inject configuration elements (either via Class constructors or function arguments) at the appropriate time.

Functionally, nothing has changed, other than putting some standards around config loading.

May fix #1010


modules/api/src/main/resources/reference.conf
```diff
@@ -194,4 +194,10 @@ vinyldns {
 
   # Global ACLs defaults to an empty list
   global-acl-rules = []
+
+  health-check-timeout = 10000
+
+  default-ttl = 7200
+
+  validate-record-lookup-against-dns-backend = false
```
pauljamescleary, line 202: these had no default values before, default values _should_ be specified in `reference.conf`

modules/api/src/main/scala/vinyldns/api/Boot.scala
```diff
@@ -37,7 +38,7 @@ import vinyldns.api.metrics.APIMetrics
 import vinyldns.api.repository.{ApiDataAccessor, ApiDataAccessorProvider, TestDataLoader}
 import vinyldns.api.route.VinylDNSService
 import vinyldns.core.VinylDNSMetrics
-import vinyldns.core.domain.backend.{BackendConfigs, BackendResolver}
+import vinyldns.core.domain.backend.BackendResolver
```
pauljamescleary, line 41: A fair amount of change in this file, but no functional changes.  Using `VinylDNSConfig` instead of arbitrarily loading values, passing new values into classes, that kind of thing.

modules/api/src/main/scala/vinyldns/api/backend/CommandHandler.scala
```diff
@@ -204,15 +204,23 @@ object CommandHandler {
       recordChangeRepo: RecordChangeRepository,
       batchChangeRepo: BatchChangeRepository,
       notifiers: AllNotifiers,
-      backendResolver: BackendResolver
+      backendResolver: BackendResolver,
+      maxZoneSize: Int
```
pauljamescleary, line 208: an example of injecting config values instead of relying on global values.  Many of the code updates in production code do the same thing

modules/api/src/main/scala/vinyldns/api/config/BatchChangeConfig.scala
```diff
@@ -0,0 +1,42 @@
...
+      "shared-approved-types",
+      "v6-discovery-nibble-boundaries"
+    ) {
+      case (limit, approvedTypes, v6) =>
+        BatchChangeConfig(
+          limit,
+          approvedTypes.flatMap(RecordType.find),
+          v6
+        )
+    }
```
pauljamescleary, line 41: the above is an example `ConfigReader` from pureconfig, this pattern is repeated below.

This is generally how we should be loading config, using implicit `ConfigReader` on companion objects everywhere.

modules/api/src/main/scala/vinyldns/api/config/ScheduledChangesConfig.scala
```diff
@@ -0,0 +1,27 @@
...
+
+import pureconfig.ConfigReader
+
+final case class ScheduledChangesConfig(enabled: Boolean)
+object ScheduledChangesConfig {
+  implicit val configReader: ConfigReader[ScheduledChangesConfig] =
+    ConfigReader.forProduct1[ScheduledChangesConfig, Boolean]("scheduled-changes-enabled")(
+      ScheduledChangesConfig(_)
+    )
+}
```
pauljamescleary, line 27: A little premature for a single boolean, but I plan on adding "auto processing" of scheduled changes shortly, so I anticipate additional config elements being read in here

modules/api/src/main/scala/vinyldns/api/config/VinylDNSConfig.scala
```diff
@@ -0,0 +1,116 @@
...
+      }
+
+    def loadFromStringListIO[A](
+        config: Config,
+        path: String
+    )(implicit cr: ConfigReader[A], classTag: ClassTag[A]): IO[List[A]] =
+      optionalStringListIO(config, path).flatMap { configKeys =>
+        configKeys.traverse(k => loadIO[A](config, s"vinyldns.$k"))
+      }
+
```
pauljamescleary, line 80: This is the primary change.  We load all config settings in here now and should for all settings moving forward.

modules/api/src/main/scala/vinyldns/api/domain/access/AccessValidations.scala
```diff
@@ -26,8 +25,10 @@ import vinyldns.core.domain.record.RecordType.RecordType
 import vinyldns.core.domain.zone.AccessLevel.AccessLevel
 import vinyldns.core.domain.zone.{ACLRule, AccessLevel, Zone}
 
-class AccessValidations(globalAcls: GlobalAcls = GlobalAcls(List.empty))
-    extends AccessValidationsAlgebra {
+class AccessValidations(
+    globalAcls: GlobalAcls = GlobalAcls(List.empty),
+    sharedApprovedTypes: List[RecordType.Value] = Nil
+) extends AccessValidationsAlgebra {
```
pauljamescleary, line 31: another example of setting injection instead of using global values

modules/api/src/main/scala/vinyldns/api/domain/access/AccessValidations.scala
```diff
@@ -234,5 +235,5 @@ class AccessValidations(globalAcls: GlobalAcls = GlobalAcls(List.empty))
       recordOwnerGroupId: Option[String]
   ): Boolean =
     recordOwnerGroupId.exists(auth.isGroupMember) ||
-      (recordOwnerGroupId.isEmpty && VinylDNSConfig.sharedApprovedTypes.contains(recordType))
```
pauljamescleary, line 237: see this global value, ugh, removed everywhere

remerle (approved):
LGTM

PR #1023, opened 2020-10-22
Update microsite, move to mdoc

pauljamescleary:
We used to rely on `tut` for docs, however it is deprecated to be replaced with mdoc.

Moved to an `mdoc` folder structure and updated all of the links (what a pain).


modules/docs/src/main/mdoc/api/approve-batchchange.md
```diff
@@ -9,9 +9,9 @@ section: "api"
 Manually approves a batch change in pending review status given the batch change ID, resulting in revalidation and
 submission for backend processing. Only system administrators (ie. support or super user) can manually review a batch
 change. In the event that a batch change is approved and still encounters non-fatal errors, it will remain in manual
-review state until a successful approval (**202** Accepted) or [rejection](../api/reject-batchchange) (**200** OK).
+review state until a successful approval (**202** Accepted) or [rejection](reject-batchchange.md) (**200** OK).
```
pauljamescleary, line 12: 99.9% of the changes were little things like this, had to add `.md` to the end of links.

pauljamescleary:
Everything works fine locally.  Have to publish on gh pages to know if it works in GH, hard to test that here.

There is a nicer "light" theme, but the css images have to be svg as far as I could gather so I couldn't use that

PR #1026, opened 2020-10-26
Enabling important features by default

TomSebty:
Fixes #.

Changes in this pull request:
- Added `approved-name-servers` and assigned a default value to it.
- Enabled the `shared` drop-down-list option for the zone in the portal by default.

### DIFF
--- a/docker/api/docker.conf
+++ b/docker/api/docker.conf
@@ -17,6 +17,8 @@
 ################################################################################################################
 vinyldns {
 
+  approved-name-servers = [ ".*" ]
+
   # configured backend providers
   backend {
     # Use "default" when dns backend legacy = true
--- a/docker/portal/application.conf
+++ b/docker/portal/application.conf
@@ -55,5 +55,8 @@ mysql {
   }
 }
 
+shared-display-enabled = true
+multi-record-batch-change-enabled = true
+
 # You generate this yourself following https://www.playframework.com/documentation/2.7.x/ApplicationSecret
 play.http.secret.key = "rpkTGtoJvLIdIV?WU=0@yW^x:pcEGyAt`^p/P3G0fpbj9:uDnD@caSjCDqA0@tB="


docker/portal/application.conf
```diff
@@ -55,5 +55,8 @@ mysql {
   }
 }
 
+shared-display-enabled = true
```
pauljamescleary, line 58: `shared-display-enabled` is only a flag in the `portal` config.

docker/api/docker.conf
```diff
@@ -17,6 +17,8 @@
 ################################################################################################################
 vinyldns {
 
+  approved-name-servers = [ ".*" ]
```
pauljamescleary, line 20: unfortunately, this will likely break functional tests.

The functional tests are run against whatever config is living in the `docker/api` folder.

Perhaps this configuration is better suited for the `build` directory, as those configs / docker images are what go into public docker hub at `docker.io`

pauljamescleary:
Thanks @TomSebty !

docker/portal/application.conf
TomSebty: Isn't this the docker config file? I mean it is inside the docker directory and that is where we added the flag to set it.

pauljamescleary: Doh, I mis-read this in the review, disregard my comment

docker/api/docker.conf
TomSebty: Yeah I saw the log of the test...

pauljamescleary: Right now, we are using the `docker` directory for our build automation as well as the quick start.  This was a fast way to show the system running without much effort

A better alternative would be to separate out the "demo" / "quickstart" into something else entirely.  This way, a user can see what a closer to real system + system config would look like.  Perhaps an "example" directory or something?

I actually have a thought out demo based on futurama I did recently that I could push up.   We could even add a few example configurations to that example, one using sqs and another without.

It does seem like it is time we take this step.  I don't think it would be hard to push this up.

TomSebty: Great thought

pauljamescleary: Ok, I will put something together and mention you on the PR.  Really want the config to make sure the config makes sense in that.

pauljamescleary:
Apologies for not getting back to this.  I have the demo mostly ready, along with a guide that follows the script I ran for the webinar.  Just need a few minutes to complete it and post it up.

remerle:
Closing due to age

PR #1035, opened 2021-03-03
Drop warning for dotted hosts on valid records

### LINKED ISSUE #1034: Exception to dotted hostname rule for ACME DNS-01 Challenge
In order to make use of ACME based SSL certificate clients, ownership of DNS records are handled through challenges involving HTTP, TLS, DNS before certificates are issued.  If for firewall reasons there is not a path via HTTP or TLS, that only leaves you with the [DNS-01 challenge mechanism](https://tools.ietf.org/html/rfc8555#section-8.4).  In short, the DNS verification method entails generating a temporary DNS TXT record `_acme-challenge.<hostname>.<domainname>` with a verification token string value.  The problem with the existing implementation of VinylDNS today is that in order to create that `_acme-challenge` record, a full zone must be created and enrolled to vinylDNS per host.  While technically possible, it's not practical.  Solving this problem will allow any VinylDNS environment to leverage the ACME protocol for automating SSL certificate creation/renewal from any ACME providing certificate authority such as Let's Encrypt.

I'd like to see an exception to the existing rule forbidding dotted hostnames for specifically that prefix on TXT records only.  This would alleviate the overhead and zone maintenance costs associated with these temporary records.  If it's something that administrators would like some kind of control over or is only available on certain DNS backends, that's ok.

The amount of overhead required to add and delete full SOA zones in addition to onboard/offboarding to vinylDNS is an option, but very significant.  Leveraging either the TLS or HTTP ACME challenge methods is an option if it is internet facing and the webserver software supports serving static directory content or ACME directly.  Both of these conditions are not universally available without creating significant single points of failure and complicated proxy setups to maintain.

pauljamescleary:
Fixes #1034 

Changes in this pull request:
- Update record view for managing records to _not_ display a warning next to records that are valid dotted hosts.
- Update the prepare portal script to do a force (needed for clean machine)
- Add queue coordinates for starting up the api locally using `reStart` - the coordinates were removed from `reference.conf` in a prior PR that looks to have broken starting up using `reStart` locally
- Added a unit test to double check we allow dotted TXT records in the API


pauljamescleary:
image

PR #1046, opened 2021-08-27
Improve messages

Aravindh-Raju:
Changes in this pull request:
- Improved messages for some of the existing ones.
- Created an object Messages.scala in vinyldns.core which holds the new messages.
- Consolidated the new messages in an object so that they are easier to manage and can be pulled into configuration.


remerle (approved):
LGTM

PR #1047, opened 2021-09-09
Add error for group creation with email that already exists

Aravindh-Raju:
Changes in this pull request:
- Error message will be displayed when we create a new group with email id that's already used by another group.
- Error message will be displayed when we update the group email id with a new one that's already used by another group.
- The error message would recommend the user to contact the existing group to be added to that group. This would reduce the confusion of having multiple groups with same email id.
- Created a new SQL query to get group by email, Created a function to display the error message, Added new test cases.
- Created a configuration whether to enforce unique email for groups or not.
- Updated the FAQ of the microsite with details about the configuration.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -35,6 +35,7 @@ Thank you! If you have contributed in any way, but do not see your name here, pl
 - Joshulyne Park
 - Nathan Pierce
 - Michael Pilquist
+- Aravindh Raju
 - Sriram Ramakrishnan
 - Khalid Reid
 - Timo Schmid
@@ -47,4 +48,3 @@ Thank you! If you have contributed in any way, but do not see your name here, pl
 - Fei Wan
 - Andrew Wang
 - Peter Willis
-- Aravindh Raju
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -47,3 +47,4 @@ Thank you! If you have contributed in any way, but do not see your name here, pl
 - Fei Wan
 - Andrew Wang
 - Peter Willis
+- Aravindh Raju
--- a/modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
@@ -25,6 +25,7 @@ import vinyldns.core.domain.zone.ZoneRepository
 import vinyldns.core.domain.membership._
 import vinyldns.core.domain.record.RecordSetRepository
 import vinyldns.core.Messages._
+import vinyldns.api.config.GroupEmailConfig
 
 object MembershipService {
   def apply(dataAccessor: ApiDataAccessor): MembershipService =
@@ -56,7 +57,7 @@ class MembershipService(
     for {
       _ <- hasMembersAndAdmins(newGroup).toResult
       _ <- groupWithSameNameDoesNotExist(newGroup.name)
-      _ <- groupWithSameEmailIdDoesNotExist(newGroup.email)
+      _ <- groupWithSameEmailIdDoesNotExist(newGroup.email, GroupEmailConfig.enforce_unique_email_address)
       _ <- usersExist(newGroup.memberIds)
       _ <- groupChangeRepo.save(GroupChange.forAdd(newGroup, authPrincipal)).toResult[GroupChange]
       _ <- groupRepo.save(newGroup).toResult[Group]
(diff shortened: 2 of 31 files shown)


AUTHORS.md
```diff
@@ -47,3 +47,4 @@ Thank you! If you have contributed in any way, but do not see your name here, pl
 - Fei Wan
 - Andrew Wang
 - Peter Willis
+- Aravindh Raju
```
remerle, line 50: These should be in alphabetical order by last name, so you can move your name up

remerle (changes requested):
For backward compatibility, we need to add a configuration flag that can toggle this functionality on and off. Take a look at the existing configuration [here](https://github.com/vinyldns/vinyldns/tree/master/modules/api/src/main/scala/vinyldns/api/config). You should be able to add something like `vinyldns.membership.groups.enforce_unique_email_address` or something similar.

By default, this flag should be set to `false` (again for backward compatibility).

AUTHORS.md
Aravindh-Raju: Sure, I'm working on the configuration. I'll commit the changes once done.

modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
```diff
@@ -242,6 +251,23 @@ class MembershipService(
       }
       .toResult
 
+  def groupWithSameEmailIdDoesNotExist(emailId: String, emailConf: Boolean): Result[Unit] =
```
remerle, line 254: It may be clearer if the parameter `emailConf` was named `enforceUniqueEmailId`

modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
```diff
@@ -264,6 +290,28 @@ class MembershipService(
       }
       .toResult
 
+  def differentGroupWithSameEmailIdDoesNotExist(
+      emailId: String,
+      groupId: String,
+      emailConf: Boolean
```
remerle, line 296: It may be clearer if the parameter `emailConf` was named `enforceUniqueEmailId`

modules/docs/src/main/mdoc/faq.md
```diff
@@ -88,3 +89,9 @@ If you use any VinylDNS tools beyond the portal you will need to provide those t
 
 ### 9. How are requests authenticated to the VinylDNS API? <a id="9"></a>
 Refer to [API Authentication](api/auth-mechanism.html).
+
+### 10. How to change email configuration for groups? <a id="10"></a>
+By default, VinylDNS is configured to disallow multiple groups with the same email address. We have found that 
+supporting multiple groups with the same email address to be confusing to many users when there are ownership questions 
+in the future. Therefore, we strongly recommend that VinylDNS deployments utilize this feature. If, however, you do not 
+wish you use this feature, you can set "vinyldns.groups.enforce-unique-email-id = false" in the configuration.
```
remerle, line 97: For backward compatibility, this feature should be off by default

```suggestion
VinylDNS can be configured to disallow multiple groups with the same email address. We have found that 
supporting multiple groups with the same email address to be confusing to many users when there are ownership questions 
in the future. Therefore, we strongly recommend that VinylDNS deployments utilize this feature. If you wish to use this feature, you can set "vinyldns.groups.enforce-unique-email-id = true" in the configuration.
```

modules/portal/public/templates/notification.html
```diff
@@ -1,5 +1,10 @@
 <div class="alert alert-box" ng-class="'alert-'+ngModel.type" role="alert">
     <button type="button" class="close" data-dismiss="alert">
         <span aria-hidden="true">×</span><span class="sr-only">Close</span>
-    </button>{{ngModel.content}}
+    </button>
+    <div>{{ngModel.content}}
+    <a ng-if="ngModel.content.includes('HTTP 409 (Conflict): Cannot create group.') || ngModel.content.includes('HTTP 409 (Conflict): Cannot update group.')"
+       style="text-decoration: none;"
+       href="https://www.vinyldns.io/faq.html#10">ℹ️</a>
+    </div>
```
remerle, line 9: I'd prefer we didn't try to check against the text that is returned. If we get around to localizing VinylDNS, then this won't work. Let's leave this change out for now.

remerle (changes requested):

modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
Aravindh-Raju: Sure, I've made the changes.

PR #1050, opened 2021-09-22
docker and make file added for functional test to execute in seperate…

Jay07GIT:
Changes in this pull request:
- Created a docker file and make file for run the functional test in the container.
- Since Functional test folder is necessary inside the container to execute the test scripts it has to be maintained inside docker folder.(where our docker file for functional test is present there function_test folder should be present).
- Since we are maintaining functional_test folder inside our container we no need to maintain the conftest.py details in dockerfile (using environmental variables).
- Now we can give $make docker_api_up command which will run the docker-up-api-server.sh file(This file points to docker-compose-func-test.yml which will execute API,SQL,BIND9, Functional_test). 
- We can give $make build-functestimage to create a seperate image which will execute dockerfile for functional tests.

Note: These changes are not a final one. we can make changes accordingly.


remerle (changes requested):
Let's add the `Dockerfile` and `Makefile` to the existing `modules/api/functional_test` directory rather than moving files around.

This PR should consist of a `Dockerfile` and, optionally, a `Makefile`. The `Makefile` would simply be the tool to provide `build-docker` and `run` targets. The former being responsible for creating a docker image, the latter for running tests in that container.

Jay07GIT:
Can I keep a docker and run.sh file in Modules>api>functional_test and funtional test scripts inside Modules>api>functional_test>functional_test.

**hierarchy:**
vinyldns
-modules
--api
---functional_test
----Dockerfile
----run.sh
----functional_test

This will avoid confusion where docker file belongs to what.

Jay07GIT:
i made the changes on the docker and make file. Now with the **makefile** we can build and run the functional test container and scripts can be execute inside it. 

I tested the code locally and its running the test cases.
image

Jay07GIT:
Now Functional test , api , local stack, bind9, mysql runs in same environment and drops after functional test scripts executed.

This can be achieved through run
$ ./bin/func-test-api-.sh

remerle:
> Can I keep a docker and run.sh file in Modules>api>functional_test and funtional test scripts inside Modules>api>functional_test>functional_test.
> 
> **hierarchy:**
> vinyldns
> -modules
> --api
> ---functional_test
> ----Dockerfile
> ----run.sh
> ----functional_test
> 
> This will avoid confusion where docker file belongs to what.
> 
> Or else if we dont want to touch the existing code structure we can go with your above comments.

I don't want to have an additional directory. The `Dockerfile` , `Makefile` and supporting scripts can be in the root of the `functional_tests` directory. This is a fairly standard convention.

remerle (changes requested):
Let's keep the folder structure intact and just add the `Dockerfile` and `Makefile`.

makefile
```diff
@@ -0,0 +1,12 @@
+#Functional test docker image name.
```
remerle, line 1: This file should be named `Makefile` by convention. Note the capitalization.

modules/api/functional_test/functional_test/live_tests/zones/update_zone_test.py
```diff
@@ -765,7 +765,6 @@ def test_user_can_update_zone_to_another_admin_group(shared_zone_test_context):
             shared_zone_test_context.ok_vinyldns_client.delete_group(group['id'], status=(200, 404))
 
 
-@pytest.mark.serial
```
remerle, line 768: Why was this removed?

Jay07GIT: Removed by mistake. Kindly ignore the changes.

makefile
Jay07GIT: Sure i will update it.

docker/functest/Dockerfile
```diff
@@ -1,14 +1,14 @@
-FROM python:2.7.15-stretch
+FROM python:2.7-buster
 
 # Install dns utils so we can run dig
 RUN apt-get update && apt-get install dnsutils -y
 
 # The run script is what actually runs our func tests
 COPY run.sh /app/run.sh
-RUN chmod a+x /app/run.sh
+RUN chmod 777 /app/run.sh
```
remerle, line 8: I know this is just the container, but `777` permissions are generally not a good idea. For executable's, this should be, at a maximum, `755` (read-only for everyone but the owner). Or even `chmod u+x` to just ensure the owner has executable rights.

docker/functest/Dockerfile
```diff
@@ -1,14 +1,14 @@
-FROM python:2.7.15-stretch
```
remerle, line 1: Do we need to even change this file? For this PR, we're not really touching this. 

Ultimately, we'll be reworking/restructuring this whole `/docker` directory as we tighten up the deployment.

Jay07GIT: Yeh, I will revert back the Python base image to 2.7.15-stretch

Jay07GIT:
Updated changes based on above PR comments. 
-Restructured `Dockerfile `and `Makefile` to docker functional test folder.
-Renamed `Makefile`.
-Revert back `docker folder`  and `update_zone.py` file to the older one .

remerle:
Let's update this once the 3.x tests are in place.

PR #1051, opened 2021-09-24
Set configurations for enforcing unique email and messages displayed to the user

Aravindh-Raju:
Changes in this pull request:
- Created a configuration whether to enforce unique email for groups or not. It can be enabled or disabled from reference.conf in api module.
- Created a configuration for updating messages displayed to the user. It can be done from reference.conf in core module.
- Consolidated all the messages displayed to the user in Messages.scala file.
- Made the messages configurable.
- Error message will be displayed when we create a new group with email id that's already used by another group.
- Error message will be displayed when we update the group email id with a new one that's already used by another group.
- The error message would recommend the user to contact the existing group to be added to that group. This would reduce the confusion of having multiple groups with same email id.
- Created a new SQL query to get group by email, Created a function to display the error message, Added new test cases.
- Updated the FAQ of the microsite with details about the email configuration.


remerle (changes requested):
As discussed offline, there are some changes we want to make here. There are also some failing tests to consider.

Aravindh-Raju:
Closing the issue [#1047](https://github.com/vinyldns/vinyldns/pull/1047) since it was merged in this PR to make some additional changes.

Aravindh-Raju:
> As discussed offline, there are some changes we want to make here. There are also some failing tests to consider.

Made changes as discussed, rebased to latest master and added functional tests.

project/Dependencies.scala
```diff
@@ -74,6 +74,8 @@ object Dependencies {
     "co.fs2"                    %% "fs2-core"                       % fs2V,
     "javax.xml.bind"            %  "jaxb-api"                       % "2.3.0",
     "javax.activation"          %  "activation"                     % "1.1.1",
+    "com.github.pureconfig"     %% "pureconfig"                     % pureConfigV,
+    "com.github.pureconfig"     %% "pureconfig-cats-effect"         % pureConfigV,
```
remerle, line 78: Why are these dependencies being specified again? They are already on line 67-68

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlGroupRepository.scala
```diff
@@ -155,6 +162,20 @@ class MySqlGroupRepository extends GroupRepository with GroupProtobufConversions
...
     }
 
+  def getGroupByEmailId(groupEmailId: String): IO[Option[Group]] =
+    monitor("repo.Group.getGroupByEmailId") {
+      IO {
+        logger.info(s"Getting group with email: $groupEmailId")
+        DB.readOnly { implicit s =>
+          GET_GROUP_BY_EMAILID
+            .bind(groupEmailId)
+            .map(toGroup(1))
```
remerle, line 172: What does this `1` mean here in `map(toGroup(1))`?

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlGroupRepository.scala
```diff
@@ -155,6 +162,20 @@ class MySqlGroupRepository extends GroupRepository with GroupProtobufConversions
       }
     }
 
+  def getGroupByEmailId(groupEmailId: String): IO[Option[Group]] =
```
remerle, line 165: Is this really getting the group by email identifier? It seems to be getting the group by email **address**. I guess it could be argued that the email address is an identifier, but in VinylDNS, identifiers are usually UUIDs/GUIDs, so we may want to avoid confusion and go with `getGroupByEmailAddress(groupEmailAddress: String)`

modules/docs/src/main/mdoc/faq.md
```diff
@@ -88,3 +89,9 @@ If you use any VinylDNS tools beyond the portal you will need to provide those t
 
 ### 9. How are requests authenticated to the VinylDNS API? <a id="9"></a>
 Refer to [API Authentication](api/auth-mechanism.html).
+
+### 10. How to change email configuration for groups? <a id="10"></a>
+VinylDNS can be configured to disallow multiple groups with the same email address. We have found that
+supporting multiple groups with the same email address to be confusing to many users when there are ownership questions
+in the future. Therefore, we strongly recommend that VinylDNS deployments utilize this feature. If you wish to use this feature,
+you can set "vinyldns.groups.enforce-unique-email-id = true" in the configuration.
```
remerle, line 97: Make sure to indicate that this is the API configuration since we also have a configuration for the Portal. It may be worth linking to [the documentation for the configuration](https://www.vinyldns.io/operator/config-api.html).

Also, make sure you include this in the example configuration in the documentation provided by [config-api.md](https://github.com/vinyldns/vinyldns/blob/85577a27f0f0232c8e2aa541a59c8235a85023e7/modules/docs/src/main/mdoc/operator/config-api.md#full-example-config)

modules/core/src/main/resources/reference.conf
```diff
@@ -0,0 +1,18 @@
+
+# Configure Messages. Override existing messages with new messages. Refer Messages.scala file for existing messages
```
remerle, line 2: Please make sure you update the example config in the documentation at [config-portal.md](https://github.com/vinyldns/vinyldns/blob/85577a27f0f0232c8e2aa541a59c8235a85023e7/modules/docs/src/main/mdoc/operator/config-portal.md#full-example-config) and [config-api.md](https://github.com/vinyldns/vinyldns/blob/85577a27f0f0232c8e2aa541a59c8235a85023e7/modules/docs/src/main/mdoc/operator/config-api.md#full-example-config)

remerle (changes requested):

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlGroupRepository.scala
Aravindh-Raju: The 1 here means ***columnIndex***. It retrieves the value of the designated column in the current row of this WrappedResultSet object as a byte array. It's mentioned at the end of the file:
`  def toGroup(colIndex: Int): WrappedResultSet => Group = res => {
    fromPB(VinylDNSProto.Group.parseFrom(res.bytes(colIndex)))
  }`
  
It retrieves the data of the group present in ***columnIndex 1*** that we got as a result while getting group by email and then parses it and returns to us the data.

All the functions that "GET" data from the database uses ***columnIndex 1*** to retrieve the value present in column index 1 as byte array and parse the data. 
For Eg: getGroupByName, getGroups, getZones etc.

| column_index | data |
| ------------- | ------------- |
| 1  | BINARY, 4074 Bytes |

project/Dependencies.scala
Aravindh-Raju: I specified it by mistake, will remove them in the next commit along with other changes.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlGroupRepository.scala
Aravindh-Raju: Done the changes in the new commit.

modules/core/src/main/resources/reference.conf
Aravindh-Raju: I have updated the `enforce-unique-email` config details in [config-api.md](https://www.vinyldns.io/operator/config-api.html) and [faq.md](https://www.vinyldns.io/faq.html). But `messages` configuration is done in `core` module `reference.conf` as discussed before, which was newly created since all the messages can only be imported and used in `core` module and cannot be imported and configured in `api` and `portal` modules. So, where should I mention about `messages` configuration? Should I create a new `config.md` file (eg) config-core.md for `core` module `reference.conf` or just mention about `messages` configuration in [faq.md](https://www.vinyldns.io/faq.html)?

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlGroupRepository.scala
remerle: As a general rule, we want to avoid "magic numbers" (i.e., unnamed numerical constants). As such, we should define a constant such as:
```scala
val DataColumnIndex = 1
```

We can always make more specific constants if there's ambiguity between queries and the `data` column index.

modules/core/src/main/resources/reference.conf
remerle: Will the configuration be picked up if it is specified in the `API` and/or `Portal` configurations? If so, then the configuration examples should be added to both the Portal and the API documentation.

remerle (changes requested):

modules/core/src/main/resources/reference.conf
Aravindh-Raju: Configuration is now specified in the `API configurations` instead of `core module` and is being picked up. Configuration examples has been added to [config-api.md](https://github.com/vinyldns/vinyldns/blob/306b3273f4574c0b37622cbfb0bc4de7c1eafadc/modules/docs/src/main/mdoc/operator/config-api.md#configure-messages) and has been referred in [faq.md](https://github.com/vinyldns/vinyldns/blob/306b3273f4574c0b37622cbfb0bc4de7c1eafadc/modules/docs/src/main/mdoc/faq.md#11-how-to-configure-messages-displayed-to-the-users-). But it cannot be picked up in `Portal configurations`, Hence did not update there.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlGroupRepository.scala
Aravindh-Raju: Defined a constant `DataColumnIndex` as suggested in every files under the `mysql` module where these "magic numbers" are used and committed the changes.

remerle (approved):

Aravindh-Raju:
Closing this as a new PR has been opened: #1291.

PR #1062, opened 2021-12-02
Transaction atomicity added in the Recordset on the zonesync

Jay07GIT:
Changes in this pull request:
- Transaction atomicity added in the Recordset on the zonesync.


remerle (changes requested):
We need to do some verification this:

1. Verify that the current execution does **not** execute the all of the DB changes in a single transaction
2. Verify that with the change, all of the updates occur within the same transaction

All database interaction should be done within the `IO` monad, so we'll need to take that into consideration.

You will likely need to instrument the JDBC driver (using a proxy driver perhaps) in order to verify that the code is working as expected.

Jay07GIT:
This PR was continued in [PR-1076](https://github.com/vinyldns/vinyldns/pull/1076), hence closed

PR #1076, opened 2021-12-20
Create transaction to maintain atomicity

Aravindh-Raju:
Refer: GH-1062

Changes in this pull request:
- Created a transaction to maintain atomicity in `ZoneSyncHandler.scala` and `RecordSetChangeHandler.scala`. Now, either everything gets executed or none gets executed.
- Verified the transaction by running locally and also by adding tests.


modules/api/src/test/scala/vinyldns/api/engine/ZoneSyncHandlerSpec.scala
```diff
@@ -25,24 +25,148 @@ import org.scalatestplus.mockito.MockitoSugar
...
+                _ <- saveRecordSets
+              } yield zoneChange.copy(
+                zone.copy(status = ZoneStatus.Active, latestSync = Some(DateTime.now)),
+                status = ZoneChangeStatus.Synced
+              )
+            } catch {
+              // Rollback changes
+              case e: Exception =>
+                db.rollbackIfActive() // no exceptions can be thrown
+                throw new Exception(e.getMessage + " Rollback completed successfully!")
```
Aravindh-Raju, line 147: When there's an exception, if any changes was made to the database it gets into the catch block and rollbacks to the previous state. Also verified this by running locally.

modules/api/src/main/scala/vinyldns/api/engine/ZoneSyncHandler.scala
```diff
@@ -143,23 +139,38 @@ object ZoneSyncHandler extends DnsConversions with Monitored {
...
+              db.begin()
+              // we want to make sure we write to both the change repo and record set repo
+              // at the same time as this can take a while
+              val saveRecordChanges = time(s"zone.sync.saveChanges; zoneName='${zone.name}'")(
+                recordChangeRepository.save(changeSet)
+              )
+              val saveRecordSets = time(s"zone.sync.saveRecordSets; zoneName='${zone.name}'")(
+                recordSetRepository.apply(changeSet)
+              )
+              db.commit() // commit a transaction
```
Aravindh-Raju, line 156: Only commits the changes to database when both `saveRecordChanges` and `saveRecordSets` executed successfully, else in case of any exception changes gets rolled back. Thus either both gets executed or nothing gets executed.

modules/api/src/test/scala/vinyldns/api/engine/ZoneSyncHandlerSpec.scala
```diff
@@ -25,24 +25,148 @@ import org.scalatestplus.mockito.MockitoSugar
...
+  // Function to test transaction by raising exception. Clone of runSync function
+  def zoneTestRunSync(
+                       recordSetRepository: RecordSetRepository,
+                       recordChangeRepository: RecordChangeRepository,
+                       zoneChange: ZoneChange,
+                       backendResolver: BackendResolver,
+                       maxZoneSize: Int,
+                       vinyldnsLoader: (Zone, RecordSetRepository) => VinylDNSZoneViewLoader =
+                       VinylDNSZoneViewLoader.apply,
+                       exceptionNumber: Int = 0
```
Aravindh-Raju, line 63: Added a parameter to raise an exception and check the working of transaction.

modules/api/src/test/scala/vinyldns/api/engine/ZoneSyncHandlerSpec.scala
```diff
@@ -25,24 +25,148 @@ import org.scalatestplus.mockito.MockitoSugar
...
+            try {
+
+              // Transaction begins
+              db.begin()
+              // we want to make sure we write to both the change repo and record set repo
+              // at the same time as this can take a while
+              val saveRecordChanges = time(s"zone.sync.saveChanges; zoneName='${zone.name}'")(
+                recordChangeRepository.save(changeSet)
+              )
+              if(exceptionNumber == 1) throw new Exception("SaveRecordChanges failed!")
```
Aravindh-Raju, line 128: Used in a test below. Raise an exception after `saveRecordChanges` to check whether any changes made before exception gets rolled back.

modules/api/src/test/scala/vinyldns/api/engine/ZoneSyncHandlerSpec.scala
```diff
@@ -25,24 +25,148 @@ import org.scalatestplus.mockito.MockitoSugar
...
+              val saveRecordSets = time(s"zone.sync.saveRecordSets; zoneName='${zone.name}'")(
+                recordSetRepository.apply(changeSet)
+              )
+              if(exceptionNumber == 2) throw new Exception("SaveRecordSets failed!")
```
Aravindh-Raju, line 132: Used in a test below. Raise an exception after `saveRecordSets` to check whether any changes made before exception gets rolled back.

modules/api/src/test/scala/vinyldns/api/engine/ZoneSyncHandlerSpec.scala
```diff
@@ -25,24 +25,148 @@ import org.scalatestplus.mockito.MockitoSugar
...
+        case Left(e: Throwable) =>
+          logger.error(
+            s"Encountered error syncing ; zoneName='${zoneChange.zone.name}'; zoneChange='${zoneChange.id}'",
+            e
+          )
+          // We want to just move back to an active status, do not update latest sync
+          zoneChange.copy(
+            zone = zoneChange.zone.copy(status = ZoneStatus.Active),
+            status = ZoneChangeStatus.Failed,
+            systemMessage = Some(e.getMessage) // check the exception occurred
```
Aravindh-Raju, line 165: Get the error message to check whether it gets into the catch block and rolls back any changes made. Used in the test below.

modules/api/src/test/scala/vinyldns/api/engine/ZoneSyncHandlerSpec.scala
```diff
@@ -495,5 +621,46 @@ class ZoneSyncHandlerSpec
...
+        testZoneChange,
+        mockBackendResolver,
+        10000,
+        (_, _) => mockVinylDNSLoader,
+        1
+      ).unsafeRunSync() // set exceptionNumber = 1 to raise an exception when saving record changes to recordChangeRepo
+
+      // Does not save changes to both recordChangeRepo and recordSetRepo as it rollbacks changes made when exception occurred
+      // Transaction saves either both record changes and record sets or saves none to database
+      result.systemMessage.get shouldBe "SaveRecordChanges failed! Rollback completed successfully!"
```
Aravindh-Raju, line 639: If there's an exception and it goes to the catch block and the rollback was successful, this should pass. If it doesn't rollback this won't pass.

modules/api/src/test/scala/vinyldns/api/engine/ZoneSyncHandlerSpec.scala
```diff
@@ -495,5 +621,46 @@ class ZoneSyncHandlerSpec
...
+        testZoneChange,
+        mockBackendResolver,
+        10000,
+        (_, _) => mockVinylDNSLoader,
+        2
+      ).unsafeRunSync() // set exceptionNumber = 2 to raise an exception when saving record sets to recordSetRepo
+
+      // Does not save changes to both recordChangeRepo and recordSetRepo as it rollbacks changes made when exception occurred
+      // Transaction saves either both record changes and record sets or saves none to database
+      result.systemMessage.get shouldBe "SaveRecordSets failed! Rollback completed successfully!"
```
Aravindh-Raju, line 660: If there's an exception and it goes to the catch block and the rollback was successful, this should pass. If it doesn't rollback this won't pass.

Aravindh-Raju:
Reference using which the changes were made:
- [Transaction management using an automatic session](https://github.com/scalikejdbc/scalikejdbc-cookbook/blob/master/en/04_transaction.md#transaction-management-using-an-automatic-session)
- [#withinTx block / session](http://scalikejdbc.org/documentation/transaction.html)
- [AutoSession to use along withinTX](http://scalikejdbc.org/documentation/auto-session.html)

Removed `localTx` transaction from `MySqlRecordChangeRepository.scala` and `MySqlRecordSetRepository.scala` to wrap them into a `withinTx` transaction in `ZoneSyncHandler.scala` by following the approach [Transaction management using an automatic session](https://github.com/scalikejdbc/scalikejdbc-cookbook/blob/master/en/04_transaction.md#transaction-management-using-an-automatic-session). Now atomicity is maintained and we can also execute them individually as we use [AutoSession](http://scalikejdbc.org/documentation/auto-session.html)

Aravindh-Raju:
Will provide details regarding how to reproduce the error, screenshots of database changes before and after implementation of transaction along with other places where non-atomic transaction takes place tomorrow.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordChangeRepository.scala
```diff
@@ -63,27 +63,25 @@ class MySqlRecordChangeRepository
...
-              changes.map { change =>
-                Seq(
-                  change.id,
-                  change.zoneId,
-                  change.created.getMillis,
-                  fromChangeType(change.changeType),
-                  toPB(change).toByteArray
-                )
-              }
+        changeSet.changes
```
remerle, line 69: Are we able to keep the local transactions? Nested transactions should work.

If this method is ever called outside of a transaction then we risk losing atomicity of the batch insert.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
```diff
@@ -154,18 +154,16 @@ class MySqlRecordSetRepository extends RecordSetRepository with Monitored {
       val deletes: Seq[Seq[Any]] = completeDeletes.map(d => Seq[Any](d.recordSet.id))
 
       IO {
-        DB.localTx { implicit s =>
```
remerle, line 157: Same here.. can we keep this `localTx`?

remerle (changes requested):
Just a couple of notes

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordChangeRepository.scala
Aravindh-Raju: I will look into using `localTx` nested transaction further. But as I've seen about using nested transactions in `scalikejdbc` there has not been much results related. There's not been enough details about nested transactions using `scalikejdbc` in their official documentation and also in other places. 

If we call this method outside of the transaction created in `zonesynchandler.scala`, it'll create a `readOnly` or `autoCommit` transaction automatically as given [here](https://github.com/scalikejdbc/scalikejdbc-cookbook/blob/master/en/04_transaction.md#transaction-management-using-an-automatic-session). But since it doesn't have rollback like `localTx`, we risk losing atomicity as you said.

Will look further into it and report back here.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
Aravindh-Raju: Updated the PR with some changes. Now, even if the method is called outside of the transaction it’s atomicity is maintained. The code is redundant, but it has a condition such that it identifies if the method is called outside of the transaction and wraps it with localTx. Else, it uses the transaction mentioned outside. Thus maintaining atomicity in both case.

remerle: Is it possible to instead have a bit more control over the transaction by controlling the connection itself?

From the caller:

```scala
using(DB(connectionPool.borrow())) { db: DB =>
  // keep the connection open while in this `using` block
  db.autoClose(false)

  db.begin()
  try {
    val saveRecordChanges = time(s"zone.sync.saveChanges; zoneName='${zone.name}'")(
      recordChangeRepository.save(changeSet)
    )
    val saveRecordSets = time(s"zone.sync.saveRecordSets; zoneName='${zone.name}'")(
      recordSetRepository.apply(changeSet)
    )

    // join together the results of saving both the record changes as well as the record sets
    for {
      _ <- saveRecordChanges
      _ <- saveRecordSets
    } yield zoneChange.copy(
      zone.copy(status = ZoneStatus.Active, latestSync = Some(DateTime.now)),
      status = ZoneChangeStatus.Synced
    )
  } catch { case e =>
    db.rollbackIfActive()
    throw e
  }
}
```

Each called method then can use `withinTx` which will fail with an `IllegalStateException` if there is no active transaction.
```scala
def save(db: DB, changeSet: ChangeSet) {
  db withinTx { implicit session =>
    // Do DB work
    changeSet.changes...
    .foreach { group => INSERT...apply() }
  }
}
```

By default, `DB.localTx` borrows a connection from the pool, then "closes" it when the execution is complete. By controlling the connection, we should be able to more cleanly control a transaction that spans repository actions.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordChangeRepository.scala
remerle: See my latest comment

remerle:
> AutoSession starts a read-only transaction session for a select statement, and an auto-commit transaction session for update statements when executing individually, and inside a `withinTx` transaction block when executing together. Thus we can now atomically make changes to the database both while executing together and also while executing them individually.

@Aravindh-Raju We want to avoid auto-commit. It can lead to unintended consequences since each statement is committed after it's executed. We want all of the changes made first, then committed **all at once**.

For example, if we have two `UPDATE` statements with auto-commit enabled, then when the first statement executes, it will be committed, then the second could execute and fail. Since the first statement is committed, those changes cannot be rolled back. For more reading, you can check out [this explanation on SO](https://stackoverflow.com/a/2950764/62195).

Take a look at my suggestion here: https://github.com/vinyldns/vinyldns/pull/1076#discussion_r782365440. I don't believe we can avoid introducing lower-level DB calls as much as we'd like to keep them confined to the `Repository` types. We need to prioritize data integrity over design purity. At some point, perhaps we can move to a DB library that has better abstractions

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
Aravindh-Raju: Thanks Ryan, I'll proceed with this and get back once completed!

Aravindh-Raju:
> > AutoSession starts a read-only transaction session for a select statement, and an auto-commit transaction session for update statements when executing individually, and inside a `withinTx` transaction block when executing together. Thus we can now atomically make changes to the database both while executing together and also while executing them individually.
> 
> @Aravindh-Raju We want to avoid auto-commit. It can lead to unintended consequences since each statement is committed after it's executed. We want all of the changes made first, then committed **all at once**.
> 
> For example, if we have two `UPDATE` statements with auto-commit enabled, then when the first statement executes, it will be committed, then the second could execute and fail. Since the first statement is committed, those changes cannot be rolled back. For more reading, you can check out [this explanation on SO](https://stackoverflow.com/a/2950764/62195).
> 
> Take a look at my suggestion here: [#1076 (comment)](https://github.com/vinyldns/vinyldns/pull/1076#discussion_r782365440). I don't believe we can avoid introducing lower-level DB calls as much as we'd like to keep them confined to the `Repository` types. We need to prioritize data integrity over design purity. At some point, perhaps we can move to a DB library that has better abstractions

Yes, I'll avoid using `auto-commit`. Didn't thought of using it, but could not find any other solution that works without `AutoSession`, therefore have to proceed with that. I'll follow your suggestion given [here](https://github.com/vinyldns/vinyldns/pull/1076#discussion_r782365440) and complete the changes.

modules/api/src/main/scala/vinyldns/api/engine/RecordSetChangeHandler.scala
```diff
@@ -50,31 +52,45 @@ object RecordSetChangeHandler {
...
-      singleChangeStatusUpdates = updateBatchStatuses(singleBatchChanges, completedState.change)
-      _ <- batchChangeRepository.updateSingleChanges(singleChangeStatusUpdates)
-    } yield completedState.change
+     recordSetRepository: RecordSetRepository,
+     recordChangeRepository: RecordChangeRepository,
+     batchChangeRepository: BatchChangeRepository,
+     conn: Backend,
+     recordSetChange: RecordSetChange
+   )(implicit timer: Timer[IO]): IO[RecordSetChange] =
+    executeWithinTransaction { db: DB => {
```
remerle, line 63: This is a lot of logic to exist within a DB transaction. We should limit the transaction scope to the fewest steps possible to ensure atomicity.

modules/api/src/main/scala/vinyldns/api/engine/RecordSetChangeHandler.scala
```diff
@@ -50,31 +52,45 @@ object RecordSetChangeHandler {
...
+        completedState <- fsm(
+          Pending(recordSetChange),
+          conn,
+          wildCardExists,
+          recordSetRepository,
+          recordChangeRepository
+        )
+        changeSet = ChangeSet(completedState.change).complete(completedState.change)
+        _ <- recordSetRepository.apply(db, changeSet)
+        _ <- recordChangeRepository.save(db, changeSet)
```
remerle, line 76: You don't need to leave these as separate steps in the `for` comprehension (that way you can limit the scope of the transaction.

modules/api/src/main/scala/vinyldns/api/engine/RecordSetChangeHandler.scala
```diff
@@ -50,31 +52,45 @@ object RecordSetChangeHandler {
...
+      } yield completedState.change
+    }.attempt.map {
+      case Left(e: Throwable) =>
+        db.rollbackIfActive() //Roll back the changes if error occurs
+        db.close() //Close DB Connection
+        throw e
+      case Right(ok) =>
+        db.commit() //Commit the changes
+        db.close() //Close DB Connection
+        ok
```
remerle, line 91: We want to avoid needing to handle this for every database transaction. This should all be rolled up into the `executeWithinTransaction` method

modules/api/src/main/scala/vinyldns/api/engine/ZoneSyncHandler.scala
```diff
@@ -82,6 +78,18 @@ object ZoneSyncHandler extends DnsConversions with Monitored {
         zoneChangeRepository.save(zoneChange)
     }
 
+  def executeWithinTransaction[A](execution: DB => A): A = {
```
remerle, line 81: This should move to its own `trait`

remerle (changes requested):
This is looking good. I just think we need to reduce the code duplication.

Take a look at [the POC I put together](https://github.com/vinyldns/vinyldns/compare/master...remerle:db-tx). It allows better encapsulation of the database transaction without leaking the details of `commit` and `rollback` to the caller.

modules/api/src/main/scala/vinyldns/api/engine/RecordSetChangeHandler.scala
Aravindh-Raju: Completed the changes in the latest commit.

modules/api/src/main/scala/vinyldns/api/engine/RecordSetChangeHandler.scala
Aravindh-Raju: Followed the POC and made the changes.

modules/api/src/main/scala/vinyldns/api/engine/RecordSetChangeHandler.scala
Aravindh-Raju: Followed the POC and made the changes accordingly.

modules/api/src/main/scala/vinyldns/api/engine/RecordSetChangeHandler.scala
```diff
@@ -67,15 +69,26 @@ object RecordSetChangeHandler {
...
       )
       changeSet = ChangeSet(completedState.change).complete(completedState.change)
-      _ <- recordSetRepository.apply(changeSet)
-      _ <- recordChangeRepository.save(changeSet)
+      _ <- saveChangeSet(recordSetRepository, recordChangeRepository, changeSet)
       singleBatchChanges <- batchChangeRepository.getSingleChanges(
         recordSetChange.singleBatchChangeIds
       )
       singleChangeStatusUpdates = updateBatchStatuses(singleBatchChanges, completedState.change)
       _ <- batchChangeRepository.updateSingleChanges(singleChangeStatusUpdates)
```
remerle, line 77: Should all of these changes happen in the same transaction? 🤔

remerle (commented):
This looks good! Thanks @Aravindh-Raju 

I think we should just include the changes to the `batch_change`/`single_change` tables as noted in my one comment below.

modules/api/src/main/scala/vinyldns/api/engine/RecordSetChangeHandler.scala
Aravindh-Raju: Yes, these should happen in same transaction. Because when there’s an exception while updating the `single-change`, recordset and recordset-change will already be `saved to database` and displayed in the respective zone as `completed`, but the batch status will be in `Pending Processing` and SQS keeps retrying to update the single change operation that failed. Therefore in DNS Changes page of our portal, the batch status will be shown as `Pending Processing`. But when we go to the respective zone the records belongs to, it will be displayed as completed or when we search in RecordSet Search the records will be displayed and they’re also stored in DB. This shouldn't happen.

If we keep them all in a single transaction, and update `single-change` fails, everything will be rollbacked and won’t be displayed anywhere until everything (save recordset, recordset-change, update single-change) is executed without any error. When error occurs nothing will be saved and SQS retries with `Pending Processing` until everything is executed without errors.

I’ve made the changes and updated the PR with everything wrapped in the same transaction to ensure proper working and maintaining atomicity. Checked the working in quickstart and it works as expected.

remerle (approved):
I think this puts us in much better shape with respect to atomicity. 👍

PR #1081, opened 2022-01-21
Maintain atomicity in membership service

Aravindh-Raju:
Changes in this pull request:

- Created a transaction to maintain atomicity in `MembershipService.scala` where group, groupChange, membership data are changed without a tranaction. Now, either everything gets executed or none gets executed.
- Verified the transaction by running locally.


remerle (approved):
LGTM 👍

PR #1087, opened 2022-02-10
Resolve deadlock

Aravindh-Raju:
Changes in this pull request:
- Remove `ON DUPLICATE KEY UPDATE` statement and add a `UPDATE` statement which may be a possible solution to resolve deadlock.

### DIFF
--- a/modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlMembershipRepository.scala
+++ b/modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlMembershipRepository.scala
@@ -23,14 +23,23 @@ import scalikejdbc._
 import vinyldns.core.domain.membership.MembershipRepository
 import vinyldns.core.route.Monitored
 
+import java.sql.SQLException
+
 class MySqlMembershipRepository extends MembershipRepository with Monitored {
   private final val logger = LoggerFactory.getLogger(classOf[MySqlMembershipRepository])
 
   private final val SAVE_MEMBERS =
     sql"""
-        | INSERT INTO membership (user_id, group_id, is_admin)
-        |      VALUES ({userId}, {groupId}, {isAdmin})
-        | ON DUPLICATE KEY UPDATE is_admin = {isAdmin}
+         | INSERT INTO membership (user_id, group_id, is_admin)
+         |      VALUES ({userId}, {groupId}, {isAdmin})
+       """.stripMargin
+
+  // Replace "ON DUPLICATE KEY UPDATE" which is used before to prevent possible deadlock
+  private final val UPDATE_MEMBERS =
+    sql"""
+         | UPDATE membership
+         |      SET is_admin = {isAdmin}
+         |      WHERE user_id = {userId} AND group_id = {groupId}
        """.stripMargin
 
   private final val BASE_GET_USERS_FOR_GROUP =
@@ -66,7 +75,17 @@ class MySqlMembershipRepository extends MembershipRepository with Monitored {
           IO {
             logger.info(s"Saving into group $groupId members $nonEmpty")
             db.withinTx { implicit s =>
-              SAVE_MEMBERS.batchByName(saveParams(nonEmpty, groupId, isAdmin): _*).apply()
+              try {
+                SAVE_MEMBERS.batchByName(saveParams(nonEmpty, groupId, isAdmin): _*).apply()
+              }
+              catch {
+                case ex: SQLException =>
+                  // Check for duplicate key exception and update the members if we get that exception
+                  // 1062 is Error Code for Duplicate key entry
+                  if(ex.getErrorCode == 1062){
+                    UPDATE_MEMBERS.batchByName(saveParams(nonEmpty, groupId, isAdmin): _*).apply()
+                  }
+              }
               memberUserIds
             }
           }


modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlMembershipRepository.scala
```diff
@@ -66,7 +75,17 @@ class MySqlMembershipRepository extends MembershipRepository with Monitored {
...
+                SAVE_MEMBERS.batchByName(saveParams(nonEmpty, groupId, isAdmin): _*).apply()
+              }
+              catch {
+                case ex: SQLException =>
+                  // Check for duplicate key exception and update the members if we get that exception
+                  // 1062 is Error Code for Duplicate key entry
+                  if(ex.getErrorCode == 1062){
+                    UPDATE_MEMBERS.batchByName(saveParams(nonEmpty, groupId, isAdmin): _*).apply()
+                  }
+              }
```
remerle, line 88: This does not provide the same functionality as the `ON DUPLICATE KEY` directive. In this case, it seems that if **any** of the users already exist, then the entire `INSERT` operation will fail and fallback to `UPDATE`. Where the original would only fall back to `UPDATE` on a per-record basis.

Instead, we should pull all known `memberUserIds` from the DB with a `SELECT` before this save operation into a value `existingUsers: Set[String]`.  Then `INSERT` any users that **do not** intersect with the `existingUsers` set and `UPDATE` all the users that **do** intersect with the `existingUsers` set.

modules/mysql/src/main/scala/vinyldns/mysql/TransactionProvider.scala
```diff
@@ -36,25 +37,43 @@ trait TransactionProvider {
...
+
+  def retry[T](times: Int)(fn: => T): T = {
+    try {
+      fn
+    } catch {
+      // Only retry transaction that has deadlock exception. Don't retry on any other exception.
+      case e: SQLException =>
+        // Error code 1213 or 1205 indicates deadlock and we must retry the transaction.
+        if (e.getErrorCode == 1213 || e.getErrorCode == 1205) {
+          logger.error(s"Encountered error executing function within a database transaction. Retrying again. Retry Count: $times")
```
remerle, line 72: This should be logged at level `WARN` since it's attempting to self-heal with a retry. It's only an error if the retries are exhausted.

modules/mysql/src/main/scala/vinyldns/mysql/TransactionProvider.scala
```diff
@@ -36,25 +37,43 @@ trait TransactionProvider {
...
-        db.commit()
-        result
-      } catch {
-        case e: Throwable =>
-          logger.error(s"Encountered error executing function within a database transaction ($txId). Rolling back transaction.", e)
-          db.rollbackIfActive()
-          throw e
-      } finally {
-        db.close()
+      retry(3) {
```
remerle, line 40: We should pull the `3` here out into a constant. It doesn't need to be configurable, just given a name like `DEADLOCK_MAX_RETRIES`

modules/mysql/src/main/scala/vinyldns/mysql/TransactionProvider.scala
```diff
@@ -36,25 +37,43 @@ trait TransactionProvider {
...
   }
+
+  def retry[T](times: Int)(fn: => T): T = {
+    try {
+      fn
+    } catch {
+      // Only retry transaction that has deadlock exception. Don't retry on any other exception.
+      case e: SQLException =>
+        // Error code 1213 or 1205 indicates deadlock and we must retry the transaction.
+        if (e.getErrorCode == 1213 || e.getErrorCode == 1205) {
```
remerle, line 71: It would be nice to pull this out into a function called `isDeadlock(sqlException: SQLException)` so that we avoid vendor-specific checks spreading throughout the code. This is just a defensive coding opportunity. That is, if we assume that all code has some probability of being copied-and-pasted, then we want to make sure the encapsulated behavior is what is copied-and-pasted, not the hard-coded vendor-specific error codes.

remerle (changes requested):

remerle (changes requested):
I forgot to leave a comment - just a few observations in comments.

From a high-level, we need to get some test coverage on the retry mechanism and ensure that it works as expected (inserting all records unless they exist, in which case they are updated).

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlMembershipRepository.scala
Aravindh-Raju: Updated the changes in latest commit.

modules/mysql/src/main/scala/vinyldns/mysql/TransactionProvider.scala
Aravindh-Raju: Updated in the latest commit.

modules/mysql/src/main/scala/vinyldns/mysql/TransactionProvider.scala
Aravindh-Raju: Updated in the latest commit.

modules/mysql/src/main/scala/vinyldns/mysql/TransactionProvider.scala
Aravindh-Raju: Updated in the latest commit.

remerle (approved):
Yay! 🎉

PR #1089, opened 2022-02-24
ACLs Should Honor Most Permissive Access Level

### LINKED ISSUE #1083: ACLs do not honor most permissive access level
## Background
Given an ACL on zone as follows:

| | |
|--|--|
|Zone|`vinyldns.io`|
|User|`leela`|
|Access Level|`read`|
|Record Mask | n/a |

If user `leela` attempts to create a record `foo` in `vinyldns.io` they will be met with a `400` error and the message
```
"User leela does not have access to create foo.vinyldns.io."
```
This is by design.

## Issue
Assume, now, that we have more than one ACL entry for user `leela`:

| | |
|--|--|
|Zone|`vinyldns.io`|
|User|`leela`|
|Access Level|`read`|
|Record Mask | n/a |

| | |
|--|--|
|Zone|`vinyldns.io`|
|User|`leela`|
|Access Level|`delete`|
|Record Mask | n/a |

Because we have a `delete` entry, user `leela` should be able to create, modify **or** delete records in the `vinyldns.io` zone. However, due to the presence of the `read` ACL entry, the user will receive the same message:
```
"User leela does not have access to create foo.vinyldns.io."
```

This is due to the code here:
https://github.com/vinyldns/vinyldns/blob/e4264fb34cc965fb7bcddaf0148472c4bf890ab0/modules/api/src/main/scala/vinyldns/api/domain/access/AccessValidations.scala#L199-L207

Specifically, the access levels are sorted and the `minimum` _(line 205)_ level is returned. Instead, the expectation is that the **most permissive** level will be returned.

The VinylDNS UI does not make it clear that each subsequent access level includes the previous level, so it's not uncommon for users to include entries for `Read`, `Write`, **and** `Delete` when only `Delete` is needed.

We need to reverse the order and take the most permissive access level and update the documentation to indicate the same. We also need to account for the `No Access` level, so please see the note about `No Access` below as part of this change.

### Important Note about `No Access`
The [public documentation](https://www.vinyldns.io/permissions.html#:~:text=accessLevel%20%2D%20what%20kinds,as%20an%20override.) does not state how permission precedence works, so there should be no expectation of behavior. 

For the different access levels, we can quantify the permissions by their type - are they allowing access, or denying access:

| Access Level | Type |
| -- | -- |
| `Read` | Allow |
| `Write` | Allow |
| `Delete` | Allow |
| `No Access` | Deny |

From a precedence point of view, it's generally the expectation that `deny` rules take precedence of `allow` rules. So, irrespective of what `allow` rules are in place, if there is a matching `deny` rule, it should take precedence.

nspadaccino:
Currently, if multiple ACL entries are associated with a user, the LEAST permissive is returned. For example, if a user has an entry with Write permission, and another with Read, that user will only be able to read.

This functionality has been changed to now return the MOST permissive rule. In the previous example, the user would be able to both read and write, because Write takes precedence by nature of being more permissive.

The NoAccess permission is treated as the highest priority, and takes precedence over other rules.

Fixes #1083

### DIFF
--- a/modules/api/src/test/functional/tests/recordsets/update_recordset_test.py
+++ b/modules/api/src/test/functional/tests/recordsets/update_recordset_test.py
@@ -821,9 +821,9 @@ def test_user_rule_priority_over_group_acl_rule(shared_zone_test_context):
 
 
 @pytest.mark.serial
-def test_more_restrictive_acl_rule_priority(shared_zone_test_context):
+def test_more_permissive_acl_rule_priority(shared_zone_test_context):
     """
-    Test more restrictive rule takes priority
+    Test more permissive rule takes priority
     """
     ok_zone = shared_zone_test_context.ok_zone
     client = shared_zone_test_context.ok_vinyldns_client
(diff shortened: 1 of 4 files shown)


remerle (approved):
LGTM 👍

Issue #1088, opened 2022-02-16
The ability to see members of other groups


David-packer:
The reason for this request is that some emails address associated with groups are no longer valid. We need a way to find other members of that group who can make this change. It is a waste of resources if we have to go to the support team each time to find out member of a group to get help for something we should be able to lookup for our selves. In my opinion, for what that is worth. :)

I had gone through the list of issues and did not see another request of this type.

remerle:
## Implementation Suggestion

Recommend that we allow all users to navigate to the group detail screen from `Groups` -> `All Groups` tab. For non-members, they will see a read-only view of the group and its membership. This should include:

* Group name
* Group email
* Member names
* Member email addresses
* Member group manager status

#### Mockup

![image](https://user-images.githubusercontent.com/1167333/155764198-545f50dd-6809-4054-9194-f41a1b8d5929.png)

## Notes
This will require allowing all authenticated users to access `GET /groups/{groupId}/members` which is currently limited to members of the group.

There is no sensitive information returned by this `GET` operation, so this should be fine. We must make sure that all write operations are still prevented. This MUST be read-only.

PR #1097, opened 2022-03-11
Allow Users to View Members of Other Groups

nspadaccino:
All logged in users can now view the group details of any group, not just groups they are part of. This allows users to see the members and admins of other groups without reaching out to support. This is read-only, users still cannot modify groups they are not part of.

Resolves https://github.com/vinyldns/vinyldns/issues/1088

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipValidations.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipValidations.scala
@@ -40,6 +40,6 @@ object MembershipValidations {
 
   def canSeeGroup(groupId: String, authPrincipal: AuthPrincipal): Either[Throwable, Unit] =
     ensuring(NotAuthorizedError("Not authorized")) {
-      authPrincipal.isGroupMember(groupId) || authPrincipal.isSystemAdmin
+      authPrincipal.isGroupMember(groupId) || authPrincipal.isSystemAdmin || true
     }
 }
--- a/modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipValidations.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipValidations.scala
@@ -23,6 +23,8 @@ import vinyldns.core.domain.membership.Group
 
 object MembershipValidations {
 
+  private val canViewGroupDetails = true
+
   def hasMembersAndAdmins(group: Group): Either[Throwable, Unit] =
     ensuring(InvalidGroupError("Group must have at least one member and one admin")) {
       group.memberIds.nonEmpty && group.adminUserIds.nonEmpty
@@ -40,6 +42,6 @@ object MembershipValidations {
 
   def canSeeGroup(groupId: String, authPrincipal: AuthPrincipal): Either[Throwable, Unit] =
     ensuring(NotAuthorizedError("Not authorized")) {
-      authPrincipal.isGroupMember(groupId) || authPrincipal.isSystemAdmin || true
+      authPrincipal.isGroupMember(groupId) || authPrincipal.isSystemAdmin || canViewGroupDetails
     }
 }
--- a/modules/api/src/test/scala/vinyldns/api/domain/membership/MembershipServiceSpec.scala
+++ b/modules/api/src/test/scala/vinyldns/api/domain/membership/MembershipServiceSpec.scala
@@ -548,11 +548,13 @@ class MembershipServiceSpec
         val result: Group = rightResultOf(underTest.getGroup(okGroup.id, okAuth).value)
         result shouldBe okGroup
       }
-
+      //TODO
       "return an error if not authorized" in {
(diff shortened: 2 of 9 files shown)


modules/api/src/test/scala/vinyldns/api/domain/membership/MembershipServiceSpec.scala
```diff
@@ -548,11 +548,13 @@ class MembershipServiceSpec
...
+      //TODO
       "return an error if not authorized" in {
         doReturn(IO.pure(Some(okGroup))).when(mockGroupRepo).getGroup(anyString)
-        val error = leftResultOf(underTest.getGroup(okGroup.id, dummyAuth).value)
-        error shouldBe a[NotAuthorizedError]
+//        val error = leftResultOf(underTest.getGroup(okGroup.id, dummyAuth).value)
+//        error shouldBe a[NotAuthorizedError]
+        val result: Group = rightResultOf(underTest.getGroup(okGroup.id, dummyAuth).value)
+        result shouldBe okGroup
       }
```
Aravindh-Raju, line 558: Since it has `TODO`, Is this test completed? If so, can we remove the comments?

modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipValidations.scala
```diff
@@ -40,6 +40,6 @@ object MembershipValidations {
 
   def canSeeGroup(groupId: String, authPrincipal: AuthPrincipal): Either[Throwable, Unit] =
     ensuring(NotAuthorizedError("Not authorized")) {
-      authPrincipal.isGroupMember(groupId) || authPrincipal.isSystemAdmin
+      authPrincipal.isGroupMember(groupId) || authPrincipal.isSystemAdmin || true
```
Aravindh-Raju, line 43: Everything looks fine and works as expected locally. 
Just a thought, wouldn’t it be good to have a variable something like `private val viewGroupDetails = true` which we can use here as:
`authPrincipal.isGroupMember(groupId) || authPrincipal.isSystemAdmin || viewGroupDetails` instead of true. So that we can know what’s the `true` is for when we revisit the code later.

nspadaccino: Good idea, I'll make that change

modules/api/src/test/scala/vinyldns/api/domain/membership/MembershipServiceSpec.scala
```diff
@@ -548,11 +548,10 @@ class MembershipServiceSpec
         val result: Group = rightResultOf(underTest.getGroup(okGroup.id, okAuth).value)
         result shouldBe okGroup
       }
-
       "return an error if not authorized" in {
         doReturn(IO.pure(Some(okGroup))).when(mockGroupRepo).getGroup(anyString)
-        val error = leftResultOf(underTest.getGroup(okGroup.id, dummyAuth).value)
-        error shouldBe a[NotAuthorizedError]
+        val result: Group = rightResultOf(underTest.getGroup(okGroup.id, dummyAuth).value)
+        result shouldBe okGroup
```
Aravindh-Raju, line 554: I think we don’t get `Not Authorized` error in `Groups` anymore since we are enabling everyone to view the group details in `MembershipValidations.scala`. Let's confirm with @remerle whether to keep this or to remove this.

modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipValidations.scala
```diff
@@ -40,6 +42,6 @@ object MembershipValidations {
 
   def canSeeGroup(groupId: String, authPrincipal: AuthPrincipal): Either[Throwable, Unit] =
     ensuring(NotAuthorizedError("Not authorized")) {
-      authPrincipal.isGroupMember(groupId) || authPrincipal.isSystemAdmin
+      authPrincipal.isGroupMember(groupId) || authPrincipal.isSystemAdmin || canViewGroupDetails
```
Aravindh-Raju, line 45: Same as the other comment I gave. Since now we are enabling everyone to view `Group Details`, this (`authPrincipal.isGroupMember(groupId) || authPrincipal.isSystemAdmin || canViewGroupDetails`) will always be `true` and we won't get `NotAuthorizedError` here anymore until we change `canViewGroupDetails` to `false` (which we won't be doing since we want everyone to see the group details). Let’s confirm with Ryan whether to keep it as it is or make any changes.

modules/api/src/test/scala/vinyldns/api/domain/membership/MembershipServiceSpec.scala
remerle: @nspadaccino Yeah, let's go ahead and remove this test since we are not actually verifying that we will `return an error if not authorized`.

remerle (changes requested):
@nspadaccino minor change

remerle (approved):
LGTM 👍

PR #1100, opened 2022-03-22
Updated wildcard filter for global recordset search with reverseFQDN 

Jay07GIT:
### Changes in this pull request:
- Updated wildcard filter for global recordset search with `reverseFQDN` using `recordset_data` table

[Issues 15](https://github.comcast.com/VinylDNS/vinyldns-team/issues/15)


modules/mysql/src/main/resources/db/migration/V3.24__RecordSetData.sql
```diff
@@ -5,19 +5,23 @@ USE ${dbName};
...
+  INDEX recordset_dataowner_group_id_index(owner_group_id),
   INDEX recordset_data_recordset_id_index(recordset_id),
   INDEX recordset_data_reverse_fqdn_index(reverse_fqdn),
   FULLTEXT INDEX recordset_data_fdqn_fulltext_index(fqdn),
-  FULLTEXT INDEX recordset_data_record_data_fulltex_index(record_data)
-);
-
+  FULLTEXT INDEX recordset_data_reverse_fqdn_fulltext_index(reverse_fqdn),
+  FULLTEXT INDEX recordset_data_record_data_fulltext_index(record_data)
+);
```
remerle, line 27: With flyway, you cannot modify existing, applied, schemas. If you want to change this table, it will have to be as a new flyway changeset.

Something like this:
```sql
ALTER TABLE recordset_data 
ADD COLUMN zone_name VARCHAR(256) AFTER zone_id;

ALTER TABLE recordset_data 
ADD COLUMN owner_group_id CHAR(36) AFTER ip;
```

This also leads me to wonder why we need to store the  zone and owner group information? We should be able to perform a `LEFT JOIN` on the `zone`

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetDataRepository.scala
```diff
@@ -57,13 +58,21 @@ class MySqlRecordSetDataRepository
       | WHERE fqdn
     """.stripMargin
 
-  private val FIND_BY_ZONEID_TYPE =
+  private val FIND_BY_ZONEID_NAME_TYPE =
     sql"""
          |SELECT data, fqdn
          |  FROM recordset_data
-         | WHERE zone_id = {zoneId} AND type = {type}
+         | WHERE zone_id = {zone_id} AND zone_name = {zone_name} AND type = {type}
```
remerle, line 65: Why do we need the `zone_name` if we have the `zone_id`?

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetDataRepository.scala
```diff
@@ -57,13 +58,21 @@ class MySqlRecordSetDataRepository
...
     """.stripMargin
 
+  private val FIND_BY_ZONEID_NAME =
+    sql"""
+         |SELECT data, fqdn
+         |  FROM recordset_data
+         | WHERE zone_id = {zone_id} AND zone_name = {zone_name}
```
remerle, line 72: Same as the previous comment. I'm not sure why the `zone_name` is necessary since the `zone_id` should uniquely identify the zone.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetDataRepository.scala
```diff
@@ -78,6 +87,14 @@ class MySqlRecordSetDataRepository
          | WHERE ip = {ip}
     """.stripMargin
 
+  private val GET_RECORDSETDATA_BY_OWNERID =
+    sql"""
+         |SELECT recordset_id
+         |  FROM recordset_data
+         |WHERE owner_group_id = {ownerGroupId}
```
remerle, line 94: This could be done with a `JOIN`

remerle (changes requested):
See my comments.

Additionally, for any methods added or updated, please include a method-level comment. We should use the javadoc standard.  There's a lack of comments throughout the code, and we should look to incrementally improve that on every commit. You can read javadocs in the Scala style guide [here](https://docs.scala-lang.org/style/scaladoc.html#objects).

The comments should include:
- A description of what the method does
- Documentation for each parameter using `@param`
- Documentation for the return value using `@return`

modules/mysql/src/main/resources/db/migration/V3.24__RecordSetData.sql
Jay07GIT: 1. I thought of JOIN queries will cause a performance issue. So i Added `record_name` and `owner_id` columns in the `recordset_data` table.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetDataRepository.scala
Jay07GIT: Sorry, Actually its `record_name`. I renamed  `zone_name` to `record_name`.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetDataRepository.scala
Jay07GIT: Actually its `record_name`. Just now i noticed the name. I renamed  `zone_name` to `record_name`.

modules/portal/conf/routes
```diff
@@ -25,6 +25,7 @@ GET           /download-creds-file/:fileName     @controllers.VinylDNS.serveCred
 POST          /regenerate-creds                  @controllers.VinylDNS.regenerateCreds
 
 GET           /api/recordsets                    @controllers.VinylDNS.listRecordSets
+GET           /api/recordsets                    @controllers.VinylDNS.listRecordSetData
```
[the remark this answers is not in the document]
Jay07GIT: Removed `line number 27`

modules/mysql/src/main/resources/db/migration/V3.24__RecordSetData.sql
```diff
@@ -18,6 +18,6 @@ CREATE TABLE recordset_data (
   INDEX recordset_data_recordset_id_index(recordset_id),
   INDEX recordset_data_reverse_fqdn_index(reverse_fqdn),
   FULLTEXT INDEX recordset_data_fdqn_fulltext_index(fqdn),
-  FULLTEXT INDEX recordset_data_record_data_fulltex_index(record_data)
-);
-
+  FULLTEXT INDEX recordset_data_reverse_fqdn_fulltext_index(reverse_fqdn),
```
remerle, line 21: Why do we need a `FULLTEXT INDEX` on the `reverse_fqdn`? If we do need it, it has to be added to the v3.26 flyway change.

remerle (changes requested):

modules/mysql/src/main/resources/db/migration/V3.24__RecordSetData.sql
Jay07GIT: Tried to implement <`match` `against`> in replacement of `like` in sql queries which will be improve performance of search with `fulltext index`. But i found that, it wont support for wildcard filters and characters used for search should be minimum 4 which will cause unnecessary confusion for the users. 

Hence i removed `full text` for `fqdn` and `reverse_fqdn`.

Jay07GIT: I tried to replace `JOIN` with [`Match` `against`] for recordset search SQL queries which uses `fulltext index`. But i found that `fulltext index` will not work with wildcard filters and it requires 4 characters minimum to search the word. This will cause unwanted confusion with the USERS in `recordset search`.

Hence i removed` fulltext index` for `fdqn` and `reverse_fdqn`.

remerle (changes requested):
Please revisit the method comments and try to be as descriptive as possible - I provided an example.

modules/mysql/src/main/resources/db/migration/V3.24__RecordSetData.sql
```diff
@@ -17,7 +17,5 @@ CREATE TABLE recordset_data (
   INDEX recordset_data_ip_index(ip),
   INDEX recordset_data_recordset_id_index(recordset_id),
   INDEX recordset_data_reverse_fqdn_index(reverse_fqdn),
-  FULLTEXT INDEX recordset_data_fdqn_fulltext_index(fqdn),
-  FULLTEXT INDEX recordset_data_record_data_fulltex_index(record_data)
-);
-
+  FULLTEXT INDEX recordset_data_record_data_fulltext_index(record_data)
```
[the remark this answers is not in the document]
Jay07GIT: Added alter query for drop full text indexes(fqdn, reverse_fqdn)

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetDataRepository.scala
```diff
@@ -202,90 +230,137 @@ class MySqlRecordSetDataRepository
...
+    }}
+
+  def getRecordSetDataList(zoneId: String, recordName: String, typ: RecordType): IO[List[RecordSet]] =
+    monitor("repo.RecordSet.getRecordSetDataList") {
+
+      /**
+        * get recordset by zone id, record name and record type from recordset_data table.
+        * @params zone ID , record name and record type.
+        * @return data, fqdn from recordset_data table.
+        */
```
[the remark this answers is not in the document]
Jay07GIT: Modified as suggested

modules/mysql/src/main/resources/db/migration/V3.26__RecordSetData_RecordName_OwnerGroup.sql
```diff
@@ -0,0 +1,12 @@
...
+USE ${dbName};
+
+ALTER TABLE recordset_data ADD COLUMN record_name VARCHAR(256) AFTER zone_id;
+CREATE INDEX recordset_data_record_name_index ON recordset_data (record_name);
+
+ALTER TABLE recordset_data ADD COLUMN owner_group_id CHAR(36) AFTER ip;
+CREATE INDEX recordset_data_owner_group_id_index ON recordset_data (owner_group_id);
+
+ALTER TABLE recordset_data DROP INDEX recordset_data_fdqn_fulltext_index;
+ALTER TABLE recordset_data DROP INDEX recordset_data_reverse_fqdn_fulltext_index;
```
remerle, line 12: We want to add a `FULLTEXT INDEX` on `record_data` here too.

```sql
CREATE FULLTEXT INDEX fulltext_record_data_index ON recordset_data_record_data (record_data);
```

remerle (changes requested):
@Jay07GIT See my notes. Let's get the SQL DDL fixed up.

Of primary concern:
- The `record_data` column should have a `FULLTEXT INDEX`
- The `data` column `BLOB` should be removed from the `recordset_data` table
    - Use `JOIN` clauses where necessary
- Do not modify existing Flyway changesets
	- If you modify the existing changesets, then Flyway will not apply the changes which is no bueno

modules/mysql/src/main/resources/db/migration/V3.26__RecordSetData_RecordName_OwnerGroup.sql
```diff
@@ -0,0 +1,12 @@
+CREATE SCHEMA IF NOT EXISTS ${dbName};
```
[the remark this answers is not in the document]
Jay07GIT: As suggested , I done the below modification

> Added `JOIN` for the `data BLOB` column and dropped from the `recordset_data` table.
> Removed the record_name and owner_id column from the `recordset_data` table. And used the same `JOIN` to get `record_name` and `owner_id` column from the `recordset` table.

modules/mysql/src/main/resources/db/migration/V3.26__RecordSetData_RecordName_OwnerGroup.sql
Jay07GIT: Since, we already have a `full text index` for `record_data` (with a typo), i dropped the full text index (`record_data`)and added a new one.

remerle:
Closing this out in favor of the updated PR #1125

PR #1103, opened 2022-03-25
Update log levels and adopt ECS

### LINKED ISSUE #1102: Make log level changes and adoption of ECS
Log levels should be based on the actual severity of the log message.
- Info only important when there's a problem? - `DEBUG` or `TRACE`
- General informational messages - `INFO`
- Exceptions and server errors - `ERROR`
- Errors that were handled - `WARN`

Currently, log levels were wrongly used with most logs being logged at `ERROR` level which should be corrected.

Once we've got the log messages at the correct levels, we should consider moving to using `Elastic Common Schema (ECS)`

Aravindh-Raju:
Fixes #1102 

Changes in this pull request:
- Change log levels based on the actual severity of the log message.
- Made changes to use ECS (Elastic Common Schema).

### DIFF
--- a/build/docker/api/logback.xml
+++ b/build/docker/api/logback.xml
@@ -1,8 +1,9 @@
 <configuration>
     <!-- Test configuration, log to console so we can get the docker logs -->
     <appender name="CONSOLE" class="ch.qos.logback.core.ConsoleAppender">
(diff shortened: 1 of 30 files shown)


remerle (approved):
Wow, meaningful log levels! I love it! 🎉

PR #1104, opened 2022-04-04
Fix Dev Deployment Failure due to Flyway DB Migration Issue

### LINKED ISSUE #1101: Fix Dev Deployment Failure due to Flyway DB Migration Issue
[Failed Deployment Job](https://ci.comcast.net/teams/vinyldns/pipelines/deployment/jobs/job-vinyldns-dev-deploy/builds/74)
[Splunk Logs](https://all-regions.ha-search.splunk.comcast.com/en-US/app/search/search?sid=1647542660.8033_5E79E30C-4ED7-4981-8441-7B127DD1475A&display.prefs.events.offset=0&dispatch.sample_ratio=1&display.page.search.mode=smart&q=search%20index%3D%22cloudservices%22%20vinyldns%20unable%20to%20start&earliest=-4h&latest=now&display.page.search.tab=events)

Error message: `VINYLDNS SERVER UNABLE TO START org.flywaydb.core.api.FlywayException: Validate failed: Detected resolved migration not applied to database: 3.24 `

A new migration file `V3.24__RecordSetData.sql` has been added to the mysql module. According to the schema version table, 3.25 was applied after 3.23 in the dev database. Thus the newly added migration files are ignored instead of being applied.
image

By using the [flyway.outOfOrder](https://flywaydb.org/documentation/configuration/parameters/outOfOrder) flag, we can allow migrations to be applied regardless of their version number, which would resolve this issue. We should make the outOfOrder flag configurable so we can just apply it to the dev environment. 

This code will be added to MySqlConnector.scala in the sql module, and will look something like this:
`// MySqlConnector.scala
val migration = Flyway
          .configure()
          .outOfOrder(true)`

nspadaccino:
This fix adds an "OutOfOrder" flag to the flyway migration, allowing migration files to be added regardless of the order of the version numbers. The flag will only be set to true if the "vinyldns_environment" variable is set to "dev". This means that the out of order flag will only apply during the dev deployment job, and will not affect other deployments.

Fixes #1101

### DIFF
--- a/build/docker/.env
+++ b/build/docker/.env
@@ -17,4 +17,4 @@ JDBC_URL=jdbc:mariadb://vinyldns-integration:19002/vinyldns?user=root&password=p
 JDBC_MIGRATION_URL=jdbc:mariadb://vinyldns-integration:19002/?user=root&password=pass
 JDBC_USER=root
 JDBC_PASSWORD=pass
-FLYWAY_OUT_OF_ORDER=true
+FLYWAY_OUT_OF_ORDER=false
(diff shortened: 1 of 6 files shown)


Aravindh-Raju (approved):

remerle (approved):
LGTM 👍

PR #1110, opened 2022-04-17
Various logging fixes

### LINKED ISSUE #864: Quiet illegal header warnings from akka-http
**Motivation**
We are seeing regular log entries like `(a.a.ActorSystemImpl:86) | Illegal header: Illegal 'authorization' header: Invalid input '/'` that are not consequential to the health of the application.  Let's see if we can quiet this somehow.

**Scope of change**
Configure akka-http and/or logback to quiet that log entry

corubba:
The goal of this PR to improve the logging in various ways. Below is a listing of the full commit messages with detailed explanations.

## Remove duplicate log output
[Appenders in logback are additive](https://logback.qos.ch/manual/configuration.html#cumulative), specifying them multiple times leads to log lines being output multiple times. It is usually enough to set the appender on the root logger, and nowhere else.
Setting the same log level on multiple package levels is also kind of unneeded, the top-most one would be enough.

## Reduce flyway log output
Let's establish a sane default by not having flyway run on full-blast debug output by default. For normal operation I found INFO to be verbose enough (as the flyway devs intended I guess).

## Disable illegal header warning from akka-http
Fixes #864

## Demote full SigV4 logging to DEBUG level
As the existing code comment said, it is for debugging; so make it DEBUG level. This means ~15 lines less logging per request. Also wrapped it in a loglevel guard to skip the string interpolation when not needed.

## Remove trailing newlines from log messages
The goal is to reduce visual jitter.

### DIFF
--- a/build/docker/api/application.conf
+++ b/build/docker/api/application.conf
@@ -327,8 +327,8 @@ akka.http {
   }
 
   parsing {
-    # Spray doesn't like the AWS4 headers
-    illegal-header-warnings = on
+    # Don't complain about the / in the AWS SigV4 auth header
+    ignore-illegal-header-for = ["authorization"]
   }
 }
 
--- a/build/docker/api/logback.xml
+++ b/build/docker/api/logback.xml
@@ -7,25 +7,13 @@
         </encoder>
     </appender>
 
-    <logger name="org.flywaydb" level="DEBUG">
-        <appender-ref ref="CONSOLE"/>
-    </logger>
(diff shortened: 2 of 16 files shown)


remerle (approved):
@corubba Thanks! This is in line with some of the other changes we've been making with logging (#1103).

Ultimately users of VinylDNS can customize their `logback.xml`, but we always prefer sensible defaults that aren't spamming the log file/shipper.

PR #1115, opened 2022-04-25
Dropped recordset BLOB hash column in recordset table

Jay07GIT:
Changes in this pull request:
- Dropped recordset BLOB `data_hash` column in `recordset` table, as its no longer use.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,7 @@ quickstart/data
 **/*cache*
 **/artifacts/
 **/.env.overrides
+**/node_modules
+.run
+**/*.zip
+**/*.log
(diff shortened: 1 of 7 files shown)


remerle:
This is a potentially breaking change. Anyone that has run `v10.5` or higher may have a `data_hash` column in their `recordset` table.

If this causes problems, workarounds include:
1. Update the column to make it nullable
    ```sql
    ALTER TABLE recordset
        MODIFY data_hash varchar(40) NULL;
    ```
2. OR, delete the column outright
    ```sql
    DROP INDEX data_hash_index ON recordset;
    
    ALTER TABLE recordset
        DROP COLUMN data_hash;
    ```

PR #1117, opened 2022-04-28
Reverse zones are updated as case insensitive

### LINKED ISSUE #957: Reverse zone not recognized as reverse if name is uppercase
**Describe the bug**
if connected zone is named ie. "0.1.10.IN-ADDR.ARPA" it is not recognized as reverse zone. PTR records are not valid  

**VinylDNS Version**
0.9.5-SNAPSHOT

**To Reproduce**
1. connect a reverse zone and name it uppercase
2. try add a new PTR rr
4. See error: "...PTR is not valid in forward lookup zone"

**Expected behavior**
both upper and lower case names should work

Jay07GIT:
Fixes #957.

Changes in this pull request:
- Reverse zone are updated with case insensitive.
- `PTR` records can be added in the `reverse_zone` irrespective of its upper and lower cases.


modules/api/src/main/scala/vinyldns/api/domain/ReverseZoneHelpers.scala
```diff
@@ -62,7 +63,7 @@ object ReverseZoneHelpers {
     }
 
   def convertPTRtoIPv4(zone: Zone, recordName: String): String = {
```
nspadaccino, line 65: I think it would be good to have some additional unit tests in `vinyldns/api/domain/ReverseZoneHelpersSpec.scala` to test the case insensitivity of this method and `convertPTRtoIPv6`

Aravindh-Raju (dismissed):

modules/api/src/main/scala/vinyldns/api/domain/ReverseZoneHelpers.scala
Jay07GIT: Added reverse zone unit test cases with uppercase as suggested.

modules/api/src/main/scala/vinyldns/api/domain/ReverseZoneHelpers.scala
```diff
@@ -62,7 +63,7 @@ object ReverseZoneHelpers {
...
-    val zoneName = zone.name.split("in-addr.arpa.")(0)
+    val zoneName = Pattern.compile("in-addr.arpa.", Pattern.CASE_INSENSITIVE).split(zone.name)(0)
```
remerle, line 66: This is not idiomatic Scala. We want to try to avoid pulling in Java libraries. To do this in Scala, you would use the `.r` function:

```scala
    val zoneName = "(?i)in-addr.arpa.".r.split(zone.name)(0)
```

However, this seems like overkill when we're just trying to remove the suffix. Instead, we can do this:
```suggestion
    val zoneName = zone.name.dropRight("in-addr.arpa.".length)
```

remerle (changes requested):

modules/api/src/main/scala/vinyldns/api/domain/ReverseZoneHelpers.scala
Jay07GIT: Updated the code as suggested.

modules/core/src/main/scala/vinyldns/core/domain/zone/Zone.scala
```diff
@@ -51,8 +51,8 @@ final case class Zone(
     isTest: Boolean = false,
     backendId: Option[String] = None
 ) {
-  val isIPv4: Boolean = name.endsWith("in-addr.arpa.")
-  val isIPv6: Boolean = name.endsWith("ip6.arpa.")
+  val isIPv4: Boolean = name.endsWith("in-addr.arpa.") || name.endsWith("IN-ADDR.ARPA.")
+  val isIPv6: Boolean = name.endsWith("ip6.arpa.") || name.endsWith("IP6.ARPA.")
```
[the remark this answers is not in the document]
Jay07GIT: Updated the code as suggested.

remerle (approved):

PR #1118, opened 2022-04-28
Get User API

### LINKED ISSUE #930: Add User Lookup API Endpoint
As a consumer of the API, I'm a member of a shared VinylDNS Zone.  I can obtain Zone, Recordset, and Change data from the shared zone.  If I'm a member of the group which owns the zone, I can query the groups and group members (User objects).  There doesn't exist a current endpoint to supply a UserID and get back details about that user unless they're a member of a group I'm presently in.

Just like there is an API endpoint to pass in a groupID to retrieve an arbitrary group, I'd like a similar one to resolve existing data obtained via Change object on a RecordSet belonging to a Zone.  It may also be useful to provide lookup methods for any other supported primary key fields such as username or email to get the ID guid.

I can make my own user lookup table based on my current group memberships, but that doesn't work very well in shared zones or long lived recordsets where users are offboarded and removed from groups.

nspadaccino:
Resolves #930

Users can now look up users from other groups, either by username or user ID


modules/api/src/main/scala/vinyldns/api/route/MembershipJsonProtocol.scala
```diff
@@ -124,4 +124,17 @@ trait MembershipJsonProtocol extends JsonValidation {
...
+      (
+        (js \ "id").default[String](UUID.randomUUID().toString),
+        (js \ "userName").optional[String],
+        (js \ "firstName").optional[String],
+        (js \ "lastName").optional[String],
+        (js \ "email").optional[String],
+        (js \ "created").optional[DateTime],
+        (js \ "lockStatus").default(LockStatus.Unlocked)
+        ).mapN(UserInfo.apply)
+  }
```
Aravindh-Raju, line 139: Everything looks fine. But I'm not sure whether we need `UserInfoSerializer` since we're not doing any validation with it unlike other Serializers. For example, `GroupSerializer` validates with the message `Missing Group.name` when we do a post request to create a group without the `name` field in the `body`.

But in this case, we're not validating anything just fetching the user data (we don't need anything in the body, just a get request to `/users/:identifier`). I think we only need serializers to validate fields in the `body`. I'm not sure about this though, let's wait for @remerle and @Jay07GIT review.

remerle (changes requested):
@nspadaccino In order to avoid leaking information, the `GET /users/<identifier>` endpoint should only return the id and username.

```json
{
  "id": "<user_id>",
  "username": "<username>"
}
```

remerle (approved):
LGTM 👍

Issue #1120, opened 2022-05-02
Validate Group Fields in API


Aravindh-Raju:
Group `name` and `email` are [required fields](https://www.vinyldns.io/api/create-group.html) and cannot be empty. `Portal` doesn't allow to create group with empty `name` or `email`. 

But while creating group through `api`, since there's no validation in place, it allows us to create group with empty name and email.

For eg: While sending POST request to `/groups` as follows, allows us to create a group with `name` and `email` field empty.

```
{
  "name": "",
  "email": "",
  "description": "Some group", 
  "members": [
    {
      "id": "professor-id"
    }
  ],
  "admins": [
    {
      "id": "professor-id"
    }
  ]
}
```

Attached screenshot of groups created through api which allows to have empty name or email or both.

<img width="1904" alt="Screenshot 2022-05-02 at 2 11 09 PM" src="https://user-images.githubusercontent.com/61419792/166208154-090c5820-d142-4502-9386-3387a7ce8b0c.png">

Aravindh-Raju:
Closing the issue as the PR: https://github.com/vinyldns/vinyldns/pull/1121 has been merged.

PR #1121, opened 2022-05-02
Validate Group Fields in API

Aravindh-Raju:
Fixes issue #1120.

Changes in this pull request:
- Added validation similar to Portal in API to throw an error message when we try to create a group with `name` and/or `email` fields empty. Now we cannot create a group with empty `name` and/or `email` fields through API and we'll get the message `Group name and email cannot be empty.` when either or both is empty.


nspadaccino (dismissed):
Looks good

Jay07GIT (dismissed):
Looks good to me.

modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
```diff
@@ -263,6 +265,16 @@ class MembershipService(
       .orFail(GroupNotFoundError(s"Group with ID $groupId was not found"))
       .toResult[Group]
 
+  // Validate group details. Group name and email cannot be empty
+  def groupValidation(group: Group): Result[Unit] = {
+    Option(group) match {
+      case Some(value) if value.name.isEmpty || value.email.isEmpty =>
```
remerle, line 271: It may be a good idea to `.trim` the values in case they're whitespace.

```suggestion
      case Some(value) if value.name.trim.isEmpty || value.email.trim.isEmpty =>
```

It's also a good idea to handle the possibility of a null value:
```suggestion
      case Some(value) if Option(value.name).forall(_.trim.isEmpty) || Option(value.email).forall(_.trim.isEmpty)=>
```

remerle (changes requested):
Minor tweak

modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
Aravindh-Raju: Made the suggested changes.

nspadaccino (approved):

PR #1123, opened 2022-05-04
refactor: allow assume role with optional externalId

pedrokiefer:
Refactor route53 to allow assuming a role, either with the default chain, or using a given key pair.


Aravindh-Raju (approved):
Tested out locally. Looks good to me!

pedrokiefer:
I'll close this in favor of #1300, which has signed commits.

Issue #237, opened 2018-09-24
Allow ACL rules for any group


rebstar6:
Zone admins are able to add ACL rules on groups to grant them limited access within a zone. However, they are only allowed to add an ACL rule for groups they are a member of, which is limiting.

We should:
- Remove the restriction that the user be in the group the ACL is being created for
- In the portal, update the groups selection in the ACL rule form to display as free text (a dropdown will be too big). Autocomplete would be ideal if possible.

britneywright:
We should break this one down

djvl:
Hi, is there any progress on this issue?

pauljamescleary:
@djvl not yet, we can prioritize this.  Is this holding you up?

Let us know if you can give some color as to how this issue is impacting you, that would be useful for prioritzation :)

jhg03a:
I'd add to this that it may not just be zone admins wanting to modify ACLs, but also record owners.  This also would include the ability to transfer record ownership from one group directly to another.

remerle:
I know that there is a new portal in the works (#968), but this seems like a limiting feature that may make sense to take care of.

remerle:
Fixed by #1127

remerle:
Only took four years! 🎉

PR #1127, opened 2022-05-17
Allow Users to See All Groups When Creating an ACL Rule

### LINKED ISSUE #1119: Users cannot include groups to which they do not belong in zone ACLs
## Description
The purpose of the ACL on zones is to enable control over who can access the zone. This could, very conceivable, include groups to which I, as a user, do not belong.

Say that I have a zone, `vinyldns.io` that is private. I and the users in the administrative group can modify the records 👍 - all good.

Now, let's say that I want to give access to the group `DB Administrators` for records that contain `^.*db.*$` but I am not in the group `DB Administrators`. At present, there is no way for my to allow access, even though I am a zone administrator, and the zone is private.

In fact, the only users that can do this would be super users.

Arguably, the only users that can truly decide whether to allow access would be the private owners. So, this puts an unnecessary burden on teams supporting VinylDNS.

## Resolution
Update the API and Portal where necessary to support/display the full list of groups in the dropdown for the zone ACL editor view.

nspadaccino:
Fixes #1119

The zone ACL editor view now displays the full list of groups in the dropdown, instead of just groups the user is part of.


remerle:
This is a duplicate of #237 so that has been fixed as well.

PR #1136, opened 2022-05-24
feat: dotted-label allowlist

### LINKED ISSUE #1129: Support for ._domainkey. CNAMEs
Most SaS uses a dotted CNAME postfixed with `._domainkey.`, currently those entries do not show up on VinylDNS. Creating a zone just for this entries feels odd. Can this entries be allowed as CNAMEs? 

I understand and agree with the idea of not allowing dotted names for most entries, and that this would be yet another exception to this rule.

I can work on a PR implementing this if needed.

pedrokiefer:
Fixes #1129.

Changes in this pull request:
- Implements basic dotted label allowlist as discussed on #1129

### DIFF
--- a/modules/api/src/test/functional/tests/recordsets/create_recordset_test.py
+++ b/modules/api/src/test/functional/tests/recordsets/create_recordset_test.py
@@ -599,6 +599,31 @@ def test_create_dotted_cname_record_fails(shared_zone_test_context):
     assert_that(error, is_(f'Record with name dot.ted and type CNAME is a dotted host which is not allowed in zone {zone["name"]}'))
 
 
+def test_create_dotted_cname_record_in_allow_list_succeeds(shared_zone_test_context):
+    """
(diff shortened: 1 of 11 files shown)


modules/api/src/test/functional/tests/recordsets/create_recordset_test.py
```diff
@@ -599,6 +599,31 @@ def test_create_dotted_cname_record_fails(shared_zone_test_context):
     assert_that(error, is_(f'Record with name dot.ted and type CNAME is a dotted host which is not allowed in zone {zone["name"]}'))
 
 
+def test_create_dotted_cname_record_in_allow_list_succeeds(shared_zone_test_context):
+    """
+    Test that creating a CNAME record set with dotted host record name returns an error.
+    """
```
Aravindh-Raju, line 605: This comment seems to be copied from the previous test. Can you please change it accordingly? Thanks!

Aravindh-Raju (changes requested):

PR #1142, opened 2022-06-01
Add pagination in group view

### LINKED ISSUE #1137: Add Pagination in Groups view
Currently we don’t have pagination in `Groups` view. We only have infinite scroll. 
Add pagination in `Groups` view similar to what we have in `Zones` view.

Aravindh-Raju:
Fixes #1137.

Changes in this pull request:
- Added pagination in `Groups` view similar to what we have in `Zones` view.

### DIFF
--- a/modules/api/src/main/resources/reference.conf
+++ b/modules/api/src/main/resources/reference.conf
@@ -110,8 +110,8 @@ vinyldns {
     limits {
       batchchange-routing-max-items-limit = 100
       membership-routing-default-max-items = 100
-      membership-routing-max-items-limit = 1000
-      membership-routing-max-groups-list-limit = 1500
+      membership-routing-max-items-limit = 100
+      membership-routing-max-groups-list-limit = 100
       recordset-routing-default-max-items= 100
       zone-routing-default-max-items = 100
       zone-routing-max-items-limit = 100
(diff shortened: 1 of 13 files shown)


nspadaccino (dismissed):
Tested locally with a connection to the actual database so I could page through all the groups, looks good

Jay07GIT (approved):
LGTM. Can be merge.

nspadaccino (commented):
I tested this again locally with a database snapshot so I could look at all of our groups, and I noticed that the groups are not in overall alphabetical order across the pages. Within one page the groups are alphabetized, but the selection of groups that are displayed on the page seems random. I expected page 1 to have groups starting with symbols or the letter "A", but instead I'm getting groups from all over the alphabet. Could you check the behavior on your side?

Aravindh-Raju:
> I tested this again locally with a database snapshot so I could look at all of our groups, and I noticed that the groups are not in overall alphabetical order across the pages. Within one page the groups are alphabetized, but the selection of groups that are displayed on the page seems random. I expected page 1 to have groups starting with symbols or the letter "A", but instead I'm getting groups from all over the alphabet. Could you check the behavior on your side?

Thanks for finding that out. I've made the changes to sort groups by `name` instead of `id`. It should be working fine as expected now. Please have a look.

nspadaccino (approved):
Looks good to me

PR #1144, opened 2022-06-10
Allow add for already exists DNS records.

### LINKED ISSUE #1138: Requested record deletes with (RecordDoesNotExist) should not prevent batch submission
When a user submits a batch change request, and one or more recordsets in the batch change result meet the below criteria, then VinylDNS should allow the batch change to complete successfully.

**Criteria**

- Requested change is a `DeleteRecordSet` change
- The DNS backend does not contain the requested record

In this case, the user's intent is to have the record removed from the DNS backend. If the record doesn't exist, then their request has already been fulfilled.

From a UI perspective, we may want to display an informational message next to the affected records saying something like:

> ℹ️ This record does not exist. No further action is required.

In short, if a user requests the deletion of a record that does not exist, VinylDNS MUST NOT treat this scenario as an error.

### LINKED ISSUE #1139: Record additions that fail with (RecordAlreadyExists) should not prevent approval after manual review
When a user submits a batch change request, and one or more recordsets in the batch change result meet the below criteria, then VinylDNS should allow the batch change to complete successfully after a manual review has been approved.

**Criteria**

- Requested change is an `Add` change
- The DNS backend already contains a record (and VinylDNS would return a `RecordAlreadyExists` error)

The `RecordAlreadyExists` error should continue to require manual review of the batch change. However, once the change is reviewed and approved, the batch change should be marked as `Completed`.

This only applies if the only error(s) in the batch are `RecordAlreadyExists` errors. If other errors exist, then the batch should behave as it does prior to this change.

### See Also
- #1138

### LINKED ISSUE #1149: Requested record additions with record data already exists should get complete without error.
When a user submits a batch change request with the existing `record set` and `record data`, then VinylDNS should allow the batch change to `complete` successfully.

From a UI perspective, we may want to display an informational message next to the affected records saying something like:

> ℹ️ This record already exists. No further action is required.

In short, if a user requests the addition of a `record set` with `record data `that already exist, VinylDNS MUST NOT treat this scenario as an error.

Follow to :
-  #1139

Jay07GIT:
**Fixes #1139 and Fixes #1149.**

- Allow add for already exists DNS records.

**Scenario:**
- Users can add the recordset in DNS change page. VinylDNS will notify with 
1. _if record data already exists_ =>  `"No further actions require."`
2. _if record data does not exists_ =>  This will go for Manual approval. `"RecordName "$name" already exists. Your request will be manually reviewed.  If you intended to update this record, you can avoid manual review by adding a DeleteRecordSet entry followed by an Add."`

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/DomainValidations.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/DomainValidations.scala
@@ -27,6 +27,10 @@ import scala.util.matching.Regex
   Object to house common domain validations
  */
 object DomainValidations {
+  val validReverseZoneFQDNRegex: Regex =
+    """^(?:([0-9a-zA-Z\-\/_]{1,63}|[0-9a-zA-Z\-\/_]{1}[0-9a-zA-Z\-\/_]{0,61}[0-9a-zA-Z\-\/_]{1}|[*.]{2}[0-9a-zA-Z\-\/_]{0,60}[0-9a-zA-Z\-\/_]{1})\.)*$""".r
+  val validForwardZoneFQDNRegex: Regex =
+    """^(?:([0-9a-zA-Z_]{1,63}|[0-9a-zA-Z_]{1}[0-9a-zA-Z\-_]{0,61}[0-9a-zA-Z_]{1}|[*.]{2}[0-9a-zA-Z\-_]{0,60}[0-9a-zA-Z_]{1})\.)*$""".r
   val validFQDNRegex: Regex =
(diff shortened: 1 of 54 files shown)


Jay07GIT:
> So I was testing locally and it seems like adding a totally new record to a zone results in a `This record already exists. No further action is required.` message. I was wondering if this is intended. Included some screenshots of that test below. Other scenarios seem to work as expected. Screen Shot 2022-07-07 at 12 22 53 PM Screen Shot 2022-07-07 at 12 23 01 PM

@nspadaccino Thanks for alerting me on this issue. 

**Reason for the issue:-** For creating a new record, _Vinyldns_ will first create a record, then it will try to validate the record already exists, then it will update the status as completed. (i.e. both already exists and newly created record has same message and same validations)

**Solution:-** Updated the message as `"No further action required."` Since its understood.

Aravindh-Raju (approved):
LGTM!

PR #1146, opened 2022-06-13
Add an Option to Filter Out PTR Zones in the Zones View

### LINKED ISSUE #1122: Hide PTR zones from zone list by default
As a user of the Vinyldns UI with large numbers of PTR zones associated with my group, I'd like a way to hide those by default to declutter the zone list.

nspadaccino:
Fixes #1122

A checkbox option to filter out PTR zones has been added in the zones view. This filter will still apply when searching for zone names


Jay07GIT (approved):
Looks good to me.

PR #1152, opened 2022-07-06
Add Unit Tests For Zone Search Case Insensitivity

### LINKED ISSUE #1143: Make Zone Search Case-Insensitive
**Description**
Currently zone search in VinylDNS is case sensitive. So if you're looking for a zone called `Testzone` and you enter `testzone` in the search box, you won't get any results. Since DNS records are not case sensitive, we can make this more intuitive for users by having the zone search ignore case.

**Fix**
Modify the `ListZones` API so the `zoneNameFilter` query parameter is case insensitive.

nspadaccino:
Fixes #1143

### DIFF
--- a/modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneRepository.scala
+++ b/modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneRepository.scala
@@ -254,7 +254,7 @@ class MySqlZoneRepository extends ZoneRepository with ProtobufConversions with M
           sb.append(withAccessorCheck)
 
           val filters = List(
-            zoneNameFilter.map(flt => s"LOWER ( z.name ) LIKE '${ensureTrailingDot(flt.replace('*', '%').toLowerCase())}'"),
+            zoneNameFilter.map(flt => s"z.name LIKE '${ensureTrailingDot(flt.replace('*', '%'))}'"),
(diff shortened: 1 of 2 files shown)


modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneRepository.scala
```diff
@@ -254,7 +254,7 @@ class MySqlZoneRepository extends ZoneRepository with ProtobufConversions with M
           sb.append(withAccessorCheck)
 
           val filters = List(
-            zoneNameFilter.map(flt => s"z.name LIKE '${ensureTrailingDot(flt.replace('*', '%'))}'"),
+            zoneNameFilter.map(flt => s"LOWER ( z.name ) LIKE '${ensureTrailingDot(flt.replace('*', '%').toLowerCase())}'"),
```
Jay07GIT, line 257: We can ignore these changes, since **zone search** is already working with case sensitive. Other wise looks good. Good that we covered tests on **zone search** with case sensitive.

Jay07GIT (changes requested):

Jay07GIT (approved):
Looks good to me

Jay07GIT (approved):

arpit4ever (approved):
LGTM

PR #1162, opened 2022-08-05
Allow All Users to View Any Shared Zone

### LINKED ISSUE #643: Allow all users to see any existing zone
We want to open up read-only access of data in VinylDNS. This scope of this card is limited to individual zones. A separate card will be created for the list zones endpoint.

Allow any user to see any existing zone they request. They should not be able to update or delete the zone.

**Additional context**
I believe the endpoints this effects are:

- get zone by ID
- get zone by Name
- list zone changes

we may want to include and return an access level check that notes if the requesting user is a super user, zone admin, has ACL rule access, or none.

nspadaccino:
Fixes #643

All users are able to make changes within shared zones, so it makes sense that they should be able to view the shared zones as well. This change allows all users to view a shared zone regardless of membership

### DIFF
--- a/modules/portal/app/views/zones/zoneTabs/manageRecords.scala.html
+++ b/modules/portal/app/views/zones/zoneTabs/manageRecords.scala.html
@@ -66,7 +66,8 @@
       <div class="vinyldns-panel-top">
         <div class="btn-group">
             <button id="refresh-records-button" class="btn btn-default" ng-click="refreshRecords()"><span class="fa fa-refresh"></span> Refresh</button>
-            <button id="create-record-button" class="btn btn-default" ng-if="canCreateRecords" ng-click="createRecord(defaultTtl)"><span class="fa fa-plus"></span> Create Record Set</button>
+<!--            <button id="create-record-button" class="btn btn-default" ng-if="canCreateRecords" ng-click="createRecord(defaultTtl)"><span class="fa fa-plus"></span> Create Record Set</button>-->
+            <button id="create-record-button" class="btn btn-default" ng-if="canReadZone" ng-click="createRecord(defaultTtl)"><span class="fa fa-plus"></span> Create Record Set</button>
             <button id="zone-sync-button" class="btn btn-default mb-control" ng-if="zoneInfo.accessLevel=='Delete'" data-toggle="modal" data-target="#mb-sync"><span class="fa fa-exchange"></span> Sync Zone</button>
         </div>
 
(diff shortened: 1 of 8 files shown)


Aravindh-Raju (commented):
Everything looks fine and I'm able to view shared zones. I just have few doubts. 

Non-owners are supposed to have `Read-only` access right? But I am able to make changes (add, delete and update records) from the zone view itself. And also, it doesn't require admin group to make changes like in DNS Changes page which must be a required field. I've attached a pic for your reference.

Screenshot 2022-08-10 at 1 09 06 PM

We only want all the users to view shared zones but continue making changes from DNS Changes right?

modules/portal/app/views/zones/zoneTabs/manageRecords.scala.html
```diff
@@ -66,7 +66,8 @@ <h3 class="panel-title">Records {{ getRecordPageTitle() }}</h3>
       <div class="vinyldns-panel-top">
         <div class="btn-group">
             <button id="refresh-records-button" class="btn btn-default" ng-click="refreshRecords()"><span class="fa fa-refresh"></span> Refresh</button>
-            <button id="create-record-button" class="btn btn-default" ng-if="canCreateRecords" ng-click="createRecord(defaultTtl)"><span class="fa fa-plus"></span> Create Record Set</button>
+<!--            <button id="create-record-button" class="btn btn-default" ng-if="canCreateRecords" ng-click="createRecord(defaultTtl)"><span class="fa fa-plus"></span> Create Record Set</button>-->
+            <button id="create-record-button" class="btn btn-default" ng-if="canReadZone" ng-click="createRecord(defaultTtl)"><span class="fa fa-plus"></span> Create Record Set</button>
```
Aravindh-Raju, line 70: Just a minor change. The other conditions from `canCreateRecords` except the condition `$scope.zoneInfo.shared` must be added to the `ng-if` because I'm able to see the `Create Record Set` button and create records when I'm given `read-only` access via ACL which shouldn’t be the expected behavior. When given read-only access, I shouldn’t be able to see the button and shouldn’t be able to create a record. The reason is, the condition `canCreateRecordsViaAcl()` was not present in `ng-if`. Changing the `ng-if` as follows solves the problem:
```
<button id="create-record-button" class="btn btn-default" ng-if="canReadZone && (zoneInfo.accessLevel == 'Delete' || canCreateRecordsViaAcl())" ng-click="createRecord(defaultTtl)"><span class="fa fa-plus"></span> Create Record Set</button>
```
We just need to add those 2 conditions to `ng-if`.
Also, we can remove the commented line.

modules/portal/public/lib/controllers/controller.records.js
```diff
@@ -369,8 +369,9 @@ angular.module('controller.records', [])
     function determineAdmin(){
         $scope.isZoneAdmin = $scope.profile.isSuper || isInAdminGroup();
         $scope.canReadZone = canReadZone();
-        $scope.canCreateRecords = $scope.zoneInfo.accessLevel == 'Delete' || $scope.zoneInfo.shared ||
-            canCreateRecordsViaAcl();
+        $scope.canCreateRecords = $scope.zoneInfo.accessLevel == 'Delete' || canCreateRecordsViaAcl() || $scope.zoneInfo.shared;
+//        $scope.canCreateRecords = $scope.zoneInfo.accessLevel == 'Delete' || canCreateRecordsViaAcl();
```
Aravindh-Raju, line 373: We can remove this commented line.

Aravindh-Raju (changes requested):

Aravindh-Raju (approved):

Jay07GIT (approved):
LGTM

arpit4ever (approved):
LGTM

PR #1163, opened 2022-08-09
Change to java8 time

### LINKED ISSUE #164: Replace joda time with java8 time
**Motivation**
Joda time was a standard for working with time before java 8 due to inefficiencies with the java 8 time support.  Java 8 has introduced a new api for time that makes joda no longer a requirement.

**Scope of change**
This touches a lot of places:

1. Remove the dependency on joda time
1. Fix all the things that no longer work.

**Notes**

* `DateTime.now` becomes `Instant.now`
* Need to check serialization everywhere, as database, protobuf, json will all be slightly different.

Aravindh-Raju:
Fixes #164 

Changes in this pull request:
- Replaced joda time with java8 time.
- Removed the dependency on joda time.

### DIFF
--- a/modules/api/src/it/scala/vinyldns/api/backend/dns/DnsBackendIntegrationSpec.scala
+++ b/modules/api/src/it/scala/vinyldns/api/backend/dns/DnsBackendIntegrationSpec.scala
@@ -16,7 +16,7 @@
 
 package vinyldns.api.backend.dns
 
-import org.joda.time.DateTime
+import java.time.Instant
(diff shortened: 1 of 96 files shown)


nspadaccino (dismissed):
Looks good to me. Tested various portal actions locally, all datetimes are in the correct format.

Jay07GIT (dismissed):
LGTM. Can be merge.

nspadaccino (approved):

PR #1165, opened 2022-08-17
Replaced orchard CIDR library to IP4s library

### LINKED ISSUE #235: Use ip4s for CIDR and IP validating

Jay07GIT:
Fixes #235.

Changes in this pull request:
- Replaced `orchard CIDR` library to `com.comcast IP4s` library.
- Upgraded `IP4s` version from `1.1.1` to `3.1.3`(latest)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -72,6 +72,9 @@ lazy val apiAssemblySettings = Seq(
       MergeStrategy.discard
     case PathList("scala", "tools", "nsc", "doc", "html", "resource", "lib", "template.js") =>
       MergeStrategy.discard
+    case "simulacrum/op.class" | "simulacrum/op$.class" | "simulacrum/typeclass$.class"
+         | "simulacrum/typeclass.class" | "simulacrum/noop.class" =>
+      MergeStrategy.discard
     case x if x.endsWith("module-info.class") => MergeStrategy.discard
     case x =>
       val oldStrategy = (assemblyMergeStrategy in assembly).value
--- a/modules/api/src/test/functional/tests/batch/create_batch_change_test.py
+++ b/modules/api/src/test/functional/tests/batch/create_batch_change_test.py
@@ -2302,6 +2302,9 @@ def test_ipv4_ptr_recordtype_add_checks(shared_zone_test_context):
 
         # delegated and non-delegated PTR duplicate name checks
         assert_successful_change_in_error_response(response[4], input_name=f"{ip4_prefix}.196", record_type="PTR", record_data="test.com.")
+        assert_failed_change_in_error_response(response[5], input_name=f"196.{ip4_zone_name}", record_type="CNAME", record_data="test.com.",
+                                               error_messages=[f'Record Name "196.{ip4_zone_name}" Not Unique In Batch Change: cannot have multiple "CNAME" records with the same name.'])
+        assert_successful_change_in_error_response(response[6], input_name=f"196.192/30.{ip4_zone_name}", record_type="CNAME", record_data="test.com.")
         assert_successful_change_in_error_response(response[7], input_name=f"{ip4_prefix}.55", record_type="PTR", record_data="test.com.")
         assert_failed_change_in_error_response(response[8], input_name=f"55.{ip4_zone_name}", record_type="CNAME", record_data="test.com.",
                                                error_messages=[f'Record Name "55.{ip4_zone_name}" Not Unique In Batch Change: cannot have multiple "CNAME" records with the same name.'])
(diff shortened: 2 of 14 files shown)


nspadaccino (commented):
Couple questions for my knowledge, no changes required

modules/api/src/test/scala/vinyldns/api/domain/ReverseZoneHelpersSpec.scala
```diff
@@ -136,8 +136,8 @@ class ReverseZoneHelpersSpec
     }
     "recordsetIsWithinCidrMask" should {
       "when testing IPv4" should {
-        "filter in/out record set based on CIDR rule of 0 (lower bound for ip4 CIDR rules)" in {
-          val mask = "120.1.2.0/0"
+        "filter in/out record set based on CIDR rule of 1 (lower bound for ip4 CIDR rules)" in {
+          val mask = "120.1.2.0/1"
```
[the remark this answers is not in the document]
Jay07GIT: Yes. `ip4s` lower bound CIDR starts from 1.

modules/api/src/test/functional/tests/zones/update_zone_test.py
```diff
@@ -318,12 +318,12 @@ def test_create_acl_user_rule_invalid_cidr_failure(shared_zone_test_context):
         "accessLevel": "Read",
         "description": "test-acl-user-id",
         "userId": "789",
-        "recordMask": "10.0.0.0/50",
+        "recordMask": "10.0.0/50",
```
Aravindh-Raju, line 321: Just for my clarification, why was the final byte/block of the IP Address removed here?

modules/api/src/test/functional/tests/recordsets/create_recordset_test.py
```diff
@@ -1665,7 +1665,7 @@ def test_create_ipv4_ptr_recordset_with_verify_in_classless(shared_zone_test_con
     try:
         new_rs = {
             "zoneId": reverse4_zone["id"],
-            "name": "196",
+            "name": "193",
```
Aravindh-Raju, line 1668: I have the same doubts that @nspadaccino mentioned in his review. In addition to that, is this change made for debugging purpose or something that was necessary to be changed?

build.sbt
```diff
@@ -72,6 +72,9 @@ lazy val apiAssemblySettings = Seq(
       MergeStrategy.discard
     case PathList("scala", "tools", "nsc", "doc", "html", "resource", "lib", "template.js") =>
       MergeStrategy.discard
+    case "simulacrum/op.class" | "simulacrum/op$.class" | "simulacrum/typeclass$.class"
```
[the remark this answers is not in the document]
Jay07GIT: `Ip4s` upgraded from 1.1.1 to 3.0.3. So there is a conflict in `simulacrum/*. class` files, because of `typelevel/simulacrum` and `mpilquist/simulacrum_2.12` dependencies. So we need to discard the MergeStrategy for corresponding files in `build.sbt`

modules/api/src/test/functional/tests/zones/update_zone_test.py
Jay07GIT: `Orchand CIDR` has different error message for `IPV4` and `IPV6` , But `ip4s CIdr` has evaluate all in one. So to make `Cidr` fail this has been removed.

Aravindh-Raju: > `Orchand CIDR` has different error message for `IPV4` and `IPV6` , But `ip4s CIdr` has evaluate all in one. So to make `Cidr` fail this has been removed.

Got it. Thanks!

modules/api/src/test/functional/tests/batch/create_batch_change_test.py
```diff
@@ -2302,9 +2302,9 @@ def test_ipv4_ptr_recordtype_add_checks(shared_zone_test_context):
 
         # delegated and non-delegated PTR duplicate name checks
         assert_successful_change_in_error_response(response[4], input_name=f"{ip4_prefix}.196", record_type="PTR", record_data="test.com.")
-        assert_successful_change_in_error_response(response[5], input_name=f"196.{ip4_zone_name}", record_type="CNAME", record_data="test.com.")
-        assert_failed_change_in_error_response(response[6], input_name=f"196.192/30.{ip4_zone_name}", record_type="CNAME", record_data="test.com.",
-                                               error_messages=[f'Record Name "196.192/30.{ip4_zone_name}" Not Unique In Batch Change: cannot have multiple "CNAME" records with the same name.'])
+        assert_failed_change_in_error_response(response[5], input_name=f"196.{ip4_zone_name}", record_type="CNAME", record_data="test.com.",
```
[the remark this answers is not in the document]
Jay07GIT: The `response[5] and [6]` from the `Cidr` is changed. So the assertions has been switched.

nspadaccino (approved):
Looks good to me

PR #1167, opened 2022-08-23
Replace await with IO

### LINKED ISSUE #464: Replace await with IO
**Motivation**
Recently there have been some transient failures in our unit tests where the result of `await`/`awaitResultOf` will fail with a `None.get` exception. These should be replaceable using the Cats `IO` library, as we've switched over.

**Scope of change**
Corresponding unit tests will have to get updated, and potentially remove deprecated functions from `CatsHelpers.scala`.

Aravindh-Raju:
Fixes #464 

Changes in this pull request:
- Replaced `await` and other related functions present in `CatsHelpers.scala` and `ResultHelpers.scala` which is not necessary with cats `IO` library as we switched over.


nspadaccino (approved):
Looks good!

arpit4ever (approved):
LGTM

PR #1168, opened 2022-08-25
Remove 'Complete' zone change status

### LINKED ISSUE #618: All completed Zone Changes are saved with the status "Synced"
**Describe the bug**
In the ZoneChangeHandler, a completed zone change is saved with the status "Synced". This doesn't seem intuitive since we also have an unused "Complete" status. Think we should decide on if we want to use that status or remove it 

**VinylDNS Version**
0.9.2-SNAPSHOT

Aravindh-Raju:
Fixes #618 

Changes in this pull request:
- Remove the zone change status `Complete` and it's occurrences since it is not being used anywhere other than tests.


nspadaccino (approved):
Looks good to me

Jay07GIT (approved):
Looks good to me.

Issue #975, opened 2020-09-04
Pull fields out of binary BLOBs


remerle:
In order to allow for metric gathering, as well as more sophisticated searching within the application, we need to move away from using BLOBs to store data. BLOBs are opaque and unqueryable.

* Any place where a BLOB is not strictly necessary (were we'd need multiple tables to house the data), pull the data from the BLOB up into fields
* Any place where a BLOB still makes sense, pull up any fields that may be important for querying (all fields where pulling them up into the table wouldn't necessitate extra tables to house the relationships)

### Impact
* This will be a breaking change which will require migration
* It will be important to drop DynamoDB support (#971) since DynamoDB cannot easily support the types of queries that we'd want to perform against the new columns. It's possible, but it'd be ugly and the wrong tool for the job.

pauljamescleary:
A few things for consideration:

1. Currently blobs are passed via commands (like `RecordSetChange`) in the message queue
2. When loading very large zones for the purposes of doing a zone sync, we load all records at one time in order to perform a massive diff operation.  This is fast for few records, but zones that have 100,000 records this can take a long time to complete.  The protobufs are excellent in this regard due to their very small size for this operation.  If we remove blobs, we need to optimize the zone sync process.

remerle:
This issue just looks to solve the opaqueness problem, so let's be sure to keep blobs where necessary for zone sync and queueing.

However, let's consider a hash-based approach to zone sync once we open up the hood. Follow up with a secondary issue to tackle that when/where it makes sense.

PR #1169, opened 2022-08-26
Added API and Updated Portal for deleted zones history

### LINKED ISSUE #617: No way to view zone change history on deleted zone
**Describe the bug**
The only endpoint we expose for zone changes is `/zones/:id/changes`. This endpoint starts by checking if the zone exists and if it doesn't it returns a 404. This makes it impossible to view the change history of a deleted zone.

Not sure if it speeds up performance by checking if the zone is there before querying dynamo, so maybe keep doing that and only don't if the user is a support user?

**VinylDNS Version**
0.9.2-SNAPSHOT

Jay07GIT:
Fixes #617 .

Changes in this pull request:
- Added API for deleted zones history (`zones/deleted/changes`)
- Made Portal changes to view deleted zones history.
- Removed Cascade from `zone_access` table. We use this on comparing `zone_change` table .
- Added zone_status column in zone_access table to identify the deleted zones and used the same in deleting the group.


modules/mysql/src/main/resources/db/migration/V3.28__AlterZoneAccess.sql
```diff
@@ -0,0 +1,8 @@
+CREATE SCHEMA IF NOT EXISTS ${dbName};
+
+USE ${dbName};
+
+ALTER TABLE zone_access
+DROP CONSTRAINT fk_zone_access,
+ADD zone_status CHAR(36) NOT NULL
```
Aravindh-Raju, line 7: Since we are adding a new column `zone_status`, I think we'll need to populate this column for existing data in our production database (other envs too), possibly using the `rsdump` utility which we used in the [BLOB issue](https://github.com/vinyldns/vinyldns/issues/975) for the same populating purpose.

modules/portal/public/lib/controllers/controller.zones.js
```diff
@@ -110,8 +113,30 @@ angular.module('controller.zones', [])
...
+                    })
+                    .catch(function (error) {
+                        handleError(error, 'zonesService::getDeletedZones-failure');
+                    });
     };
 
+    function updateDeletedZoneDisplay (deletedZones) {
+        $scope.deletedZones = deletedZones;
+        $scope.DeletedZonesLoaded = true;
+        $log.log("Displaying my Deleted zones: ", $scope.deletedZones);
```
Aravindh-Raju, line 132: I'm thinking if we need to have `log` statements which will be displayed in production too. It may not be a great idea to expose `log` messages in our production code. Ideally, we should remove such log messages with a build script before deployment. I think we can have a separate PR to handle all the `logs` in `portal` module after further discussion.

Aravindh-Raju (commented):
Everything works fine as expected locally. Is it possible to increase code coverage further? I see few warnings by `codecov` in the lines that were added as not covered by tests. I'm seeing that in `Files Changed` column at the files `MySqlZoneChangeRepository.scala`, `ZoneProtocol.scala` and `ZoneService.scala`. Code coverage will be increased well enough if we are able to cover those lines in tests.

modules/mysql/src/main/resources/db/migration/V3.28__AlterZoneAccess.sql
Jay07GIT: Yes. Thats right. We need to populate the data for existing record by using separate utility. We will try to do this by `rsdump`. Noted.

modules/portal/public/lib/controllers/controller.zones.js
Jay07GIT: Sure. Will discuss about it.

Jay07GIT: Portal logs updated to debug.

Aravindh-Raju (changes requested):
Functionality looks good. Just need a few changes.

modules/portal/public/lib/controllers/controller.zones.js
```diff
@@ -110,8 +113,30 @@ angular.module('controller.zones', [])
             .catch(function (error) {
                 handleError(error, 'zonesService::getZones-failure');
             });
+
+        zonesService
+                    .getDeletedZones(zonesPaging.maxItems, undefined, $scope.query)
```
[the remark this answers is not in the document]
Jay07GIT: updated the code.

Aravindh-Raju (approved):
Tested it out locally. LGTM!

JoshSEdwards:
Tested it locally and it looked good to me. Only one weird thing I noticed (which may be desirable?) is that you can connect and abandon the same zone over and over again, and it appears multiple times in the My Abandoned Zones tab. Each time the zone is newly connected it gets a new UUID. Not sure if that's by design.

JoshSEdwards (approved):

nspadaccino (commented):
One potential issue I noticed is that if I abandon a zone with a specific admin group, and then delete that group, the abandon history for that zone gets removed, and does not come back if I create the group again. I'm assuming that happens because the zone deletes are tied to a group ID, and deleting and adding a group with the same name changes the ID. So I'm not sure if that behavior could be changed.

Jay07GIT:
SInce faced conflicts and git sign issue. This PR has been replicated to [New PR](https://github.com/vinyldns/vinyldns/pull/1284)

PR #1171, opened 2022-08-30
Override ZoneConnection toString

### LINKED ISSUE #408: Ignore secret in ZoneConnection keys when stringifying
**Motivation**
Sensitive connection information should be masked in the logs. Currently `ZoneConnection` doesn't have an override `toString` method to encrypt the connection information, so printing either the `ZoneConnection` or `Zone` in logs will store the keys.

**Scope of change**
One approach could be to associate a cryptography with the ZoneConnection and override the `toString` method.

Another, simpler approach, is to just remove the key information.

---
Edit: Should override toString in ZoneConnection to print name/keyName/primaryServer. Omit key. See toString override in Zone for example. Validate that Zone toString has the connection in this format.

Aravindh-Raju:
Fixes #408 

Changes in this pull request:
- Excluded zone connection key and algorithm to prevent accidental logging of sensitive info by overriding `toString` in ZoneConnection. Validated that it doesn't show up in any log statements and also that Zone `toString` has the zone connection in this format.

### DIFF
--- a/modules/api/src/test/scala/vinyldns/api/domain/zone/ZoneConnectionValidatorSpec.scala
+++ b/modules/api/src/test/scala/vinyldns/api/domain/zone/ZoneConnectionValidatorSpec.scala
@@ -245,8 +245,8 @@ class ZoneConnectionValidatorSpec
     "Zone Connection toString" should {
       "not display key and algorithm" in {
         zc.toString shouldBe "ZoneConnection: [name=\"zc.\"; keyName=\"zc.\"; primaryServer=\"10.1.1.1\"; ]"
-
-        // verify the same while displaying connection and transferConnection in Zone
+      }
+      "not display key and algorithm while displaying connection and transferConnection of a Zone" in {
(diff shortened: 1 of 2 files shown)


nspadaccino (approved):
Looks good

modules/api/src/test/scala/vinyldns/api/domain/zone/ZoneConnectionValidatorSpec.scala
```diff
@@ -241,5 +241,15 @@ class ZoneConnectionValidatorSpec
         underTest.isValidBackendId(Some("bad")) shouldBe left
       }
     }
+
+    "Zone Connection toString" should {
+      "not display key and algorithm" in {
+        zc.toString shouldBe "ZoneConnection: [name=\"zc.\"; keyName=\"zc.\"; primaryServer=\"10.1.1.1\"; ]"
+
+        // verify the same while displaying connection and transferConnection in Zone
+        val zoneString = s"""Zone: [id="${testZone.id}"; name="vinyldns."; account="system"; adminGroupId="system"; status="Active"; shared="false"; connection="Some(ZoneConnection: [name="vinyldns."; keyName="vinyldns."; primaryServer="10.1.1.1"; ])"; transferConnection="Some(ZoneConnection: [name="vinyldns."; keyName="vinyldns."; primaryServer="10.1.1.1"; ])"; reverse="false"; isTest="false"; created="${testZone.created}"; ]"""
```
Jay07GIT, line 250: This test is not related to `"not display key and algorithm"` tests. Can you please add this as a separate one.

eg : `"displaying connection and transferConnection"`

Aravindh-Raju: Made the change in latest commit.

nspadaccino (approved):
lgtm

Jay07GIT:
LGTM.

Jay07GIT (approved):
LGTM. Can Merge to Main.

PR #1172, opened 2022-09-01
Correctly handle SPF record data 

### LINKED ISSUE #570: SPF records exceeding 255 characters do not fail properly
**Describe the bug**
When creating an SPF record with too many characters, the record fails in processing (an UPDATE is issued) as opposed to failing up front on record submission. This is the same issue encountered for TXT records in #563 

**VinylDNS Version**
8.0.0

**To Reproduce**

Go to a zone
Click on Create RecordSet
Choose SPF record
Enter any name for the record name
Enter the following long value for the record data...
Click the Save button
Record is submitted, and fails in verification
```
This is an example of a really long string that I wrote let us see how many characters can actually flow through to the backend dns server before the entire thing fails and blows up maybe in processing maybe up front but that really doesnt matter as the most important point is that we have to exceed 300 characters in the entire string so that we can try and force some kind of issue to happen in the backend itself and 65000 characters is indeed a lot of characters I dont think I will get that far but let us see
```

Expected behavior
VinylDNS should split the very long string into 255 character chunks. The entire string has to be less than 65000 characters (including extra quotes and spaces).

The very long string above should be processed properly, and return as a list of 255 character strings instead.

The dnsjava library can accept a String[] array during creation. We should use that constructor for SPFRecord instead.

Aravindh-Raju:
Fixes #570 

Changes in this pull request:
- Made changes similar to the PR: #568 which fixes a similar issue for TXT records. Group SPF record data as a list of 255 character length strings.

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/backend/dns/DnsConversions.scala
+++ b/modules/api/src/main/scala/vinyldns/api/backend/dns/DnsConversions.scala
@@ -278,7 +278,7 @@ trait DnsConversions {
 
   def fromSPFRecord(r: DNS.SPFRecord, zoneName: DNS.Name, zoneId: String): RecordSet =
     fromDnsRecord(r, zoneName, zoneId) { data =>
-      List(SPFData(data.getStrings.asScala.mkString(",")))
+      List(SPFData(data.getStrings.asScala.mkString))
     }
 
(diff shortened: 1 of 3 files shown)


nspadaccino (approved):
Tested locally, I was able to create SPF records with record data over 255 chars in length. Looks good

Jay07GIT (approved):
This issue working as expected in API. Looks good to me. Can be merge.

PR #1177, opened 2022-09-09
Remove portal logs from browser console

### LINKED ISSUE #1176: Turn off logging for portal
We currently display the `logs` added at the portal module in production too.  It may not be a great idea to expose/display log messages in the browser console at production.

Solution:
- Replace the `$log.log` with `$log.debug` and then turn off the logging by adding the [$logProvider](https://docs.angularjs.org/api/ng/provider/$logProvider) provider in the config directive of AngularJS. We can turn it on when needed for local development and testing purpose to see the logs.

Aravindh-Raju:
Fixes #1176 

Changes in this pull request:
-  Replaced `$log.log` with `$log.debug` to make use of the `$logProvider.debugEnabled()` method to disable portal logs in production and other environments.

### DIFF
--- a/modules/portal/public/app.js
+++ b/modules/portal/public/app.js
@@ -13,7 +13,7 @@ angular.module('vinyldns', [
             });
         $animateProvider
             .classNameFilter(/toshow/);
-        //turning off $log
+        // turning off $log. Change to true for local development and testing
         $logProvider.debugEnabled(false);
     })
     .controller('AppController', function ($scope, $timeout, profileService, utilityService) {
--- a/modules/portal/public/lib/controllers/controller.zones.js
+++ b/modules/portal/public/lib/controllers/controller.zones.js
@@ -70,7 +70,7 @@ angular.module('controller.zones', [])
     });
 
     $scope.canAccessGroup = function(groupId) {
-         return $scope.myGroupIds !== "undefined" &&  $scope.myGroupIds.indexOf(groupId) > -1;
+         return $scope.myGroupIds !== undefined &&  $scope.myGroupIds.indexOf(groupId) > -1;
     };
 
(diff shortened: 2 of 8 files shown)


modules/portal/public/lib/controllers/controller.zones.js
```diff
@@ -70,7 +70,7 @@ angular.module('controller.zones', [])
     });
 
     $scope.canAccessGroup = function(groupId) {
-         return $scope.myGroupIds !== "undefined" &&  $scope.myGroupIds.indexOf(groupId) > -1;
+         return $scope.myGroupIds !== undefined &&  $scope.myGroupIds.indexOf(groupId) > -1;
```
Aravindh-Raju, line 73: Changed the condition here as comparing with `"undefined"`(string undefined) will always be `true` and results in an error while using `indexOf`. It maybe added as a string by mistake.

Jay07GIT (approved):
LGTM.

nspadaccino (approved):
Looks good

modules/portal/public/app.js
```diff
@@ -13,7 +13,7 @@ angular.module('vinyldns', [
             });
         $animateProvider
             .classNameFilter(/toshow/);
-        //turning off $log
+        // turning off $log. Change to true for local development and testing
```
JoshSEdwards, line 16: Everything looks good, and this PR definitely removes logging from the browser console. Updating this to true for debugging doesn't seem to reenable it though

JoshSEdwards (dismissed):
Approving, as long as we don't foresee needing the enable logging in the browser console for testing.

arpit4ever (approved):
LGTM

arpit4ever (approved):
LGTM

PR #1178, opened 2022-09-09
Restrict IPV4 address as CNAME in DNS changes.

### LINKED ISSUE #843: Disallow IPv4 addresses as CNAME record data in batch change interface
We currently allow users to submit IP addresses that are formatted as fully-qualified domain names (FQDNs) in CNAME record data, which is incorrect since mappings from a record name → IP address should be an A record.

Example of bad data:
```
localhost.some.zone  IN  CNAME  127.0.0.1
```

We need to:
- Add an extra validation in our `BatchChangeValidations` to verify that the `cname` data is _not_ simply an IP address with a terminating dot. Note: A regex like `^[0-9]\.[0-9]\.[0-9]+\.[0-9+]$` may be useful.
- Add a new error message indicating that IP addresses are not valid CNAME targets

-----
Good candidates for files to change:
- Adding new validation: `BatchChangeValidations.scala`
- Adding new error message: `DomainValidationErrors.scala`
- Adding unit tests: `BatchChangeValidationsSpec.scala`
- Adding end-to-end tests: `create_batch_change_test.py`

Jay07GIT:
Fixes #843.

Changes in this pull request:
- Restrict IPV4 address as CNAME in batch changes interface.

### DIFF
--- a/modules/api/src/test/scala/vinyldns/api/domain/DomainValidationsSpec.scala
+++ b/modules/api/src/test/scala/vinyldns/api/domain/DomainValidationsSpec.scala
@@ -79,24 +79,23 @@ class DomainValidationsSpec
 
   property("Shortest fqdn name should be valid") {
     val fqdn = Fqdn("a.")
-    validateCName(fqdn) shouldBe valid
+    validateCname(fqdn, false) shouldBe valid
   }
 
   property("Ip address in cname should be invalid") {
     val fqdn = Fqdn("1.2.3.4")
-    println(validateCName(fqdn))
-    validateCName(fqdn) shouldBe invalid
+    validateCname(fqdn, false) shouldBe invalid
   }
 
   property("Longest fqdn name should be valid") {
     val fqdn = Fqdn(("a" * 50 + ".") * 5)
-    validateCName(fqdn) shouldBe valid
+    validateCname(fqdn, false) shouldBe valid
   }
 
   property("fqdn name should pass property-based testing") {
     forAll(domainGenerator) { domain: String =>
       val domains= Fqdn(domain)
-      whenever(validateCName(domains).isValid) {
+      whenever(validateHostName(domains).isValid) {
         domains.fqdn.length should be > 0
         domains.fqdn.length should be < 256
         (domains.fqdn should fullyMatch).regex(validFQDNRegex)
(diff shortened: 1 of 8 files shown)


modules/api/src/test/scala/vinyldns/api/domain/DomainValidationsSpec.scala
```diff
@@ -78,6 +78,55 @@ class DomainValidationsSpec
     validateHostName("asterisk.domain*.name.") shouldBe invalid
   }
 
+  property("Shortests fqdn name should be valid") {
```
nspadaccino, line 81: typo, "Shortests" should be "Shortest"

nspadaccino (commented):
Works locally for batch changes and DNS Changes through the portal. But I did notice that I was able to make a cname with an ipv4 address as the record data via the Manage Records tab in a zone, so I think it would be good to restrict that as well.

Jay07GIT:
> record

I noticed the same. Thought to work on new PR for zone page record set. 
Thanks for the suggestion.

modules/api/src/test/scala/vinyldns/api/domain/DomainValidationsSpec.scala
Jay07GIT: Corrected typo.

Aravindh-Raju (commented):
Same as what Nick suggested. Looks good to me other than that!

nspadaccino (dismissed):
Looks good!

JoshSEdwards (dismissed):

Aravindh-Raju (approved):
Checked it locally and it works as expected. Looks good to me!

modules/core/src/main/scala/vinyldns/core/domain/DomainValidationErrors.scala
```diff
@@ -52,6 +52,11 @@ final case class InvalidDomainName(param: String) extends DomainValidationError
       "joined by dots, and terminated with a dot."
 }
 
+final case class InvalidIPv4CName(param: String) extends DomainValidationError {
+  def message: String =
+    s"""Invalid Cname: "$param", valid cname should not be IP address"""
```
Aravindh-Raju, line 57: A small suggestion, can we update the error message to say it's for `CNAME record data`?
Like:
```
Invalid CNAME record data: "$param", valid CNAME record data should not be IP address.
```

nspadaccino (dismissed):
Something I noticed: While the validation works with the POST /zones/batchrecordchanges route, there isn't validation on the POST /zones/<zoneID>/recordsets route, so I was able to use that to create a CNAME with an IPv4 record data. I think we should have the check on that route as well.

modules/core/src/main/scala/vinyldns/core/domain/DomainValidationErrors.scala
nspadaccino: I second Aravindh's suggestion

arpit4ever (approved):
LGTM

modules/core/src/main/scala/vinyldns/core/domain/DomainValidationErrors.scala
Jay07GIT: updated the error message

nspadaccino:
> Something I noticed: While the validation works with the POST /zones/batchrecordchanges route, there isn't validation on the POST /zones//recordsets route, so I was able to use that to create a CNAME with an IPv4 record data. I think we should have the check on that route as well.

We can have this in a new PR

nspadaccino (approved):

PR #1184, opened 2022-09-19
restrict IPv4 as record data in Cname validation

Jay07GIT:
Fixes #.

Changes in this pull request:
- Restrict IPv4 as record data in Cname validation.

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetValidations.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetValidations.scala
@@ -20,13 +20,14 @@ import cats.syntax.either._
 import vinyldns.api.Interfaces._
 import vinyldns.api.backend.dns.DnsConversions
 import vinyldns.api.config.HighValueDomainConfig
+import vinyldns.api.domain.DomainValidations.validateIpv4Address
 import vinyldns.api.domain._
 import vinyldns.core.domain.DomainHelpers._
 import vinyldns.core.domain.record.RecordType._
 import vinyldns.api.domain.zone._
 import vinyldns.core.domain.auth.AuthPrincipal
 import vinyldns.core.domain.membership.Group
-import vinyldns.core.domain.record.{RecordType, RecordSet}
+import vinyldns.core.domain.record.{RecordSet, RecordType}
 import vinyldns.core.domain.zone.Zone
 import vinyldns.core.Messages._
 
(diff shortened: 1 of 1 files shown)


JoshSEdwards (commented):
I don't have approval permissions, but the change looks good to me!

modules/api/src/test/scala/vinyldns/api/domain/record/RecordSetValidationsSpec.scala
```diff
@@ -413,6 +413,10 @@ class RecordSetValidationsSpec
         val error = leftValue(cnameValidations(invalid, List(), okZone))
         error shouldBe an[InvalidRequest]
       }
+      "return an InvalidRequest if a cname record set name is IPv4 address" in {
+        val error = leftValue(cnameValidations(cname.copy(name = "1.2.3.4"), List(), okZone))
+        error shouldBe an[InvalidRequest]
+      }
```
Aravindh-Raju, line 419: The test doesn't verify the error that was added to check for IP address in cname `record data`. It checks for cname `record name`. Can you please update the test as follows:
```
      "return an RecordSetValidation if a cname record data is an IPv4 address" in {
        val error = leftValue(cnameValidations(cname.copy(records = List(CNAMEData(Fqdn("1.2.3.4")))), List(), okZone))
        error shouldBe an[RecordSetValidation]
      }
```

modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetValidations.scala
```diff
@@ -164,6 +165,15 @@ object RecordSetValidations {
         noConsecutiveDots(newRecordSet.records.head.toString)
       )
     }
+    val isNotIPv4inCname = {
+      ensuring(
+        RecordSetValidation(
+          s"""Invalid CNAME: ${newRecordSet.records.head.toString.dropRight(1)}, valid cname cannot not be an IP address"""
```
Aravindh-Raju, line 171: Just a suggestion. Can we update the error message as the following:
```
Invalid CNAME record data: ${newRecordSet.records.head.toString.dropRight(1)}, valid CNAME record data cannot be an IP address.
```
This makes it more clear that the error message was for CNAME `record data` and not for record name.

Aravindh-Raju (changes requested):
Just a few changes. Looks good otherwise!

modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetValidations.scala
Jay07GIT: Updated the error message.

modules/api/src/test/scala/vinyldns/api/domain/record/RecordSetValidationsSpec.scala
Jay07GIT: Updated the tests.

nspadaccino (approved):
LGTM

nspadaccino (changes requested):
One of the merge commits is unsigned

Jay07GIT:
> One of the merge commits is unsigned

I rebased the commit with signed. Can you try to merge now?

Jay07GIT:
Opened new [PR](https://github.com/vinyldns/vinyldns/pull/1234) due to unsign issue in old commit.

PR #1187, opened 2022-09-23
Allow dotted hosts creation using config

Aravindh-Raju:
Changes in this pull request:
- Added a config with settings which are used to allow the creation of dotted hosts.
- Details about the config can be found in `modules/docs/src/main/mdoc/operator/config-api.md`.

### DIFF
--- a/modules/api/src/main/resources/reference.conf
+++ b/modules/api/src/main/resources/reference.conf
@@ -92,18 +92,16 @@ vinyldns {
 
   # approved zones, individual users, users in groups and record types that are allowed for dotted hosts
   dotted-hosts = {
-      zone-list = [
-      "parent.com*" # for local testing with wildcard zones (allows parent.com and child.parent.com zones)
-      ]
-      allowed-user-list = [
-      "ok" # for local testing (username)
-      ]
-      allowed-group-list = [
-      "dummy-group*" # for local testing (group name)
-      ]
-      allowed-record-type = [
-      "CNAME" # for local testing
-      ]
+    allowed-settings = [
+      {
+      # for wildcard zones. Settings will be applied to all matching zones
+      type = "auth-configs"
+      zone = "*ent.com*."
+      allowed-user-list = ["ok"]
+      allowed-group-list = ["dummy-group*"]
+      allowed-record-type = ["CNAME"]
+      }
+    ]
   }
 
   # color should be green or blue, used in order to do blue/green deployment
--- a/modules/api/src/main/resources/reference.conf
+++ b/modules/api/src/main/resources/reference.conf
@@ -90,20 +90,19 @@ vinyldns {
     "ns1.parent.com."
   ]
 
-  # approved zones/domains and users that are allowed to create dotted hosts
+  # approved zones, individual users, users in groups and record types that are allowed for dotted hosts
   dotted-hosts = {
       zone-list = [
-      "dummy.", # for local testing
-      "*ent.com", # for local testing with wildcard zones (allows parent.com and child.parent.com zone)
+      "parent.com*" # for local testing with wildcard zones (allows parent.com and child.parent.com zones)
       ]
       allowed-user-list = [
-      "testuser" # for local testing
+      "ok" # for local testing (username)
       ]
       allowed-group-list = [
-      "test-group" # for local testing
+      "dummy-group*" # for local testing (group name)
       ]
       allowed-record-type = [
-      "A" # for local testing
+      "CNAME" # for local testing
       ]
   }
 
--- a/modules/api/src/main/resources/reference.conf
+++ b/modules/api/src/main/resources/reference.conf
@@ -100,6 +100,7 @@ vinyldns {
       allowed-user-list = ["ok"]
       allowed-group-list = ["dummy-group"]
       allowed-record-type = ["CNAME"]
+      allowed-dots-limit = 3
       },
       {
       # for wildcard zones. Settings will be applied to all matching zones
@@ -107,6 +108,7 @@ vinyldns {
       allowed-user-list = ["sharedZoneUser"]
       allowed-group-list = ["history-group1"]
       allowed-record-type = ["A"]
+      allowed-dots-limit = 3
       }
     ]
   }
--- a/modules/api/src/main/resources/reference.conf
+++ b/modules/api/src/main/resources/reference.conf
@@ -97,18 +97,18 @@ vinyldns {
       {
       # for wildcard zones. Settings will be applied to all matching zones
       zone = "*ent.com*."
-      allowed-user-list = ["ok"]
-      allowed-group-list = ["dummy-group"]
-      allowed-record-type = ["CNAME"]
-      allowed-dots-limit = 3
+      user-list = ["ok"]
+      group-list = ["dummy-group"]
+      record-types = ["CNAME"]
+      dots-limit = 3
       },
       {
       # for wildcard zones. Settings will be applied to all matching zones
       zone = "dummy*."
-      allowed-user-list = ["sharedZoneUser"]
-      allowed-group-list = ["history-group1"]
-      allowed-record-type = ["A"]
-      allowed-dots-limit = 3
+      user-list = ["sharedZoneUser"]
+      group-list = ["history-group1"]
+      record-types = ["A"]
+      dots-limit = 3
       }
     ]
   }
--- a/modules/api/src/main/resources/reference.conf
+++ b/modules/api/src/main/resources/reference.conf
@@ -90,7 +90,7 @@ vinyldns {
     "ns1.parent.com."
   ]
 
-  # approved zones, individual users, users in groups and record types that are allowed for dotted hosts
+  # approved zones, individual users, users in groups, record types and no.of.dots that are allowed for dotted hosts
   dotted-hosts = {
     # for local testing
     allowed-settings = [
--- a/modules/api/src/main/resources/reference.conf
+++ b/modules/api/src/main/resources/reference.conf
@@ -99,6 +99,12 @@ vinyldns {
       allowed-user-list = [
       "testuser" # for local testing
       ]
+      allowed-group-list = [
+      "test-group" # for local testing
+      ]
+      allowed-record-type = [
+      "A" # for local testing
+      ]
   }
 
   # color should be green or blue, used in order to do blue/green deployment
--- a/modules/api/src/main/resources/reference.conf
+++ b/modules/api/src/main/resources/reference.conf
@@ -90,10 +90,16 @@ vinyldns {
     "ns1.parent.com."
   ]
 
-  # approved zones/domains that are allowed to create dotted hosts
-  zone-list = [
-    "ok." # for local testing
-  ]
+  # approved zones/domains and users that are allowed to create dotted hosts
+  dotted-hosts = {
+      zone-list = [
+      "dummy.", # for local testing
+      "*ent.com", # for local testing with wildcard zones (allows parent.com and child.parent.com zone)
+      ]
+      allowed-user-list = [
+      "testuser" # for local testing
+      ]
+  }
 
   # color should be green or blue, used in order to do blue/green deployment
   color = "green"
--- a/modules/api/src/main/resources/reference.conf
+++ b/modules/api/src/main/resources/reference.conf
@@ -96,7 +96,6 @@ vinyldns {
     allowed-settings = [
       {
       # for wildcard zones. Settings will be applied to all matching zones
-      type = "auth-configs"
       zone = "*ent.com*."
       allowed-user-list = ["ok"]
       allowed-group-list = ["dummy-group"]
@@ -104,7 +103,6 @@ vinyldns {
       },
       {
       # for wildcard zones. Settings will be applied to all matching zones
-      type = "auth-configs"
       zone = "dummy*."
       allowed-user-list = ["sharedZoneUser"]
       allowed-group-list = ["history-group1"]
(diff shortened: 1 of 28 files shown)


Jay07GIT:
On creating the dotted hosts A record in the zones, I could see weird error message Can you please update this.

Steps to reproduce: 
1. Create dummy. zone.
2. Create dotted hots with the record `ok.ok.` ->  record name, record type ->  A, Ip Address -> 9.5.6.7

image

Aravindh-Raju:
> Can you please update this message. image

Updated the message, only indicates that the record is a dotted host now and removed the other contents.

Aravindh-Raju:
> On creating the dotted hosts A record in the zones, I could see weird error message Can you please update this.
> 
> Steps to reproduce:
> 
> 1. Create dummy. zone.
> 2. Create dotted hots with the record `ok.ok.` ->  record name, record type ->  A, Ip Address -> 9.5.6.7
> 
> image

Thanks for finding that out. It was because of the dot at the end of the record name. Made the required changes. Works as expected now.

nspadaccino (commented):
Couple more things I noticed:
• Currently there seems to be no limit on how many dots could be in a dotted host. For example I could create a record "foo.foo.foo.foo.foo." in zone "ok." Depending on requirements, maybe we could have this be a configurable value
• In the case of wildcard zones, we could have conflicts in the access rules. For example, if we have dotted host config with zone = "*ent.com." and config with zone = "child.parent.com.", which would take precedence for zone "child.parent.com." ? Does the order of the config matter?
• It seems like if allowed-group-list and allowed-user-list are both left empty, any user can create a dotted record in the zone. Is this what we want, or should we allow nobody if those fields are both blank? 
• It looks like if allowed-record-type is left empty, users can only create A records, while other record types fail, which doesn't seem like the intended behavior

modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetValidations.scala
```diff
@@ -90,6 +90,68 @@ object RecordSetValidations {
...
+  def isDotted(
+     newRecordSet: RecordSet,
+     zone: Zone,
+     existingRecordSet: Option[RecordSet] = None,
+     recordFqdnDoesNotExist: Boolean,
+     isRecordTypeAndUserAllowed: Boolean
+  ): Either[Throwable, Unit] =
+    ensuring(
+      InvalidRequest(
+        s"Record with fqdn '${newRecordSet.name}.${zone.name}' cannot be created. " +
```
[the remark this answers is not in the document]
Aravindh-Raju: Steps to create this exception:
1. Add dotted hosts config for the zone as `"*ent.com"`.
2. Create two zones `child.parent.com` and `parent.com`. Create a dotted host record in `child.parent.com` zone for (eg) record name = `test.dotted`. Then create a new dotted host record in `parent.com` zone with the same name `test.dotted.child`. This will throw the above error as the record with same type and FQDN already exist in `child.parent.com` zone.

modules/api/src/test/scala/vinyldns/api/domain/record/RecordSetValidationsSpec.scala
```diff
@@ -184,34 +186,82 @@ class RecordSetValidationsSpec
...
+        val test = aaaa.copy(name = "this.is.a.failure.")
+        leftValue(isDotted(test, okZone, None, false, true)) shouldBe an[InvalidRequest]
+      }
+
+      "return a failure for any record that is a dotted host if user or record type is not allowed" in {
+        val test = aaaa.copy(name = "this.is.a.failure." + okZone.name)
+        leftValue(isDotted(test, okZone, None, true, false)) shouldBe an[InvalidRequest]
+      }
+
+      "return a failure for a dotted record name that matches the zone name" in {
```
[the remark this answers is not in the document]
Aravindh-Raju: I should’ve removed that test case. We should allow creation of records with record name same as the zone name. We already do that before these changes which we can see when we run locally. There is no validation not to be the same. So, I think we can allow that as before. I'll remove the test case as it wasn’t necessary and added accidentally .

modules/docs/src/main/mdoc/operator/config-api.md
```diff
@@ -536,7 +536,57 @@ v6-discovery-nibble-boundaries {
...
+}
+```
+
+In the above, the dotted hosts can be created only in the zone `dummy.` and zones matching `*ent.com.` (parent.com., child.parent.com.)
+
+Also, it must satisfy the allowed users or group users and record type of the respective zone to create a dotted host.
+
+For eg, we can't create a dotted host with `CNAME` record type in the zone `dummy.` as it's not in `allowed-record-type`.
+And the user `professor` can't create a dotted host in the zone `dummy.` as the user is not in `allowed-user-list` or 
+`allowed-group-list`.
```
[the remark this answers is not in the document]
Aravindh-Raju: Yes. I've made the changes in the latest commit. Added about not being part of `dummy-group`. Please have a look.

modules/api/src/test/functional/tests/recordsets/create_recordset_test.py
```diff
@@ -524,8 +526,35 @@ def test_create_dotted_a_record_not_apex_fails(shared_zone_test_context):
 
     zone_name = shared_zone_test_context.parent_zone["name"]
     error = client.create_recordset(dotted_host_a_record, status=422)
-    assert_that(error, is_("Record with name " + dotted_host_a_record["name"] + " and type A is a dotted host which "
-                                                                                "is not allowed in zone " + zone_name))
+    assert_that(error, is_("Record type is not allowed or the user is not authorized to create a dotted host in the "
+                           "zone '" + zone_name + "'"))
+
+
+def test_create_dotted_a_record_succeeds_if_all_dotted_hosts_config_satisfies(shared_zone_test_context):
```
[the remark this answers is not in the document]
Aravindh-Raju: Made the change.

modules/api/src/test/functional/tests/recordsets/create_recordset_test.py
```diff
@@ -581,22 +610,52 @@ def test_create_dotted_a_record_apex_with_trailing_dot_succeeds(shared_zone_test
...
+        "ttl": 500,
+        "records": [{"cname": "foo.bar."}]
+    }
+
+    error = client.create_recordset(dotted_host_cname_record, status=422)
+    assert_that(error, is_("Record type is not allowed or the user is not authorized to create a dotted host in the "
+                           "zone '" + zone["name"] + "'"))
+
+
+def test_create_dotted_cname_record_succeeds_if_all_dotted_hosts_config_satisfies(shared_zone_test_context):
```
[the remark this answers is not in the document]
Aravindh-Raju: Made the change.

modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetService.scala
```diff
@@ -169,6 +195,136 @@ class RecordSetService(
...
+      else {
+        false
+      }
+    }
+    else{
+      false
+    }
+  }
+
+  // Check if user is allowed to create dotted hosts using the users present in dotted hosts config
```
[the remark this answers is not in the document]
Aravindh-Raju: Made the change.

Aravindh-Raju:
> Couple more things I noticed:
> 
> • Currently there seems to be no limit on how many dots could be in a dotted host. For example I could create a record "foo.foo.foo.foo.foo." in zone "ok." Depending on requirements, maybe we could have this be a configurable value
> 
> • In the case of wildcard zones, we could have conflicts in the access rules. For example, if we have dotted host config with zone = "*ent.com." and config with zone = "child.parent.com.", which would take precedence for zone "child.parent.com." ? Does the order of the config matter?
> 
> • It seems like if allowed-group-list and allowed-user-list are both left empty, any user can create a dotted record in the zone. Is this what we want, or should we allow nobody if those fields are both blank?
> 
> • It looks like if allowed-record-type is left empty, users can only create A records, while other record types fail, which doesn't seem like the intended behavior

1. Sure, I'll add a config for number of dots allowed for each zones in the `dotted-host-config` and update the PR.
2. No, the order of the config doesn’t matter. In the end, we group all the zones together (both wildcard and individual) and process them. So, the order doesn't matter. But I noted that, currently if a zone is present in both `wildcard` and as `individual` in the `dotted-host-config`, it combines both of their properties instead of just taking the individual zone in precedence. Will look into it.
3. No, it doesn't allow any user to create a dotted record in the zone if `allowed-group-list` and `allowed-user-list` are left empty. I checked it once again today and if they're left empty, it didn’t allow anyone to create a dotted host record in that zone, which is the intended behaviour. Did it allow any user to create dotted host record for you if left empty? Can you please try again once more with latest commit?
4. Same as above. If `allowed-record-type` is left empty, user cannot create dotted host record of any type. It works fine when I ran locally and I'm able to create record of other types without any problem. Not sure what the problem is.

modules/api/src/main/scala/vinyldns/api/config/DottedHostsConfig.scala
```diff
@@ -0,0 +1,31 @@
+/*
+ * Copyright 2018 Comcast Cable Communications Management, LLC
```
deepakmohanakrishnan07, line 2: Shall we change the copyright to have 2022 year in it?

modules/docs/src/main/mdoc/operator/config-api.md
```diff
@@ -551,6 +551,7 @@ Note the following:
 5. If the user is either in `allowed-user-list` or `allowed-group-list`, they are allowed to create a dotted host. It is
 not necessary for the user to be in both `allowed-user-list` and `allowed-group-list`.
 6. The record types which are allowed while creating a dotted host is added to the `allowed-record-type`.
+7. The number of dots allowed in a record name for a zone is given in `allowed-dots-limit`.
```
nspadaccino, line 554: I think we should specify here that the number of allowed dots is set to zero (i.e. no dotted hosts allowed) if a value is not provided. We should also specify what happens if the other config values are not provided

modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetValidations.scala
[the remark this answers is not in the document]
nspadaccino: Thanks, makes sense. I was able to reproduce the error

nspadaccino:
> 3. No, it doesn't allow any user to create a dotted record in the zone if `allowed-group-list` and `allowed-user-list` are left empty. I checked it once again today and if they're left empty, it didn’t allow anyone to create a dotted host record in that zone, which is the intended behaviour. Did it allow any user to create dotted host record for you if left empty? Can you please try again once more with latest commit?

I tried this again and it works as you said. I probably missed something in my test config

nspadaccino:
> 4. Same as above. If `allowed-record-type` is left empty, user cannot create dotted host record of any type. It works fine when I ran locally and I'm able to create record of other types without any problem. Not sure what the problem is.
 
Tried this again and it works as you said.

modules/docs/src/main/mdoc/operator/config-api.md
Aravindh-Raju: Updated the doc with those details. Please have a look at it.

modules/api/src/main/scala/vinyldns/api/config/DottedHostsConfig.scala
Aravindh-Raju: Contacted the open source team as suggested. Will make the required changes once we hear from them.

modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetValidations.scala
```diff
@@ -278,6 +357,29 @@ object RecordSetValidations {
       .leftMap(errors => InvalidRequest(errors.toList.map(_.message).mkString(", ")))
   }
 
+  def checkAllowedDots(allowedDotsLimit: Int, recordSet: RecordSet, zone: Zone): Either[Throwable, Unit] = {
+    ensuring(
+      InvalidRequest(
+        s"RecordSet with name ${recordSet.name} has more dots than that is allowed in config for this zone " +
```
Jay07GIT, line 363: This Message looks good for dots limit greater than 1. If the dots limits is set to 0, `Record with name "" and type "" is a dotted host which is not allowed in zone ""` could be valid error message.

modules/api/src/main/resources/reference.conf
```diff
@@ -90,6 +90,29 @@ vinyldns {
...
+      zone = "*ent.com*."
+      allowed-user-list = ["ok"]
+      allowed-group-list = ["dummy-group"]
+      allowed-record-type = ["CNAME"]
+      allowed-dots-limit = 3
+      },
+      {
+      # for wildcard zones. Settings will be applied to all matching zones
+      zone = "dummy*."
+      allowed-user-list = ["sharedZoneUser"]
```
Jay07GIT, line 108: A minor change can be do.

In `allowed-user-list` allowed is already mentioned above in config. So the key name can be `user-list`. What do you think?

If it's fine, we can do the similar changes for other key words in dotted hosts config as well.

Aravindh-Raju: Made the changes as suggested. Please have a look.

modules/api/src/main/scala/vinyldns/api/config/DottedHostsConfig.scala
nspadaccino: I think we can update the copyrights in another PR so we can get this merged

nspadaccino (approved):
Looks good to me!

PR #1190, opened 2022-09-28
Dotted hosts portal config

Jay07GIT:
Fixes #.

Changes in this pull request:
- Allow Dotted hosts configuration for zone and ACL using portal.


JoshSEdwards (dismissed):

modules/portal/public/lib/controllers/controller.manageZones.js
```diff
@@ -276,6 +277,8 @@ angular.module('controller.manageZones', [])
             $log.log('recordsService::getZone-success');
             $scope.zoneInfo = response.data.zone;
             $scope.updateZoneInfo = angular.copy($scope.zoneInfo);
+            $log.log('wqewweq::getZone-success', $scope.updateZoneInfo);
+            $log.log('sfsafd::getZone-success', $scope.zoneInfo);
```
Aravindh-Raju, line 281: These 2 log messages seems to be used for testing. Can we remove these or change the messages like (` $log.log('refreshZone:: updateZoneInfo', $scope.updateZoneInfo);`) if it is necessary?

Aravindh-Raju (changes requested):
Few minor changes. Looks good otherwise!

Aravindh-Raju (changes requested):
I think we should also add the new fields `allowDottedHosts`, `allowDottedLimits` in our api docs: https://www.vinyldns.io/api/zone-model.html

PR #1192, opened 2022-10-14
Refactor crypto be strongly typed

### LINKED ISSUE #193: Refactor crypto be strongly typed and enforce encryption at rest
**Motivation**
For existing fields that are encrypted (user.secret and zoneConnection.key), these fields are Stringly typed.  There is no way to understand their disposition when we save / retrieve the values.

We should have compile time / type checks on the disposition of these values so they are handled properly in the code and we do not accidentally surface a value decrypted when it should not be.

**Scope of change**
1. Create a strongly typed value to represent a `String` as being encrypted or decrypted.
1. Update the `ZoneConnection` type to use the new encrypted type
1. Update the `User` type to use the new encrypted type

**Design Notes**

Some sketches from a design perspective...

```scala
// I thought of having a Decrypted here as well, but in reality we should never have the 
// data classes contain decrypted data.  This makes sure we don't accidentally surface decrypted data
// We need a way to pull the decrypted value out of the Encrypted instance though elsewhere
final case class Encrypted private (value: String) extends AnyVal

object Encrypted { 
  def apply(crypto: CryptoAlgebra, value: String): Encrypted = new Encrypted(crypto.encrypt(value))
  def decrypt(crypto: CryptoAlgebra, x: Encrypted): String = crypto.decrypt(x.value)
}
```

Aravindh-Raju:
Fixes #193.

Changes in this pull request:
- Created a strongly typed value to represent a `String` as being encrypted or decrypted.
- Updated the `ZoneConnection` type to use the new encrypted type.
- Updated the `User` type to use the new encrypted type.

### DIFF
--- a/modules/api/src/it/scala/vinyldns/api/domain/record/RecordSetServiceIntegrationSpec.scala
+++ b/modules/api/src/it/scala/vinyldns/api/domain/record/RecordSetServiceIntegrationSpec.scala
@@ -34,7 +34,7 @@ import vinyldns.api.engine.TestMessageQueue
 import vinyldns.mysql.TransactionProvider
 import vinyldns.core.TestMembershipData._
 import vinyldns.core.TestZoneData.testConnection
-import vinyldns.core.domain.{Fqdn, HighValueDomainError}
+import vinyldns.core.domain.{Encrypted, Fqdn, HighValueDomainError}
 import vinyldns.core.domain.auth.AuthPrincipal
 import vinyldns.core.domain.backend.{Backend, BackendResolver}
 import vinyldns.core.domain.membership.{Group, GroupRepository, User, UserRepository}
@@ -63,8 +63,8 @@ class RecordSetServiceIntegrationSpec
 
   private var testRecordSetService: RecordSetServiceAlgebra = _
 
-  private val user = User("live-test-user", "key", "secret")
-  private val user2 = User("shared-record-test-user", "key-shared", "secret-shared")
+  private val user = User("live-test-user", "key", Encrypted("secret"))
+  private val user2 = User("shared-record-test-user", "key-shared", Encrypted("secret-shared"))
   private val group = Group(s"test-group", "test@test.com", adminUserIds = Set(user.id))
   private val group2 = Group(s"test-group", "test@test.com", adminUserIds = Set(user.id, user2.id))
   private val sharedGroup =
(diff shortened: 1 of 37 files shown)


modules/api/src/main/scala/vinyldns/api/route/DnsJsonProtocol.scala
```diff
@@ -129,12 +130,22 @@ trait DnsJsonProtocol extends JsonValidation {
     override def toJson(a: Algorithm): JValue = JString(a.name)
   }
 
+  case object EncryptedSerializer extends ValidationSerializer[Encrypted] {
```
nspadaccino, line 133: Do we not have unit tests for DnsJsonProtocol? It seems like we do for other serializers, I think that would be a good addition

nspadaccino (commented):
Looks good, tested locally via API and portal, I am able authenticate with test creds. One comment about unit tests

modules/api/src/it/scala/vinyldns/api/domain/record/RecordSetServiceIntegrationSpec.scala
```diff
@@ -33,7 +33,7 @@ import vinyldns.api.domain.zone._
 import vinyldns.api.engine.TestMessageQueue
 import vinyldns.mysql.TransactionProvider
 import vinyldns.core.TestZoneData.testConnection
-import vinyldns.core.domain.{Fqdn, HighValueDomainError}
+import vinyldns.core.domain.{Encrypted, Fqdn, HighValueDomainError}
```
JoshSEdwards, line 36: Is this needed? It doesn't seem to be used in this file.

JoshSEdwards (approved):
Looks good to me!

modules/api/src/it/scala/vinyldns/api/domain/record/RecordSetServiceIntegrationSpec.scala
Aravindh-Raju: I haven't completed changes to this file at that point in time. Made the required changes now.

modules/api/src/main/scala/vinyldns/api/route/DnsJsonProtocol.scala
Aravindh-Raju: Added tests for `EncryptedSerializer` to verify it's working. Please have a look.

JoshSEdwards (dismissed):

Jay07GIT (dismissed):
Looks good to me.

JoshSEdwards:
Looks good to me. Tested a bunch of things locally using the portal and the API, and didn't see any weird behavior.

JoshSEdwards (approved):

arpit4ever (dismissed):
Tested locally. Everything looks good.

arpit4ever (approved):
LGTM

PR #1194, opened 2022-10-25
Fix dotted hosts update (PR #1187)

Aravindh-Raju:
Fixes #1187 

Changes in this pull request:
- When we `update` a created dotted host, we currently get an error and doesn’t allow us to `update`. The problem is with the validation which is added to check the conflict while `creating` dotted hosts. That validation is not necessary to be in `update` since we already handle it while `creating` and recordset name cannot be changed during `update`. 
- Removed the validation from `update` now and it works as expected.
- Added tests to verify the working.


nspadaccino (approved):
Looks good to me

JoshSEdwards (approved):
Looks good to me -- I was able to update a dotted host record in my local environment 👍

PR #1195, opened 2022-10-26
Hotfix: Raise Limit of Maximum Items in a Group List to 3000

### LINKED ISSUE #1196: User is missing the last page of groups in zone admin group dropdown
User was trying to change the ownership of a zone, and could not see the last page of groups in the dropdown selector. This is due to the configured limit of 1500 items. We need to increase the limit. This value should be made configurable in the deployment repo.

nspadaccino:
Fixes #1196

Currently we are hitting the 1500 item limit for groups that can be displayed in a dropdown. This fix increases the limit to 3000.


JoshSEdwards (approved):
looks good 👍

arpit4ever (approved):
LGTM

PR #1204, opened 2022-11-15
Added metrics api for zone change failure and record change failure

### LINKED ISSUE #439: Add backend failure metric
We currently do not expose any metrics/health in our system when a record change or zone change fails.  We store the failure in our database, but there is no way to independently monitor that other than querying the database on a regular basis

If a record or zone change fails, report that as an error metric (using monitor like everywhere else).  This way, we can monitor it in real-time via prometheus and detect the failure.

1. Monitor record change failures
2. Monitor zone change failures

Jay07GIT:
Fixes #439.

Changes in this pull request:

> Added metrics api for zone change failure and record change failure.
```
/metrics/health/zonechangesfailure
/metrics/health/recordsetchangesfailure
```

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/record/ListRecordSetChangesResponse.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/record/ListRecordSetChangesResponse.scala
@@ -17,7 +17,7 @@
 package vinyldns.api.domain.record
 
 import vinyldns.api.domain.zone.RecordSetChangeInfo
-import vinyldns.core.domain.record.ListRecordSetChangesResults
+import vinyldns.core.domain.record.{ListRecordSetChangesResults, RecordSetChange}
 
 case class ListRecordSetChangesResponse(
     zoneId: String,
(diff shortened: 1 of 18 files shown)


nspadaccino (approved):

nspadaccino (changes requested):
Looks like the first commit is still unsigned

Jay07GIT:
> Looks like the first commit is still unsigned

Signed all commits.

Aravindh-Raju (approved):
Looks good to me!

nspadaccino (approved):
Looks good!

nspadaccino:
New API endpoints are working as expected. A few thoughts though:
• It looks like both APIs return a list of failed changes without any sort of grouping or filtering. I feel like that would be a very large result if we called recordsetchangesfailure in prod. I was thinking it may make more sense to have these as zone routes, unless that would be counterintuitive to the goal of this PR, which is monitoring.
• If this structure makes the most sense for monitoring, I think we should still have some filtering capability like the global recordset search API https://www.vinyldns.io/api/list-recordsets-global.html

Thoughts?

JoshSEdwards (approved):

Jay07GIT:
It's a good idea to add a filter in API. We will open a new PR for it. Thanks for your idea.

PR #1205, opened 2022-11-17
Make the Number of Groups Displayed in a List a Configurable Value

### LINKED ISSUE #1200: Make Groups maxItems a Configurable Value
Currently the maximum number of groups the portal can display in a dropdown list is a hardcoded value in service.groups.js. We have a config value that should dictate that, which is MEMBERSHIP_ROUTING_MAX_GROUPS_LIST_LIMIT in LimitsConfig.scala, but the value is not used in the frontend. We should pass that value to the groups service.

nspadaccino:
Fixes #1200

Added membership-routing-max-groups-list-limit config value to the portal config. The value is now used in the frontend to manage how many groups to display in the owner group dropdown lists.


modules/portal/app/models/Meta.scala
```diff
@@ -33,6 +34,7 @@ object Meta {
       config.getOptional[Int]("batch-change-limit").getOrElse(1000),
       config.getOptional[Long]("default-ttl").getOrElse(7200L),
       config.getOptional[Boolean]("manual-batch-review-enabled").getOrElse(false),
-      config.getOptional[Boolean]("scheduled-changes-enabled").getOrElse(false)
+      config.getOptional[Boolean]("scheduled-changes-enabled").getOrElse(false),
+      config.getOptional[Int]("api.limits.membership-routing-max-groups-list-limit").getOrElse(2500)
```
Aravindh-Raju, line 38: Just for my clarification, why are we setting the value as `2500` instead of the default `3000` mentioned in the config (similar to that of `batch-change-limit`)?

modules/portal/public/lib/controllers/controller.groups.js
```diff
@@ -28,6 +28,7 @@ angular.module('controller.groups', []).controller('GroupsController', function
     $scope.ignoreAccess = false;
     $scope.hasGroups = false;
     $scope.query = "";
+    $scope.maxGroupItemsDisplay = 2000;
```
Aravindh-Raju, line 31: Should we set the default value of 3000 defined in the config here too, to make it uniform everywhere?

Aravindh-Raju (changes requested):

JoshSEdwards (approved):
LGTM

Jay07GIT (approved):
LGTM

Aravindh-Raju (approved):
LGTM!

arpit4ever (approved):
LGTM

PR #1207, opened 2022-11-21
Dependencies upgraded for both mac M1 and Intel chip

Jay07GIT:
Changes in this pull request:
- Dependencies upgraded for protobuf and Play. This changes will work for both M1 and Intel chips.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -113,8 +113,8 @@ lazy val coreBuildSettings = Seq(
   // do not use unused params as NoOpCrypto ignores its constructor, we should provide a way
   // to write a crypto plugin so that we fall back to a noarg constructor
   scalacOptions ++= scalacOptionsByV(scalaVersion.value).filterNot(_ == "-Ywarn-unused:params"),
-  PB.targets in Compile := Seq(PB.gens.java("2.6.1") -> (sourceManaged in Compile).value),
-  PB.protocVersion := "-v261"
+  PB.targets in Compile := Seq(PB.gens.java("3.21.7") -> (sourceManaged in Compile).value),
+  PB.protocVersion := "-v3.21.7"
 )
 
 lazy val corePublishSettings = Seq(
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -2,7 +2,7 @@ logLevel := Level.Warn
 
 resolvers += "Flyway" at "https://flywaydb.org/repo"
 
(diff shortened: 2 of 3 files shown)


build.sbt
```diff
@@ -116,8 +116,8 @@ lazy val coreBuildSettings = Seq(
   // do not use unused params as NoOpCrypto ignores its constructor, we should provide a way
   // to write a crypto plugin so that we fall back to a noarg constructor
   scalacOptions ++= scalacOptionsByV(scalaVersion.value).filterNot(_ == "-Ywarn-unused:params"),
-  PB.targets in Compile := Seq(PB.gens.java("2.6.1") -> (sourceManaged in Compile).value),
-  PB.protocVersion := "-v261"
+  PB.targets in Compile := Seq(PB.gens.java("3.21.7") -> (sourceManaged in Compile).value),
+  PB.protocVersion := "-v3.21.7"
```
JoshSEdwards, line 120: Recommend removing `-v` from this line, as it results in this warning:

```
[warn] The PB.protocVersion setting no longer requires '-v` prefix. The prefix has been automatically removed. This warning will turn into an error in a future version.
```

project/plugins.sbt
```diff
@@ -2,7 +2,7 @@ logLevel := Level.Warn
 
 resolvers += "Flyway" at "https://flywaydb.org/repo"
 
-addSbtPlugin("com.thesamet" % "sbt-protoc" % "0.99.18")
+addSbtPlugin("com.thesamet" % "sbt-protoc" % "1.0.6")
```
JoshSEdwards, line 5: I think Sivaranjani and I both found that we also needed to up sbt-plugin (on line 23) to version `2.8.18` to get the portal to compile.

JoshSEdwards:
No update to the SBT version in this PR? I can't get it to work unless I change sbt.version to 1.7.2 in project/build.properties

Also, if we're updating the code to add support for M1 chips, should that section be removed from the Developer Guide now?

project/plugins.sbt
Jay07GIT: Could you please reconfirm this . Because i have been tested in M4. Its working good.

JoshSEdwards:
Just retested locally and everything is working well with no additional changes. Branch looks good to me

JoshSEdwards (approved):

PR #1214, opened 2023-01-03
Add zone sync scheduler

### LINKED ISSUE #1203: Allow Zone Syncs to be scheduled/repeating
Some zones go long periods without being synced, which can cause confusion about what records actually exist on the backend. Often we will only sync a zone when a user has an issue with a DNS change due to the zone being out of sync with the backend. 

We could mitigate this issue by allowing zone syncs to be scheduled regularly. We could allow this to be done by superusers via the manage zone tab in the portal, or through deployment configuration. If done through the portal, it could have a similar interface to scheduling DNS changes.

Aravindh-Raju:
Fixes #1203.

Changes in this pull request:
- Allow zone syncs to be scheduled for zones through portal from the `manage zone` tab.

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneProtocol.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneProtocol.scala
@@ -44,6 +44,7 @@ case class ZoneInfo(
                      adminGroupName: String,
                      latestSync: Option[Instant],
                      backendId: Option[String],
+                     recurrenceSchedule: Option[String],
                      accessLevel: AccessLevel
                    )
 
@@ -70,6 +71,7 @@ object ZoneInfo {
       adminGroupName = groupName,
       latestSync = zone.latestSync,
       backendId = zone.backendId,
+      recurrenceSchedule = zone.recurrenceSchedule,
       accessLevel = accessLevel
     )
 }
@@ -90,6 +92,7 @@ case class ZoneSummaryInfo(
                             adminGroupName: String,
                             latestSync: Option[Instant],
                             backendId: Option[String],
+                            recurrenceSchedule: Option[String],
                             accessLevel: AccessLevel
                           )
 
@@ -111,6 +114,7 @@ object ZoneSummaryInfo {
       adminGroupName = groupName,
       latestSync = zone.latestSync,
       zone.backendId,
+      recurrenceSchedule = zone.recurrenceSchedule,
       accessLevel = accessLevel
     )
 }
--- a/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneProtocol.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneProtocol.scala
@@ -45,6 +45,7 @@ case class ZoneInfo(
                      latestSync: Option[Instant],
                      backendId: Option[String],
                      recurrenceSchedule: Option[String],
+                     scheduleRequestor: Option[String],
                      accessLevel: AccessLevel
                    )
 
@@ -72,6 +73,7 @@ object ZoneInfo {
       latestSync = zone.latestSync,
       backendId = zone.backendId,
       recurrenceSchedule = zone.recurrenceSchedule,
+      scheduleRequestor = zone.scheduleRequestor,
       accessLevel = accessLevel
     )
 }
@@ -93,6 +95,7 @@ case class ZoneSummaryInfo(
                             latestSync: Option[Instant],
                             backendId: Option[String],
                             recurrenceSchedule: Option[String],
+                            scheduleRequestor: Option[String],
                             accessLevel: AccessLevel
                           )
 
@@ -115,6 +118,7 @@ object ZoneSummaryInfo {
       latestSync = zone.latestSync,
       zone.backendId,
       recurrenceSchedule = zone.recurrenceSchedule,
+      scheduleRequestor = zone.scheduleRequestor,
       accessLevel = accessLevel
     )
 }
--- a/modules/api/src/main/scala/vinyldns/api/Boot.scala
+++ b/modules/api/src/main/scala/vinyldns/api/Boot.scala
@@ -53,7 +53,7 @@ object Boot extends App {
 
   private val logger = LoggerFactory.getLogger("Boot")
 
-  // Create a ScheduledExecutorService with a single thread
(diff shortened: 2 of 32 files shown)


nspadaccino:
Could you post a screenshot of what the scheduler should look like? Running locally with a superuser account I can see the `Schedule Zone Sync` text bu
Screenshot 2023-02-14 at 11 36 08 AM
t not any input fields in the manage zone tab

Aravindh-Raju:
> Could you post a screenshot of what the scheduler should look like? Running locally with a superuser account I can see the `Schedule Zone Sync` text but not any input fields in the manage zone tab Screenshot 2023-02-14 at 11 36 08 AM

It works as expected for me. As discussed, try clearing browsing data and reload the page to see if it appears. Else, try opening in a different browser like Safari or Firefox to see if it's visible.

Screenshots for reference:

Screenshot 2023-02-15 at 10 13 26 AM
Screenshot 2023-02-15 at 10 10 57 AM
Screenshot 2023-02-15 at 10 12 00 AM
Screenshot 2023-02-15 at 10 12 57 AM
Screenshot 2023-02-15 at 10 13 13 AM
Screenshot 2023-02-15 at 10 13 19 AM

nspadaccino (commented):
Functionality looks good locally, I think we'll need to test longer term schedules in the staging environment though. Some comments on styling:
Screenshot 2023-02-16 at 2 03 10 PM

I think it would make sense for the day, hour, and minute selectors to be dropdowns instead of fixed-size boxes, I think it would looks cleaner and easier to read after making a selection.
It looks like the day, hour, and minute selectors are not in line with the initial selector, they're slightly higher.
Also, I think we should make the box corners angular like the rest of the zone info boxes, instead of being rounded.

Aravindh-Raju:
> Functionality looks good locally, I think we'll need to test longer term schedules in the staging environment though. Some comments on styling: Screenshot 2023-02-16 at 2 03 10 PM
> 
> I think it would make sense for the day, hour, and minute selectors to be dropdowns instead of fixed-size boxes, I think it would looks cleaner and easier to read after making a selection. It looks like the day, hour, and minute selectors are not in line with the initial selector, they're slightly higher. Also, I think we should make the box corners angular like the rest of the zone info boxes, instead of being rounded.

Made all the requested changes (align the selectors, made the boxes square instead of rounded ones) except for making selectors to be made as dropdown. 

I am not able to do that for two reasons:
1. Since a library is used for the scheduling inputs, I cannot change them to dropdown.
2. As they are multi select fields (more than one option can be selected, like schedule for every 5 mins and 10 mins), I'm not sure if we could change that to a dropdown, as drop-downs usually are single select types.

We could schedule zone sync directly through api by setting up any cron expression, if there is any difficulties with scheduling with portal. Maybe we could open a separate PR for this next. We could try changing this portal library (this is the only working on that I've found) or try creating the whole thing ourselves.

JoshSEdwards:
> > Functionality looks good locally, I think we'll need to test longer term schedules in the staging environment though. Some comments on styling: Screenshot 2023-02-16 at 2 03 10 PM
> > I think it would make sense for the day, hour, and minute selectors to be dropdowns instead of fixed-size boxes, I think it would looks cleaner and easier to read after making a selection. It looks like the day, hour, and minute selectors are not in line with the initial selector, they're slightly higher. Also, I think we should make the box corners angular like the rest of the zone info boxes, instead of being rounded.
> 
> Made all the requested changes (align the selectors, made the boxes square instead of rounded ones) in the latest commit, except for making selectors to be made as dropdown.
> 
> I am not able to do that for two reasons:
> 
> 1. Since a library is used for the scheduling inputs, I cannot change them to dropdown.
> 2. As they are multi select fields (more than one option can be selected, like schedule for every 5 mins and 10 mins), I'm not sure if we could change that to a dropdown, as drop-downs usually are single select types.
> 
> We could schedule zone sync directly through api by setting up any cron expression (for eg: `"recurrenceSchedule" = "0/5 0 0 ? * * *"`). So if there is any difficulties with scheduling with portal we can manage it through api. Maybe we could open a separate PR for this next. We could try changing this portal library (this is the only working one that I've found though) or try creating the whole thing ourselves.

I think it definitely looks better than before now that the edges aren't rounded and the boxes are more aligned. This is me being super nitpicky, but it looks like the boxes are still a *tiny* bit misaligned. Maybe one or two pixels too high:

Screen Shot 2023-02-17 at 11 46 47 AM

As for the dropdowns, I think it makes sense to leave them as multi-select boxes. Seems like a lot of extra work rewriting a library just to make a dropdown that allows you to select multiple times when this works fine already. The API is a good idea too though.

nspadaccino:
> We could schedule zone sync directly through api by setting up any cron expression (for eg: `"recurrenceSchedule" = "0/5 0 0 ? * * *"`). So if there is any difficulties with scheduling with portal we can manage it through api. Maybe we could open a separate PR for this next. We could try changing this portal library (this is the only working one that I've found though) or try creating the whole thing ourselves.

Gotcha, that makes sense about the selector boxes, we can leave them as-is then. And I agree with Josh, the API sounds like a good idea

JoshSEdwards:
Tested again this morning and the CSS looks good 👍

JoshSEdwards (approved):

JoshSEdwards (approved):

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneProtocol.scala
```diff
@@ -44,6 +44,8 @@ case class ZoneInfo(
                      adminGroupName: String,
                      latestSync: Option[Instant],
                      backendId: Option[String],
+                     recurrenceSchedule: Option[String],
```
arpit4ever, line 47: @Aravindh-Raju 
Looks like recurrenceSchedule can be any string.
We need to add guard rails around what type of values can be accepted.

Aravindh-Raju: Thanks for finding that out. Added validation for this and now only valid cron strings will be allowed. Else we get the error: `Invalid cron expression. Please enter a valid cron expression in 'recurrenceSchedule'.` Attaching the screenshot for reference.

Screenshot 2023-02-23 at 11 27 36 PM
Screenshot 2023-02-23 at 11 28 37 PM

arpit4ever (approved):
LGTM

nspadaccino (approved):

PR #1219, opened 2023-02-03
Fix batch change validation bug

### LINKED ISSUE #1217: Resolve batch change validation bug
When an add and delete is submitted for the same record in a batch and the record originally does not exist, the RecordDoesNotExist error "validation_errors" is applied to the add record line instead of the delete record line, triggering the error: `["Record \"dummy.net.\" Does Not Exist: cannot delete a record that does not exist."]`. This error makes the batch impossible to approve.

Aravindh-Raju:
Fixes #1217.

Changes in this pull request:
- When user submit add and delete for non-existent record with same record data in the batch, it displays a validation error.
- When user submit add and delete for non-existent record with different record data in the batch, it just creates the record and omits the delete request as changes made for the issue [Requested record deletes with (RecordDoesNotExist) should not prevent batch submission #1138](https://github.com/vinyldns/vinyldns/issues/1138)

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeConverter.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeConverter.scala
@@ -35,9 +35,9 @@ import java.net.InetAddress
 class BatchChangeConverter(batchChangeRepo: BatchChangeRepository, messageQueue: MessageQueue)
     extends BatchChangeConverterAlgebra {
 
-  private val NonExistentRecordDeleteMessage: String = "This record does not exist." +
+  private val nonExistentRecordDeleteMessage: String = "This record does not exist." +
     "No further action is required."
-  private val NonExistentRecordDataDeleteMessage: String = "Record data entered does not exist." +
+  private val nonExistentRecordDataDeleteMessage: String = "Record data entered does not exist." +
     "No further action is required."
   private val failedMessage: String = "Error queueing RecordSetChange for processing"
   private val logger = LoggerFactory.getLogger(classOf[BatchChangeConverter])
@@ -129,7 +129,7 @@ class BatchChangeConverter(batchChangeRepo: BatchChangeRepository, messageQueue:
           change match {
             case _: SingleDeleteRRSetChange if change.recordSetId.isEmpty =>
               // Mark as Complete since we don't want to throw it as an error
-              change.withDoesNotExistMessage(NonExistentRecordDeleteMessage)
+              change.withDoesNotExistMessage(nonExistentRecordDeleteMessage)
             case _ =>
               // Failure here means there was a message queue issue for this change
               change.withFailureMessage(failedMessage)
@@ -142,7 +142,7 @@ class BatchChangeConverter(batchChangeRepo: BatchChangeRepository, messageQueue:
   def storeQueuingFailures(batchChange: BatchChange): BatchResult[Unit] = {
     // Update if Single change is Failed or if a record that does not exist is deleted
     val failedAndNotExistsChanges = batchChange.changes.collect {
-      case change if change.status == SingleChangeStatus.Failed || change.systemMessage.contains(NonExistentRecordDeleteMessage) => change
+      case change if change.status == SingleChangeStatus.Failed || change.systemMessage.contains(nonExistentRecordDeleteMessage) => change
     }
     batchChangeRepo.updateSingleChanges(failedAndNotExistsChanges).as(())
   }.toBatchResult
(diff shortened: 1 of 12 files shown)


nspadaccino (approved):
Looks good to me, verified locally that delete+add with the same record data throws the new validation error, and delete+add with different record data passes

JoshSEdwards:
Maybe I'm doing something wrong since Nick was able to validate it, but shouldn't this batch change throw a validation error based on the statement, "When user submit add and delete for non-existent record with same record data in the batch, it displays a validation error." ?

image

Instead, it just created an A record for `doesntexist.ok`

JoshSEdwards:
Okay, Nick let me know what I was doing wrong, which was that I wasn't entering the IP address in the DeleteRecordSet's record data. Now that I added that, I'm seeing the error, so it looks good to me 👍 

image

JoshSEdwards (approved):

nspadaccino:
I compared a few DNS change operations between this branch and master, here are my results:
Screenshot 2023-02-16 at 12 34 46 PM

The changes for delete + add look good and seem to match the requirements. However I think that the behavior for `Delete record that already exists w/ incorrect record data` may be a bug. 
If I have a record called `foo.ok. -> A -> 3.3.3.3` and issue a delete on `foo.ok. -> A -> 4.4.4.4`, this used to generate an error, whereas now the change goes through and displays as an Update on `foo.ok.` in the zone details view. Is this behavior intended?

Aravindh-Raju:
> I compared a few DNS change operations between this branch and master, here are my results: Screenshot 2023-02-16 at 12 34 46 PM
> 
> The changes for delete + add look good and seem to match the requirements. However I think that the behavior for `Delete record that already exists w/ incorrect record data` may be a bug. If I have a record called `foo.ok. -> A -> 3.3.3.3` and issue a delete on `foo.ok. -> A -> 4.4.4.4`, this used to generate an error, whereas now the change goes through and displays as an Update on `foo.ok.` in the zone details view. Is this behavior intended?

Yes, the behaviour is intended. We ultimately wanted to eliminate all the `RecordDoesNotExist` and `NonExistentRecordData` errors while submitting a `Delete` request as per the issue: [Requested record deletes with (RecordDoesNotExist) should not prevent batch submission](https://github.com/vinyldns/vinyldns/issues/1138), as it has been a blocker for batch changes when it can be handled without an error. So displaying the error is kind of pointless as they wanted the record/record data `deleted` anyways. 

And in the case you mentioned above, it is considered as `update` in `zone details view`  because a record and record data already exist and a delete is submitted for that record with non-existent record data. Only complete removal of a record (when only record name is specified and record data isn't specified) is displayed as `delete` in `zone details view`. Else, for an individual `IP Address` deletion of a record (even if it is non-existent), it is considered as `update`.

If you maybe want to change the `Change Type` for this non-existent data delete to something like `Non-existent Delete` or something like that, I can do it.  But `Update` seems fine to me. Also, if you want to add back that error (`Delete record that already exists w/ incorrect record data`), I'll do that too. Let me know your thoughts on this.

nspadaccino:
> If you maybe want to change the `Change Type` for this non-existent data delete to something like `Non-existent Delete` or something like that, I can do it. But `Update` seems fine to me. Also, if you want to add back that error (`Delete record that already exists w/ incorrect record data`), I'll do that too. Let me know your thoughts on this.

Ok thanks Aravindh, that explanation makes sense! I think we can keep it as an update in that case, but I'm thinking we may want a different message in the Additional Info column to notify the user in case they typed the wrong IP address in the record data or something. We can discuss on monday though to get the rest of the team's thoughts

Aravindh-Raju:
> > If you maybe want to change the `Change Type` for this non-existent data delete to something like `Non-existent Delete` or something like that, I can do it. But `Update` seems fine to me. Also, if you want to add back that error (`Delete record that already exists w/ incorrect record data`), I'll do that too. Let me know your thoughts on this.
> 
> Ok thanks Aravindh, that explanation makes sense! I think we can keep it as an update in that case, but I'm thinking we may want a different message in the Additional Info column to notify the user in case they typed the wrong IP address in the record data or something. We can discuss on monday though to get the rest of the team's thoughts

I have updated the message (`Record data entered does not exist. No further action is required.`) as requested in the `Additional Info` column. Attaching the screenshot for reference:
Screenshot 2023-02-21 at 3 15 08 PM

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeConverter.scala
```diff
@@ -30,11 +30,14 @@ import vinyldns.core.domain.zone.Zone
...
 import vinyldns.core.queue.MessageQueue
+import java.net.InetAddress
 
 class BatchChangeConverter(batchChangeRepo: BatchChangeRepository, messageQueue: MessageQueue)
     extends BatchChangeConverterAlgebra {
 
-  private val notExistCompletedMessage: String = "This record does not exist." +
+  private val nonExistentRecordDeleteMessage: String = "This record does not exist." +
+    "No further action is required."
+  private val nonExistentRecordDataDeleteMessage: String = "Record data entered does not exist." +
```
nspadaccino, line 40: Small nitpick here, but can we have a space after the period in the first string? Same with line 38

nspadaccino (dismissed):
Looks good, one small styling comment

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeConverter.scala
Aravindh-Raju: Made the requested change.

nspadaccino (approved):

arpit4ever (approved):
tested various add/delete record sets. Looks good.

Issue #644, opened 2019-05-22
Allow super users to update record owner group


britneywright:
**Is your feature request related to a problem? Please describe.**
We restricted super users' ability to modify records in zones where they are not in the owner group. We do want to allow super users the ability to assign records to owner groups though.

**Describe the solution you'd like**
Allow a super user to update a record if the only change is the record owner group. They do not have to be a member of the owner group they are assigning to.

**Additional context**

remerle:
This can help us now as we need to work around this limitation by assigning ourselves to the owner group temporarily, then making the mod.

Marking as `ready`

PR #1222, opened 2023-02-20
Allow Super Users to Update Record Owner Group in Shared Zone

nspadaccino:
Copy of https://github.com/vinyldns/vinyldns/pull/1174

Fixes https://github.com/vinyldns/vinyldns/issues/644

Changes in this pull request:
Super users can now update a record's owner group in a shared zone if the owner group is the only attribute being changed. This prevents super users from having to add themselves to a group temporarily to make that update.


arpit4ever (approved):
LGTM

PR #1226, opened 2023-02-27
Fix zone change history bug

### LINKED ISSUE #1225: Bug in zone change history
**Describe the bug**

When we create a group, associate it with a zone, then dissociate it with that zone by assigning another group and finally delete that group, we see the error `Group with ID was not found`. This is because the group doesn't exist anymore and when we try to get the group details from [here](https://github.com/vinyldns/vinyldns/pull/1158/files#diff-45c6ef99b2637a3bb16d1cb9acf492b48cd4409ca931ec5f8f5caadb1c6f4309R351-R362) using the group ID, we get the error. When we cannot get a group with group ID here, that means the group have been deleted. We should handle that case without throwing an error.

Screenshot 2023-02-27 at 11 19 54 AM

Aravindh-Raju:
Fixes #1225.

Changes in this pull request:
- Added the message `Group has been deleted now.` in the table to handle the error.


nspadaccino (approved):
Looks good! Seeing the updated message in the group history tab, no longer seeing an error

PR #1229, opened 2023-03-02
Added ownership transfer validation in API and Portal

### LINKED ISSUE #1218: Add ownership transfer in api for shared zones

We were reached out by a team's automation which creates VMs. When it does create VMs, it creates DNS records via Vinyl API in public DNS zones, which are then owned by his team. If requesters then rename or abandon their VMs, they then have to reach out to the VM team to make those updates on the recordset since they're owned by them. 

Add a record owner group transfer option in api, similar to the one present in portal.

Jay07GIT:
Fixes #1218.

Changes in this pull request:

Added ownership transfer validation in API and Portal

List of Validations added: (Specific to shared zone)
1. User can claim their own group for the unowned record by their own. Group will be Auto-assigned to the record, once user requested.
2. User can request their own group for the record already assigned. Once requested, the request will be send to the existing assigned group members(Approvers). The request will be notified in email.
3. Approver can Approve/Reject the request. 
        1. If the Approver approved the request, the requested group will be assign to the particular record.
        2. If the Approver rejected the request, the existing group assigned will be remains same.
        

Steps to test:

1. Create 2 user. **Professor**, **fry**.
2. Create 2 groups for each members.
3. Create a shared zone. 
4. Now switch the use and try to give request and approve.(consider one as requester and one as approver.)

Not included in this PR:
Ownership transfer history and filter records with ownership transfer status will be complete in separate PR.


Aravindh-Raju (changes requested):
A personal opinion: Instead of adding the group name in the `Ownership Transfer Status` column, I think we can just have Approved, Cancelled, Pending Action and Rejected. We can confirm with others on their thoughts and proceed further on this.

Few things I noticed:
1. When a request is made for an unowned record, it is auto approved and claimed without any approval by anyone. So, I don’t think we need the message “Approved by Admin”, since no one is approving the request.

2. When a request is cancelled by the requestor, it shows “Cancelled by 'current-owner-group-name'“, but it should show the requestor’s group name as he’s the one cancelling. It should show “Cancelled by <requestor-selected-owner-group-name>“.

3. When a request is made, it shows “Pending Review to 'current-owner-group-name'“. I think it would be better if we can change that to “Pending action by 'current-owner-group-name'“.

4. When a request is approved by the current owner, it shows “Approved by 'new-owner-group-name'“, but it should show the approver’s group name as he’s the one approving. It should show “Approved by 'approver-owner-group-name'“.

5. When a request is made, it goes to Pending action state. Then when a superuser update the requested group for that record instead of the actual approver, it still shows “Approved by 'new-owner-group-name'“. I think in this case it should say “Approved by admin”, similar to when we request for an unowned record.

5. This is a doubt. When we approve or reject or cancel a record, the corresponding ownership transfer status stays unless an update is made to that record?

Jay07GIT:
> A personal opinion: Instead of adding the group name in the `Ownership Transfer Status` column, I think we can just have Approved, Cancelled, Pending Action and Rejected. We can confirm with others on their thoughts and proceed further on this.
> 
> Few things I noticed:
> 
> 1. When a request is made for an unowned record, it is auto approved and claimed without any approval by anyone. So, I don’t think we need the message “Approved by Admin”, since no one is approving the request.
> 2. When a request is cancelled by the requestor, it shows “Cancelled by 'current-owner-group-name'“, but it should show the requestor’s group name as he’s the one cancelling. It should show “Cancelled by 'requestor-selected-owner-group-name'“.
> 3. When a request is made, it shows “Pending Review to 'current-owner-group-name'“. I think it would be better if we can change that to “Pending action by 'current-owner-group-name'“.
> 4. When a request is approved by the current owner, it shows “Approved by 'new-owner-group-name'“, but it should show the approver’s group name as he’s the one approving. It should show “Approved by 'approver-owner-group-name'“.
> 5. When a request is made, it goes to Pending action state. Then when a superuser update the requested group for that record instead of the actual approver, it still shows “Approved by 'new-owner-group-name'“. I think in this case it should say “Approved by admin”, similar to when we request for an unowned record.
> 6. This is a doubt. When we approve or reject or cancel a record, the corresponding ownership transfer status stays unless an update is made to that record?

Updated the PR with the above comments.

arpit4ever:
Closing this PR due to unsigned commit on this one.
New PR: https://github.com/vinyldns/vinyldns/pull/1348

PR #1233, opened 2023-03-09
Add config for zone sync scheduler

### LINKED ISSUE #1232: Fix automated sync on every api servers
**Describe the bug**

- Currently the zone sync scheduler that was added recently performs automated sync on every server/instance when there is more than one server configured for the api. Add a config to limit the automated sync to be performed at a single api server/instance.

Aravindh-Raju:
Fixes #1232.

Changes in this pull request:
- Add a config for zone sync scheduler to perform automated sync only on a single api server/instance instead of being done on all the api servers/instances.


JoshSEdwards (approved):
LGTM

Tested locally and it worked fine. Need to test this in the dev environment to be sure though

nspadaccino (approved):
Looks good

PR #1234, opened 2023-03-15
Restrict IPv4 as Cname in manage records

### LINKED ISSUE #1181: Disallow IPv4 addresses as CNAME record data in Manage Zones records page.
We currently allow users to submit IP addresses that are formatted as fully-qualified domain names (FQDNs) in CNAME record data, which is incorrect since mappings from a record name → IP address should be an A record.

Add an extra validation in our **RecordSetValidations** to verify that the **cname** data is not simply an IP address with a terminating dot.

Jay07GIT:
Fixes #1181.

- Restrict IPv4 as record data in Cname validation.


JoshSEdwards (approved):
LGTM

Tested locally and saw the error as expected when trying to add a CNAME with IPv4 record data from the zone page

nspadaccino (approved):

arpit4ever (approved):
LGTM

PR #1235, opened 2023-03-17
Add logs and time converter for automated zone syncs

Aravindh-Raju:
- Zone change seems to be saved multiple times in the zone change table while performing automated sync. Adding logs to debug. 
- Added a modal to convert the local time entered to it's equivalent UTC time. The modal can be viewed by clicking `Convert to UTC time` below the scheduler.

### DIFF
--- a/modules/portal/public/lib/controllers/controller.manageZones.js
+++ b/modules/portal/public/lib/controllers/controller.manageZones.js
@@ -338,6 +338,7 @@ angular.module('controller.manageZones', ['angular-cron-jobs'])
         zoneHistoryPaging = pagingService.resetPaging(zoneHistoryPaging);
          function success(response) {
             $log.log('zonesService::getZoneChanges-success');
+            $log.log('zonesService::getZoneChanges: ', response.data.zoneChanges);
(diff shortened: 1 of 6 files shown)


modules/portal/public/lib/controllers/controller.manageZones.js
```diff
@@ -99,6 +101,23 @@ angular.module('controller.manageZones', ['angular-cron-jobs'])
...
+    $scope.openTimeConverter = function() {
+        $("#time_converter_modal").modal("show");
+    };
+
+    $scope.getLocalTimeZone = function() {
+        return new Date().toLocaleString('en-us', {timeZoneName:'short'}).split(' ')[3];
+    };
+
+    $scope.getUtcTime = function() {
+        $scope.utcTime = moment($scope.time, 'hh:mm A').utc().format('hh:mm A');
```
arpit4ever, line 113: Since zone sync scheduler is in 24hr format, it will be useful if convert to UTC also uses 24hr format.

arpit4ever (approved):
One comment otherwise looks good.

nspadaccino (approved):

PR #1239, opened 2023-03-21
Updates to zone sync scheduler

Aravindh-Raju:
Changes in this pull request:
- Minor updates to zone sync scheduler: Added 24hr format to time picker. Added minutes to increment by 5. Added trailing zero before minute values in scheduler dropdown. Hidden calendar as it is not necessary. Removed unnecessary logs.

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/backend/CommandHandler.scala
+++ b/modules/api/src/main/scala/vinyldns/api/backend/CommandHandler.scala
@@ -164,7 +164,6 @@ object CommandHandler {
       message.command match {
         case sync: ZoneChange
             if sync.changeType == ZoneChangeType.Sync || sync.changeType == ZoneChangeType.AutomatedSync || sync.changeType == ZoneChangeType.Create =>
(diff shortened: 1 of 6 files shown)


nspadaccino (commented):
One suggested change:
The UTC time converter has you enter the local time in 24 hour format, but returns the UTC time in 12 hour format, which you then have to convert back to 24 hour format to put into the scheduler. Could we just return the UTC time in 24 hour format instead? That way everything related to the scheduler will be in 24 hour format to reduce confusion

Aravindh-Raju:
> One suggested change: The UTC time converter has you enter the local time in 24 hour format, but returns the UTC time in 12 hour format, which you then have to convert back to 24 hour format to put into the scheduler. Could we just return the UTC time in 24 hour format instead? That way everything related to the scheduler will be in 24 hour format to reduce confusion

Yes, that would make it a lot easier. Made the change.

nspadaccino (approved):
Looks good to me!

arpit4ever (approved):
LGTM

PR #1246, opened 2023-04-11
Add record history in global recordset search

### LINKED ISSUE #1242: Record Change History in Global Recordset Search

Associating the Record Change History from within a zone to record lookups via the Record-Set Search would make searching individual record histories much easier. 

Add a column in Global Recordset Search which can be opened to view the complete change history of a particular record which is being searched.

Aravindh-Raju:
Fixes #1242.

Changes in this pull request:
- Display a particular record's history from global recordset search page.
- Made changes to display zone's admin group as a record's owner group for the records in private zone.


modules/api/src/main/scala/vinyldns/api/route/RecordSetRouting.scala
```diff
@@ -225,7 +225,34 @@ class RecordSetRoute(
...
+    } ~
+    path("recordsetchange" / "history") {
+      (get & monitor("Endpoint.listRecordSetChangeHistory")) {
+        parameters("startFrom".as[Int].?, "maxItems".as[Int].?(DEFAULT_MAX_ITEMS), "fqdn".as[String].?, "recordType".as[String].?) {
+          (startFrom: Option[Int], maxItems: Int, fqdn: Option[String], recordType: Option[String]) =>
+            handleRejections(invalidQueryHandler) {
+              val errorMessage = if(fqdn.isEmpty && recordType.isEmpty) {
+                "Record fqdn and record type cannot be empty"
+              } else {
+                s"maxItems was $maxItems, maxItems must be between 0 exclusive " +
```
nspadaccino, line 245: I noticed when testing the API that if I put in the wrong parameters, for example "type" instead of "recordType", the maxitems message gets triggered even though that isn't the real error. So I think the conditional should be updated to handle that case.

modules/api/src/main/scala/vinyldns/api/route/RecordSetRouting.scala
```diff
@@ -225,7 +225,34 @@ class RecordSetRoute(
...
+        }
+      }
+    } ~
+    path("recordsetchange" / "history") {
+      (get & monitor("Endpoint.listRecordSetChangeHistory")) {
+        parameters("startFrom".as[Int].?, "maxItems".as[Int].?(DEFAULT_MAX_ITEMS), "fqdn".as[String].?, "recordType".as[String].?) {
+          (startFrom: Option[Int], maxItems: Int, fqdn: Option[String], recordType: Option[String]) =>
+            handleRejections(invalidQueryHandler) {
+              val errorMessage = if(fqdn.isEmpty && recordType.isEmpty) {
+                "Record fqdn and record type cannot be empty"
```
nspadaccino, line 243: Could this instead say "recordType and fqdn cannot be empty" to match the parameter names?

modules/api/src/test/functional/tests/recordsets/list_recordset_changes_test.py
```diff
@@ -174,3 +206,100 @@ def test_list_recordset_changes_max_items_boundaries(shared_zone_test_context):
...
+
+def test_list_recordset_history_success(shared_zone_test_context):
+    """
+    Test recordset history succeeds with membership auth for member of admin group
+    """
+    client = shared_zone_test_context.history_client
+    fqdn = "test-create-cname-ok.system-test-history1."
+    type = "CNAME"
+    response = client.list_recordset_change_history(fqdn, type, status=200)
+    check_change_history_response(response, fqdn, type, recordChanges=True, startFrom=False, nextId=False)
```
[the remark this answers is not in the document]
Aravindh-Raju: The earlier checks were just some auth checks and hence I didn't add in them. But now, I added the check response in successful auth test. In auth fails, we get 401 and 403 and so we cannot add this check.

modules/api/src/main/scala/vinyldns/api/route/RecordSetRouting.scala
Aravindh-Raju: It was due to the `&&` in the if condition. I changed it now and we get the expected error when we type in a wrong parameter ("type" instead of "recordType")

modules/api/src/main/scala/vinyldns/api/route/RecordSetRouting.scala
Aravindh-Raju: Made the change.

Jay07GIT (approved):
LGTM

JoshSEdwards:
Tested locally. I really like the implementation of this one! It looks good. Only issue I noticed is after you press View History, the Sync Zone button appears on the Global RecordSet Search page.

Aravindh-Raju:
> Tested locally. I really like the implementation of this one! It looks good. Only issue I noticed is after you press View History, the Sync Zone button appears on the Global RecordSet Search page.

Thanks for finding that out. Seems like `Sync Zone` is never used in `Global Recordset Search` page. The `ng-if` condition is never satisfied before I made some changes here to get zone details. I guess the `Sync Zone` button is added here by mistake and seems to be copied from `Zones` page. I removed the button now. Please have a look.

JoshSEdwards:
> > Tested locally. I really like the implementation of this one! It looks good. Only issue I noticed is after you press View History, the Sync Zone button appears on the Global RecordSet Search page.
> 
> Thanks for finding that out. Seems like `Sync Zone` is never used in `Global Recordset Search` page. The `ng-if` condition is never satisfied before I made some changes here to get zone details. I guess the `Sync Zone` button is added here by mistake and seems to be copied from `Zones` page. I removed the button now. Please have a look.

Looks good now! 😄

JoshSEdwards (approved):

nspadaccino (approved):

arpit4ever (approved):
LGTM

PR #1251, opened 2023-04-26
Enable debug logging for hikariCP

nspadaccino:
Hikari pool stats are now logged on the console. Every 30 seconds the stats get output like this:
`{"@timestamp":"2023-04-26T00:17:26.667Z", "log.level":"DEBUG", "message":"mysqlDbPool - Pool stats (total=10, active=0, idle=10, waiting=0)", "ecs.version": "1.2.0","service.name":"vinyldns-portal","service.node.name":"vinyldns-portal","event.dataset":"vinyldns-portal","process.thread.name":"mysqlDbPool housekeeper","log.logger":"com.zaxxer.hikari.pool.HikariPool"}`

### DIFF
--- a/build/docker/api/application.conf
+++ b/build/docker/api/application.conf
@@ -188,9 +188,9 @@ vinyldns {
       name = ${?DATABASE_NAME}
       driver = "org.mariadb.jdbc.Driver"
       driver = ${?JDBC_DRIVER}
-      migration-url = "jdbc:mariadb://localhost:19002/?user=root&password=pass"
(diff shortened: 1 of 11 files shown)


Jay07GIT:
This looks good. But my wish is to identify the root cause for the time out(in which request the timeout occurs.) . Can we add a logs in the case of timeout exception , from which function(query) the timeout occurs.?

Aravindh-Raju (dismissed):

arpit4ever (approved):
LGTM

JoshSEdwards (approved):
LGTM

PR #1252, opened 2023-05-08
Add new record types in batch change

### LINKED ISSUE #1248: Support SRV and NAPTR Records via Batch Change
Currently batch changes only support the following record types: A, AAAA, CNAME, PTR, TXT, MX

A team looking to provision their zones to be managed with Vinyl. Those zones contain many SRV and NAPTR records, and having these supported via the batch change interface, and CSV import in particular, would make the process much smoother.

Aravindh-Raju:
Fixes #1248 and #1255.

Changes in this pull request:
- Add NAPTR, SRV and NS record types in batch change.
- Add support for the above record types through csv.


modules/api/src/main/scala/vinyldns/api/route/BatchChangeJsonProtocol.scala
```diff
@@ -249,8 +249,23 @@ trait BatchChangeJsonProtocol extends JsonValidation {
...
+          "Missing BatchChangeInput.changes.record.weight and " +
+          "Missing BatchChangeInput.changes.record.port and " +
+          "Missing BatchChangeInput.changes.record.target"
+      )
+      case NAPTR => js.required[NAPTRData](
+        "Missing BatchChangeInput.changes.record.order and " +
+          "Missing BatchChangeInput.changes.record.preference and " +
+          "Missing BatchChangeInput.changes.record.flags and " +
+          "Missing BatchChangeInput.changes.record.service and " +
+          "Missing BatchChangeInput.changes.record.regexp and " +
```
nspadaccino, line 264: regexp is optional right? Or is it required via the API?

modules/docs/src/main/resources/microsite/static/dns-changes-csv-sample.csv
```diff
@@ -6,6 +6,11 @@ Add,AAAA,test3.example.com.,,fd69:27cc:fe91::60
 Add,CNAME,test4.example.com.,,test0.example.com.
 Add,PTR,192.0.2.193,200,test.example.com.
 Add,TXT,test5.example.com,7200,example text
+Add,MX,test16.ok.,7200,1 text.com
+Add,NS,test17.ok.,7200,172.17.42.1
+Add,NAPTR,test18.ok.,7200,10 20 s SIPS+D2T _si._tc.ac-ot1.wdv.test.net.
```
nspadaccino, line 11: So in the case where the record data has multiple fields like for NAPTR and SRV, the values should be space separated?

nspadaccino (changes requested):
Tested out locally and the functionality looks good. Just one thing: for NAPTR records, the only validation we have on the Flags input is that it should be less than two characters. I am able to submit flags other than U, S, A, or P (such as "F" or "1"), but when I do, the record change fails with no message. I think if those flags are the only valid ones, we should validate the input to make sure the flag is one of those

modules/api/src/main/scala/vinyldns/api/route/BatchChangeJsonProtocol.scala
Aravindh-Raju: Regexp was optional, but the field was made required while making NAPTR changes via Zones view as seen in `DNSJsonProtocol.scala` - https://github.com/vinyldns/vinyldns/blob/master/modules/api/src/main/scala/vinyldns/api/route/DnsJsonProtocol.scala#L524-L528

So, I made it required in `BatchChangeJsonProtocol.scala` too. The `regexp` field is required but the data is optional and can be left empty.

modules/docs/src/main/resources/microsite/static/dns-changes-csv-sample.csv
Aravindh-Raju: Yes, in the case where the record data has multiple fields, the values are space separated. But if we want to have some other delimiter instead of space, I can do that.

Aravindh-Raju:
> Tested out locally and the functionality looks good. Just one thing: for NAPTR records, the only validation we have on the Flags input is that it should be less than two characters. I am able to submit flags other than U, S, A, or P (such as "F" or "1"), but when I do, the record change fails with no message. I think if those flags are the only valid ones, we should validate the input to make sure the flag is one of those

Added validation for NAPTR flags in api and converted the text box to dropdown in portal with U, S, A, and P values.

modules/docs/src/main/resources/microsite/static/dns-changes-csv-sample.csv
```diff
@@ -6,6 +6,11 @@ Add,AAAA,test3.example.com.,,fd69:27cc:fe91::60
 Add,CNAME,test4.example.com.,,test0.example.com.
 Add,PTR,192.0.2.193,200,test.example.com.
 Add,TXT,test5.example.com,7200,example text
+Add,MX,test16.ok.,7200,1 text.com
+Add,NS,test17.ok.,7200,172.17.42.1
+Add,NAPTR,test18.ok.,7200,10 20 s SIPS+D2T _si._tc.ac-ot1.wdv.test.net.
+Add,NAPTR,test20.ok.,7200,10 20 s SIPS+D2T !^.*$!sip:jd@corpxyz.com! _si._tc.ac-ot1.wdv.test.net.
```
nspadaccino, line 12: With the new validation for the Flags field, the flags have to be uppercase if you are importing from csv, so we should change that here

nspadaccino:
Another thing I noticed is that if you enter a string in the regex field and the string does not begin and end with an "!", the record change passes validation but fails, so I think we should add some input validation for the regex field

modules/docs/src/main/resources/microsite/static/dns-changes-csv-sample.csv
Aravindh-Raju: Made the changes

Aravindh-Raju:
> Another thing I noticed is that if you enter a string in the regex field and the string does not begin and end with an "!", the record change passes validation but fails, so I think we should add some input validation for the regex field

Looks like regexp should begin and end with '!' or can be empty. Added the validation now.

nspadaccino:
Could we also enforce the input validation when creating NAPTR records from the zone view? Unless you think that would be better in another PR since this PR is about batch changes

Aravindh-Raju:
> Could we also enforce the input validation when creating NAPTR records from the zone view? Unless you think that would be better in another PR since this PR is about batch changes

I think it would be better to open a separate PR, incase we need to refer that individually in the future. I'll open one soon with the changes.

JoshSEdwards (approved):

nspadaccino (approved):
Looks good to me

PR #1262, opened 2023-06-01
Fix bugs in Zone view record fields

### LINKED ISSUE #1261: Fix bugs in zone view record fields
**Describe the bug**

1. There is a bug with PTR record validation while making changes through zone view. It should display the validation error `At least one fully qualified domain name is required` when we submit without any value. But due to a wrong value in `name` attribute it wasn't working.
2. After creating NAPTR, DS and SSHFP records, when we click on `View created recordset` option in zones view and also when we click on `delete recordset` option, the dropdown fields should be readonly and disabled. But currently it can be changed and not disabled.

Aravindh-Raju:
Fixes #1261 

Changes in this pull request:
- Fix few bugs present in `Zone` view record fields.


nspadaccino (approved):
Looks good!

arpit4ever (approved):
LGTM

PR #1279, opened 2023-07-11
Fix group history and zone update bug

### LINKED ISSUE #1269: Bug with accessing group changes
Certain groups display a 500 error when accessed via the portal or API. This error happens when the group contains a group change which has a user ID that is no longer in the users table. Trying to get the user which no longer exists causes a chain reaction making both the group and the particular group change inaccessible. While users should never be removed from the users table, this has happened a few times. We need to add a check in the relevant membership methods which prevents an error if a user id in the group change is not found.

Aravindh-Raju:
Fixes #1269.

Changes in this pull request:
- Added a default value `unknown user` when a user is no longer present in database. Fixed a bug with group change `description`.
- Fixed a bug with zone sync scheduler while updating zone.


JoshSEdwards:
Will this not work retroactively on the group already generating the error? I just tested it locally with a prod snapshot and it's still throwing a 500. I'll send more details on Slack

Aravindh-Raju:
> Will this not work retroactively on the group already generating the error? I just tested it locally with a prod snapshot and it's still throwing a 500. I'll send more details on Slack

The problem seems to be with a change in group description. Since group description is optional, they updated it to `None` which wasn't covered in the conditional before. Made the required change now and seems to work as expected. Please have a look.

nspadaccino (approved):
Tested locally with prod proxy, problem groups no longer throw 500 errors and groups changes are visible. I can also update zones without a zone sync scheduler error being thrown. Looks good

arpit4ever (approved):
LGTM
Tested locally with prod DB.

PR #1284, opened 2023-07-24
list abandoned zones

### LINKED ISSUE #617: No way to view zone change history on deleted zone

Jay07GIT:
Fixes #617

Changes in this pull request:

- Added API for deleted zones history (zones/deleted/changes)
- Made Portal changes to view deleted zones history.
- Removed Cascade from zone_access table. We use this on comparing zone_change table .
- Added zone_status column in zone_access table to identify the deleted zones and used the same in deleting the group.

### DIFF
--- a/modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneChangeRepository.scala
+++ b/modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneChangeRepository.scala
@@ -206,7 +206,7 @@ class MySqlZoneChangeRepository
             }
 
           val deletedZonesWithStartFrom: List[ZoneChange] = startFrom match {
-            case Some(zoneName) => results.dropWhile(_.zone.name != zoneName)
+            case Some(zoneId) => results.dropWhile(_.zone.id != zoneId)
             case None => results
           }
 
@@ -214,7 +214,7 @@ class MySqlZoneChangeRepository
 
           val (newResults, nextId) =
             if (deletedZonesWithMaxItems.size > maxItems)
-              (deletedZonesWithMaxItems.dropRight(1), deletedZonesWithMaxItems.lastOption.map(_.zone.name))
+              (deletedZonesWithMaxItems.dropRight(1), deletedZonesWithMaxItems.lastOption.map(_.zone.id))
             else (deletedZonesWithMaxItems, None)
 
           ListDeletedZonesChangeResults(
--- a/modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneChangeRepository.scala
+++ b/modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneChangeRepository.scala
@@ -17,10 +17,13 @@
 package vinyldns.mysql.repository
 
 import cats.effect.IO
+
 import java.time.Instant
 import java.time.temporal.ChronoUnit
 import org.slf4j.LoggerFactory
 import scalikejdbc._
+import vinyldns.core.domain.auth.AuthPrincipal
+import vinyldns.core.domain.membership.User
 import vinyldns.core.domain.zone._
 import vinyldns.core.protobuf._
 import vinyldns.core.route.Monitored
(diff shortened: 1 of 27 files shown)


modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneChangeRepository.scala
```diff
@@ -110,6 +127,106 @@ class MySqlZoneChangeRepository
...
+
+          val zoneNotInZoneChange: List[ZoneChange] =
+            zoneChangeResults.filter(z=> !zoneResults.map(_.name).contains(z.zone.name) && z.zone.status != ZoneStatus.Active)
+
+          val deletedZoneResults: List[ZoneChange] =
+            zoneNotInZoneChange.filter(_.zone.status.equals(ZoneStatus.Deleted)).distinct.sortBy(_.zone.updated).reverse
+
+          val results: List[ZoneChange] =
+            if (zoneNameFilter.nonEmpty) {
+              deletedZoneResults.filter(r => r.zone.name.contains(zoneNameFilter.getOrElse("not found")))
```
Aravindh-Raju, line 203: I see zoneNameFilter and pagination are not covered by tests. We can add those tests in `MySqlZoneChangeRepositoryIntegrationSpec.scala` and increase the code coverage.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneChangeRepository.scala
```diff
@@ -110,6 +127,106 @@ class MySqlZoneChangeRepository
...
+            } else {
+              deletedZoneResults
+            }
+
+          val deletedZonesWithStartFrom: List[ZoneChange] = startFrom match {
```
Aravindh-Raju, line 208: Seems like pagination is not working as expected. Maybe a problem with how `startfrom` is used here. I have attached screenshots for reference. There were 4 items before pagination. But while paginating, I get the same results in every page. Needs to be fixed.

Screenshot 2023-08-02 at 12 34 36 PM
Screenshot 2023-08-02 at 12 37 41 PM
Screenshot 2023-08-02 at 12 37 46 PM

Aravindh-Raju (changes requested):

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlZoneChangeRepository.scala
Jay07GIT: Startfrom changed from zone name to zone id. Now its working fine.

arpit4ever (approved):
LGTM
Tested (GUI & API) functionality locally.

nspadaccino (changes requested):
I'm seeing a bug in the zones view.
If I log in as one user, create a group, and connect a zone, then log in with another user not in that group, I am able to see the zone in My Zones, and I am able to click on it. When I click on it, I get a 403 error (since that user doesn't have permission to view that private zone).
Screenshot 2023-08-16 at 11 19 09 AM
Other groups' private zones shouldn't appear in My Zones tab

nspadaccino:
Also, it seems like wildcard search isn't working correctly for abandoned zones.
In My Zones, I can search for "o*" and see the zone "ok". But when I search for "o*" in abandoned zones, I get no results.
Screenshot 2023-08-16 at 11 30 24 AM
Screenshot 2023-08-16 at 11 30 32 AM
Screenshot 2023-08-16 at 11 30 40 AM
Screenshot 2023-08-16 at 11 30 47 AM

Jay07GIT:
> Also, it seems like wildcard search isn't working correctly for abandoned zones. In My Zones, I can search for "o*" and see the zone "ok". But when I search for "o*" in abandoned zones, I get no results. Screenshot 2023-08-16 at 11 30 24 AM Screenshot 2023-08-16 at 11 30 32 AM Screenshot 2023-08-16 at 11 30 40 AM Screenshot 2023-08-16 at 11 30 47 AM

Yes. its right . Since we used scala filter we dont need wildcard for seach. we can jus give "o" to search ok. zones.

Jay07GIT:
> I'm seeing a bug in the zones view. If I log in as one user, create a group, and connect a zone, then log in with another user not in that group, I am able to see the zone in My Zones, and I am able to click on it. When I click on it, I get a 403 error (since that user doesn't have permission to view that private zone). Screenshot 2023-08-16 at 11 19 09 AM Other groups' private zones shouldn't appear in My Zones tab

Oh yes. Updated the code. Now its working fine.

arpit4ever (approved):
LGTM

nspadaccino (approved):
looks good

PR #1304, opened 2023-09-25
Added api for record set count and displayed in portal

### LINKED ISSUE #1244: Display total record count in Manage Records tab in Zone view
Display the total number of records in a zone in the Manage Records tab. This will help us estimate how long a zone sync would take.

Jay07GIT:
Fixes #1244.

Changes in this pull request:
- Added api for record set count and displayed in portal

Steps: 
Login to the vinyldns 
zones > manage records
image

### DIFF
--- a/modules/api/src/test/functional/tests/recordsets/get_recordset_test.py
+++ b/modules/api/src/test/functional/tests/recordsets/get_recordset_test.py
@@ -88,6 +88,21 @@ def test_get_recordset_doesnt_exist(shared_zone_test_context):
     client.get_recordset(shared_zone_test_context.ok_zone["id"], "123", status=404)
 
 
+def test_get_recordsetcount(shared_zone_test_context):
+    """
+    Test getting recordset count for a valid zoneid should return 200
+    """
+    client = shared_zone_test_context.ok_vinyldns_client
+    client.get_recordset_count(shared_zone_test_context.ok_zone["id"],status=200)
+
+
+def test_get_recordsetcount_error(shared_zone_test_context):
+    """
+    Test getting recordset count for a invalid zoneid should return 404
+    """
+    client = shared_zone_test_context.ok_vinyldns_client
+    client.get_recordset_count("999",status=404)
+
 @pytest.mark.serial
 def test_at_get_recordset(shared_zone_test_context):
     """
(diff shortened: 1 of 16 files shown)


modules/portal/public/lib/controllers/controller.records.spec.js
```diff
@@ -91,7 +91,7 @@ describe('Controller: RecordsController', function () {
 
     it('deleteSshfp should keep at least one sshfp object', function() {
         this.scope.currentRecord.sshfpItems = [{algorithm: '1', type: '', fingerprint: ''},
-            {algorithm: '2', type: '', fingerprint: ''}];
+            {algorithm: '2', type: '', fideleteSshfpngerprint: ''}];
```
arpit4ever, line 94: possible typo.

Jay07GIT: Oops. Happened by mistake. Updated the code.

modules/api/src/test/functional/tests/recordsets/get_recordset_test.py
```diff
@@ -88,7 +88,60 @@ def test_get_recordset_doesnt_exist(shared_zone_test_context):
...
+
+
+def test_get_recordset_count_error(shared_zone_test_context):
+    """
+    Test getting recordset count for a invalid zoneid should return 404
+    """
+    client = shared_zone_test_context.ok_vinyldns_client
+    client.get_recordset_count("999",status=404)
+
 @pytest.mark.serial
```
Aravindh-Raju, line 106: Just a doubt, why do we have `@pytest.mark.serial` here and not on the other tests? Maybe it has been misplaced while recreating the PR. If so, we can remove that.

Aravindh-Raju (changes requested):
Looks good to me except for a minor change.

Jay07GIT:
> Looks good, but I'm wondering if we should also include documentation for the new api as part of this PR so we don't forget about it

Can we add a separate PR for Docs since we can have only api and portal changes here. I will open a PR and add it in comments.

arpit4ever:
> Looks good, but I'm wondering if we should also include documentation for the new api as part of this PR so we don't forget about it
Logged Issue
https://github.com/vinyldns/vinyldns/issues/1309

modules/api/src/test/functional/tests/recordsets/get_recordset_test.py
Aravindh-Raju: Jay mentioned `@pytest.mark.serial` is added to avoid conflict while creating new records. We can proceed with the merge if everything is alright.

arpit4ever: Sounds good.

arpit4ever (approved):
LGTM

nspadaccino (approved):

PR #1313, opened 2023-10-04
Migration 3.29 Record Change History Fields

nspadaccino:
Update the migration to check if the new columns already exist before adding them

### DIFF
--- a/modules/mysql/src/it/resources/application.conf
+++ b/modules/mysql/src/it/resources/application.conf
@@ -7,10 +7,10 @@ mysql {
     # JDBC Settings, these are all values in scalikejdbc-config, not our own
     # these must be overridden to use MYSQL for production use
     # assumes a docker or mysql instance running locally
-    name = "vinyldns2"
+    name = "vinyldns"
     driver = "org.mariadb.jdbc.Driver"
     migration-url = "jdbc:mariadb://"${mysql.endpoint}"/"
-    url = "jdbc:mariadb://"${mysql.endpoint}"/vinyldns2"
+    url = "jdbc:mariadb://"${mysql.endpoint}"/vinyldns"
     user = "root"
     password = "pass"
     flyway-out-of-order = false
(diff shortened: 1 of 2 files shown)


JoshSEdwards (approved):
LGTM

modules/mysql/src/main/resources/db/migration/V3.29__AddRecordChangeFields.sql
```diff
@@ -2,7 +2,27 @@ CREATE SCHEMA IF NOT EXISTS ${dbName};
...
+        EXECUTE stmt;
+        DEALLOCATE PREPARE stmt;
+    END IF;
+END $$
+
+DELIMITER ;
+
+CALL AddColumnIfNotExists('record_type');
+CALL AddColumnIfNotExists('fqdn');
+
```
Aravindh-Raju, line 26: Looks good. But I'm not sure this migration will run since `V3.29__AddRecordChangeFields.sql` must be already applied to the other environments and altering the file `V3.29__AddRecordChangeFields.sql` would have no effect. And as far as I know, versioned flyway migrations once applied can't be changed. So this change wouldn't be considered. We may have to undo this migration and create a new migration with the above changes. 
For more details about versioned migrations and undoing versioned migrations:
https://documentation.red-gate.com/fd/migrations-184127470.html#:~:text=They%20are%20applied%20in%20order,every%20time%20their%20checksum%20changes.

We can try merging and running it. If it works well, fine. Else we can try the above approach that I found.

modules/mysql/src/it/resources/application.conf
```diff
@@ -7,10 +7,10 @@ mysql {
     # JDBC Settings, these are all values in scalikejdbc-config, not our own
     # these must be overridden to use MYSQL for production use
     # assumes a docker or mysql instance running locally
-    name = "vinyldns2"
+    name = "vinyldns"
```
Aravindh-Raju, line 10: Just for my clarification, is there a reason why this is changed?

Aravindh-Raju (approved):

PR #1319, opened 2023-10-11
Update migration 3.30 to use drop foreign key instead of drop constraint

### LINKED ISSUE #1318: SQL Syntax error in migration 3.30
Migration 3.30 was added as part of https://github.com/vinyldns/vinyldns/pull/1284
Flyway error when validating migration 3.30:

Caused by: java.sql.SQLSyntaxErrorException: (conn=547326) You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'CONSTRAINT fk_zone_access,
Caused by: org.mariadb.jdbc.internal.util.exceptions.MariaDbSqlException: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'CONSTRAINT fk_zone_access,
Caused by: java.sql.SQLException: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'CONSTRAINT fk_zone_access,
ADD zone_status CHAR(36) NOT NULL' at line 1

The migration should be using DROP FOREIGN KEY instead of DROP CONSTRAINT

nspadaccino:
Fixes #1318

### DIFF
--- a/modules/mysql/src/main/resources/db/migration/V3.30__AddStatusZoneAccess.sql
+++ b/modules/mysql/src/main/resources/db/migration/V3.30__AddStatusZoneAccess.sql
@@ -3,6 +3,5 @@ CREATE SCHEMA IF NOT EXISTS ${dbName};
 USE ${dbName};
 
 ALTER TABLE zone_access
-DROP CONSTRAINT fk_zone_access,
-ADD zone_status CHAR(36) NOT NULL
-;
+DROP FOREIGN KEY fk_zone_access,
+ADD COLUMN zone_status CHAR(36) NOT NULL;


modules/mysql/src/main/resources/db/migration/V3.30__AddStatusZoneAccess.sql
```diff
@@ -3,6 +3,5 @@ CREATE SCHEMA IF NOT EXISTS ${dbName};
 USE ${dbName};
 
 ALTER TABLE zone_access
-DROP CONSTRAINT fk_zone_access,
-ADD zone_status CHAR(36) NOT NULL
-;
\ No newline at end of file
+DROP FOREIGN KEY fk_zone_access,
+ADD COLUMN zone_status CHAR(36) NOT NULL;
```
Aravindh-Raju, line 7: We would mostly like get an error if this migration file is already attempted to be applied in a database. We can only make changes to an existing migration file as long as it wasn't run on a database as far as I know.

nspadaccino: Testing out on a copy of the dev database, I was able to use the flyway CLI to manually fix the error, so I think we should be good

arpit4ever (approved):
LGTM

PR #1320, opened 2023-10-13
Display action buttons if there's ACL access

### LINKED ISSUE #1316: Users are not able to view Edit/Delete buttons on records which they have ACLs for
This applies to both Shared and Private zones.

If a user or group has an ACL rule giving them write/delete access to some or all records in a zone, then they should be able to go to that zone view in the portal and see Edit/Delete buttons next to the records they have access to.

Currently, ACLs just give access via batch change, and users cannot update the records they have ACL access to via the zone view unless they are also in the zone owner group.

Aravindh-Raju:
Fixes #1316 and #1322.

Changes in this pull request:
- Display the `Update` and `Delete` buttons in Zones view when a user has Write and Delete ACL rules applied. Only the `Update` button will be displayed if there's only `Write` ACL rule. Both `Update` and `Delete` will be displayed if there's `Delete` ACL rule.


nspadaccino (commented):
This works for private zones, but for shared zones the change makes the buttons available for all unowned records. When we opened up the ability to view shared zones to all users, we kept it so that they wouldn't be able to update or delete records from that view https://github.com/vinyldns/vinyldns/pull/1162. 

If we want to stick with that choice, then if a group has an ACL granting delete access for just A records in  a shared zone, they should just see the buttons for A records. Same goes for ACL on a specific record mask.

However, maybe this is something we want to change. Maybe limiting access to the update and delete buttons in shared zones is too restrictive?

Aravindh-Raju:
> This works for private zones, but for shared zones the change makes the buttons available for all unowned records. When we opened up the ability to view shared zones to all users, we kept it so that they wouldn't be able to update or delete records from that view #1162.
> 
> If we want to stick with that choice, then if a group has an ACL granting delete access for just A records in a shared zone, they should just see the buttons for A records. Same goes for ACL on a specific record mask.
> 
> However, maybe this is something we want to change. Maybe limiting access to the update and delete buttons in shared zones is too restrictive?

I've made changes to hide the action buttons in shared zones for now. We shall discuss things we want to do about shared zones as discussed and open a separate PR for that.

nspadaccino:
Created issue https://github.com/vinyldns/vinyldns/issues/1322 to show buttons in shared zones

nspadaccino (dismissed):
Looks good

nspadaccino (approved):
Tested locally, looks good, ACL rules are making the proper buttons show up in private zones, and users can now edit and delete unowned records in shared zones directly via zone view

PR #1321, opened 2023-10-18
Updated record set ownergroup id as none when zone is private

### LINKED ISSUE #1317: Users can lose access to certain records if a zone is changed from shared to private
Scenario:

- A shared zone is added to vinyl
- A user from group A adds a record "foo" to that zone. That record is owned by group A.
- The zone is later changed to from shared to private, with the owner group being set to group B.
- Users from group B expect to be able to update any of the records in the zone, since they own the zone. However, record "foo" is still owned by group A, even though it doesn't say so when looking at the zone view in the portal. 
- If a user from group B tries to update or delete record "foo", they will see the error `HTTP 422 (Unprocessable Entity): User not in record owner group with id \"<groupID of group A>\"`

If a zone is changed from shared to private, all the records contained in that zone should have their owner group assigned to the new zone admin group

Jay07GIT:
Fixes #1317.

Changes in this pull request:
- Updated record set owner group id as none when zone is private
-

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetService.scala
@@ -145,7 +145,7 @@ class RecordSetService(
       _ <- unchangedZoneId(existing, recordSet).toResult
       change <- RecordSetChangeGenerator.forUpdate(existing, recordSet, zone, Some(auth)).toResult
       // because changes happen to the RS in forUpdate itself, converting 1st and validating on that
-      rsForValidations = change.recordSet
+      rsForValidations = if(!zone.shared) change.recordSet.copy(ownerGroupId = None) else change.recordSet
       superUserCanUpdateOwnerGroup = canSuperUserUpdateOwnerGroup(existing, recordSet, zone, auth)
       _ <- isNotHighValueDomain(recordSet, zone, highValueDomainConfig).toResult
       _ <- canUpdateRecordSet(auth, existing.name, existing.typ, zone, existing.ownerGroupId, superUserCanUpdateOwnerGroup).toResult
@@ -184,6 +184,7 @@ class RecordSetService(
       ).toResult
       _ <- if(existing.name == rsForValidations.name) ().toResult else if(allowedZoneList.contains(zone.name)) checkAllowedDots(allowedDotsLimit, rsForValidations, zone).toResult else ().toResult
       _ <- if(allowedZoneList.contains(zone.name)) isNotApexEndsWithDot(rsForValidations, zone).toResult else ().toResult
+
       _ <- messageQueue.send(change).toResult[Unit]
     } yield change
 


nspadaccino (commented):
Using this scenario:
```
A shared zone is added to vinyl
A user from group A adds a record "foo" to that zone. That record is owned by group A.
The zone is later changed to from shared to private, with the owner group being set to group B.
```

Testing with these changes, a user from group B will be able to successfully update record "foo". And when you search "foo" in the recordset search, it shows up as being owned by group B. So the functionality in the portal looks good.
However, in the database, the record in the recordset table still shows group A as the owner_group_id, even though it says group B in the portal, which is a bit confusing.

I think as part of these changes we should update the entries in the database as well. Could we have the owner_group_id column update when the zone access type is changed

Jay07GIT:
> Using this scenario:
> 
> ```
> A shared zone is added to vinyl
> A user from group A adds a record "foo" to that zone. That record is owned by group A.
> The zone is later changed to from shared to private, with the owner group being set to group B.
> ```
> 
> Testing with these changes, a user from group B will be able to successfully update record "foo". And when you search "foo" in the recordset search, it shows up as being owned by group B. So the functionality in the portal looks good. However, in the database, the record in the recordset table still shows group A as the owner_group_id, even though it says group B in the portal, which is a bit confusing.
> 
> I think as part of these changes we should update the entries in the database as well. Could we have the owner_group_id column update when the zone access type is changed

Nice finding. I noticed it. This is done intentionally because if if we update the DB then when user wants to change zone from private to shared, then the pre - owned records will be updated. It might get confused again right? If we still want to update the DB, we can.

Jay07GIT:
> Using this scenario:
> 
> ```
> A shared zone is added to vinyl
> A user from group A adds a record "foo" to that zone. That record is owned by group A.
> The zone is later changed to from shared to private, with the owner group being set to group B.
> ```
> 
> Testing with these changes, a user from group B will be able to successfully update record "foo". And when you search "foo" in the recordset search, it shows up as being owned by group B. So the functionality in the portal looks good. However, in the database, the record in the recordset table still shows group A as the owner_group_id, even though it says group B in the portal, which is a bit confusing.
> 
> I think as part of these changes we should update the entries in the database as well. Could we have the owner_group_id column update when the zone access type is changed

Nice finding. I noticed it. This is done intentionally because if if we update the DB then when user wants to change zone from private to shared, then the pre - owned records will be updated. It might get confused again right? 

If we still want to update the DB, we can.

modules/api/src/it/scala/vinyldns/api/domain/record/RecordSetServiceIntegrationSpec.scala
```diff
@@ -659,6 +660,16 @@ class RecordSetServiceIntegrationSpec
         Some(group2.id)
     }
 
+    "update successfully for private zone if user is updating record set when record set is already owned by another group were user is not a part of" in {
```
Aravindh-Raju, line 663: A small typo here. Should be `where` instead of `were`.

modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetService.scala
```diff
@@ -184,6 +184,7 @@ class RecordSetService(
       ).toResult
       _ <- if(existing.name == rsForValidations.name) ().toResult else if(allowedZoneList.contains(zone.name)) checkAllowedDots(allowedDotsLimit, rsForValidations, zone).toResult else ().toResult
       _ <- if(allowedZoneList.contains(zone.name)) isNotApexEndsWithDot(rsForValidations, zone).toResult else ().toResult
+
```
Aravindh-Raju, line 187: Can remove this new line here.

modules/api/src/test/scala/vinyldns/api/domain/record/RecordSetServiceSpec.scala
```diff
@@ -887,6 +887,51 @@ class RecordSetServiceSpec
       result.changeType shouldBe RecordSetChangeType.Update
       result.status shouldBe RecordSetChangeStatus.Pending
     }
+
+    "return the recordSet change as the result for private zone if user is updating record set " +
+      "when record set is already owned by another group were user is not a part of" in {
```
Aravindh-Raju, line 892: Typo here too. `were` should be replaced by `where`.

Aravindh-Raju (changes requested):
Few changes, looks good otherwise!

Also, as nick mentioned, we could update the group id to the db too if it doesn't involve too much of an effort. Else, we can do that in a separate PR.

modules/api/src/test/scala/vinyldns/api/domain/record/RecordSetServiceSpec.scala
Jay07GIT: Updated the code.

modules/api/src/it/scala/vinyldns/api/domain/record/RecordSetServiceIntegrationSpec.scala
Jay07GIT: Updated the code.

PR #1327, opened 2023-11-08
Hide group change history for non-members

### LINKED ISSUE #1326: Hide group change history for non-members
Currently group change history tab in portal and the route in api allows all users to get a group's change history. Ideally only the members, admins, super users and support users should be able to see these details.

Hide group change history for users not being part of the group and is not a super or support user.

Aravindh-Raju:
Fixes #1326. 

Changes in this pull request:
- Hidden group change history for non-members in portal. Made restrictions to not access group change history through api for non-members.


nspadaccino (approved):
Looks good!

PR #1328, opened 2023-11-08
add group id in user response

### LINKED ISSUE #1334: Add UserId field in User details response API
Changes to be do:-
1. Add UserId field in User details response API.

Currently we are not able to get group id where user located in. This API will helps to get the group id of the User.

Jay07GIT:
Fixes #1334.

Changes in this pull request:
- Added group id in user details response API.

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
@@ -362,6 +362,15 @@ class MembershipService(
       .orFail(UserNotFoundError(s"User $userIdentifier was not found"))
       .toResult[User]
 
+
+  def getGroupByUser(userIdentifier: String, authPrincipal: AuthPrincipal): Result[UserResponseInfo] =
+      for{
+    userId <- getUser(userIdentifier,authPrincipal) .map(_.id)
+        user <- getUser(userIdentifier,authPrincipal)
+    group =  membershipRepo.getGroupsForUser(userId).unsafeRunSync()
+  } yield UserResponseInfo(user.id, Some(user.userName), group)
+
+
   def getUsers(
       userIds: Set[String],
       startFrom: Option[String] = None,
--- a/modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
@@ -364,11 +364,11 @@ class MembershipService(
 
 
   def getGroupByUser(userIdentifier: String, authPrincipal: AuthPrincipal): Result[UserResponseInfo] =
-      for{
-    userId <- getUser(userIdentifier,authPrincipal) .map(_.id)
+    for{
+        userId <- getUser(userIdentifier,authPrincipal) .map(_.id)
         user <- getUser(userIdentifier,authPrincipal)
-    group =  membershipRepo.getGroupsForUser(userId).unsafeRunSync()
-  } yield UserResponseInfo(user.id, Some(user.userName), group)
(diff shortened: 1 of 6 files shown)


modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
```diff
@@ -362,6 +362,15 @@ class MembershipService(
       .orFail(UserNotFoundError(s"User $userIdentifier was not found"))
       .toResult[User]
 
+
+  def getGroupByUser(userIdentifier: String, authPrincipal: AuthPrincipal): Result[UserResponseInfo] =
+      for{
+    userId <- getUser(userIdentifier,authPrincipal) .map(_.id)
+        user <- getUser(userIdentifier,authPrincipal)
+    group =  membershipRepo.getGroupsForUser(userId).unsafeRunSync()
```
Aravindh-Raju, line 370: We should avoid using `unsafeRunSync()`. Should find a way to get the groups without having to use that.

modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
```diff
@@ -362,6 +362,15 @@ class MembershipService(
       .orFail(UserNotFoundError(s"User $userIdentifier was not found"))
       .toResult[User]
 
+
+  def getGroupByUser(userIdentifier: String, authPrincipal: AuthPrincipal): Result[UserResponseInfo] =
```
Aravindh-Raju, line 366: Since we are getting groups we could change the function name to `getGroupForUser`.

Aravindh-Raju (changes requested):

modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
Jay07GIT: Changed the name to getUserDetails.

modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
Jay07GIT: Removed unsafeRunSync().

modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
```diff
@@ -362,6 +362,14 @@ class MembershipService(
       .orFail(UserNotFoundError(s"User $userIdentifier was not found"))
       .toResult[User]
 
+
+  def getUserDetails(userIdentifier: String, authPrincipal: AuthPrincipal): Result[UserResponseInfo] =
+    for{
+        user <- getUser(userIdentifier,authPrincipal)
+        group <-  membershipRepo.getGroupsForUser(user.id).toResult[Set[String]]
+    } yield UserResponseInfo(user.id, Some(user.userName), group)
```
Aravindh-Raju, line 370: The new function is not covered by tests. Tests for these should be added in `MembershipServiceSpec.scala`.

Aravindh-Raju (commented):
Test should be added. Looks good otherwise while tested locally!

modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipService.scala
Jay07GIT: Added unit tests for `getUserDetails`

nspadaccino (approved):
Looks good, users route is returning groups list as expected. Would we also be able to add the groups info to the `users/lookupuser` route?

arpit4ever (approved):
LGTM

Aravindh-Raju (approved):
LGTM!

PR #1333, opened 2023-12-12
Hide zone change for users not in admin group

### LINKED ISSUE #1332: Hide zone change for users not in admin group
Currently zone change history tab in portal and the route in api allows all users to get a zone's change history. Ideally only the members and admins of the zone's admin group, super users and support users should be able to see these details.

Hide zone change history for users not being part of the admin group and is not a super or support user.

Aravindh-Raju:
Fixes #1332

Changes in this pull request:
- Hidden zone change history for non-members in portal. Made restrictions to not access zone change history through api for users of the zone who are not part of the zone's admin group or a support/super user.


nspadaccino:
Currently normal users aren't able to access private zones owned by other groups in the portal and API at all, so this change may be redundant

Aravindh-Raju:
> Currently normal users aren't able to access private zones owned by other groups in the portal and API at all, so this change may be redundant

This is for users who have ACL rules applied. Even if they are applied a read only ACL rule, they can still see zone details and zone changes while they should only see records and record changes and not anything related to zones. As ACL rules are only for records and even with delete ACL rule applied they can't make any changes to zones. So I thought of hiding those from them.

arpit4ever (approved):
LGTM

PR #1335, opened 2024-01-10
Filter batch changes with their status

### LINKED ISSUE #1245: Add Search and Filter to DNS Changes All Requests tab
Currently it's very difficult to track down a specific DNS change in the All Requests tab if you don't know the change ID. Being able to search/filter by submitter and datetime range would make this much easier. We could also add a filter all the change statuses to find failed and pending processing changes.

Aravindh-Raju:
Fixes #1245.

Changes in this pull request:
- Users can now filter DNS Changes by batch change status in the `All Requests` tab.


nspadaccino (changes requested):
Schema version needs to be updated so we can test in the prod snapshot. We'll also need a script to populate the batch status values in the table

Aravindh-Raju:
Removed the filter from portal and disabled it from api route for now until we populate the column as discussed in the standup.

nspadaccino (approved):
Looks good, batch statuses are getting updated in the db when I ran locally

arpit4ever (approved):
LGTM

PR #1338, opened 2024-01-30
Fix record change history view in portal

### LINKED ISSUE #1337: Fix issues with record change history column
**Describe the bug**

- The column name seems to be hidden in the portal as it fails some condition. Make the column name always visible. 
- Make the `View History` button always visible. But disable if the correct permissions are not met.

Aravindh-Raju:
Fixes #1337.

Changes in this pull request:
- Fixed the column name by making it permanent. 
- Allowed every users to view the history of unowned records.
- For records in the private zones, only the users who belong to the owning group can view the history.
- The button will be disabled if the users don't have the right permissions.


nspadaccino (approved):
Looks great, tested against prod snapshot

PR #1342, opened 2024-02-08
Display info message for failed single changes

### LINKED ISSUE #1340: Improve Batch Change Error Messages
Currently, if a subset of record changes in a batch change fail, the batch change view will only display those changes as failed, with no extra information about the nature of the failures. 
In the zone view, users can see the more detail error messages. These errors are typically one of the following:

Failed validating update to DNS for change "xxxxxxx": "xxxxxx": This record set is out of sync with the DNS backend; sync this zone before attempting to update this record set. 

Failed validating update to DNS for change "xxxxxxx": "xxxxxx": Incompatible record in DNS. 

Failed validating update to DNS for change "xxxxxxx": "xxxxxx": referral

We should have the batch change view also display these messages. We could also parse these messages to provide more info on how to proceed. For the first two messages, the solution would be a zone sync. For the last message (referral), we should let the user know that there is a conflict due to the record being added having the same name as an NS record in the same zone.

Aravindh-Raju:
Fixes #1340.

Changes in this pull request:
- Made info message for failed single changes to be displayed in the portal for that batch change.
- Changed the existing message to a more informational one.

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/access/AccessValidations.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/access/AccessValidations.scala
@@ -38,6 +38,13 @@ class AccessValidations(
         .isGroupMember(zone.adminGroupId) || userHasAclRules(auth, zone)
     )
 
+  def canSeeZoneChange(auth: AuthPrincipal, zone: Zone): Either[Throwable, Unit] =
+    ensuring(
+      NotAuthorizedError(s"User ${auth.signedInUser.userName} cannot access zone '${zone.name}' changes")
+    )(
+      auth.isSystemAdmin || zone.shared || auth.isGroupMember(zone.adminGroupId)
+    )
+
   def canChangeZone(
       auth: AuthPrincipal,
       zoneName: String,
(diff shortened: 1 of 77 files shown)


modules/api/src/main/scala/vinyldns/api/engine/RecordSetChangeHandler.scala
```diff
@@ -35,6 +35,11 @@ object RecordSetChangeHandler extends TransactionProvider {
   private implicit val cs: ContextShift[IO] =
     IO.contextShift(scala.concurrent.ExecutionContext.global)
 
+  private val outOfSyncFailureMessage: String = "This record set is out of sync with the DNS backend; sync this zone before attempting to update this record set."
+  private val incompatibleRecordFailureMessage: String = "Incompatible record in DNS."
+  private val syncZoneMessage: String = "This record set is out of sync with the DNS backend. Sync this zone before attempting to update this record set."
+  private val recordConflictMessage: String = "Conflict due to the record being added having the same name as an NS record in the same zone."
```
nspadaccino, line 41: During testing I found that this also happens for deletes and updates, not just adds. So I think we can change the message to:

"Conflict due to the record having the same name as an NS record in the same zone. Please create the record using the DNS service the NS record has been delegated to (ex. AWS r53), or use a different record name."

nspadaccino (commented):
Changes look good, one comment

nspadaccino (approved):
LGTM

arpit4ever (approved):
LGTM

PR #1343, opened 2024-02-12
Improve CSV Import Error Message

### LINKED ISSUE #1341: Enhace CSV Import error messages
Currently, if a user attempts to import a CSV file and the import fails, Vinyl will throw a generic error message saying:

`Import failed. Not a valid file. File should be of ‘.csv’ type.`

This will happen even if the file is indeed a csv file. Typically the issue is that the CSV header is misconfigured or missing. We should improve the error handling to let users know when their header is wrong, and provide the correct header in the error message so they can fix it.

nspadaccino:
Fixes #1341

CSV batch change import error message has been improved, there is now an error for the file type and and error for the header being incorrect.
The scala ImportError variable has been removed since it wasn't being used anywhere.

### DIFF
--- a/modules/core/src/main/scala/vinyldns/core/Messages.scala
+++ b/modules/core/src/main/scala/vinyldns/core/Messages.scala
@@ -64,9 +64,6 @@ object Messages {
   // Error displayed when NSData field is not a positive integer
   val NSDataError = "NS data must be a positive integer"
 
-  // Error displayed when importing files other than .csv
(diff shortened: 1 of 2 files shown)


Aravindh-Raju (approved):
Tested locally. LGTM!

Jay07GIT (approved):
Looks Perfect to me.

arpit4ever (approved):
LGTM

PR #1348, opened 2024-03-01
Ownership transfer for records

Jay07GIT:
Fixes https://github.com/vinyldns/vinyldns/issues/1218.

Added ownership transfer validation in API and Portal

List of Validations added: (Specific to shared zone)

User can claim their own group for the unowned record by their own. Group will be Auto-assigned to the record, once user requested.
User can request their own group for the record already assigned. Once requested, the request will be send to the existing assigned group members(Approvers). The request will be notified in email.
Approver can Approve/Reject the request.
1. If the Approver approved the request, the requested group will be assign to the particular record.
2. If the Approver rejected the request, the existing group assigned will be remains same.
Steps to test:

Create 2 user. **_Professor, fry_**.
Create 2 groups for each members.
Create a shared zone.
Now switch the use and try to give request and approve.(consider one as requester and one as approver.)

Not included in this PR:
Ownership transfer history and filter records with ownership transfer status will be complete in separate PR.


Jay07GIT:
SInce faced unsigned issue in the [PR commit](https://github.com/vinyldns/vinyldns/pull/1229/). Opened this as a new one and closing the existing PR.

nspadaccino (changes requested):
I noticed a few bugs:

• When the user is part of the shared zone's owner group, the ownership transfer button does not appear on any of the records in the zone
Screenshot 2024-03-11 at 11 38 04 PM
• Columns look squished on default zoom in web browser (record type and ttl values are running over into two lines)
• Transfer request can be submitted even if no group was chosen
• Once a transfer request has been raised/approved in a zone, record updates in that zone cannot be submitted. When the "Yes" button is clicked, nothing happens
Screenshot 2024-03-11 at 11 41 21 PM
• On a record which has had a successful ownership transfer, when you click the update button with a non-owner user, the record owner group box is blank
Screenshot 2024-03-11 at 11 43 11 PM
• When you go to request an ownership transfer of an owned record, both boxes are blank
Screenshot 2024-03-11 at 11 45 04 PM
•

nspadaccino:
Tested with the latest code, all my previous comments have been resolved. I found one more bug:

When I go to close an ownership transfer request, the Record Owner Group Status box is blank, and I am able to submit the request without choosing anything.
Screenshot 2024-03-19 at 9 50 16 AM

Jay07GIT:
@nspadaccino Fixed the issue.

arpit4ever (dismissed):
Ownership transfer functionality looks good.

modules/api/src/main/resources/reference.conf
```diff
@@ -161,7 +161,7 @@ vinyldns {
     }
   }
 
-  notifiers = []
+  notifiers = ["email"]
```
Aravindh-Raju, line 164: I don't see this config in `application.conf`. We will have to add it there with the right details as mentioned in the [operator guide](https://www.vinyldns.io/operator/config-api.html#notifiers) before or after deploying for it to work.

modules/api/src/main/scala/vinyldns/api/notifier/email/EmailNotifier.scala
```diff
@@ -81,10 +78,52 @@ class EmailNotifier(config: EmailNotifierConfig, session: Session, userRepositor
...
+      }
       case _ => IO.unit
     }
+  }
 
+  def sendRecordSetOwnerTransferNotification(rsc: RecordSetChange): IO[Unit] = {
+    val toUser =
+      for{
+        group  <- groupRepository.getGroup(rsc.recordSet.ownerGroupId.getOrElse("<none>"))
+        member <- userRepository.getUser(group.map(_.memberIds.head).getOrElse("<none>"))
```
Aravindh-Raju, line 92: Just for my clarification, we will always get the first user/member of the group from the previous step here, since we are using `memberIds.head`. So I guess only the first user/member will receive the email. Is this the intended behaviour?

modules/api/src/main/scala/vinyldns/api/notifier/email/EmailNotifier.scala
```diff
@@ -81,10 +78,52 @@ class EmailNotifier(config: EmailNotifierConfig, session: Session, userRepositor
...
+        user  <- userRepository.getUser(member.get.id)}
+      yield user
+
+    val cCUser =
+      for{
+        group  <- groupRepository.getGroup(rsc.recordSet.recordSetGroupChange.map(_.requestedOwnerGroupId.getOrElse("<none>")).getOrElse("<none>"))
+        member <- userRepository.getUser(group.map(_.memberIds.head).getOrElse("<none>"))
```
Aravindh-Raju, line 99: Same doubt as above here.

Aravindh-Raju (approved):
Few things to test out in dev env related to email notifier. Looks good otherwise!

nspadaccino:
Two things I found when testing:
Screenshot 2024-04-23 at 3 48 25 PM
It looks like the Type and TTL columns don't have enough space

Screenshot 2024-04-23 at 4 03 13 PM
This error message should be changed to "Transfer status is required" or something similar

Otherwise all the functionality looks good

modules/api/src/main/resources/reference.conf
Jay07GIT: Right! . We refers `reference.conf` only, not using email notifier flag from `application.conf`. 
I removed the `email` from `reference.conf`, Since we not use in localhost.

modules/api/src/main/scala/vinyldns/api/notifier/email/EmailNotifier.scala
Jay07GIT: This has been intentionally done for test the email. But, I guess we can good to test in dev So, I updated the code as expected.

modules/api/src/main/scala/vinyldns/api/notifier/email/EmailNotifier.scala
Jay07GIT: Answered in above comments.

nspadaccino (approved):
Looks good, all my previous findings were resolved

arpit4ever (approved):
LGTM

PR #1357, opened 2024-04-18
Remove unnecessary parameters from "listRecordSetChanges"

### LINKED ISSUE #1356: Remove unnecessary parameters from "listRecordSetChanges"
**Describe the bug**

The parameters `fqdn` and `recordType` seem to have added by mistake to `listRecordChanges` route as seen here: [RecordsetRouting](https://github.com/vinyldns/vinyldns/blob/master/modules/api/src/main/scala/vinyldns/api/route/RecordSetRouting.scala#L233). Those parameters are only used by `listRecordSetChangeHistory` [route](https://github.com/vinyldns/vinyldns/blob/master/modules/api/src/main/scala/vinyldns/api/route/RecordSetRouting.scala#L252) and have no usage in `listRecordChanges`. 

Those parameters needs to be removed to avoid any confusion in the future.

Aravindh-Raju:
Fixes #1356.

Changes in this pull request:
- Removed the parameters `fqdn` and `recordType` from the `listRecordSetChanges` route as the only parameters in this route are `startFrom` and `maxItems`. The `fqdn` and `recordType` parameters are only used in `listRecordSetChangeHistory` route.
- Reuse `toFqdn` method to save the fqdn to database.

### DIFF
--- a/modules/api/src/test/scala/vinyldns/api/domain/record/RecordSetServiceSpec.scala
+++ b/modules/api/src/test/scala/vinyldns/api/domain/record/RecordSetServiceSpec.scala
@@ -1937,19 +1937,22 @@ class RecordSetServiceSpec
       val completeRecordSetChanges: List[RecordSetChange] =
         List(pendingCreateAAAA, pendingCreateCNAME, completeCreateAAAA, completeCreateCNAME)
 
+      doReturn(IO.pure(Some(zoneActive)))
+        .when(mockZoneRepo)
+        .getZone(zoneActive.id)
       doReturn(IO.pure(ListRecordSetChangesResults(completeRecordSetChanges)))
         .when(mockRecordChangeRepo)
-        .listRecordSetChanges(zoneId = Some(okZone.id), startFrom = None, maxItems = 100, fqdn = None, recordType = None)
+        .listRecordSetChanges(zoneId = Some(zoneActive.id), startFrom = None, maxItems = 100, fqdn = None, recordType = None)
       doReturn(IO.pure(ListUsersResults(Seq(okUser), None)))
         .when(mockUserRepo)
         .getUsers(any[Set[String]], any[Option[String]], any[Option[Int]])
 
       val result: ListRecordSetChangesResponse =
-        underTest.listRecordSetChanges(Some(okZone.id), authPrincipal = okAuth).value.unsafeRunSync().toOption.get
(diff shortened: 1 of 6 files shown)


Jay07GIT (approved):
LGTM.

modules/api/src/test/scala/vinyldns/api/domain/record/RecordSetServiceSpec.scala
```diff
@@ -1937,19 +1937,22 @@ class RecordSetServiceSpec
       val completeRecordSetChanges: List[RecordSetChange] =
         List(pendingCreateAAAA, pendingCreateCNAME, completeCreateAAAA, completeCreateCNAME)
 
+      doReturn(IO.pure(Some(zoneActive)))
+        .when(mockZoneRepo)
+        .getZone(zoneActive.id)
       doReturn(IO.pure(ListRecordSetChangesResults(completeRecordSetChanges)))
         .when(mockRecordChangeRepo)
-        .listRecordSetChanges(zoneId = Some(okZone.id), startFrom = None, maxItems = 100, fqdn = None, recordType = None)
+        .listRecordSetChanges(zoneId = Some(zoneActive.id), startFrom = None, maxItems = 100, fqdn = None, recordType = None)
```
arpit4ever, line 1945: `fqdn` and `recordType` are removed. Tests should reflect the same.

Aravindh-Raju: The `listRecordSetChanges` we see here is a method in our mysql module which is used to retrieve both `recordsetchanges` and `recordchangehistory` as they both return list of recordset changes. We are required in unit test to pass these parameters even though there is a default value set.

But the route and it's corresponding service method don’t have these parameters anymore. It's also removed from the test method: https://github.com/vinyldns/vinyldns/pull/1357/files#diff-82570c9fb20890f64d82bbdfa19c5933d39037e68789e03b1bcb76dc7e73e70eL764

Aravindh-Raju: Both the service method and mysql method has the name `listRecordSetChanges`. The service method which corresponds to the route doesn't have these params anymore. These parameters are required in the mysql method though.

arpit4ever (approved):
LGTM

nspadaccino (approved):

PR #1368, opened 2024-05-23
Update get record change route

### LINKED ISSUE #1367: Update Get Recordset Change Route
Currently the Get Recordset Change has a Segment which is of no use. It's not passed on to any service methods or used anywhere. We should remove that to avoid any confusion.

For eg in the below api, the segment next to `recordsets` is of no use. It can be anything like dummy, test or whatever and it'll still work as it is not used anywhere or provide any use:
```
https://localhost:9000/zones/3db0d8y4-967c-46fa-9560-560b11m5cb38/recordsets/{this can be anything. it doesn't matter}/changes/09950016-14bd-469a-bc62-9e8b0zlka69
```

Aravindh-Raju:
Fixes #1367.

Changes in this pull request:
- Removed the unnecessary segment to avoid any confusion as it's not used anywhere. Change the route a bit to match the other recordsetchange routes.

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetService.scala
@@ -567,7 +567,7 @@ class RecordSetService(
     for {
       zone <- getZone(zoneId)
       change <- recordChangeRepository
-        .getRecordSetChange(zone.id, changeId)
+        .getRecordSetChange(changeId)
         .orFail(
           RecordSetChangeNotFoundError(
             s"Unable to find record set change with id $changeId in zone ${zone.name}"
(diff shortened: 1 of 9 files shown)


nspadaccino (approved):
Looks good, tested API locally

nspadaccino:
The clients that use this route will have to be updated as well:
https://github.com/vinyldns/vinyldns-python/blob/master/src/vinyldns/client.py#L602
https://github.com/vinyldns/go-vinyldns/blob/main/vinyldns/recordsets.go#L248
https://github.com/vinyldns/vinyldns-java/blob/master/src/main/java/io/vinyldns/java/model/record/set/GetRecordSetChangeRequest.java#L19
https://github.com/vinyldns/terraform-provider-vinyldns/blob/master/vinyldns/resource_record_set.go#L393
https://github.com/vinyldns/vinyldns-js/blob/master/src/urls.js#L62
https://github.com/vinyldns/vinyldns-cli/blob/main/src/record_sets.go#L54

PR #1374, opened 2024-07-19
Fix for latency issue in batch change upload

### LINKED ISSUE #1375: Latency issue in batch change upload causes 503 time out in api.
**Describe the bug**
Latency issue in batch change upload causes 503 time out in api.

Root Cause:
Since the batch upload will take much time to give response from the SQL side, our akka server gets 503 timeout, but in backend we gets the successful response.

Jay07GIT:
Fixes #1375.

Changes in this pull request:
- Fix for latency issue in batch change upload.

Root Cause:
Since the batch upload will take much time to give response from the SQL side, our akka server gets `503 timeout`, but in backend we gets the successful response.

Fix: 
Increased the akka http timeout to `60seconds` (default is 20s) and keeps the portal in loading state untill we gets the response from api.

### DIFF
--- a/build/docker/api/application.conf
+++ b/build/docker/api/application.conf
@@ -350,6 +350,7 @@ akka.http {
     # The time period within which the TCP binding process must be completed.
     # Set to `infinite` to disable.
     bind-timeout = 5s
+    request-timeout = 60s
 
     # Show verbose error messages back to the client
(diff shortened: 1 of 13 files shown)


nspadaccino (approved):
LGTM

PR #1376, opened 2024-08-05
Optimize batch change

Aravindh-Raju:
Changes in this pull request:
- Reduced the use of calling certain functions in multiple places.
- Added a few configs and change to batch change structure to help with timeouts.
- Removed hikari logs as it keeps pushing trace level logs every few seconds making it hard to debug code related logs.


nspadaccino (approved):
LGTM

PR #1361, opened 2024-05-07
Update info messages

Aravindh-Raju:
Fixes #1360.

Changes in this pull request:
- When a record data doesn't exist, updated the case to display the below message:
`The record data entered doesn't exist. Please enter the correct record data or leave the field empty if it's a delete operation.`

PR #1382, opened 2024-08-21
Revert "Update info messages"

Aravindh-Raju:
Reverts vinyldns/vinyldns#1361

Since we already display a message when record data entered does not exist `ℹ️ Record data entered does not exist. No further action is required.`, I don't think we need this here. As this is not the right place to display this message and should only display message we might get from DNS servers (like zone out of sync, record conflict, etc).


Aravindh-Raju:
Closing this as reverting still doesn't solve the issue. Will look more into it.

PR #1383, opened 2024-08-21
added batch-change-limit params in portal config

Jay07GIT:
Changes
- Added batch-change-limit params in portal config

### DIFF
--- a/modules/portal/conf/application.conf
+++ b/modules/portal/conf/application.conf
@@ -88,6 +88,9 @@ api {
   }
 }
 
+# Batch change limit
+batch-change-limit = 1000
+
 http.port = 9001
 http.port = ${?PORTAL_PORT}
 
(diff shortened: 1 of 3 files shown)


arpit4ever (approved):
LGTM

modules/portal/conf/application.conf
```diff
@@ -88,6 +88,9 @@ api {
   }
 }
 
+# Batch change limit
```
Aravindh-Raju, line 91: Just for my clarification, shouldn't we add this inside api { limits {} } in the conf? Just a couple of lines above here: https://github.com/vinyldns/vinyldns/pull/1383/files#diff-521ead018f215072950ade8af5030c97838cb065d2bde54423bd2cf39852b258R87? Why is only this config outside of that limits block?

Aravindh-Raju (approved):
Looks good!

modules/portal/conf/application.conf
Jay07GIT: Good finding. We use the same setup in deployment configs. So i had given it outside the limits{}

JoshSEdwards (approved):
LGTM

PR #1389, opened 2024-09-02
Fix batch update logic

### LINKED ISSUE #1388: Bug in batch update
**Describe the bug**

When a user submits a Add + Delete for a record that doesn't already exist, we currently just `Add` the record in the batch and ignore the `Delete`. 

Similarly when a user submits a Add + Delete for a record that already exist, with a record data in Delete that doesn't exist for the record, it should ideally ignore the Delete and just process the `Add` to update the existing record. But we get the following error right now:
```
The record data entered doesn't exist. Please enter the correct record data or leave the field empty if it's a delete operation.
```

This shouldn't be the case and results in batch performing in weird ways sometimes. The logic should be updated to process this without throwing an error.

Attaching screenshot for reference.
Screenshot 2024-09-02 at 3 21 28 PM

Aravindh-Raju:
Fixes #1388.

Changes in this pull request:
- Now when a user submits Add + Delete for a record that already exist with a record data in Delete that doesn't exist, we ignore the Delete and just process the Add to update the record. Which is the ideal behaviour.

Attaching screenshot for reference.
Screenshot 2024-09-02 at 3 23 40 PM

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/engine/RecordSetChangeHandler.scala
+++ b/modules/api/src/main/scala/vinyldns/api/engine/RecordSetChangeHandler.scala
@@ -195,10 +195,14 @@ object RecordSetChangeHandler extends TransactionProvider {
           else AlreadyApplied(change) // we did not find the record set, so already applied
 
         case RecordSetChangeType.Sync =>
-          Failure(
-            change,
-            outOfSyncFailureMessage
-          )
+          if (existingRecords.nonEmpty) {
+            Failure(
+              change,
+              outOfSyncFailureMessage
+            )
+          } else {
+            AlreadyApplied(change)
+          }
       }
     }
 
--- a/modules/api/src/main/scala/vinyldns/api/engine/RecordSetChangeHandler.scala
+++ b/modules/api/src/main/scala/vinyldns/api/engine/RecordSetChangeHandler.scala
@@ -38,7 +38,6 @@ object RecordSetChangeHandler extends TransactionProvider {
   private val outOfSyncFailureMessage: String = "This record set is out of sync with the DNS backend; sync this zone before attempting to update this record set."
   private val incompatibleRecordFailureMessage: String = "Incompatible record in DNS."
   private val syncZoneMessage: String = "This record set is out of sync with the DNS backend. Sync this zone before attempting to update this record set."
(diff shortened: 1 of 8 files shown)


modules/api/src/main/scala/vinyldns/api/engine/RecordSetChangeHandler.scala
```diff
@@ -38,7 +38,6 @@ object RecordSetChangeHandler extends TransactionProvider {
   private val outOfSyncFailureMessage: String = "This record set is out of sync with the DNS backend; sync this zone before attempting to update this record set."
   private val incompatibleRecordFailureMessage: String = "Incompatible record in DNS."
   private val syncZoneMessage: String = "This record set is out of sync with the DNS backend. Sync this zone before attempting to update this record set."
-  private val wrongRecordDataMessage: String = "The record data entered doesn't exist. Please enter the correct record data or leave the field empty if it's a delete operation."
```
Jay07GIT, line 41: Do we want to add a tests for error message?

Aravindh-Raju: The error message previously added was removed. 

If it's the other error messages, there are already tests for `out of sync` messages we have here. And for `incompatibleRecordFailureMessage`, I am not sure we can produce this locally. This comes off from DNS server.

nspadaccino:
One thing I noticed while testing:

I created the record foo2.ok in the local environment. I then manually removed it from the database, so the record still exists on the backend.
When I try to update the record, I get the correct zone out of sync error:
Screenshot 2024-09-10 at 3 45 59 PM

But when I try to just delete the record, it passes with a "record does not exist" message.
Screenshot 2024-09-10 at 3 46 38 PM

Is there a way to also provide the zone out of sync error for independent deletes?

Aravindh-Raju:
> One thing I noticed while testing:
> 
> I created the record foo2.ok in the local environment. I then manually removed it from the database, so the record still exists on the backend. When I try to update the record, I get the correct zone out of sync error: Screenshot 2024-09-10 at 3 45 59 PM
> 
> But when I try to just delete the record, it passes with a "record does not exist" message. Screenshot 2024-09-10 at 3 46 38 PM
> 
> Is there a way to also provide the zone out of sync error for independent deletes?

Handled this case in the latest commits.

nspadaccino (approved):
looks good, getting the out of sync message when I try to delete a record i manually removed from the database

PR #1393, opened 2024-09-16
Reset the batch file name when re-upload the same file again

### LINKED ISSUE #1394: Re-upload the same file again in batch change should take the latest file in batch change
**Describe the bug**
When user re-upload the same file again in batch change by making changes, latest file is not picking by the batch change in vinyl.

**Steps to reproduce.**
Click upload csv in batch change
select the file vinyldns_batch.csv and click upload
modify the csv
click upload csv>> select the same file >> click upload.
_Application will still shows the old csv data._

Jay07GIT:
Fixes #1394.

Changes in this pull request:
- Reset the batch file name when re-upload the same file with the changes again.

### DIFF
--- a/modules/portal/public/lib/dns-change/dns-change-new.controller.js
+++ b/modules/portal/public/lib/dns-change/dns-change-new.controller.js
@@ -162,16 +162,32 @@
                 $scope.alerts.push(alert);
             }
 
+            function resetFileInput() {
+              $scope.csvInput = null;
+              var inputElement = document.getElementById('batchChangeCsv');
+              if (inputElement) {
+                inputElement.value = null;
+              }
+              if ($scope.createBatchChangeForm && $scope.createBatchChangeForm.batchChangeCsv) {
+                $scope.createBatchChangeForm.batchChangeCsv.$setViewValue(null);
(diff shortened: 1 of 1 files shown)


modules/portal/public/lib/dns-change/dns-change-new.controller.js
```diff
@@ -162,16 +162,32 @@
...
                 }, function(error) {
                     $scope.alerts.push({type: 'danger', content: error});
                 });
 
                 function parseFile(file, batchChangeLimit) {
                   return $q(function(resolve, reject) {
-                    if (!file.name.endsWith('.csv')) {
+                     if (!file || !file.name) {
+                        $log.debug('No file selected or file has no name property');
+                      }
```
Aravindh-Raju, line 189: Just for my clarification, will the if block code gets executed in some case? A file name can never be empty by default and we are going to upload some kind of file for sure. And if the file is not of csv type, we throw the error in if else block. So is there any reason this if block is added?

Aravindh-Raju (approved):
LGTM!

nspadaccino (approved):
lgtm

PR #1398, opened 2024-09-30
Fix batch delete

Aravindh-Raju:
Changes in this pull request:
- When there is a record that exist in Vinyl and we issue a batch change with Delete for that record with a non-existing record data, it throws `out of sync` error. But it should not throw any error and process with info `Record data does not exist. No action is required`. Fixed that in this PR.


arpit4ever (approved):
Updated error looks good.
Approved.

nspadaccino (approved):
LGTM

PR #1400, opened 2024-10-02
Mysql 8.0 compatibility

nspadaccino:
In mysql 8.0, "group" and "groups" are reserved keywords, so they cannot be referenced as columns or tables without using backtick quotes. I have updated that everywhere those terms are used in queries and migrations. It seems that those are the only reserved words we currently use in the database.


Aravindh-Raju (commented):
Tested locally. Everything works as expected.

Can we add backticks for groups in this file as well: https://github.com/vinyldns/vinyldns/blob/master/modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlGroupRepositoryIntegrationSpec.scala#L53

arpit4ever (approved):
LGTM

PR #1401, opened 2024-10-03
Added search by group name in DNS change for api and portal.

### LINKED ISSUE #1407: Add search by group name in DNS change for api and portal.
Add search by group name in DNS change for api and portal.

Now user not able to view the DNS change based on the groups. With this feature user can able to see, which group member done DNS change.

Jay07GIT:
Fixes #1407 .

Changes in this pull request:
- Added search by group name in All DNS change for api and portal.

image


Jay07GIT:
> Screenshot 2024-10-17 at 2 15 18 PM
> Submitter has a column which makes it easier for us to see and search for that submitter. I think it might be good to add a group name column so that it'll be easier to see that and use it for the new search by group name. We can wait for others to see if we should do this.

Resolved the issue in new commit

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -596,8 +601,11 @@ class BatchChangeService(
     val startDateTime = if(dateTimeStartRange.isDefined && dateTimeStartRange.get.isEmpty) None else dateTimeStartRange
     val endDateTime = if(dateTimeEndRange.isDefined && dateTimeEndRange.get.isEmpty) None else dateTimeEndRange
     for {
+      mId <- membershipService.listMyGroups(groupName, None, maxItems,auth,false,false).
```
nspadaccino: mId is computed unconditionally but only used when groupName.isDefined. This adds a membership-service call to every batch-change listing and couples the two. Can we move the lookup inside if (groupName.isDefined) so it only runs when needed?

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
```diff
@@ -596,8 +601,11 @@ class BatchChangeService(
...
+        map(_.groups.map(_.members.map(_.id).mkString("', '")).mkString).getOrElse("None").toBatchResult
```
nspadaccino: Two problems on this line, both fixed by the same change:

1. Substring over-match: listMyGroups filters names with name.contains(...) (MembershipService.scala:239), so "search by group X" resolves members of every group you're in whose name contains X, not just the group named X. That substring behavior is intended for the groups-list UI, so the fix belongs here, not in listMyGroups.
2. ID corruption: the outer .mkString has no separator, so when more than one group comes back the per-group ID strings glue at the seam — groups a,b and c,d produce "a', 'bc', 'd" -> bc.user_id in ('a', 'bc', 'd'), dropping b/c and querying a non-existent bc. Single-group happens to work, which is why the tests miss it.

Filtering to the exact group name and flattening fixes both:

`  .map(_.groups.filter(_.name == groupName.getOrElse("")).flatMap(_.members.map(_.id)).mkString("', '"))`

Exact-match scopes results to the named group, and the single .mkString("', '") removes the seam-merge. Was exact-match the intent here? (Access scope is fine either way — ignoreAccess=false keeps it limited to the caller's own groups.) Worth adding a two-matching-groups test to lock this in.

nspadaccino: `.getOrElse("None")` hides membership errors here. A failed listMyGroups becomes the string "None", which turns into bc.user_id in ('None') and returns empty instead of erroring. Let's keep it in the EitherT chain so failures propagate.

nspadaccino (changes requested):

modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeService.scala
Jay07GIT: Updated in latest commit

PR #1415, opened 2024-11-08
Query updates for MySQL 8

nspadaccino:
"groups" is now a reserved keyword, so there must be backticks around every usage of that keyword


JoshSEdwards (approved):
lgtm

arpit4ever (approved):
LGTM

PR #1417, opened 2024-11-15
Fix batch status display

Aravindh-Raju:
Changes in this pull request:
- Reverted the changes made to the get batch change and get batch changes method. Removed the unpopulated `batch_status` column values with previous way of getting batch status. Note: Even after the new column is populated completely, we can still use the previous method and use the column only for filtering as basically both would result in the same batch status.
- Removed batch status field from BatchChange case class as it's not really necessary to have it as we already have those in other type while sending the response. Note: As in the previous point, we could have batch status field in Batch Change and make use of the column. But it's not necessary.


nspadaccino (approved):
Looks good, when I tested with the prod snapshot it resolved the issue

PR #1418, opened 2024-11-20
Updated email config with password authentication

### LINKED ISSUE #1419: Enable password authentication for email
Enable password authentication for smtp host when smtp `auth` is true in config

Jay07GIT:
Fixes #1419 .

Changes in this pull request:

- Updated email config with password authentication when smtp auth is enabled in config


Jay07GIT:
> Looks good, but I was wondering if the default application.conf files should be changed along with this?

Its good idea. But i guess we can have in [api-config docs](https://github.com/vinyldns/vinyldns/blob/master/modules/docs/src/main/mdoc/operator/config-api.md) about the email authentication details instead of adding in app.conf.

PR #1420, opened 2024-11-20
Fix records repetition in recordset search

### LINKED ISSUE #1412: Remove duplicate records in output from searches
**Describe the bug**
1. A description of the bug (expected behavior vs actual behavior)
        Anytime a searched "Record Name" (or RecordSet) has more than one "Record Data" result, the results are displayed multiple times. i.e. if there are x Record Data results, the same, exact information is repeatedly displayed x times (in the UI and API). This make parsing the results very difficult, due to having to handle the duplicated data.

3. The VinylDNS version which contains the bug
Current version, as of 2024-11-06 (and it's been like this for a while, now). 
As I recall, years ago, this didn't happen (back in 2019?)

4. Any steps to reproduce (if we can't reproduce it, we can't fix it!)
     NS record types have 5 "Record Data" matches - and the exact same information is repeated 5 times
Example: (search for)  *.10.dnsx.comcast.net (and add filter "NS")

5. Any other helpful information (stack trace, log messages, screenshots, etc)
Nothing to add, here. 
As mentioned, back in 2019, I don't recall this happening. The results would only be displayed on time.

Aravindh-Raju:
Fixes #1412.

Changes in this pull request:
- Made query change to fix duplicate records being retrieved.

Steps taken to reproduce and fix the error locally:
- In the RecordSetService.scala file under the searchRecordset function, we have `useRecordSetCache`. When working locally it's turned off, but in production, it may be turned on, which is what is responsible for duplicate records being displayed. So we should set it as `true` to reproduce this error locally.
- Once the above is done, just run the platform. In an existing record, add another record data. Then go to Global RecordSet Search and search for the record. You will see two rows for the same record with same values.
- Then we can add the changes in this PR to the code and repeat the above two. This time only one row will be displayed for a record even if it has multiple record data which is the intended behaviour.


nspadaccino (approved):
looks good!

PR #1421, opened 2024-12-02
Search Groups by user name in All groups

### LINKED ISSUE #1422: Search Groups by user name in All groups tab.
Search Groups by `user name` in All groups tab. This helps user to search the group names based on the Userid or user name

Jay07GIT:
Fixes #1422 .

Changes in this pull request:
- Search Groups by user name in All groups tab.


nspadaccino (commented):
Functionality seems good, but one bug I saw was when I try to search by user with a wildcard (asterisk), I get a `HTTP 401 (Unauthorized): Authentication Failed: Request signature could not be validated `
Seems like wildcard search works when you just type a partial search, but not with an asterisk

modules/portal/public/lib/controllers/controller.groups.js
```diff
@@ -154,33 +155,60 @@ angular.module('controller.groups', []).controller('GroupsController', function
...
         allGroupsPaging = pagingService.resetPaging(allGroupsPaging);
-
-        groupsService
-            .getGroupsAbridged(groupsPaging.maxItems, undefined, false, $scope.query)
+        const groupsSearchByUser = [];
+        if($scope.isSearchByUser){
+        try {
+            if($scope.query==""){
+                throw new Error("User name should not empty.");
+            }
```
Aravindh-Raju, line 163: We can remove this try, alert and catch here and just use the angularjs `ng-required` as follows:
```
<input ng-model="query" type="text" class="form-control all-group-search-text"  placeholder="{{!isSearchByUser ? 'Group Name' : 'User Name'}}" ng-required="isSearchByUser"/>   
```

This avoids us having an alert in portal which we don't have so far and just gives it in place. Attaching a screenshot for reference:
Screenshot 2024-12-12 at 11 00 52 AM

Aravindh-Raju (commented):
A minor change. Functionality looks good otherwise.

nspadaccino (commented):
asterisk search does work now with other characters, except when I enter just an asterisk or just a `%` in the search box and hit enter, I get a 404 error:
`HTTP 404 (Not Found): <!DOCTYPE html> <html lang="en"> <head> <title>Action Not Found</title> <link rel="shortcut icon" href="data:image/`
It should probably return all results, or a warning saying you need at least one other character for wildcard search to work

Jay07GIT:
> asterisk search does work now with other characters, except when I enter just an asterisk or just a `%` in the search box and hit enter, I get a 404 error: `HTTP 404 (Not Found): <!DOCTYPE html> <html lang="en"> <head> <title>Action Not Found</title> <link rel="shortcut icon" href="data:image/` It should probably return all results, or a warning saying you need at least one other character for wildcard search to work

Resolved the issue in latest commit

modules/portal/public/lib/controllers/controller.groups.js
Jay07GIT: Updated the code as suggested

modules/portal/public/lib/controllers/controller.groups.js
```diff
@@ -154,33 +155,65 @@ angular.module('controller.groups', []).controller('GroupsController', function
...
         allGroupsPaging = pagingService.resetPaging(allGroupsPaging);
-
-        groupsService
-            .getGroupsAbridged(groupsPaging.maxItems, undefined, false, $scope.query)
+        const groupsSearchByUser = [];
+        userNameQuery = "";
+        if($scope.isSearchByUser){
+        try {
+            if ($scope.query === "%" || $scope.query === "*"){
+                throw new Error("User name should at least one other character for wildcard search");
```
nspadaccino: *should contain

nspadaccino (changes requested):
Some testing notes:
* Entering a blank username should reset the search, just like when normally searching groups. Right now it prevents you from entering blank in the form when search by user is selected. Also this displays a browser alert, which I don't think we use elsewhere in the portal.
* Search by user checkbox should be inside the same row as the search box, similar to the checkboxes for zone search
* Starting user search with a *, e.g. "*bender", returns "HTTP 401 (Unauthorized): Authentication Failed: Request signature could not be validated"
* Autocomplete is broken for user and group search
* Selecting search by user checkbox in All groups, then navigating to My Groups and submitting a blank search returns error "can't access property "endsWith", $scope.query is undefined"
* Search by user being selected in All Groups breaks My Groups search in general. Say I search by user for "bender", and that returns "bendergroup". Then when I navigate to My Groups, it displays the default results for My Groups but the search by user is still active. It should ideally be cleared when you go back to My Groups.

modules/portal/public/lib/controllers/controller.groups.js
Jay07GIT: @nspadaccino Resolved all the queries.

modules/portal/public/lib/controllers/controller.groups.js
```diff
@@ -151,37 +207,103 @@ angular.module('controller.groups', []).controller('GroupsController', function
...
+                userNameQuery = $scope.query;
+            }
+            function success(response) {
+                $log.debug('profileService::getZoneUserDataByName-success');
+                $scope.response = response.data;
+                const groupMap = $scope.response.groupMap || {};
+                const groupIds = Object.keys(groupMap);
+                $log.debug("getGroupsByUser:groupIds: ", groupIds);
+                $log.debug("getGroupsByUser:groupMap:", groupMap);
+                groupIds.forEach((groupId) => {
```
nspadaccino: Possible race condition here: groupsSearchByUser is populated asynchronously, and then updateAllGroupDisplay (line 255) is called without waiting for the promises to resolve.

modules/portal/public/lib/controllers/controller.groups.js
```diff
@@ -151,37 +207,103 @@ angular.module('controller.groups', []).controller('GroupsController', function
...
+                updateAllGroupDisplay(groupsSearchByUser);
+            }
+            return profileService
+                .getUserDataById(userNameQuery)
+                .then(success)
+                .catch(function (error) {
+                    handleError(error, 'profileService::getZoneUserDataById-failure');
+                });
+        } catch (error) {
+            alert(error.message);
```
nspadaccino: Nitpick, but the rest of the file does alerts using `$scope.alerts.push(alert);`, so we should probably stick to that instead of `alert()`

modules/portal/app/views/groups/groups.scala.html
```diff
@@ -134,34 +134,50 @@
...
+                                                <div class="checkbox" style="margin-top:4px; margin-bottom:0;">
+                                                    <label>
+                                                        <input type="checkbox" ng-model="isSearchByUser">
+                                                        Search by User
+                                                    </label>
+                                                </div>
+                                            </form>
+                                        </div>
                                     </div>
+                                </div>
```
nspadaccino, line 179: I think this </div> is extra, already covered on line 181

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlUserRepository.scala
```diff
@@ -64,7 +64,7 @@ class MySqlUserRepository(cryptoAlgebra: CryptoAlgebra)
     sql"""
          | SELECT data
          |   FROM user
-         |  WHERE ? IN(id, user_name)
+         |    WHERE id = {id} OR user_name LIKE {userName}
```
nspadaccino, line 67: By changing this to LIKE {userName}, this will affect other methods that use this query as well, like MembershipService.getUser. I think keeping this as-is and adding another method/query like searchUsersByName(query: String) to do fuzzy-matching would be better. 
(Related to line 173)

nspadaccino (changes requested):
Couple more comments

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlUserRepository.scala
Jay07GIT: @nspadaccino Function `getUserByIdOrName` is not used elsewhere in the code. Although it exists in `getUser`, it is not actively used, so I have reused it.

modules/portal/app/views/groups/groups.scala.html
Jay07GIT: Resolved

modules/portal/public/lib/controllers/controller.groups.js
Jay07GIT: Resolved

modules/portal/public/lib/controllers/controller.groups.js
Jay07GIT: Resolved

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlUserRepository.scala
nspadaccino: Agreed getUserByIdOrName itself has a single caller. But the endpoint it backs, `GET /api/users/:id`, is also used by getZoneUser and getAclUser in controller.manageZones.js for exact id->userName resolution. Those happen to keep working because they pass UUIDs (the id = match wins), but we've now got one single-resource endpoint doing both exact lookup and fuzzy LIKE search, with .first() and no ORDER BY deciding the winner when multiple usernames match.

Could we split the fuzzy path into its own method/query so getUserByIdOrName keeps its exact-match contract? If you'd rather not, at minimum let's add ORDER BY so the multi-match result is deterministic, plus a test for it.

modules/portal/public/lib/controllers/controller.groups.js
```diff
@@ -151,37 +207,100 @@ angular.module('controller.groups', []).controller('GroupsController', function
...
     };
 
-    $scope.refresh = function () {
-        groupsPaging = pagingService.resetPaging(groupsPaging);
-        allGroupsPaging = pagingService.resetPaging(allGroupsPaging);
+$scope.refresh = function () {
+    groupsPaging = pagingService.resetPaging(groupsPaging);
+    allGroupsPaging = pagingService.resetPaging(allGroupsPaging);
+    const groupsSearchByUser = [];
+    userNameQuery = "";
```
nspadaccino: Minor: 
userNameQuery is assigned with no var/let/const, so it leaks to window. Please scope it `let userNameQuery = "";`

modules/portal/public/lib/controllers/controller.groups.js
```diff
@@ -151,37 +207,100 @@ angular.module('controller.groups', []).controller('GroupsController', function
             });
     };
 
-    $scope.refresh = function () {
-        groupsPaging = pagingService.resetPaging(groupsPaging);
-        allGroupsPaging = pagingService.resetPaging(allGroupsPaging);
+$scope.refresh = function () {
+    groupsPaging = pagingService.resetPaging(groupsPaging);
+    allGroupsPaging = pagingService.resetPaging(allGroupsPaging);
+    const groupsSearchByUser = [];
```
nspadaccino: Minor:
const groupsSearchByUser = [] is never read — it's shadowed by the .then((groupsSearchByUser) => …) parameter on :248. Dead code, can be removed

modules/portal/public/lib/controllers/controller.groups.js
```diff
@@ -151,37 +207,100 @@ angular.module('controller.groups', []).controller('GroupsController', function
...
+    userNameQuery = "";
+    if ($scope.isSearchByUser) {
+        try {
+            if ($scope.query === "%" || $scope.query === "*") {
```
nspadaccino, line 225: The empty / lone-wildcard validation still only hits $log.error, so the user gets no feedback. Should probably push the message onto $scope.alerts like the rest of the controller does, right now nothing surfaces to the UI.

modules/portal/public/lib/controllers/controller.groups.js
```diff
@@ -151,37 +207,100 @@ angular.module('controller.groups', []).controller('GroupsController', function
...
+            } else if ($scope.query.endsWith("%") || $scope.query.endsWith("*")) {
+                userNameQuery = $scope.query.substring(0, $scope.query.length - 1);
+            } else if ($scope.query.startsWith("%") || $scope.query.startsWith("*")) {
+                userNameQuery = $scope.query.substring(1);
+            } else {
+                userNameQuery = $scope.query;
+            }
+
+            function success(response) {
+                $log.debug('profileService::getZoneUserDataByName-success');
```
nspadaccino: update log label

modules/portal/public/lib/controllers/controller.groups.js
```diff
@@ -151,37 +207,100 @@ angular.module('controller.groups', []).controller('GroupsController', function
...
+                    $log.debug('getGroupsByUser:refresh-success', groupsSearchByUser);
+                    updateAllGroupDisplay(groupsSearchByUser);
+                });
+            }
+
+            return profileService
+                .getUserDataById(userNameQuery)
+                .then(success)
+                .catch(function (error) {
+                    handleError(error, 'profileService::getZoneUserDataById-failure');
```
nspadaccino: update log label

nspadaccino (changes requested):

modules/portal/public/lib/controllers/controller.groups.js
Jay07GIT: Updated in latest commit

modules/portal/public/lib/controllers/controller.groups.js
Jay07GIT: Updated in latest commit

modules/portal/public/lib/controllers/controller.groups.js
Jay07GIT: Updated in latest commit

modules/portal/public/lib/controllers/controller.groups.js
Jay07GIT: Updated in latest commit

modules/portal/public/lib/controllers/controller.groups.js
Jay07GIT: Updated in latest commit

modules/portal/app/views/groups/groups.scala.html
```diff
@@ -134,35 +134,50 @@
...
+                                            <button id="refresh-group-button"
+                                                    class="btn btn-default"
+                                                    ng-click="refresh();">
+                                                <span class="fa fa-refresh"></span> Refresh
+                                            </button>
+                                        </div>
+                                        <div class="pull-right">
+                                            <form ng-submit="refresh()" style="text-align:right;">
+                                                <!-- SEARCH BOX -->
+                                                <div class="input-group" style="width:250px; margin-left:auto;">
```
arpit4ever, line 157: Search box style should be same for myGroups and allGroups.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlUserRepository.scala
```diff
@@ -180,11 +180,18 @@ class MySqlUserRepository(cryptoAlgebra: CryptoAlgebra)
    */
   def getUserByIdOrName(userIdentifier: String): IO[Option[User]] =
     monitor("repo.User.getUser") {
+      val userInfo=
```
arpit4ever, line 183: Search for `*prof*` not captured.

modules/portal/app/views/groups/groups.scala.html
```diff
@@ -134,35 +134,50 @@
...
-                                            <div class="input-group">
-                                                <span class="input-group-btn">
-                                                    <button id="group-search-button" type="submit" class="btn btn-primary btn-left-round">
-                                                        <span class="fa fa-search"></span>
-                                                    </button>
-                                                </span>
-                                                <input id="group-search-text" ng-model="query" type="text" class="form-control"  placeholder="Group Name"/>
-                                            </div>
-                                        </form>
+                                    <div class="clearfix">
```
arpit4ever, line 140: Do we need same `clearfix` for myGroups ?

arpit4ever (changes requested):
Key observation:

- *professor works

Failed search

- professor* doesn't work
- prof* doesn't work
- *prof* doesn't work

Failed scenario: 
- On All groups, select Search by User and then click Refresh button. Error Please enter a user name to search for groups
- On All groups, select Search by User and navigate to MyGroups page and click Refresh button. Error Please enter a user name to search for groups

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlUserRepository.scala
Jay07GIT: image
Working for me. Can you reconfirm?

modules/portal/app/views/groups/groups.scala.html
Jay07GIT: Updated the same form class for both myGroups and allGroups. Styling is required to align the “Search by User” checkbox.
`style="text-align:right;"`

modules/portal/app/views/groups/groups.scala.html
Jay07GIT: image
Yeh required. Removing it causes the overlap.

arpit4ever: `clearfix` is present for AllGroups
Same is missing for myGroups.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlUserRepository.scala
nspadaccino: @Jay07GIT Can you take another look at this?   

MySqlUserRepository.getUserByIdOrName has exactly one caller:

  - MembershipService.getUser (MembershipService.scala:361)

  and that method is used in two production paths:

  1. getUserDetails → the API route GET /users/{id} (MembershipRouting.scala:215) — this is your new feature, expected.
  2. UserAccountAccessor / the portal's user-resolution path — not part of this feature.

More importantly, the "used elsewhere or not" question isn't really what the concern is about. The problem is the semantic change: this rewrites WHERE ? 
IN(id, user_name) (exact match) into user_name LIKE '%…%' (substring) followed by .first(). So an exact-identifier lookup can now silently resolve to a different user whose name is a superstring of the input, and .first() picks an arbitrary one among matches. That would be a correctness/security regression for any caller expecting exact resolution — and this codebase clearly treats these as exact lookups (note UserAccountAccessor.get even falls back to a separate getUserByName). Even if this method were only used by the new search, broadening its match semantics in place wouldn't be safe.

Suggested fix: add a dedicated search method for the wildcard/substring case and leave getUserByIdOrName doing exact matching. That keeps the group-search-by-user feature working without changing behavior for existing callers.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlUserRepository.scala
nspadaccino: The wildcard-stripping leaks into the exact-id lookup. userInfo is bound to both id = {id} and user_name LIKE %{userName}%, so stripping, which only makes sense for the name-search path, silently mutates the id path. An identifier like "abc*" becomes id = 'abc', i.e. the last character is chopped before the exact-id comparison. This matters because getUserByIdOrName is also reached by UserAccountAccessor and GET /users/{id}, not just this feature; those callers expect exact id-or-name resolution and never asked for wildcard semantics.

The stripping is also asymmetric: it's if/else if and removes only a single leading or trailing char, so "\*abc\*" strips just the trailing * → "abc*", which then matches nothing by id and gets LIKE %abc*% by name.

Please move both the stripping and the LIKE into a dedicated search method and leave getUserByIdOrName doing exact matching (WHERE ? IN(id, user_name)).

modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipProtocol.scala
```diff
@@ -103,17 +103,17 @@ object UserInfo {
 }
 
 case class UserResponseInfo(
-   id: String,
-   userName: Option[String] = None,
-   groupId: Set[String] = Set.empty
- )
+      id: String,
+      userName: Option[String] = None,
+      groupMap: Map[String, String] = Map.empty
```
nspadaccino: This is a backward-incompatible change to the public GET /users/{id} response. The route (MembershipRouting.scala:213) completes UserResponseInfo directly and there's no custom serializer, so json4s renders the case-class fields as-is — the field name is the JSON key. So the response changes in two ways at once:
  - key groupId → groupMap
  - type array → object: "groupId": ["g1","g2"] becomes "groupMap": {"g1":"Group One","g2":"Group Two"}

The PR's own portal code (controller.groups.js reading search.groupMap) is updated, so it works end-to-end here, but any external client/script reading groupId on this documented endpoint breaks silently.

If the group names are what you need, prefer an additive change — keep groupId and add groupMap — rather than renaming/retyping in place.

modules/portal/public/lib/services/groups/service.groups.js
```diff
@@ -41,7 +41,16 @@ angular.module('service.groups', [])
 
         this.getGroup = function (id) {
             var url = '/api/groups/' + id;
-            return $http.get(url);
+            let loader = $("#loader");
+            loader.modal({
```
nspadaccino: This identical loader block is copy-pasted into six methods. Extract a withLoader($http.get(url)) helper instead of duplicating.

modules/portal/public/lib/controllers/controller.groups.js
```diff
@@ -102,7 +103,78 @@ angular.module('controller.groups', []).controller('GroupsController', function
...
+        open: function () {
+            $(this).removeClass("ui-corner-all").addClass("ui-corner-top");
+        },
+        close: function () {
+            $(this).removeClass("ui-corner-top").addClass("ui-corner-all");
+        }
+    });
+
+    // Autocomplete text-highlight
+    $.ui.autocomplete.prototype._renderItem = function(ul, item) {
```
nspadaccino: This assigns to the shared jQuery UI prototype, so whichever controller's _renderItem ran last wins for every autocomplete on the page — group, zone (controller.zones.js:167), and recordset (recordsets.controller.js:134) search all hit the same slot. That matters here because the three implementations are not equivalent:
  - This PR's version is the hardened one — it HTML-escapes the label ($("<div>").text(...).html()) and escapes the regex term before building the highlight RegExp.
  - The existing zone/recordset versions do String(item.label).replace(new RegExp(this.term, "gi"), "<b>$&</b>"), which is raw interpolation into HTML (XSS-prone) and an unescaped `this.term`.
   
So depending on controller load/init order, the group search may silently get one of the unsafe renderers, or your safe renderer may replace theirs. The right fix is to stop mutating the global prototype and scope the renderer to this widget instance instead, e.g. $("#all-group-search-text").autocomplete("instance")._renderItem = … (or pass it via the widget's options).

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlUserRepository.scala
Jay07GIT: Resolved in latest commits

modules/portal/public/lib/services/groups/service.groups.js
Jay07GIT: Resolved in latest commits

modules/api/src/main/scala/vinyldns/api/domain/membership/MembershipProtocol.scala
Jay07GIT: Resolved in latest commits

modules/portal/public/lib/controllers/controller.groups.js
Jay07GIT: Resolved in latest commits

PR #1423, opened 2024-12-03
Ownership transfer Bug Fix

nspadaccino:
Bug: as of version 0.21.1, when a user tries to update a record in a private zone from the zone view, it fails with a `422: Cannot update RecordSet OwnerShip Status when zone is not shared.` error.

This can be worked around by updating the record via batch change. Also, once that particular record has been updated via batch change, it can be updated in the zone view as well from then on.

It seems like the issue is that in a private zone, the existing record's recordSetGroupChange is simply set to `None`, whereas the updated record's is set to `Some(OwnerShipTransfer: [ownerShipTransferStatus=\"None\"; requestedOwnerGroupId=\"None\"])`.  This causes the unchangedRecordSetOwnershipStatus check to fail. I have updated the check to pass when the existing record's recordSetGroupChange is set to `None`, since that seems to apply only in this case. 

Once the record is successfully updated, then on future updates both the existing and updated records recordSetGroupChange value will be set to `Some(OwnerShipTransfer: [ownerShipTransferStatus=\"None\"; requestedOwnerGroupId=\"None\"])`, so the `updates.recordSetGroupChange == existing.recordSetGroupChange` check will pass as expected.

@Jay07GIT, let me know if you see a better way to fix this, or if this seems OK.

### DIFF
--- a/modules/api/src/it/scala/vinyldns/api/domain/record/RecordSetServiceIntegrationSpec.scala
+++ b/modules/api/src/it/scala/vinyldns/api/domain/record/RecordSetServiceIntegrationSpec.scala
@@ -128,7 +128,10 @@ class RecordSetServiceIntegrationSpec
     RecordSetStatus.Active,
     Instant.now.truncatedTo(ChronoUnit.MILLIS),
     None,
-    List(AAAAData("fd69:27cc:fe91::60"))
+    List(AAAAData("fd69:27cc:fe91::60")),
+    recordSetGroupChange =
+      Some(OwnerShipTransfer(ownerShipTransferStatus = OwnerShipTransferStatus.None,
+        requestedOwnerGroupId = None))
   )
   private val subTestRecordA = RecordSet(
     zone.id,
--- a/modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetService.scala
@@ -27,6 +27,7 @@ import vinyldns.core.domain.zone.{Zone, ZoneCommandResult, ZoneRepository}
 import vinyldns.core.queue.MessageQueue
 import cats.data._
 import cats.effect.IO
+import org.slf4j.{Logger, LoggerFactory}
 import org.xbill.DNS.ReverseMap
 import vinyldns.api.config.{ZoneAuthConfigs, DottedHostsConfig, HighValueDomainConfig}
 import vinyldns.api.domain.DomainValidations.{validateIpv4Address, validateIpv6Address}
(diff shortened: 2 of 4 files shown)


Jay07GIT:
> Bug: as of version 0.21.1, when a user tries to update a record in a private zone from the zone view, it fails with a `422: Cannot update RecordSet OwnerShip Status when zone is not shared.` error.
> 
> This can be worked around by updating the record via batch change. Also, once that particular record has been updated via batch change, it can be updated in the zone view as well from then on.
> 
> It seems like the issue is that in a private zone, the existing record's recordSetGroupChange is simply set to `None`, whereas the updated record's is set to `Some(OwnerShipTransfer: [ownerShipTransferStatus=\"None\"; requestedOwnerGroupId=\"None\"])`. This causes the unchangedRecordSetOwnershipStatus check to fail. I have updated the check to pass when the existing record's recordSetGroupChange is set to `None`, since that seems to apply only in this case.
> 
> Once the record is successfully updated, then on future updates both the existing and updated records recordSetGroupChange value will be set to `Some(OwnerShipTransfer: [ownerShipTransferStatus=\"None\"; requestedOwnerGroupId=\"None\"])`, so the `updates.recordSetGroupChange == existing.recordSetGroupChange` check will pass as expected.
> 
> @Jay07GIT, let me know if you see a better way to fix this, or if this seems OK.

Its looks prefect since we cant update the existing `recordSetGroupChange`.

Jay07GIT:
> Bug: as of version 0.21.1, when a user tries to update a record in a private zone from the zone view, it fails with a `422: Cannot update RecordSet OwnerShip Status when zone is not shared.` error.
> 
> This can be worked around by updating the record via batch change. Also, once that particular record has been updated via batch change, it can be updated in the zone view as well from then on.
> 
> It seems like the issue is that in a private zone, the existing record's recordSetGroupChange is simply set to `None`, whereas the updated record's is set to `Some(OwnerShipTransfer: [ownerShipTransferStatus=\"None\"; requestedOwnerGroupId=\"None\"])`. This causes the unchangedRecordSetOwnershipStatus check to fail. I have updated the check to pass when the existing record's recordSetGroupChange is set to `None`, since that seems to apply only in this case.
> 
> Once the record is successfully updated, then on future updates both the existing and updated records recordSetGroupChange value will be set to `Some(OwnerShipTransfer: [ownerShipTransferStatus=\"None\"; requestedOwnerGroupId=\"None\"])`, so the `updates.recordSetGroupChange == existing.recordSetGroupChange` check will pass as expected.
> 
> @Jay07GIT, let me know if you see a better way to fix this, or if this seems OK.

Its looks prefect since we cant update the existing `recordSetGroupChange`.

PR #1425, opened 2024-12-07
updated ownership transfer email content

Jay07GIT:
Fixes #.

Changes in this pull request:
- Updated ownership transfer email content
-


nspadaccino (commented):
Could we also have the email include a link to the transfer or zone so the recipient doesn't need to find it in the portal? And if the request is pending review, some instructions on how to approve would be good

nspadaccino (approved):
lgtm

PR #1431, opened 2024-12-19
Zone Dump CSV Export

### LINKED ISSUE #1430: Dump Zone to CSV

Add a button in the portal zone view that will output all the records in the zone to a CSV file.

nspadaccino:
Fixes #1430

## Zone dump: CSV export of zone records

Adds an **Export as CSV** button to the zone records view that downloads all recordsets in the zone as CSV (`recordset_id, fqdn, record_type, ttl, record_data`). The export pages through the zone client-side via the existing `listRecordSetsByZone` API, converting each page to CSV as it arrives, then triggers a browser download.

  ### Notes
  - **Memory**: each page is converted and dropped, so we never hold the full recordset list and the assembled CSV at once; keeps very large zones (~300k records) from crashing the tab. Also removed an O(n²) concat in the paging loop.
  - **UX**: the button disables and shows a spinner while the export runs.
  - **Known limit**: still latency-bound (API caps page size at 100, so ~150 sequential requests for 15k records). Faster pages need an API change; this can be a follow-up.

  ### Testing
  - [ ] Export a small zone; verify CSV contents and filename (`<zone>.csv`).
  - [ ] Export a large zone; verify the spinner shows and the button re-enables on
        success and error.


Aravindh-Raju (commented):
SOA record data has key-value pairs. But only the values are displayed in record_data column of csv. I think it'll be the case for any record type having key-value pairs. We should fix that probably.

Rest all looks fine.

PR #1433, opened 2025-01-12
Updated validation for ownership transfer status in api

Jay07GIT:
Fixes #.

Changes in this pull request:
Changes in API
- User cannot able to submit "pending review" in Ownership transfer status
- If requested owner group already a member of ownergroup, cannot submit the request again.
- User cannot able to submit "invalid status" in Ownership transfer status

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetService.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/record/RecordSetService.scala
@@ -147,13 +147,14 @@ class RecordSetService(
       _ <- messageQueue.send(change).toResult[Unit]
     } yield change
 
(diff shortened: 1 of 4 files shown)


nspadaccino:
When testing locally, I found a few scenarios that need changing.

It seems I am still able to submit PendingReview and AutoApproved as valid ownerShipTransferStatus, I think those should both be disallowed. 

When I try to submit a Requested status for an unowned record, it goes to PendingReview instead of AutoApproved. 

I am also able to sumbit a Cancelled request for a record already in an Approved state, which also sets the ownership back to unowned. Cancelled should only be allowed for a record in PendingReview state.

Sokitha-21 (approved):

PR #1437, opened 2025-01-31
Create Zone -- APIs

### LINKED ISSUE #1455: Allow user to create zones in vinyldns
Allow users to create zones in vinyldns. Tsis supports the users to create the zones in DNS backend and manage the records.

The API uri of the providers and fields should be dynamic, hence we can connect any providers. URI and fields should be config based.

Allow user to access both API and portal for zone creation.

Jay07GIT:
Fixes #1455.

Changes in this pull request:
User can generate the New zone in vinyldns.

Create- Post request :- `http://localhost:9000/zones/generate`
Update- Put request :- `http://localhost:9000/zones/generate`
Delete- Delete request :- `http://localhost:9000/zones/generate/<"zone_id">`
List- Get Request :- `http://localhost:9000/zones/generate/info?maxItems=100&nameFilter=<"name">&ignoreAccess=false `
List by name- Get Request :- `http://localhost:9000/zones/generate/name/"zone_name"`
List by ID- Get Request :- `http://localhost:9000/zones/generate/id/"zone_name"`

Create Zone request - 
for powerdns
```
{
    "groupId": "groupid",
    "provider": "powerdns",
    "zoneName": "mttest1.example.org.", 
    "email": "test@test.com",
    "providerParams":{
        "kind": "Native", 
        "nameservers": ["172.17.42.1.", "ns1.mttest1.example.org."]}
        "masters": ["172.17.42.1", "172.17.42.2"]
    }
}
```
for bind -  "ns_ipaddress": ["172.17.42.1"], use the same to test the connection
```
{
    "groupId": "groupid",
    "provider": "bind",
    "zoneName": "test.com.",
    "providerParams":{
        "nameservers": ["172.17.42.1"],
        "admin_email": "admin@test.com",
        "ttl": 3600,
        "refresh":  6048000,
        "retry": 86400,
        "expire":  24192000,
        "negative_cache_ttl":  6048000
    }
}
```
To test the bind server zone creation in vinyldns ,

Comment the image in the file `quickstart/docker-compose.xml`
#image : "vinyldns/build:base-test-integration-${VINYLDNS_IMAGE_VERSION}"

run 
```
./quickstart/quickstart-vinyldns.sh --clean
./quickstart/quickstart-vinyldns.sh --deps-only 
```

Update zone request:- 
```
{
    "groupId": "b82c2a36-a973-4abd-b47d-0711b105c0d5",
    "email":"test@test.com",
    "provider": "powerdns",
     "zoneName": "mttest13.example.org.",
     "providerParams":{
         "kind": "Master"
      }
}
```

Dependency PR: (Merge before this PR)
 for the protobuf update.:- https://github.com/vinyldns/vinyldns/pull/1207
 for the logs.:- https://github.com/vinyldns/vinyldns/pull/1446

To do
Update zone for the test bind server.


modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneService.scala
```diff
@@ -135,6 +159,394 @@ class ZoneService(
       _ <- messageQueue.send(deleteZoneChange).toResult[Unit]
     } yield deleteZoneChange
 
+  def getGenerateZoneByName(zoneName: String, auth: AuthPrincipal): Result[GenerateZone] =
+    for {
+      generateZone <- getGenerateZoneByNameOrFail(ensureTrailingDot(zoneName))
```
JoshSEdwards, line 165: I think the generated zone read paths are missing the access check this PR introduced. `canSeeGenerateZone` exists, but `getGenerateZoneByName` and `getGeneratedZoneById` return the generated zone directly. Can we require `canSeeGenerateZone(auth, generatedZone)` before returning those reads?

I also don't love that some methods and vars are missing the D at the end of generated.

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneService.scala
```diff
@@ -135,6 +159,394 @@ class ZoneService(
...
+          authPrincipal,
+          nameFilter,
+          startFrom,
+          maxItems,
+          ignoreAccess
+        )
+        generatedZones = listZonesResult.generatedZones
+        groupIds = generatedZones.map(_.groupId).toSet
+        groups <- groupRepository.getGroups(groupIds)
+        generateZoneSummaryInfos = generateZoneSummaryInfoMapping(generatedZones, authPrincipal, groups)
```
JoshSEdwards, line 532: This looks like it fetches all matching generated zones first and only computes `accessLevel` after the data has already been loaded into the response model. The repository signature accepts `authPrincipal` and `ignoreAccess` but the MySQL implementation doesn't use either one to filter results. We should probably enforce access before building the response.

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneService.scala
```diff
@@ -135,6 +159,394 @@ class ZoneService(
...
+
+      zoneGenerateResponse = ZoneGenerationResponse(
+        responseCode = Some(responseCode),
+        status = Some(dnsConnResponse.getResponseMessage),
+        message = Some(responseJson),
+        changeType = GenerateZoneChangeType.Update
+      )
+      zoneToUpdate = existingGeneratedZone.copy(
+        email = request.email,
+        groupId = request.groupId,
```
JoshSEdwards, line 299: This allows the generated-zone update flow to reassign ownership without checking access to the target group. The normal zone update path checks both the current admin group and the requested new admin group, so this should probably follow the same rules.

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneValidations.scala
```diff
@@ -54,6 +54,11 @@ class ZoneValidations(syncDelayMillis: Int) {
       (rule.groupId ++ rule.userId).size == 1
     }
 
+  def isValidGenerateZoneConn(responseCode : Int, responseMsg : String): Either[Throwable, Unit] =
+    ensuring(InvalidRequest(responseMsg)) {
+      responseCode <= 400
```
JoshSEdwards: This should probably be an explicit 2xx check, or at least `< 400` so it doesn't include a 400 response as a success.

modules/core/src/main/scala/vinyldns/core/domain/zone/Zone.scala
```diff
@@ -123,8 +140,78 @@ object Zone {
     )
   }
 }
+final case class GenerateZone(
```
JoshSEdwards, line 143: The core `Zone.scala` file now contains a lot of generated zone and provider-integration code. `GenerateZone`, generated-zone request/response transport models, json4s payloads, provider endpoint templates, API keys, and provider config parsing.

Feels like feature-specific integration is leaking into the canonical zone domain model here. Does it make more sense to move the generated zone models/config into a dedicated module or file so the core zone domain stays focused?

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlGenerateZoneRepository.scala
```diff
@@ -0,0 +1,257 @@
...
+  def getGenerateZoneById(id: String): IO[Option[GenerateZone]] =
+    monitor("repo.ZoneJDBC.getGenerateZoneById") {
+      IO {
+        DB.readOnly { implicit s =>
+          getGenerateZoneByIdInSession(id)
+        }
+      }
+    }
+
+  def listGenerateZones(
```
JoshSEdwards, line 156: `listGenerateZones` and `listGeneratedZonesByAdminGroupIds` look like they're auth-aware from the arguments they take, but the MySQL implementations don't use `authPrincipal` or `ignoreAccess` anywhere. We should either use them, or if we're intentionally keeping auth out of this remove them from the args list.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlGenerateZoneRepository.scala
```diff
@@ -0,0 +1,257 @@
...
+                       ): IO[ListGeneratedZonesResults] =
+    monitor("repo.ZoneJDBC.listGeneratedZones") {
+      IO {
+        DB.readOnly { implicit s =>
+          val sb = new StringBuilder
+          sb.append(BASE_GENERATE_ZONE_SEARCH_SQL)
+
+          val filters = if (zoneNameFilter.isDefined && (zoneNameFilter.get.takeRight(1) == "." || zoneNameFilter.get.contains("*"))) {
+            List(
+              zoneNameFilter.map(flt => s"gz.name LIKE '${ensureTrailingDot(flt.replace('*', '%'))}'"),
```
JoshSEdwards: This query is being built by interpolating request-derived values directly into SQL. `zoneNameFilter` and `startFrom` are turned into SQL fragments here and then appended into the final query with `filters.mkString(" AND ")`.

Can we use bind parameters for these values instead? Even if similar patterns exist elsewhere, this is new code and should avoid carrying that risky design pattern forward.

JoshSEdwards (changes requested):
A few comments, mostly around auth, but also I'm a little concerned with mixing all the new generated-zone code in the same files as existing zone code, especially Zone.scala. The feature is spread through routing, service, core domain, config loading, and persistence, all without a clear boundary. I realize that's asking for a big refactor, so I'm open to hearing why it should remain as-is.

modules/core/src/main/scala/vinyldns/core/domain/zone/Zone.scala
nspadaccino: Yeah it definitely makes sense to refactor this into its own module

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneService.scala
```diff
@@ -417,6 +843,35 @@ class ZoneService(
       }
       .toResult
 
+  private def generateZoneDoesNotExist(zoneName: String): Either[Throwable, Unit] = {
+    val existingZoneOpt: Option[GenerateZone] =
+      generateZoneRepository.getGenerateZoneByName(zoneName).unsafeRunSync()
```
Sokitha-21: unsafeRunSync() is called inside a method that is already executing within a for. Calling unsafeRunSync() on a thread that is already managed by the runtime can cause a deadlock.

modules/api/src/main/scala/vinyldns/api/route/ZoneRouting.scala
```diff
@@ -143,6 +142,91 @@ class ZoneRoute(
...
+                  complete(StatusCodes.OK, result)
+                }
+              }
+            }
+          }
+        }
+      }
+    } ~
+    path("zones" /"generate"/ "name" / Segment) { zoneName =>
+      get & monitor("Endpoint.getGenerateZoneByName")
```
Sokitha-21: The braces {} are missing. This means the route is not restricted to GET requests and the monitor is never applied.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlGenerateZoneRepository.scala
```diff
@@ -0,0 +1,257 @@
...
+          if(adminGroupIds.nonEmpty) {
+            val groupIds = adminGroupIds.map(x => "'" + x + "'").mkString(",")
+            sb.append(s" WHERE admin_group_id IN ($groupIds) ")
+          } else {
+            sb.append(s" WHERE admin_group_id IN ('') ")
+          }
+
+          if(startFrom.isDefined){
+            sb.append(" AND ")
+            sb.append(s"gz.name > '${startFrom.get}'")
```
Sokitha-21: startFrom is interpolated directly as a string literal and adminGroupIds values are wrapped in single quotes and joined into the query. Can we use bind parameters as suggested before instead of manual string construction.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlGenerateZoneRepository.scala
```diff
@@ -0,0 +1,257 @@
...
+      monitor("repo.generateZone.save") {
+        IO {
+            DB.localTx { implicit s =>
+              PUT_GENERATE_ZONE
+              .bindByName(
+                  'id -> generateZone.id,
+                  'name -> generateZone.zoneName,
+                  'provider -> generateZone.provider,
+                  'adminGroupId -> generateZone.groupId,
+                  'response -> toPB(generateZone.response.get).toByteArray,
```
Sokitha-21: response is Option[ZoneGenerationResponse]. Calling .get will throw NoSuchElementException at runtime whenever a GenerateZone is saved without a response.

modules/api/src/main/scala/vinyldns/api/route/ZoneRouting.scala
```diff
@@ -143,6 +142,91 @@ class ZoneRoute(
...
+      (put & monitor("Endpoint.updateGeneratedZone")) {
+        authenticateAndExecuteWithEntity[GenerateZone, ZoneGenerationInput](
+          (authPrincipal, generateZone) =>
+            zoneService.handleUpdateGeneratedZoneRequest(generateZone, authPrincipal)
+        ) { response =>
+          complete(StatusCodes.Accepted, response)
+        }
+      }
+    } ~
+    path("zones" / "generate" / Segment) { id =>
```
Sokitha-21: Segment acts as a wildcard and can match routes like info, allowedDNSProviders and nameservers causing requests to hit the wrong route and return 405 Method Not Allowed.
Can we consider moving the specific routes above the wildcard Segment routes so they are matched first.

Sokitha-21 (changes requested):

project/plugins.sbt
```diff
@@ -20,7 +20,7 @@ addSbtPlugin("com.typesafe.sbt" % "sbt-ghpages" % "0.6.3")
 
 addSbtPlugin("de.heikoseeberger" % "sbt-header" % "5.5.0")
 
-addSbtPlugin("com.typesafe.play" % "sbt-plugin" % "2.7.4")
+addSbtPlugin("com.typesafe.play" % "sbt-plugin" % "2.8.18")
```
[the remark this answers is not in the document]
Jay07GIT: @arpit4ever These deps update are part of the news deps added. So its require to run this PR success.

modules/api/src/main/scala/vinyldns/api/route/ZoneRouting.scala
Jay07GIT: Resolved.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlGenerateZoneRepository.scala
Jay07GIT: Resolved.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlGenerateZoneRepository.scala
Jay07GIT: Resolved.

modules/api/src/main/scala/vinyldns/api/route/ZoneRouting.scala
Jay07GIT: Resolved.

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneService.scala
Jay07GIT: Resolved.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlGenerateZoneRepository.scala
Jay07GIT: Resolved.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlGenerateZoneRepository.scala
Jay07GIT: Resolved. Removed the authPrincipal which is not in use.

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneValidations.scala
Jay07GIT: Resolved

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneService.scala
Jay07GIT: Resolved

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneService.scala
Jay07GIT: Added the required code.

modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneService.scala
Jay07GIT: Done the changes

modules/core/src/main/scala/vinyldns/core/domain/zone/Zone.scala
Jay07GIT: Since we do not have a separate Connect Zone file, we commonly use the Zone file for that purpose. Therefore, it would be consistent to place generateZone within zones as well, since it is part of the Zone configuration.

Sokitha-21 (approved):
LGTM

modules/api/src/main/resources/application.conf
```diff
@@ -50,6 +50,217 @@ vinyldns {
         class-name = "vinyldns.api.backend.dns.DnsBackendProviderLoader"
         settings = {
           legacy = false
+          dns-provider-api = {
+            providers = {
+              bind = {
+                api-key = "bind-api-key"
+                endpoints = {
+                  create-zone = "http://localhost:19000/api/zones/generate"
+                  delete-zone = "http://localhost:19000/api/zones/delete?zoneName={{zoneName}}"
```
arpit4ever, line 59: Do we need query parameter zoneName on delete-zone API since body already has the same parameter.

build/docker/api/application.conf
```diff
@@ -50,6 +50,16 @@ vinyldns {
         class-name = "vinyldns.api.backend.dns.DnsBackendProviderLoader"
         settings = {
           legacy = false
+          dns-provider-api ={
```
JoshSEdwards: This configuration is mounted by the quickstart API, but the loader reads dns-provider-api.providers; this block still uses the old create-zone-uri/api-key shape, so the provider map is empty. The URLs also use localhost even though Bind and PowerDNS run in separate containers. Please update this config to the current provider schema and container hostnames, and add a quickstart smoke test.

modules/mysql/src/main/resources/db/migration/V3.33__GenerateZone.sql
```diff
@@ -0,0 +1,18 @@
...
+*/
+CREATE TABLE generate_zone (
+  id CHAR(36) NOT NULL,
+  name VARCHAR(256) NOT NULL,
+  provider VARCHAR(256) NOT NULL,
+  admin_group_id CHAR(36) NOT NULL,
+  response BLOB NOT NULL,
+  data BLOB NOT NULL,
+  PRIMARY KEY (id),
+  INDEX generate_zone_name_index (name),
```
JoshSEdwards: The service checks for an existing name before calling save, but this migration only adds a non-unique index and the upsert is keyed by id. Two concurrent creates can therefore persist the same generated-zone name, after which name reads and grouped listings are ambiguous. Please enforce uniqueness on name and translate duplicate-key failures into the conflict response.

modules/api/src/main/scala/vinyldns/api/domain/zone/GenerateZoneService.scala
```diff
@@ -0,0 +1,598 @@
...
+      // so an existing generate_zone record must not block connecting to that zone.
+      _ <- zoneDoesNotExist(request.zoneName)
+
+      // Send request
+      _ <- logger.info(s"Request: provider=${request.provider}, path=$endpoint, request=$requestJsonOpt").toResult
+      dnsProviderConn <- createConnection(endpoint).toResult
+      dnsConnResponse <- createDnsZoneService(Encryption.decrypt(crypto, providerConfig.apiKey), "create-zone", requestJsonOpt, dnsProviderConn).toResult
+
+      // Process response
+      responseCode = dnsConnResponse.getResponseCode
```
JoshSEdwards: These provider calls perform blocking connection and stream I/O inside the request IO, which runs on the API general executor. A provider that reaches the 30-second read timeout can consume request threads and affect unrelated endpoints. Please move this work to a dedicated blocking pool or use a nonblocking client, preferably behind one shared operation helper.

modules/core/src/main/scala/vinyldns/core/domain/zone/Zone.scala
```diff
@@ -228,7 +231,8 @@ final case class LegacyDnsBackend(
 final case class ConfiguredDnsConnections(
     defaultZoneConnection: ZoneConnection,
     defaultTransferConnection: ZoneConnection,
-    dnsBackends: List[LegacyDnsBackend]
+    dnsBackends: List[LegacyDnsBackend],
+    dnsProviderApiConnection : DnsProviderApiConnection
```
JoshSEdwards: ConfiguredDnsConnections now owns generated-zone provider configuration, coupling the canonical zone model/config loader to this feature. Please move the generated-provider config parsing into its own module and inject it into GenerateZoneService independently.

JoshSEdwards (changes requested):

PR #1450, opened 2025-07-11
Implemented Zone Specific Filtering to Separate Dotted Host Record Histories.

### LINKED ISSUE #1442: Record Change History output combines history for records with the same FQDN
This is a corner case relating to dotted host records. Typically two records cannot share an FQDN, but they can if one record is a dotted host. For example, you can have:
`FQDN=foo.child.parent.com, record_name=foo, zone=child.parent.com`
and
`FQDN=foo.child.parent.com, record_name=foo.child, zone=parent.com`

Currently, if you look up the history for foo.child.parent.com, either through the portal or the API, you will see the history of both records mixed together. 

To have the histories be separate, the relevant SQL query should be filtering on both the FQDN and the zone_id, not just the FQDN

Sokitha-21:
Fixes #1442.

Changes in this pull request:
- In the portal, appended the zone name to each of the suggestions displayed during the search operation.
- Passed the zoneId along with the selected FQDN to fetch the correct record set.
- In the SQL part, filter by both FQDN and zone_id.
- Dependency Issue #1462 (In the SQL part, the issue was fixed by escaping the underscores in the query so that they are treated as literal characters during database searches.)


Jay07GIT (dismissed):
Tested the PR , Looks good to me.

image

JoshSEdwards:
### :tada: **Snyk checks have passed. No issues have been found so far.**

:white_check_mark: **security/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/comcast-vinyldns/pr-checks/150d5bf1-cc43-43f0-a2b4-726273287bb6?refs=16b2899b98286d25c8b2166f755207adaa4a7e7c&source=prChecksComment)

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetCacheRepository.scala
```diff
@@ -289,17 +290,22 @@ class MySqlRecordSetCacheRepository
         val maxPlusOne = maxItems.map(_ + 1)
         val wildcardStart = raw"^\s*[*%](.+[^*%])\s*$$".r
 
-        // setup optional filters
+        def escapeSqlLike(input: String): String =
+          input.replace("\\", "\\\\").replace("_", "\\_").replace("%", "\\%").replace("*", "%").replace(".", "%")
```
nspadaccino, line 294: Why are we replacing "." with a wildcard?

Sokitha-21: When searching a record by name a “.” is added at the end of the name during query formation. Because of this FQDN filtering only works in some cases. For an instance, if the name is “ok” it filters correctly as its FQDN is ok., but if the name is “already-exists” it fails since the FQDN is already-exists.zonename. Replacing “.” with % resolves this as we are filtering using LIKE.

PR #1468, opened 2025-10-27
Create Zone -- Portal changes

### LINKED ISSUE #1455: Allow user to create zones in vinyldns

Sokitha-21:
Fixes https://github.com/vinyldns/vinyldns/issues/1455.

prerequisites PR - https://github.com/vinyldns/vinyldns/pull/1437

To test the PR - merge the above branch into this and test.

Changes in this pull request:

Create a New Zone from the portal
View, delete, update the generated zone from the portal


modules/portal/conf/routes
```diff
@@ -38,6 +41,11 @@ PUT           /api/zones/:id                     @controllers.VinylDNS.updateZon
 DELETE        /api/zones/:id                     @controllers.VinylDNS.deleteZone(id: String)
 POST          /api/zones/:id/sync                @controllers.VinylDNS.syncZone(id: String)
 
+GET           /api/zones/generate/nameservers    @controllers.VinylDNS.getNameservers
+POST          /api/zones/generate                @controllers.VinylDNS.generateZone
```
JoshSEdwards, line 45: The UI submits `PUT /api/zones/generate` in `/modules/portal/public/lib/services/zones/service.zones.js:141` but we're only routing GET, POST, and DELETE for generated zones. I think we need a `PUT /api/zones/generate` route and a proxy method, otherwise the update path will 404.

modules/portal/public/lib/controllers/controller.zones.js
```diff
@@ -82,6 +165,117 @@ angular.module('controller.zones', [])
...
+        }
+        return zonesService.getCreateZoneTemplate(provider).then(function(results) {
+            const createZoneFields = JSON.parse(results.data["request-templates"]["create-zone"]);
+            const updateZoneFields = results.data["request-templates"]["update-zone"];
+            $scope.UpdateRequiredFields = results.data["required-fields"]["update-zone"];
+            const updateZone = JSON.parse(updateZoneFields);
+            $scope.updateZoneTemplate = updateZone;
+            $scope.isEditMode = true;
+
+            $scope.zoneFields = Object.entries(createZoneFields).map(([label, config]) => {
```
JoshSEdwards: This is the update zone flow, yet we're displaying `createZoneFields` instead of `updateZoneFields` to the user. Should this be changes to `$scope.zoneFields = Object.entries(updateZone).map(([label, config])`?

modules/portal/public/lib/controllers/controller.zones.js
```diff
@@ -82,6 +165,117 @@ angular.module('controller.zones', [])
...
+                    value,
+                    field,
+                    editable: true
+                };
+            });
+        });
+    };
+
+    $scope.isDynamicZoneFieldRequired = function(fieldName) {
+      return $scope.requiredFields && $scope.requiredFields.indexOf(fieldName) !== -1;
```
JoshSEdwards: The zone creation modal `modules/portal/public/templates/zonecreation-modal.html` uses this method for both the update and create paths, but the required-field lookup is pinned to the create fields `$scope.requiredFields`. This should be made mode-aware and read `$scope.UpdateRequiredFields`. Right now the update required-field list is loaded but never used anywhere.

modules/portal/conf/application.conf
```diff
@@ -86,6 +86,71 @@ api {
     zone-routing-default-max-items = 100
     zone-routing-max-items-limit = 100
   }
+  dns-provider-portal-fields = {
```
JoshSEdwards, line 103: Can we avoid re-declaring provider metadata in the portal here? #1437 already establishes this contract on the API side (`modules/api/src/main/resources/application.conf:42`). Is it needed in both application.conf?

modules/portal/public/lib/controllers/controller.zones.js
JoshSEdwards, line 144: This needs to rewire `generateZone-search-text` as well, otherwise you'll check Search by Admin Group on the Created Zones tab, and autocomplete will still show zone names instead of group names.

modules/portal/public/lib/controllers/controller.zones.js
```diff
@@ -312,6 +578,93 @@ angular.module('controller.zones', [])
             });
     };
 
+    $scope.getUpdatePayload = function (zone) {
```
JoshSEdwards: Similar to my comment on #1437 about separating generated zones with clear boundaries, I think this controller file has gotten way too large for the size of the feature. This file alone almost doubled in length, and this duplicate `getUpdatePayload()` definition (there's another on line 212) is a good signal that this controller is too crowded to keep track of. I think we should probably split generated-zone behavior into its own module/controller.

JoshSEdwards (changes requested):
Found a few bugs, and I think a refactor to go along with the one in #1437 would be a good idea. Generated zones is a meaty enough feature that it deserves to live in its own files rather than entangled with existing zone code.

modules/portal/public/lib/services/zones/service.zones.js
```diff
@@ -79,17 +79,75 @@ angular.module('service.zones', [])
...
+            return $http.get(url);
+        }
+
+        this.getAllowedDNSProviders = function() {
+            var url = "/config/allowedDNSProviders";
+            return $http.get(url);
+        }
+
+        this.getCreateZoneTemplate = function(provider) {
+            var url = "config/createZoneTemplate/" + provider;
```
Sokitha-21: The URL appears to be missing a leading /.

modules/portal/conf/routes
```diff
@@ -17,6 +17,9 @@ GET           /dnschanges                        @controllers.FrontendController
 GET           /dnschanges/new                    @controllers.FrontendController.viewNewBatchChange
 GET           /dnschanges/:id                    @controllers.FrontendController.viewBatchChange(id: String)
 GET           /recordsets                        @controllers.FrontendController.viewRecordSets
+GET           /config/allowedDNSProviders        @controllers.FrontendController.getAllowedDNSProviders
+GET           /config/createZoneTemplate/:provider   controllers.FrontendController.getCreateZoneTemplate(provider: String)
```
Sokitha-21: This route is missing the @ prefix before controllers.FrontendController

modules/portal/public/lib/controllers/controller.zones.js
```diff
@@ -312,6 +578,93 @@ angular.module('controller.zones', [])
...
+                $("#zone_creation_modal").modal("hide");
+                $scope.zoneError = true;
+                handleError(error, 'zonesService::generateZone-failure');
+                $scope.processing = false;
+              });
+        }
+    };
+
+    $scope.deleteCreatedZone = function () {
+        if (!confirm("Are you sure you want to delete this zone?")) return;
```
Sokitha-21: Browser `confirm()` is not recommended as it blocks the event loop. Suggest to use an angular modal pattern instead.

modules/portal/public/lib/controllers/controller.zones.js
```diff
@@ -59,13 +72,83 @@ angular.module('controller.zones', [])
...
+        $scope.createZone.providerParams = {};
+        $scope.createZone.providerParams.nameservers = [];
+        $scope.nameserverSelection = {};
+        $scope.createZone.ns_ipaddress = [];
+    };
+
+    $scope.viewCreatedZone = function (createdZone) {
+        $log.debug(createdZone)
+        $scope.createZone = angular.copy(createdZone);
+        $scope.isEditMode = false;
```
Sokitha-21: `viewCreatedZone` resets `isEditMode` but never clears `$scope.isDelete ` unlike `updateCreatedInfo`. As a result, `isDelete` may retain its previous value when opening a zone via `viewCreatedZone` causing the delete button to be shown or hidden incorrectly.

modules/portal/public/lib/controllers/controller.zones.js
```diff
@@ -312,6 +578,93 @@ angular.module('controller.zones', [])
...
+                    $scope.zoneError = true;
+                    handleError(error, 'zonesService::generateZone-failure');
+                    $scope.processing = false;
+                });
+        }else {
+        // update zone service
+            const payload = $scope.getUpdatePayload($scope.createZone);
+            zonesService.updateGeneratedZone(payload)
+              .then(function (response) {
+                $("#zone_creation_modal").modal("hide");
```
Sokitha-21: The create path inside `addZoneCreation` calls `$timeout($scope.refreshGeneratedZones(), 1000)` after success.The update (else) path is missing that same refreshGeneratedZones() call, so the edited zone's data stays stale in the table after saving.

modules/portal/conf/application.conf
Sokitha-21: Yes, both are needed. The API config holds backend substitution templates and endpoints for calling DNS providers whereas the portal config holds UI field descriptors for rendering form fields. There's no shared config mechanism between the two Play/Akka processes at runtime, so they can't reference each other's config. The provider names overlap, but the structure and purpose are entirely different.

PR #1485, opened 2026-02-05
Added Export to CSV Option for RecordSet Search Results

### LINKED ISSUE #1410: Add "Export to csv" option, for search results
Instead of only displaying search results to the UI, provide the option to export all results to a downloadable .csv file. 

Not all users are comfortable with the API and this feature would enable UI users greater flexibility to perform bulk audits and changes.

The format of the exported csv should include all of the same information that's displayed, currently, to the UI. Any additional details could also be included, if possible; perhaps with a check list of "fields" to include in the output - as well as the format of the columns.
Let the user decide the formatting, to enable their ability to more easily use the csv file(s) for bulk-csv uploads.

Sokitha-21:
VDNS-273

Fixes #1410.

Enhanced the Global RecordSet Search to support CSV export enabling users to download all matching records with configurable fields, formatting and optional additional details via a selectable checklist making audits and bulk CSV uploads easier for UI users.

### DIFF
--- a/modules/portal/app/views/recordsets/recordSets.scala.html
+++ b/modules/portal/app/views/recordsets/recordSets.scala.html
@@ -108,6 +108,17 @@
                             </div>
                         </div>
                     </div>
+                    <div class="dropdown export-dropdown" ng-show="isRecordSearchTriggered" ng-click="$event.stopPropagation()">
+                        <button class="btn btn-primary dropdown-toggle" ng-click="toggleExportOptions($event)">
+                            Export CSV ▾
+                        </button>
+                        <div class="dropdown-menu dropdown-menu-export" ng-if="showExportOptions" ng-click="$event.stopPropagation()">
+                            <label ng-repeat="(key, label) in HEADER_MAP">
+                            <input type="checkbox" ng-model="selectedFields[key]"> {{label}}
+                            </label>
+                            <button class="btn btn-success btn-sm" ng-click="exportWithSelection()">Download</button>
+                        </div>
+                    </div>
                 </div>
             </div>
 
--- a/modules/portal/app/views/recordsets/recordSets.scala.html
+++ b/modules/portal/app/views/recordsets/recordSets.scala.html
@@ -73,7 +73,7 @@
         <!-- START RECORD TABLE -->
         <div class="panel panel-default">
             <div class="panel-heading">
-                <div class="row">
+                <div class="row" style="display: flex;">
                     <div class="col-md-10">
                         <span class="panel-title bolder-font-weight">Records</span>
                     </div>
--- a/modules/portal/app/views/recordsets/recordSets.scala.html
+++ b/modules/portal/app/views/recordsets/recordSets.scala.html
@@ -2,8 +2,9 @@
 
 @content = {
     <!-- PAGE CONTENT -->
+    
     <div class="right_col" role="main">
-
+        
         <!-- START BREADCRUMB -->
         <ul class="breadcrumb">
             <li><a href="/">Home</a></li>
@@ -73,7 +74,7 @@
         <!-- START RECORD TABLE -->
         <div class="panel panel-default">
             <div class="panel-heading">
-                <div class="row" style="display: flex;">
+                <div class="row">
                     <div class="col-md-10">
                         <span class="panel-title bolder-font-weight">Records</span>
                     </div>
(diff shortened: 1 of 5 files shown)


arpit4ever:
> @Sokitha-21 Code changes looks good.
> 
> I have few observation.
> 
> 1. Owner group name not updating in the csv.
> 2. Name field is not understandable in the csv. Instead we can give record id which helps user.
>    [already_ - recordsets.csv](https://github.com/user-attachments/files/25390513/already_.-.recordsets.csv)
> 3. We can give select all option in download csv filter.

Additional recommendation:

- The Download button should only be enabled when at least one field is selected. Currently, if no fields are selected, clicking the button downloads a blank CSV.
- By default all fields should be selected.
- During testing, only 101 records were downloaded even when search returned 200+ records across pages.

modules/portal/app/views/recordsets/recordSets.scala.html
```diff
@@ -108,6 +108,17 @@ <h5>PTR Records</h5>
                             </div>
                         </div>
                     </div>
+                    <div class="dropdown export-dropdown" ng-show="isRecordSearchTriggered" ng-click="$event.stopPropagation()">
```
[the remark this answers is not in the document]
nspadaccino: Also like I said in https://github.com/vinyldns/vinyldns/pull/1485/changes#r2855089566, we could restructure this to use a bootstrap dropdown

nspadaccino (changes requested):

modules/portal/public/lib/recordset/recordsets.controller.js
```diff
@@ -203,8 +292,173 @@
...
+                        .catch(function (error) {
+                            if (error && Object.keys(error).length > 0) {
+                                handleError(error, 'recordExport-failure');
+                            }
+                            hideLoader();
+                        });
+                }, 0);
+            };
+            
+            function toCSVCell(value) {
```
nspadaccino, line 436: Quote-escaping is correct here, but still vulnerable to CSV formula injection, since any field whose value begins with =, +, -, @, or a tab/return is executed as a formula when the CSV is opened in Excel. 

Before quoting, check if the value starts with a dangerous character, and if it does, put an apostrophe in front of it so it is not executed as a formula

modules/portal/public/lib/recordset/recordsets.controller.js
```diff
@@ -203,8 +292,173 @@
...
+                        document.body.removeChild(link);
+
+                        hideLoader();
+                    }
+
+                    fetchAllPages()
+                        .then(loadPrivateZoneOwners)
+                        .then(downloadCsv)
+                        .catch(function (error) {
+                            if (error && Object.keys(error).length > 0) {
```
nspadaccino: Gating on object.keys can swallow errors that occur in the CSV-building path. Instead, just report any non-null error.
`  .catch(function (error) {
      if (error) {
          handleError(error, 'recordExport-failure');
      }
      hideLoader();
  });`

modules/portal/public/lib/recordset/recordsets.controller.js
```diff
@@ -203,8 +292,173 @@
...
+
+                            if (!recordSets.length) return;
+
+                            allRecords.push(...recordSets);
+
+                            if (!newNextId || visitedNextIds.has(newNextId)) return;
+
+                            visitedNextIds.add(newNextId);
+
+                            return fetchAllPages(newNextId);
```
nspadaccino, line 368: Would it be possible to indicate progress here? Like, besides just the loading spinner, maybe we could display how many records/pages have been exported? Mostly relevant for very large zones where users may be waiting for a while. Not a blocking issue.

modules/portal/public/lib/recordset/recordsets.controller.js
```diff
@@ -203,8 +292,173 @@
...
+                    }
+
+                    function downloadCsv() {
+                        if (!allRecords.length) {
+                            hideLoader();
+                            return;
+                        }
 
+                        let filename;
+                        if (recordType) {
```
nspadaccino: These two conditionals are essentially the same, could we condense this?

modules/portal/public/css/vinyldns.css
```diff
@@ -138,6 +138,22 @@ a.action-link {
     margin-left: 10px;
 }
 
+.export-dropdown {
+    position: relative;
+    display: inline-block;
+}
+
+.dropdown-menu-export{
```
nspadaccino: Is this CSS used anywhere? If not, we should remove it. Same for `.export-dropdown label`

modules/portal/public/lib/recordset/recordsets.controller.js
```diff
@@ -203,8 +292,173 @@
...
+                                : `recordsets.csv`;
+                        }
 
+                        const rows = [];
+                        rows.push(buildHeaders());
+
+                        allRecords.forEach(r => {
+                            rows.push(buildRow(r));
+                        });
+                        csvContent = rows.join('\n');
```
nspadaccino: Prepend UTF-8 BOM (Byte order mark) here to prevent excel from corrupting any non-ascii bytes. Important for TXT record content.
` csvContent = '\uFEFF' + rows.join('\n');`

nspadaccino (commented):
Couple small changes needed

nspadaccino (dismissed):
LGTM

modules/portal/public/lib/recordset/recordsets.controller.js
```diff
@@ -198,7 +286,160 @@
...
+                            const newNextId = response.data.nextId;
+                            if (!recordSets.length) return;
+                            allRecords.push(...recordSets);
+                            if (!newNextId || visitedNextIds.has(newNextId)) return;
+                            visitedNextIds.add(newNextId);
+                            return fetchAllPages(newNextId);
+                        });
+                    }
+
+                    function loadPrivateZoneOwners() {
```
JoshSEdwards, line 374: `loadPrivateZoneOwners()` issues one additional zone-details request for every distinct private zone, all at once. This happens even when “Owner Group Name” is deselected. A broad search spanning thousands of zones could generate thousands of requests, heavily load the API, or cause the export to appear hung.

At minimum, we should skip these requests unless `selectedFields.ownerGroupName` is enabled; ideally, use owner information already supplied by the search API or retrieve it in bulk.

modules/portal/public/lib/recordset/recordsets.controller.js
```diff
@@ -99,12 +117,81 @@
...
+                if (!records || !records.length) return '';
+                switch (type) {
+                    case 'A':
+                    case 'AAAA':
+                    return records.map(r => r.address).join('\n');
+
+                    case 'CNAME':
+                    return records.map(r => r.cname).join('\n');
+
+                    case 'TXT':
```
JoshSEdwards, line 151: Right now getRecordData() handles every supported record type except SPF, even though SPF records use the same text field as TXT. Exporting a search containing SPF records will silently produce empty Record Data cells for them. We should add handle SPF alongside TXT and add coverage for it.

JoshSEdwards (changes requested):

modules/portal/app/views/recordsets/recordSets.scala.html
```diff
@@ -2,8 +2,9 @@
 
 @content = {
     <!-- PAGE CONTENT -->
+    
     <div class="right_col" role="main">
-
+
```
arpit4ever, line 7: remove whitespace

JoshSEdwards (dismissed):
LGTM

arpit4ever (changes requested):
The export CSV button should be placed closer to `matches found` rather than the search field.
Search field is for input. 
Matches found is the output and we want to export the found records.

PR #1486, opened 2026-02-13
Add ability to search for FQDNs in "Record Data" field

### LINKED ISSUE #1411: Add ability to search for FQDNs in "Record Data" field (in addition to the "FQDN" field)
Users often do not know the CNAMEs associated with A records, and they only know the A records.
Adding the ability to search for the FQDN (of an A record) in the "Record Data" field  (of a CNAME record) will connect searches across the two types of records in vinylDNS.
Currently, users must already know the CNAMEs, in order to search for them. However, teams and time (re-orgs, etc) can result in the loss of the knowledge about CNAMEs. Users often only know the IPs, when they start their lookups in DNS/vinylDNS. As a result, people are blind to the CNAME records. Adding this search feature would prevent this loss of knowledge and record management.

Sokitha-21:
Fixes #1411.
Jira Ticket: VDNS-306

Changes in this pull request:

Currently recordSet search only matches against the FQDN field of records. 
As a result:
- Searching for an A record FQDN returns the A record.
- It does not return any CNAME records that point to that A record.
- Users must already know the CNAME alias to find it.

Now enhanced the search functionality to also match the searched FQDN against the Record Data field of CNAME records. When searching by FQDN:
- The system continues to search across all record types using the fqdn field.
- Additionally for CNAME records the search also checks the Record Data field.
- If the searched value matches either:
      The record FQDN or
      The CNAME target value
  the record is returned.
If matches exist in both, both results are returned.

### DIFF
--- a/modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlRecordSetCacheRepositoryIntegrationSpec.scala
+++ b/modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlRecordSetCacheRepositoryIntegrationSpec.scala
@@ -22,6 +22,7 @@ import org.scalatest._
 import org.scalatest.matchers.should.Matchers
 import org.scalatest.wordspec.AnyWordSpec
 import scalikejdbc.DB
+import vinyldns.core.domain.Fqdn
 import vinyldns.core.domain.record._
 import vinyldns.core.domain.record.RecordType._
 import vinyldns.core.domain.zone.Zone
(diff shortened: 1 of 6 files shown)


JoshSEdwards:
I took a look at the changes and the implementation looks good. One thing we should tighten before merge is automated coverage around the new search path.

Recommend adding them in the existing MySQL integration specs:

  - `modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlRecordSetRepositoryIntegrationSpec.scala`
  - `modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlRecordSetCacheRepositoryIntegrationSpec.scala`

  Some ideas for what to add under the current "list record sets" sections to cover:

  1. Searching by FQDN returns both direct FQDN matches and records whose record_data points to that FQDN.
  2. Searching by FQDN returns records from multiple record types.
  3. A search that matches both `fqdn` and `record_data_value` in a single record only appears once (deduplication).
  4. Pagination (maxItems/nextId) still behaves correctly with the new union query.
  5. recordTypeFilter still applies correctly when the new path is active.

PR #1487, opened 2026-02-18
Properly parse and display multi-value TXT records in batch changes

Sokitha-21:
Fixes #1482.
Jira Ticket: VDNS-259

Changes in this pull request:

When submitting multi-value TXT records in batch changes with newline-separated values the system did not properly expand them into separate records so it appeared like: ANSWER SECTION:
txttest.shared. 7200 IN TXT ""thisisoneline",\010"thisistwoline",\010thing=value,\010thing2=anotherthing,".

The issue has been fixed by:
- Adding TXT record expansion logic to split multiline input into separate batch change entries before submission.
- Updating the HTML template to group multiple TXT values of the same record into a single row and display them as bullet-pointed values.

### DIFF
--- a/modules/portal/app/views/dnsChanges/dnsChangeDetail.scala.html
+++ b/modules/portal/app/views/dnsChanges/dnsChangeDetail.scala.html
@@ -128,7 +128,9 @@
                             </tr>
                             </thead>
                             <tbody>
-                            <tr ng-repeat="change in batch.changes|filter:query" ng-class="{changeError: change.outstandingErrors}">
+                            <tr ng-repeat="change in batch.changes|filter:query" 
+                                ng-hide="change.type=='TXT' && $index > 0 && batch.changes[$index - 1].zoneId === change.zoneId && batch.changes[$index - 1].recordName === change.recordName && batch.changes[$index - 1].type === change.type"
+                                ng-class="{changeError: change.outstandingErrors}">
                                 <td ng-bind="change.changeType"></td>
                                 <td class="wrap-long-text">{{change.inputName}}</td>
                                 <td class="wrap-long-text">{{change.recordName}}</td>
@@ -145,7 +147,14 @@
                                         {{change.record.ptrdname}}
                                     </td>
                                     <td ng-if="change.type=='TXT'" class="wrap-long-text">
-                                        {{change.record.text}}
+                                        <ul class="table-cell-list">
+                                        <li class="wrap-long-text"ng-repeat="c in batch.changes | filter:query" 
+                                            ng-if="c.zoneId === change.zoneId && 
+                                                    c.recordName === change.recordName && 
+                                                    c.type === change.type">
+                                            {{c.record.text}}
+                                        </li>
+                                        </ul>
                                     </td>
                                     <td ng-if="change.type=='MX'">
                                         <ul>
--- a/modules/portal/app/views/dnsChanges/dnsChangeDetail.scala.html
+++ b/modules/portal/app/views/dnsChanges/dnsChangeDetail.scala.html
@@ -129,7 +129,8 @@
                             </thead>
                             <tbody>
                             <tr ng-repeat="change in batch.changes|filter:query" 
-                                ng-hide="change.type=='TXT' && $index > 0 && batch.changes[$index - 1].zoneId === change.zoneId && batch.changes[$index - 1].recordName === change.recordName && batch.changes[$index - 1].type === change.type"
+                                ng-hide="change.type=='TXT' && $index > 0 && batch.changes[$index - 1].zoneId === change.zoneId && batch.changes[$index - 1].recordName === change.recordName
+                                 && batch.changes[$index - 1].type === change.type && batch.changes[$index - 1].changeType === change.changeType"
                                 ng-class="{changeError: change.outstandingErrors}">
                                 <td ng-bind="change.changeType"></td>
                                 <td class="wrap-long-text">{{change.inputName}}</td>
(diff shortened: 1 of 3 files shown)


modules/portal/public/lib/dns-change/dns-change-new.controller.js
```diff
@@ -103,6 +103,24 @@
                                 payload.changes[i] = newEntry;
                             }
                         }
+                        if(entry.type == 'TXT'){
```
arpit4ever: Basic DeleteRecordSet for TXT record with empty record data doesn't work.

modules/portal/public/lib/dns-change/dns-change-new.controller.js
```diff
@@ -103,6 +103,24 @@
...
+                            var splitValues = entry.record.text
+                                .split(/\r?\n/)
```
arpit4ever: Also, Delete + Add gets combined. 

Image
Image

arpit4ever (changes requested):
While adding multiline TXT record works, few other basic functions failed.

modules/portal/public/lib/dns-change/dns-change-new.controller.js
```diff
@@ -103,6 +103,24 @@
...
+                                .map(function(v) { return v.trim(); })
+                                .filter(function(v) { return v.length > 0; });
+                            if (splitValues.length > 1) {
+                                entry.record.text = splitValues[0];
+                                for (var j = 1; j < splitValues.length; j++) {
+                                    payload.changes.splice(i + j, 0, {
```
nspadaccino: So `i` in the outer loop is looping over payload.changes. But on this line we are splicing new changes into payload.changes, so the outer loop will also iterate over those spliced changes unnecessarily. Outside this inner for loop, we could advance `i `past the new entries so the outer loop doesn't look at them.

modules/portal/public/lib/dns-change/dns-change-new.controller.js
```diff
@@ -78,7 +78,7 @@
                 $scope.processing = true;
 
                 var payload = $scope.newBatch;
-
+
```
nspadaccino: remove trailing whitespace from this line

modules/portal/app/views/dnsChanges/dnsChangeDetail.scala.html
```diff
@@ -128,7 +128,9 @@ <h3 class="panel-title">Changes</h3>
                             </tr>
                             </thead>
                             <tbody>
-                            <tr ng-repeat="change in batch.changes|filter:query" ng-class="{changeError: change.outstandingErrors}">
+                            <tr ng-repeat="change in batch.changes|filter:query" 
+                                ng-hide="change.type=='TXT' && $index > 0 && batch.changes[$index - 1].zoneId === change.zoneId && batch.changes[$index - 1].recordName === change.recordName && batch.changes[$index - 1].type === change.type"
```
nspadaccino: Add a check for changeType to the conditional here to fix Add+Delete getting messed up

nspadaccino (changes requested):

modules/portal/app/views/dnsChanges/dnsChangeDetail.scala.html
```diff
@@ -145,7 +148,16 @@ <h3 class="panel-title">Changes</h3>
                                         {{change.record.ptrdname}}
                                     </td>
                                     <td ng-if="change.type=='TXT'" class="wrap-long-text">
-                                        {{change.record.text}}
+                                        <ul class="table-cell-list">
+                                        <li class="wrap-long-text"ng-repeat="c in batch.changes | filter:query"
```
nspadaccino: typo, there should be a space between "wrap-long-text" and ng-repeat

modules/portal/app/views/dnsChanges/dnsChangeDetail.scala.html
```diff
@@ -128,7 +128,10 @@ <h3 class="panel-title">Changes</h3>
                             </tr>
                             </thead>
                             <tbody>
-                            <tr ng-repeat="change in batch.changes|filter:query" ng-class="{changeError: change.outstandingErrors}">
+                            <tr ng-repeat="change in batch.changes|filter:query"
```
nspadaccino: The ng-hide compares against the unfiltered `batch.changes[$index - 1]`, but $index comes from the filtered list `batch.changes | filter:query`, so when a filter/search is active the two index spaces diverge and the wrong rows get hidden/shown. They only line up when no search filter is active. 

You can alias the filtered collection and index that:
```
  <tr ng-repeat="change in batch.changes | filter:query as filteredChanges"
      ng-hide="... filteredChanges[$index - 1] ...">
```

modules/portal/app/views/dnsChanges/dnsChangeDetail.scala.html
nspadaccino: For each TXT row rendered, this inner ng-repeat re-scans all of batch.changes (plus runs filter:query a second time) to gather that group's values, which is O(n^2) complexity.

It would be better to precompute the TXT grouping once in the dns-change-detail.controller.js when the batch loads, then have the view read from that map instead of re-scanning batch.changes for every row

nspadaccino (changes requested):

nspadaccino (approved):
LGTM

arpit4ever (approved):
LGTM

arpit4ever (changes requested):
One of the standard flow failed.
Scenario:
- User already has a txttest.shared. TXT record.
- User tries to create a new DNS change with just `Add` for new record data.
- Batch will throw an error stating record name already exists. 
- User add an entry for `DeleteRecordSet` for the TXT record. Now the batch should continue since it include Add & Delete (may not be in order). On current branch, this scenario fails.

arpit4ever (approved):
LGTM
Previous failing scenario must have been a local env issue. After resetting local env, the scenario works. 
PR approved.

PR #1494, opened 2026-03-06
Validate zoneId in Create RecordSet API

### LINKED ISSUE #1490: zoneId in Create RecordSet API Not Properly Validated
VDNS-307

Description:
When creating a RecordSet via the API (POST /zones/{zoneId}/recordsets), the zoneId is present both in the URI and in the request body. Currently, the implementation only considers the zoneId from the request body and does not validate that it matches the zoneId in the URI. 

Steps to Reproduce:
- Make a POST request to /zones/{zoneId}/recordsets with a different zoneId in the request body.
- Observe that the record is created using the body’s zoneId, not the URI’s.

Impact:
- Potential for records to be created in unintended zones.

Potential Solution:
- The zoneId in the URI should be treated as the primary source of truth.
- The zoneId in the request body should be marked optional.
- If both are provided, the API should validate that they match. If they do not match, the API should return a 400 Bad Request error.

Sokitha-21:
VDNS-307

Fixes #1490.

This PR fixes an issue in the Create RecordSet API (POST /zones/{zoneId}/recordsets) where the zoneId provided in the request body was used without validating it against the zoneId specified in the URI.

Changes in this pull request:
- The zoneId from the URI is used for processing the request.
- If a zoneId is provided in the request body and it does not match the URI zoneId, the API returns a 400 Bad Request error.

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/route/DnsJsonProtocol.scala
+++ b/modules/api/src/main/scala/vinyldns/api/route/DnsJsonProtocol.scala
@@ -40,6 +40,7 @@ trait DnsJsonProtocol extends JsonValidation {
   val dnsSerializers = Seq(
     CreateZoneInputSerializer,
     UpdateZoneInputSerializer,
+    CreateRecordSetInputSerializer,
     ZoneConnectionSerializer,
     AlgorithmSerializer,
     EncryptedSerializer,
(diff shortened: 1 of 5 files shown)


modules/api/src/main/scala/vinyldns/api/route/RecordSetRouting.scala
```diff
@@ -92,7 +93,14 @@ class RecordSetRoute(
   val recordSetRoute: Route = path("zones" / Segment / "recordsets") { zoneId =>
     (post & monitor("Endpoint.addRecordSet")) {
       authenticateAndExecuteWithEntity[ZoneCommandResult, RecordSet](
-        (authPrincipal, recordSet) => recordSetService.addRecordSet(recordSet, authPrincipal)
+        (authPrincipal, recordSet) =>   
+          recordSet match {
+            case badRs if badRs.zoneId.nonEmpty && badRs.zoneId != zoneId =>
```
JoshSEdwards: I don’t think this implements the part of #1490 that makes the body `zoneId` optional. `authenticateAndExecuteWithEntity` deserializes the body into a `RecordSet` before this match runs, and `RecordSetSerializer` currently requires `zoneId`. That means a request that omits `zoneId` will fail with `Missing RecordSet.zoneId` before we can copy the URI value onto the record set.

 Could we use a create-specific input model with an optional `zoneId`, validate it against the URI when it's present, and then construct the `RecordSet` using the URI value? That would represent the create request accurately and make the URI’s authority explicit. It would also be helpful to add a test showing that creation succeeds when the body omits `zoneId`.

modules/api/src/main/scala/vinyldns/api/route/RecordSetRouting.scala
```diff
@@ -92,7 +93,14 @@ class RecordSetRoute(
...
+              Left(InvalidRequest("zoneId in URI and body must match")).toResult
```
JoshSEdwards: This currently returns 422 because `InvalidRequest` maps to `UnprocessableEntity` in this route, and the new tests encode that behavior. However, both #1490 and the PR description say that a URI/body `zoneId` mismatch should return 400 Bad Request.

Could we resolve that discrepancy before merging? If 400 is the intended API behavior, this needs to use an error that maps to 400 and the tests should expect it. If 422 is intentional and consistent with the API’s existing validation behavior, then the issue and PR description should be updated so clients have one clear expectation.

JoshSEdwards (changes requested):

JoshSEdwards (dismissed):
LGTM

modules/api/src/main/scala/vinyldns/api/route/DnsJsonProtocol.scala
```diff
@@ -142,6 +143,61 @@ trait DnsJsonProtocol extends JsonValidation {
         ).mapN(UpdateZoneInput.apply)
   }
 
+  case object CreateRecordSetInputSerializer extends ValidationSerializer[vinyldns.api.route.CreateRecordSetInput] {
```
nspadaccino: This serializer is a near-verbatim copy of RecordSetSerializer above. The TTL bounds, name checks, extractRecords, and the CNAME multi-record rule are all duplicated, and the only real difference is zoneId being `.optional` instead of `.required`. This will drift the moment someone changes record-level validation in one place but not the other. Could we share the validation instead?

modules/api/src/main/scala/vinyldns/api/route/DnsJsonProtocol.scala
```diff
@@ -142,6 +143,61 @@ trait DnsJsonProtocol extends JsonValidation {
...
+        (js \ "status").default(RecordSetStatus, RecordSetStatus.Pending),
+        (js \ "created").default[Instant](Instant.now.truncatedTo(ChronoUnit.MILLIS)),
+        (js \ "updated").optional[Instant],
+        recordType.andThen(extractRecords(_, js \ "records")),
+        (js \ "id").default[String](UUID.randomUUID().toString),
+        (js \ "account").default[String]("system"),
+        (js \ "ownerGroupId").optional[String],
+        (js \ "recordSetGroupChange").optional[OwnershipTransfer],
+        (js \ "fqdn").optional[String]
+        ).mapN(vinyldns.api.route.CreateRecordSetInput.apply)
```
nspadaccino: `CreateRecordSetInput` is in this same package (vinyldns.api.route), so the fully-qualified name is redundant here and on the fromJson/toJson signatures. Plain `CreateRecordSetInput` reads cleaner.

modules/api/src/test/functional/tests/recordsets/create_recordset_test.py
```diff
@@ -57,6 +58,61 @@ def test_create_recordset_with_dns_verify(shared_zone_test_context):
...
+        url,
+        "POST",
+        client.headers,
+        json.dumps(new_rs),
+        status=400
+    )
+
+    assert_that(error, is_("zoneId in URI and body must match"))
+
+def test_create_recordset_omitted_zoneid(shared_zone_test_context):
```
nspadaccino, line 92: This creates a recordset with status=202 but never deletes it. Every other create test in this file wraps the create in try/finally and deletes on teardown (e.g. lines 51–58). Without cleanup this leaves a record in the shared zone, which pollutes other tests and can make reruns flaky. Please wrap it in the same pattern, waiting for the change to complete and then deleting.

nspadaccino (changes requested):

nspadaccino (approved):
LGTM

arpit4ever (approved):
LGTM

Tested both POST & PUT.
POST {{HTTP_TYPE}}://{{API_URL}}:{{API_PORT}}/zones/{{zone_id}}/recordsets
PUT {{HTTP_TYPE}}://{{API_URL}}:{{API_PORT}}/zones/{{zone_id}}/recordsets/{{record_id}}

JoshSEdwards (approved):
LGTM

PR #1495, opened 2026-03-17
Fix zone list pagination on MySQL 8.0 by adding explicit ORDER BY

nspadaccino:
- Zone list queries (listZones, listZonesByAdminGroupIds) relied on GROUP BY z.name       
  implicitly sorting results, which MySQL 8.0 no longer guarantees. This caused keyset
  pagination (z.name > startFrom) to return non-deterministic result sets, producing short  
  pages that ended pagination early and silently dropped zones from responses.
  - Added explicit ORDER BY z.name ASC to both queries.
  - Added integration tests that page through zones using nextId from each response and     
  assert it is present on every intermediate page.

Root cause: https://dev.mysql.com/blog-archive/removal-of-implicit-and-explicit-sorting-for-group-by/

Without deterministic ordering, LIMIT 101 returns an arbitrary subset of zones, the      
  z.name > startFrom cursor skips unpredictable ranges, and any page that happens to return
  <= maxItems rows terminates pagination.

### DIFF
--- a/modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlZoneRepositoryIntegrationSpec.scala
+++ b/modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlZoneRepositoryIntegrationSpec.scala
@@ -826,6 +826,82 @@ class MySqlZoneRepositoryIntegrationSpec
         .nextId shouldBe None
     }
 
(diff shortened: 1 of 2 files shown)


JoshSEdwards (approved):
LGTM

Jay07GIT (approved):
LGTM

PR #1496, opened 2026-03-26
Consolidate messages and make them configurable

### LINKED ISSUE #1292: Consolidate messages and make them configurable

- Consolidate all the message present in the platform to `Messages.scala` file to make it easier to locate any message we want.
- Make the messages configurable, so that people can change the messages according to their needs. Also, we can make the change to messages displayed immediately without having to make code changes and redeploying the code again.

Sokitha-21:
VDNS-405

Fixes https://github.com/vinyldns/vinyldns/issues/1292.

Changes in this pull request:

- Consolidated all the messages present in the platform to a single file Messages.scala making it easier to locate any messages we want.
- Added a configuration to enable changing the existing messages without having to change the code. Open source users can change the messages displayed as per their need using this config. Also, we can change any message immediately from this config instead of changing the code and cutting a new release.

### DIFF
--- a/modules/api/src/main/resources/application.conf
+++ b/modules/api/src/main/resources/application.conf
@@ -33,10 +33,24 @@ vinyldns {
   use-recordset-cache = ${?USE_RECORDSET_CACHE}
   load-test-data = false
   load-test-data = ${?LOAD_TEST_DATA}
-  # should be true while running locally or when we have only one api server/instance, for zone sync scheduler to work
+  # Should be true while running locally or when we have only one api server/instance, for zone sync scheduler to work
+  # Should be set to true only on a single server/instance else automated sync will be performed at every server/instance
   is-zone-sync-schedule-allowed = true
-  # should be set to true only on a single server/instance else automated sync will be performed at every server/instance
   is-zone-sync-schedule-allowed = ${?IS_ZONE_SYNC_SCHEDULE_ALLOWED}
+
+  # Configure Messages. Override existing messages with new messages. Refer Messages.scala file for existing messages
+  # Have a look on placeholders while making changes, else you may face errors
+  messages = [
+    # Place the existing message present at Messages.scala file in 'text' and the new message in 'override-text'
+    {
+      text = "Dummy message."
+      override-text = "Replacing dummy message with this."
+    },
+    {
+      text = "Cannot create group. A group, %s, is already associated with the email address %s. Please contact %s to be added to the group."
+      override-text = "Cannot create group. A group, %s, is already associated with the email address %s. Please contact %s to be added to the group. Visit FAQ for more information."
+    }
+  ]
 
   # configured backend providers
   backend {
--- a/modules/api/src/main/resources/application.conf
+++ b/modules/api/src/main/resources/application.conf
@@ -33,12 +33,13 @@ vinyldns {
   use-recordset-cache = ${?USE_RECORDSET_CACHE}
   load-test-data = false
   load-test-data = ${?LOAD_TEST_DATA}
-  # Should be true while running locally or when we have only one api server/instance, for zone sync scheduler to work
-  # Should be set to true only on a single server/instance else automated sync will be performed at every server/instance
+  # Enable only for local or single api server/instance to allow zone sync scheduler to run.
   is-zone-sync-schedule-allowed = true
+  # In multi-instance setups, keep this false to avoid duplicate automated sync across instances.
   is-zone-sync-schedule-allowed = ${?IS_ZONE_SYNC_SCHEDULE_ALLOWED}
 
-  # Configure Messages. Override existing messages with new messages. Refer Messages.scala file for existing messages
+  # Configure Messages. Override existing messages with new messages.
+  # Refer to path: vinyldns/modules/core/src/main/scala/vinyldns/core/Messages.scala file for existing messages.
   # Have a look on placeholders while making changes, else you may face errors
   messages = [
     # Place the existing message present at Messages.scala file in 'text' and the new message in 'override-text'
--- a/modules/core/src/main/scala/vinyldns/core/Messages.scala
+++ b/modules/core/src/main/scala/vinyldns/core/Messages.scala
@@ -205,7 +205,7 @@ object Messages {
 
   val OwnershipTransferAlreadyOwnedErrorMsg: String = "Record owner group with id %s already owns the record, new request is not needed.".orConfig
 
-  val OwnershipTransferInvalidPendingReviewErrorMsg: String = "Invalid ownership transfer status: %s".orConfig
+  val OwnershipTransferInvalidPendingReviewErrorMsg: String = "Invalid Ownership transfer status: %s".orConfig
 
   val OwnershipTransferInvalidApprovalErrorMsg: String = "Unable to %s the Ownership transfer status for the record: None".orConfig
 
--- a/modules/core/src/main/scala/vinyldns/core/Messages.scala
+++ b/modules/core/src/main/scala/vinyldns/core/Messages.scala
@@ -130,7 +130,7 @@ object Messages {
 
   val DeleteRecordDataDoesNotExistErrorMsg: String = "Record data %s does not exist for \"%s\".".orConfig
 
-  val NotAuthorizedErrorMsg: String = "User \"%s\" is not authorized. Contact %s owner group: %s at %s to make DNS changes.".orConfig
+  val NotAuthorizedErrorMsg: String = "The record \"%s\" is owned by the [%s] group(%s). Only members of this group may update the record. Please contact them for assistance: %s.".orConfig
 
   val InvalidIPv4CNameErrorMsg: String = "Invalid Cname: \"%s\", Valid CNAME record data should not be an IP address".orConfig
 
@@ -139,6 +139,13 @@ object Messages {
   val InvalidReverseCnameErrorMsg: String = "Invalid Cname: \"%s\", valid cnames must be letters, numbers, underscores, and hyphens, joined by dots, and terminated with a dot.".orConfig
 
   val InvalidUpdateRequestErrorMsg: String = "Cannot perform request for the record \"%s\". Add and Delete for the record with same record data exists in the batch.".orConfig
+
+  /* Deprecated errors */
+  val ExistingMultiRecordErrorMsg: String = "RecordSet with name %s and type %s cannot be updated in a single Batch Change because it contains multiple DNS records (%s).".orConfig
+
+  val NewMultiRecordErrorMsg: String = "Multi-record recordsets are not enabled for this instance of VinylDNS. Cannot create a new record set with multiple records for inputName %s and type %s.".orConfig
+
+  val UserIsNotAuthorizedMsg: String = "User \"%s\" is not authorized.".orConfig
 
   /* RecordSetValidations.scala */
   val RecordNameFilterErrorMsg: String = "Record Name Filter field must contain at least two letters or numbers and cannot have wildcard at both the start and end.".orConfig
@@ -585,4 +592,5 @@ object Messages {
   val nonExistentRecordDeleteMessage = "This record does not exist. No further action is required."
 
   val nonExistentRecordDataDeleteMessage = "Record data entered does not exist. No further action is required."
+
 }
--- a/modules/core/src/main/scala/vinyldns/core/Messages.scala
+++ b/modules/core/src/main/scala/vinyldns/core/Messages.scala
@@ -16,74 +16,571 @@
 
 package vinyldns.core
 
+import vinyldns.core.config.MessagesConfig._
+import vinyldns.core.config.Message
+
 object Messages {
 
-  // Error displayed when less than two letters or numbers is filled in Record Name Filter field in RecordSetSearch page
-  val RecordNameFilterError =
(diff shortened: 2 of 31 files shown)


modules/api/src/main/resources/application.conf
```diff
@@ -33,11 +33,25 @@ vinyldns {
...
   load-test-data = false
   load-test-data = ${?LOAD_TEST_DATA}
-  # should be true while running locally or when we have only one api server/instance, for zone sync scheduler to work
+  # Should be true while running locally or when we have only one api server/instance, for zone sync scheduler to work
+  # Should be set to true only on a single server/instance else automated sync will be performed at every server/instance
   is-zone-sync-schedule-allowed = true
-  # should be set to true only on a single server/instance else automated sync will be performed at every server/instance
   is-zone-sync-schedule-allowed = ${?IS_ZONE_SYNC_SCHEDULE_ALLOWED}
 
+  # Configure Messages. Override existing messages with new messages. Refer Messages.scala file for existing messages
```
JoshSEdwards: Recommend adding the full path to Messages.scala in the repo here

modules/api/src/main/resources/application.conf
```diff
@@ -33,11 +33,25 @@ vinyldns {
   use-recordset-cache = ${?USE_RECORDSET_CACHE}
   load-test-data = false
   load-test-data = ${?LOAD_TEST_DATA}
-  # should be true while running locally or when we have only one api server/instance, for zone sync scheduler to work
+  # Should be true while running locally or when we have only one api server/instance, for zone sync scheduler to work
+  # Should be set to true only on a single server/instance else automated sync will be performed at every server/instance
```
JoshSEdwards: Looks like a duplicate message with slightly different wording -- I'd pick just one.

modules/core/src/main/scala/vinyldns/core/Messages.scala
```diff
@@ -16,74 +16,571 @@
...
+
+  val CnameAtZoneApexErrorMsg: String = "CNAME cannot be the same name as zone \"%s\".".orConfig
+
+  val RecordManualReviewErrorMsg: String = "Record set with name \"%s\" requires manual review.".orConfig
+
+  val UnsupportedOperationErrorMsg: String = "%s is not yet implemented/supported in VinylDNS.".orConfig
+
+  val DeleteRecordDataDoesNotExistErrorMsg: String = "Record data %s does not exist for \"%s\".".orConfig
+
+  val NotAuthorizedErrorMsg: String = "User \"%s\" is not authorized. Contact %s owner group: %s at %s to make DNS changes.".orConfig
```
JoshSEdwards: Arpit suggested earlier this month to change this error message to be clearer since we get a lot of support questions about it. Seems like this is a good PR to make that change in. His suggestion (with some of my own tweaking) was:

```
The record <record name> is owned by the <group hyperlink> group. Only the members of this group may update the record. Please contact them for assistance: <group email>
```

modules/core/src/main/scala/vinyldns/core/domain/SingleChangeError.scala
```diff
@@ -64,14 +63,11 @@ object DomainValidationErrorType extends Enumeration {
       case _: RecordDoesNotExist => RecordDoesNotExist
       case _: InvalidUpdateRequest => InvalidUpdateRequest
       case _: CnameIsNotUniqueError => CnameIsNotUniqueError
-      case _: UserIsNotAuthorized => UserIsNotAuthorized
```
JoshSEdwards, line 67: Were these three error types intentionally removed? This might break historical batch changes because `SingleChangeError.errorType` is persisted as a string in protobuf (`setErrorType(error.errorType.toString)`) and read back with `DomainValidationErrorType.withName(...)`. If we remove an enum member, the old stored strings become unreadable.

https://github.com/vinyldns/vinyldns/blob/main/modules/core/src/main/scala/vinyldns/core/protobuf/BatchChangeProtobufConversions.scala#L102

I'd recommend keeping these enums to make sure old batch changes' error deserialization is not broken.

JoshSEdwards (changes requested):
Overall changes look good, and this will certainly make it easier to maintain and update our messaging. I noted just a few minor concerns to take a look at. Let me know if you have any questions -- the comment on `SingleChangeError.scala` is the most substantive to address.

modules/api/src/main/resources/application.conf
```diff
@@ -33,11 +33,26 @@ vinyldns {
   use-recordset-cache = ${?USE_RECORDSET_CACHE}
   load-test-data = false
   load-test-data = ${?LOAD_TEST_DATA}
-  # should be true while running locally or when we have only one api server/instance, for zone sync scheduler to work
+  # Enable only for local or single api server/instance to allow zone sync scheduler to run.
   is-zone-sync-schedule-allowed = true
-  # should be set to true only on a single server/instance else automated sync will be performed at every server/instance
+  # In multi-instance setups, keep this false to avoid duplicate automated sync across instances.
```
JoshSEdwards, line 38: These are still two comments explaining the same single-instance requirement, just reworded and split around the setting. Recommend consolidating them into one concise comment so the configuration guidance isn’t duplicated.

modules/core/src/main/scala/vinyldns/core/Messages.scala
```diff
@@ -16,76 +16,581 @@
...
+
+  val CnameAtZoneApexErrorMsg: String = "CNAME cannot be the same name as zone \"%s\".".orConfig
+
+  val RecordManualReviewErrorMsg: String = "Record set with name \"%s\" requires manual review.".orConfig
+
+  val UnsupportedOperationErrorMsg: String = "%s is not yet implemented/supported in VinylDNS.".orConfig
+
+  val DeleteRecordDataDoesNotExistErrorMsg: String = "Record data %s does not exist for \"%s\".".orConfig
+
+  val NotAuthorizedErrorMsg: String = "The record \"%s\" is owned by the [%s] group(%s). Only members of this group may update the record. Please contact them for assistance: %s.".orConfig
```
JoshSEdwards, line 133: This currently produces `[group name] group(/groups/id)`, which isn’t valid Markdown, and the portal renders these batch errors with plain Angular interpolation, so users will see the brackets and URL as text rather than a hyperlink. Recommend either adding link-aware rendering and emitting a valid link, or keeping this message plain text without the pseudo-link syntax.

JoshSEdwards (changes requested):

modules/core/src/main/scala/vinyldns/core/Messages.scala
```diff
@@ -16,76 +16,581 @@
...
+  val OwnershipTransferInvalidCancelErrorMsg: String = "Unable to cancel the Ownership transfer. Current status: %s".orConfig
+
+  val OwnershipTransferUnauthorisedCancelErrorMsg: String = "Unauthorised to Cancel the ownership transfer".orConfig
+
+  val OwnershipTransferUnauthorizedChangeErrorMsg: String = "Unauthorized to change ownership transfer status to '%s'".orConfig
+
+
+
+  /* BatchChangeErrors.scala */
+  val BatchChangeNotFoundErrorMsg: String = "Batch change with id %s cannot be found".orConfig
```
arpit4ever: Change this to:
Batch change with id %s not found.

modules/core/src/main/scala/vinyldns/core/config/MessagesConfig.scala
```diff
@@ -0,0 +1,30 @@
...
+import pureconfig._
+import pureconfig.ConfigSource
+import pureconfig.generic.auto._
+import pureconfig.error.ConfigReaderFailures
+
+final case class Message(text: String, overrideText: String)
+final case class MessagesConfig(messages: List[Message])
+
+object MessagesConfig {
+  val appConfigSource: ConfigObjectSource = ConfigSource.resources("application.conf")
```
nspadaccino: Loading application.conf directly off the classpath bypasses the normal typesafe-config stack: no reference.conf fallback, no merge of the operator's external
-Dconfig.file, no env substitution. Operators using the standard external-config deployment model get no message overrides at all; only messages baked into
a bundled application.conf on the classpath are honored. Separately, messages is a required field with no default, so any module whose classpath application.conf lacks vinyldns.messages (portal, tests, r53/sqs standalone) yields a Left. Add a messages = [] default via a core reference.conf and load through the standard stack

modules/core/src/main/scala/vinyldns/core/Messages.scala
```diff
@@ -16,76 +16,581 @@
...
-
-  val EmailValidationErrorMsg = "Please enter a valid Email. Valid domains should end with"
-
-  val InvalidEmailValidationErrorMsg = "Please enter a valid Email."
-
-  val DotsValidationErrorMsg = "Please enter a valid Email. Number of dots allowed after @ is"
+  // Getting the messages present in the config file
+  val configMessages: Map[String, Message] = messages.right.toOption match {
+    case Some(value) => value.messages.map(value => value.text -> value).toMap
+    case None => Map("error" -> Message("config-not-found", "config-not-found"))
```
nspadaccino: On config-load failure the ConfigReaderFailures is discarded and silently replaced with a dummy entry, so an operator with a malformed messages block gets no
signal that none of their overrides loaded; every message just falls back to its hardcoded default. At minimum log the Left here so it's diagnosable; consider failing fast on a malformed config rather than degrading silently.

modules/core/src/main/scala/vinyldns/core/Messages.scala
```diff
@@ -16,76 +16,581 @@
...
+  }
+
+  // Checking if a message is present in config file and overriding the existing message
+  implicit class MessagesStringExtension(existingMessage: String) {
+    def orConfig: String =
```
nspadaccino: configMessages(existingMessage) is looked up three times after contains. Use a single configMessages.get(existingMessage) match. Also Message.overrideText is a plain String; if an operator omits override-text in a messages entry, pureconfig fails the entire MessagesConfig load, which then gets swallowed to the dummy map (see above), silently dropping every override. Make overrideText an Option[String] and treat absent as "no override."

nspadaccino (changes requested):

modules/core/src/main/scala/vinyldns/core/Messages.scala
```diff
@@ -16,76 +16,581 @@
...
+
+  val PendingUpdateErrorMsg: String = "RecordSet with id %s, name %s and type %s currently has a pending change".orConfig
+
+  val RecordNameLengthErrorMsg: String = "record set name %s is too long".orConfig
+
+  val ReverseLookupErrorMsg: String = "%s is not valid in reverse lookup zone.".orConfig
+
+  val InvalidPtrErrorMsg: String = "PTR is not valid in forward lookup zone".orConfig
+
+  val InvalidRequestErrorMsg: String = "Record with fqdn '%s.%s' cannot be created. " +
```
nspadaccino: .orConfig binds to the second string literal only, not the whole concatenation:
`"Record with fqdn '%s.%s' cannot be created. " + "Please check ... there.".orConfig`
So the config text key an operator must supply to override this message is just the trailing fragment "Please check ... there.", and an override replaces only that tail while the "Record with fqdn ..." prefix stays hardcoded. Same bug at MoreDotsThanAllowedErrorMsg (line 2400) and InvalidQueueNameErrorMsg (line
2787). Wrap the full concatenation in parens: ("a" + "b").orConfig.

modules/core/src/main/scala/vinyldns/core/Messages.scala
```diff
@@ -16,76 +16,581 @@
...
+
+  val MissingDSDigestMsg: String = "Missing DS.digest".orConfig
+
+  val DigestConvertMsg: String = "Could not convert digest to valid hex".orConfig
+
+  val MissingTXTTextMsg: String = "Missing TXT.text".orConfig
+
+  val TXTRecordValidationMsg: String = "TXT record must be less than 64764 characters".orConfig
+
+  val NSDataErrorMsg: String = "NS data must absolute".orConfig //"NS data must be a positive integer"
```
nspadaccino: I think the message should remain "NS data must be a positive integer"

modules/api/src/main/resources/application.conf
```diff
@@ -33,11 +33,27 @@ vinyldns {
...
+  # zone sync scheduler to run. Keep false in multi-instance setups to avoid
+  # duplicate automated syncs across instances.
   is-zone-sync-schedule-allowed = true
-  # should be set to true only on a single server/instance else automated sync will be performed at every server/instance
   is-zone-sync-schedule-allowed = ${?IS_ZONE_SYNC_SCHEDULE_ALLOWED}
 
+  # Configure Messages. Override existing messages with new messages.
+  # Refer to path: vinyldns/modules/core/src/main/scala/vinyldns/core/Messages.scala file for existing messages.
+  # Have a look on placeholders while making changes, else you may face errors
+  messages = [
```
nspadaccino: There's no test exercising the override mechanism itself: MessagesConfig loading, .orConfig with an override present vs. absent, the missing-key fallback, or an override that breaks a %d/%b placeholder. Can you add a small core unit test covering override-present / override-absent / malformed-placeholder.

nspadaccino (changes requested):

PR #1515, opened 2026-06-16
Paginate zone sync database load

nspadaccino:
## Summary

  - Paginates the VinylDNS database load during zone sync to prevent JDBC                                                                                      
    socket timeouts on large zones (100k+ records).
    Uses the existing cursor-based pagination in `listRecordSets` with a                                                                                       
    configurable page size (default 5000).                                                                                                                     
  - Splits the single write transaction into batches of 1000 changes each,                                                                                     
    reducing lock hold time and deadlock risk during sync.                                                                                                     
  - Adds `zone-sync-page-size` config option and per-page/per-batch logging                                                                                    
    for operational visibility.

### DIFF
--- a/modules/api/src/main/resources/reference.conf
+++ b/modules/api/src/main/resources/reference.conf
@@ -63,6 +63,7 @@ vinyldns {
 
   zone-sync-page-size = 5000 # number of records per page when loading zones from the database during sync
 
+
   queue {
     class-name = "vinyldns.sqs.queue.SqsMessageQueueProvider"
 
--- a/modules/api/src/main/resources/reference.conf
+++ b/modules/api/src/main/resources/reference.conf
@@ -61,6 +61,8 @@ vinyldns {
 
   max-zone-size = 60000 # number of records that can be in a zone
 
+  zone-sync-page-size = 5000 # number of records per page when loading zones from the database during sync
+
   queue {
     class-name = "vinyldns.sqs.queue.SqsMessageQueueProvider"
 
--- a/modules/api/src/main/scala/vinyldns/api/config/ServerConfig.scala
+++ b/modules/api/src/main/scala/vinyldns/api/config/ServerConfig.scala
@@ -36,12 +36,13 @@ final case class ServerConfig(
                                useRecordSetCache: Boolean,
                                loadTestData: Boolean,
                                isZoneSyncScheduleAllowed: Boolean,
+                               zoneSyncPageSize: Int,
                              )
 object ServerConfig {
 
   import ZoneRecordValidations.toCaseIgnoredRegexList
 
-  implicit val configReader: ConfigReader[ServerConfig] = ConfigReader.forProduct13[
+  implicit val configReader: ConfigReader[ServerConfig] = ConfigReader.forProduct14[
     ServerConfig,
     Int,
     Int,
@@ -55,7 +56,8 @@ object ServerConfig {
     Boolean,
     Boolean,
     Boolean,
-    Boolean
+    Boolean,
+    Int
   ](
     "health-check-timeout",
     "default-ttl",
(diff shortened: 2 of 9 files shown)


modules/api/src/main/resources/reference.conf
```diff
@@ -61,6 +61,9 @@ vinyldns {
 
   max-zone-size = 60000 # number of records that can be in a zone
 
+  zone-sync-page-size = 5000 # number of records per page when loading zones from the database during sync
```
JoshSEdwards, line 64: Just flagging that once this is merged, we'll need a deployment PR to add this new param to application.conf

modules/api/src/main/scala/vinyldns/api/engine/ZoneSyncHandler.scala
```diff
@@ -142,27 +142,30 @@ object ZoneSyncHandler extends DnsConversions with Monitored with TransactionPro
...
+                s"batchSize=$syncBatchSize; zoneChange='${zoneChange.id}'"
+            )
 
-              // join together the results of saving both the record changes as well as the record sets
-              for {
-                _ <- saveRecordChanges
-                _ <- saveRecordSets
-                _ <- saveRecordSetDatas
-              } yield zoneChange.copy(
+            changeBatches.zipWithIndex.foldLeft(IO.unit) { case (prevIO, (batch, idx)) =>
```
JoshSEdwards: This currently commits each batch in a separate transaction, so it looks like if a later batch fails, the earlier record changes and record sets remain applied even though the sync is marked failed and the zone returns to Active. Let me know if I'm misunderstanding this.

The rollback tests also call the copied testRunSyncFunc, which still contains the old single-transaction implementation. So this is completely bypassed in the regression tests.

I think we'd need to introduce some sort of resumable/checkpointed sync model if we want to do this.

modules/api/src/main/scala/vinyldns/api/config/ServerConfig.scala
```diff
@@ -36,12 +36,13 @@ final case class ServerConfig(
                                useRecordSetCache: Boolean,
                                loadTestData: Boolean,
                                isZoneSyncScheduleAllowed: Boolean,
+                               zoneSyncPageSize: Int,
```
JoshSEdwards, line 40: We're currently not protecting against zero or negative page sizes. A value of zero would cause the repository's look-ahead query to return an empty page with no next cursor, so a populated zone would be treated as empty and produce a full zone create diff. We should validate this as strictly positive in the config reader to make sure we cover invalid values.

JoshSEdwards (changes requested):

modules/api/src/main/resources/reference.conf
nspadaccino: I'll follow up with a deployment PR to add zone-sync-page-size to application.conf once this merges

modules/api/src/main/scala/vinyldns/api/config/ServerConfig.scala
nspadaccino: Fixed. Added an emap on the config reader so zone-sync-page-size must be strictly positive; zero or negative now fails config load with a clear message (must be a positive integer).

modules/api/src/main/scala/vinyldns/api/engine/ZoneSyncHandler.scala
nspadaccino: Rather than take on a resumable/checkpointed model in this PR, I've reverted the write side back to a single atomic transaction, restoring the all-or-nothing rollback behavior. The regression tests now mirror the real runSync path again. The read-side pagination in ZoneViewLoader, which is what actually addresses the JDBC socket timeouts on large zones, is retained, so the core fix stands. We can revisit checkpointed writes separately if lock hold time becomes a concern.

JoshSEdwards (approved):
LGTM

PR #1516, opened 2026-06-17
Tighten zone authorization checks

JoshSEdwards:
## Summary
- enforce zone visibility when looking up a zone by name or listing failed zone/record changes
- prevent global record set searches from returning zones the caller cannot access
- validate that a deleted record set belongs to the requested zone
- share MySQL zone-accessor filtering logic across zone and recordset repositories

## API behavior changes
- `GET /zones/name/{zoneName}` now returns `403` when the caller cannot access the zone instead of returning zone details.
- `GET /zones/deleted` only lists deleted zones visible to the caller.
- `GET /zones/{zoneId}/failed` only lists failed zone changes for zones visible to the caller.
- `GET /recordsets` global searches now omit recordsets in zones the caller cannot access.
- `GET /recordsets/{recordSetId}/failed` only returns failed recordset changes for zones visible to the caller.
- `DELETE /zones/{zoneId}/recordsets/{recordSetId}` now verifies the recordset belongs to the requested zone before authorizing deletion.

## Testing
- sbt -Dsbt.log.noformat=true -Dsbt.supershell=false "project api" "testOnly vinyldns.api.domain.record.RecordSetServiceSpec vinyldns.api.domain.zone.ZoneServiceSpec"
- sbt -Dsbt.log.noformat=true -Dsbt.supershell=false "mysql/IntegrationTest/testOnly vinyldns.mysql.repository.MySqlRecordSetRepositoryIntegrationSpec vinyldns.mysql.repository.MySqlRecordSetCacheRepositoryIntegrationSpec" (compiled, test execution requires local MariaDB on localhost:19002)

VDNS-352

### DIFF
--- a/modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlRecordSetRepositoryIntegrationSpec.scala
+++ b/modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlRecordSetRepositoryIntegrationSpec.scala
@@ -725,8 +725,11 @@ class MySqlRecordSetRepositoryIntegrationSpec
         id = UUID.randomUUID().toString
       )
 
+      val allAccessChange = makeTestAddChange(allAccessRecord, allAccessZone)
+      val noAccessChange = makeTestAddChange(noAccessRecord, noAccessZone)
+
       saveZones(Seq(allAccessZone, noAccessZone))
-      insert(List(makeTestAddChange(allAccessRecord, allAccessZone), makeTestAddChange(noAccessRecord, noAccessZone)))
+      insert(List(allAccessChange, noAccessChange))
 
       val found = repo
         .listRecordSets(
@@ -742,7 +745,7 @@ class MySqlRecordSetRepositoryIntegrationSpec
         )
         .unsafeRunSync()
 
-      found.recordSets should contain theSameElementsAs List(recordSetWithFQDN(allAccessRecord, allAccessZone))
+      found.recordSets should contain theSameElementsAs List(recordSetWithFQDN(allAccessChange.recordSet, allAccessZone))
     }
     "return no recordsets when no zoneId or recordNameFilter are given" in {
       val found =
--- a/modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlRecordSetRepositoryIntegrationSpec.scala
+++ b/modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlRecordSetRepositoryIntegrationSpec.scala
@@ -22,9 +22,11 @@ import org.scalatest._
 import org.scalatest.matchers.should.Matchers
 import org.scalatest.wordspec.AnyWordSpec
 import scalikejdbc.DB
+import vinyldns.core.TestMembershipData.{okGroup, okUser}
 import vinyldns.core.domain.record._
 import vinyldns.core.domain.record.RecordType._
-import vinyldns.core.domain.zone.Zone
+import vinyldns.core.domain.auth.AuthPrincipal
+import vinyldns.core.domain.zone.{Zone, ZoneACL}
 import vinyldns.mysql.TestMySqlInstance
 import vinyldns.mysql.repository.MySqlRecordSetRepository.PagingKey
 import vinyldns.mysql.TransactionProvider
@@ -41,6 +43,7 @@ class MySqlRecordSetRepositoryIntegrationSpec
   import vinyldns.core.TestRecordSetData._
   import vinyldns.core.TestZoneData._
   private val repo = TestMySqlInstance.recordSetRepository.asInstanceOf[MySqlRecordSetRepository]
+  private val zoneRepo = TestMySqlInstance.zoneRepository
 
   override protected def beforeEach(): Unit = clear()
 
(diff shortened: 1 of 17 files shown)


modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
```diff
@@ -292,6 +298,23 @@ class MySqlRecordSetRepository extends RecordSetRepository with Monitored {
...
+    val accessors = buildZoneSearchAccessorList(authPrincipal.signedInUser, authPrincipal.memberGroupIds)
+    sqls"zone_id IN (SELECT za.zone_id FROM zone_access za WHERE za.accessor_id IN ($accessors))"
+  }
+
+  private def buildZoneSearchAccessorList(user: User, groupIds: Seq[String]): Seq[String] = {
+    val allAccessors = user.id +: groupIds
+
+    if (allAccessors.length > MAX_ACCESSORS) {
+      logger.warn(
+        s"User ${user.userName} with id ${user.id} is in more than $MAX_ACCESSORS groups, no all zones maybe returned!"
```
nspadaccino: Typo, "maybe" -> "may be"

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetCacheRepository.scala
```diff
@@ -405,6 +411,23 @@ class MySqlRecordSetCacheRepository
       }
     }
 
+  private def buildAccessFilter(authPrincipal: AuthPrincipal): SQLSyntax = {
```
nspadaccino, line 411: buildAccessFilter and buildZoneSearchAccessorList are duplicated across MySqlRecordSetRepository.scala and MySqlRecordSetCacheRepository.scala (and MySqlZoneRepository though that wasn't touched in this PR). Not blocking, but I think it would be a good idea to extract this logic as a shared helper

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetCacheRepository.scala
```diff
@@ -405,6 +411,23 @@ class MySqlRecordSetCacheRepository
...
+    val accessors = buildZoneSearchAccessorList(authPrincipal.signedInUser, authPrincipal.memberGroupIds)
+    sqls"recordset.zone_id IN (SELECT za.zone_id FROM zone_access za WHERE za.accessor_id IN ($accessors))"
+  }
+
+  private def buildZoneSearchAccessorList(user: User, groupIds: Seq[String]): Seq[String] = {
+    val allAccessors = user.id +: groupIds
+
+    if (allAccessors.length > MAX_ACCESSORS) {
+      logger.warn(
+        s"User ${user.userName} with id ${user.id} is in more than $MAX_ACCESSORS groups, no all zones maybe returned!"
```
nspadaccino: Typo, "maybe" -> "may be"

modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlRecordSetRepositoryIntegrationSpec.scala
```diff
@@ -650,6 +657,42 @@ class MySqlRecordSetRepositoryIntegrationSpec
         .map(r => recordSetWithFQDN(r, okZone))
         .reverse
     }
+    "omit recordsets from groups if the user has more than 30 groups when doing a global search" in {
```
nspadaccino, line 660: Can we also add a test case asserting that recordsets from shared zones are listed even when the user has no membership? That will test the "EVERYONE" path

nspadaccino (changes requested):
Just a few changes. Also, could you update the PR description with the APIs that will have their behavior changed?

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetRepository.scala
JoshSEdwards: Fixed in 26e039e1. I moved the shared accessor list logic into a `MySqlAccessors` helper to centralize the repeat code and made the warning message clearer.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetCacheRepository.scala
JoshSEdwards: Agreed, done in 26e039e1. I extracted the shared accessor list behavior into `MySqlAccessors` and wired the zone, zone-change, recordset, and recordset-cache repositories through it. I left the SQL predicates local to each repo since the table aliases differ.

modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetCacheRepository.scala
JoshSEdwards: Fixed in 26e039e1 with the shared `MySqlAccessors` helper.

modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlRecordSetRepositoryIntegrationSpec.scala
JoshSEdwards: Added coverage in 26e039e1 for global recordset searches returning records from zones shared with everyone when the user has no group memberships. I added the same case for both `MySqlRecordSetRepositoryIntegrationSpec` and `MySqlRecordSetCacheRepositoryIntegrationSpec`.

nspadaccino (approved):
Looks good

PR #1517, opened 2026-06-17
Sanitize portal autocomplete rendering

JoshSEdwards:
## Summary
- Sanitize portal autocomplete result rendering for group and zone searches by treating labels as text and only highlighting matched substrings
- Render group change messages as plain text with preserved line breaks in the group detail view
- Add unit coverage for the sanitized autocomplete rendering and change message formatting

## Testing
- `cd modules/portal && npx grunt unit`
- Passed: 174 tests, 0 failures

VDNS-353

### DIFF
--- a/modules/portal/public/lib/controllers/controller.groups.js
+++ b/modules/portal/public/lib/controllers/controller.groups.js
@@ -31,6 +31,41 @@ angular.module('controller.groups', []).controller('GroupsController', function 
     $scope.validEmailDomains= [];
     $scope.maxGroupItemsDisplay = 3000;
 
+    function escapeRegExp(value) {
+        return value.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
+    }
+
+    function renderAutocompleteItem(ul, item, term) {
+        var label = String(item.label || "");
+        var $item = $("<li></li>").data("ui-autocomplete-item", item.value);
+        var $content = $("<div></div>");
+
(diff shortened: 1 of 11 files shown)


modules/portal/public/lib/controllers/controller.groups.js
```diff
@@ -31,6 +31,41 @@ angular.module('controller.groups', []).controller('GroupsController', function
     $scope.validEmailDomains= [];
     $scope.maxGroupItemsDisplay = 3000;
 
+    function escapeRegExp(value) {
```
nspadaccino: Perhaps escapeRegExp and renderAutocompleteItem could be extracted as shared helpers since they are being used across multiple files?

nspadaccino (commented):
It seems like this vulnerability is also in recordsets.controller.js at line 134. That should probably be addressed in this PR as well

modules/portal/public/lib/controllers/controller.groups.js
JoshSEdwards: Agreed, done in 811e3ebd. I extracted the autocomplete escaping/rendering into a shared vinyldnsAutocomplete helper and updated the groups, zones, and recordset autocomplete renderers to use it.

Also updated the global recordset search autocomplete to use the same safe renderer and added a regression spec.

nspadaccino (approved):
LGTM

PR #1518, opened 2026-06-17
Reject custom zone connections

JoshSEdwards:
## Summary
- reject custom zone connection and transfer connection inputs during zone creation
- reject updates that introduce or modify custom zone connections while allowing removals
- add zone service and validation coverage for the new invalid request behavior

## Testing
- `sbt -batch "api/testOnly vinyldns.api.domain.zone.ZoneValidationsSpec vinyldns.api.domain.zone.ZoneServiceSpec"`
- passed: 115 tests, 0 failures

VDNS-356

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneValidations.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneValidations.scala
@@ -23,11 +23,15 @@ import java.time.temporal.ChronoUnit
 import vinyldns.api.Interfaces.ensuring
 import vinyldns.core.domain.membership.User
 import vinyldns.core.domain.record.RecordType
-import vinyldns.core.domain.zone.{ACLRule, Zone, ZoneACL}
+import vinyldns.core.domain.zone.{ACLRule, Zone, ZoneACL, ZoneConnection}
 
 import scala.util.{Failure, Success, Try}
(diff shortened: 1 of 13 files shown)


modules/api/src/main/scala/vinyldns/api/domain/zone/ZoneValidations.scala
```diff
@@ -37,6 +41,22 @@ class ZoneValidations(syncDelayMillis: Int) {
       case _ => Right(())
     }
 
+  def noCustomZoneConnections(
+      connection: Option[ZoneConnection],
+      transferConnection: Option[ZoneConnection]
+  ): Either[Throwable, Unit] =
+    ensuring(customConnectionError)(connection.isEmpty && transferConnection.isEmpty)
+
+  def noCustomZoneConnectionUpdates(newZone: Zone, existingZone: Zone): Either[Throwable, Unit] =
```
nspadaccino, line 51: I think a comment documenting the functionality would be helpful here, like "allows removing a pre-existing connection, but rejects introducing or modifying one"

JoshSEdwards: Added in 51986625

nspadaccino (approved):
LGTM

PR #1523, opened 2026-06-30
Generate API credentials with SecureRandom

nspadaccino:
## Summary

  API access/secret keys were generated via `RandomStringUtils.randomAlphanumeric(20)`,
  backed by a predictable JVM-default PRNG (SRC-021). `User.generateKey` now draws from
  a `SecureRandom`; the alphanumeric set and 20-char length are unchanged, so format
  stays compatible while entropy becomes cryptographically strong. Both keys flow
  through `generateKey`, so one fix covers both.

  ## Tests

  New `UserSpec`: keys are 20-char alphanumeric and unique; default source is a
  `SecureRandom` (guards against PRNG fallback); `generateKey` derives output solely
  from the supplied source; `regenerateCredentials()` rotates both keys.

  ## Note

  Existing keys should be rotated operationally (self-service via `regenerateCredentials()`);
  this PR does not force a mass invalidation.

### DIFF
--- a/modules/core/src/main/scala/vinyldns/core/domain/membership/User.scala
+++ b/modules/core/src/main/scala/vinyldns/core/domain/membership/User.scala
@@ -16,7 +16,8 @@
 
 package vinyldns.core.domain.membership
 
-import java.util.UUID
+import java.security.SecureRandom
+import java.util.{Random, UUID}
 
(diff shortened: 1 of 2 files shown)


JoshSEdwards (approved):
LGTM

PR #1524, opened 2026-06-30
Fix SQL injection in MySQL repository list/search queries

nspadaccino:
## Summary

  Fixes authenticated SQL injection in six MySQL repository list/search paths that
  built queries by interpolating request-derived values directly into the SQL
  string and executing via `SQL(query)`, bypassing ScalikeJDBC parameterization.
  Any authenticated user could break out of the quoted literals to bypass
  per-tenant ACL/accessor scoping or exfiltrate arbitrary tables via
  `UNION`/boolean/time-based payloads (zone, group, user, and batch-change data —
  including encrypted connection/credential material).

  The fix is mechanical and behavior-preserving: every tainted value is now a bound
  parameter.

  ## Changes

  | Method | File | Bound values |
  |---|---|---|
  | `listZones` | `MySqlZoneRepository` | `nameFilter` (LIKE), `startFrom`, `LIMIT`, reverse-zone RLIKE |
  | `listZonesByAdminGroupIds` | `MySqlZoneRepository` | `adminGroupIds` (IN list), `startFrom`, `LIMIT` |
  | `getGroupsByName(String)` | `MySqlGroupRepository` | `nameFilter` (LIKE), via the `sql"…"` interpolator |
  | `listDeletedZones` | `MySqlZoneChangeRepository` | `nameFilter` (LIKE) |
  | `getUsers` | `MySqlUserRepository` | `startFrom`, `LIMIT` |
  | `getBatchChangeSummaries` | `MySqlBatchChangeRepository` | `userId`, `userName`, `approvalStatus`, `batchStatus`, date-time range |

  ## Notes

  - **Wildcard semantics unchanged and documented** at each LIKE site: `*` remains
    the user-facing wildcard (mapped to SQL `%`), otherwise prefix match. Literal
    `%`/`_` are not yet escaped — that's a behavior change left as a follow-up, not
    required to close the injection.
  - **Date-time range** is bound as strings rather than parsed to `Instant`. This
    fully closes the injection without touching the `BatchChangeRepository` trait
    or service layer; typed parsing/validation upstream remains a worthwhile
    defense-in-depth follow-up.

  ## Testing

  - Added injection regression tests for group search (`' OR '1'='1`, quote-bearing
    input, stacked statements asserting the table survives) and batch-change
    summaries (`userName` and date-range payloads).
  - Full `mysql` integration suite passes against a live MariaDB: **148 tests, 0
    failures**, including the existing wildcard/filter/pagination assertions that
    pin the preserved behavior.

### DIFF
--- a/modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlGroupRepository.scala
+++ b/modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlGroupRepository.scala
@@ -187,14 +187,12 @@ class MySqlGroupRepository extends GroupRepository with GroupProtobufConversions
     monitor("repo.Group.getGroupByName") {
       IO {
         logger.debug(s"Getting groups with name: $nameFilter")
-        // Wildcard semantics: '*' is the user-facing wildcard (mapped to SQL '%'),
-        // otherwise prefix match. The pattern is bound, never interpolated into SQL.
-        val pattern =
-          if (nameFilter.contains('*')) nameFilter.replace('*', '%')
-          else s"$nameFilter%"
+        // '*' is the only user-facing wildcard; literal LIKE metacharacters are
+        // escaped and the pattern is bound, never interpolated into SQL.
+        val pattern = LikePattern.prefix(nameFilter)
 
         DB.readOnly { implicit s =>
-          sql"SELECT data FROM `groups` WHERE name LIKE $pattern"
+          sql"SELECT data FROM `groups` WHERE name LIKE $pattern ESCAPE '\\'"
             .map(toGroup(1))
             .list()
             .apply()
--- a/modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlGroupRepository.scala
+++ b/modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlGroupRepository.scala
@@ -187,19 +187,14 @@ class MySqlGroupRepository extends GroupRepository with GroupProtobufConversions
     monitor("repo.Group.getGroupByName") {
       IO {
         logger.debug(s"Getting groups with name: $nameFilter")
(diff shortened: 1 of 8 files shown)


modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlGroupRepository.scala
```diff
@@ -187,19 +187,14 @@ class MySqlGroupRepository extends GroupRepository with GroupProtobufConversions
...
-        val sb = new StringBuilder
-        sb.append(initialQuery)
-        val groupsLike = if (nameFilter.contains('*')) {
-          s" LIKE '${nameFilter.replace('*', '%')}'"
-        } else {
-          s" LIKE '$nameFilter%'"
-        }
-        sb.append(groupsLike)
-        val query = sb.toString()
+        // Wildcard semantics: '*' is the user-facing wildcard (mapped to SQL '%'),
```
JoshSEdwards: `LIKE` filters still treat `%` and `_` as SQL wildcards even though the comments say SQL metacharacters are handled literally. If `*` is the only user-facing wildcard, can we centralize that translation in a small helper that turns `*` into `%`, escapes literal `%`/`_`, and uses `LIKE ? ESCAPE '\\'` everywhere?

nspadaccino: Done. Added a LikePattern helper (modules/mysql/.../repository/LikePattern.scala) that centralizes the translation: * -> %, and literal `%/_/\ `get backslash-escaped. All three user-facing sites (getGroupsByName, listDeletedZones, listZones) now bind the pattern and use `LIKE ... ESCAPE '\\'.`

JoshSEdwards (approved):
LGTM

PR #1525, opened 2026-06-30
Enforce SigV4 freshness and mandatory signed headers

nspadaccino:
### Problem
  `Aws4Authenticator.authenticateReq` only computed and compared the request
  signature. Two AWS SigV4 spec checks were missing:

  1. **No freshness window.** The parsed `X-Amz-Date`/`Date` was used solely to
     format the string-to-sign — never compared against the current time. A
     captured, fully-signed request verified **forever**.
  2. **Attacker-chosen `SignedHeaders`.** Nothing required `host` (mandatory per
     spec) or `x-amz-date` to be in the signed set, and declared-but-absent signed
     headers were logged and silently dropped rather than rejected. A signature
     could be bound to method/path/body only — no host, no date.

  **Impact (High):** an attacker who captures one signed request (TLS-terminating
  proxy, logs with full headers, MITM on a misconfigured client) can replay it
  indefinitely against state-changing routes (`POST /zones`, `DELETE /zones/:id`,
  batch changes, …). Because `host` need not be signed, a request signed for one
  deployment is valid against any other deployment sharing the same user secret.

  ### Fix
  `authenticateReq` now runs request-metadata validation **before** the signature
  comparison and rejects on any failure (`AuthRejected`):

  - **Freshness:** reject if `|requestDate − now| > 15 min`.
  - **Scope-date binding:** the credential-scope `yyyyMMdd` must equal the request
    date in UTC.
  - **Mandatory signed headers:** require `host`; require `x-amz-date` when an
    `X-Amz-Date` header is present.
  - **Declared-but-missing headers:** every header named in `SignedHeaders` must
    actually be on the request.
  - A missing/unparseable date now rejects instead of throwing on `.get`.

  Signature math for valid requests is unchanged; the rejection reason returned to
  clients is the existing generic `Request signature could not be validated` (no
  leak of which check failed).

  ### Testing
  `Aws4AuthenticatorSpec` now builds genuinely valid signatures via the
  authenticator's own primitives and asserts:
  - a fresh, fully-signed request is **accepted**;
  - an otherwise-valid signature is **rejected** when stale, when the scope date is
    wrong, when `host`/`x-amz-date` are unsigned, when a declared header is absent,
    and when the date is missing.

  All 18 tests in `Aws4AuthenticatorSpec` + `VinylDNSAuthenticatorSpec` pass.

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/route/Aws4Authenticator.scala
+++ b/modules/api/src/main/scala/vinyldns/api/route/Aws4Authenticator.scala
@@ -22,7 +22,7 @@ import javax.crypto.spec.SecretKeySpec
 import javax.crypto.{Mac, SecretKey}
 
 import akka.http.scaladsl.model.HttpRequest
-import java.time.{Instant, ZoneId, ZonedDateTime}
+import java.time.{Duration, Instant, ZoneId, ZonedDateTime}
 import java.time.format.DateTimeFormatter
 import org.slf4j.LoggerFactory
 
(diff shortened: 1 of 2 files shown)


JoshSEdwards (approved):
LGTM

PR #1527, opened 2026-07-22
Batch recordset_data writes during zone sync

nspadaccino:
## Summary

  Speeds up the recordset-cache write during zone sync by batching it.
  `MySqlRecordSetCacheRepository.save` previously inserted `recordset_data`
  one row at a time, each in its own transaction, so `rewriteBatchedStatements`
  never kicked in. On a 300k-record zone this phase (`zone.sync.saveRecordSetDatas`)
  was ~267s of a ~448s sync — the single largest cost.

  `save` now collects all deletes and inserts and runs them as grouped batch
  statements (`batch`/`batchByName`, groups of 1000) within one transaction,
  the same pattern `MySqlRecordSetRepository` already uses. Expected to bring
  the cache write down toward the record-set repo's ballpark (tens of seconds).

  ## Changes

  - **Batch the cache write.** Single-row, per-transaction inserts → grouped
    `batch`/`batchByName` inside one `withinTx`.
  - **Correctness fix.** Inserts/updates previously executed eagerly as side
    effects *outside* the caller's transaction. They now run inside the
    returned `IO`, so cache writes are part of the sync transaction.
    (`ZoneSyncHandler` already sequences the save, so production is correct;
    the integration tests were relying on the old side-effect behavior and
    are updated to sequence the IOs.)
  - **Null `ip` binding.** Bind `ip` as an `Option` instead of `orNull`; a raw
    null is silently dropped by `batchByName`, which would have prevented all
    non-IP record types (CNAME/NS/TXT/SOA/MX/…) from being written.

  ## Testing

  - Full mysql integration suite: 278/278; mysql unit: 24/24; api
    `ZoneSyncHandlerSpec`: 18/18.
  - New integration tests: >1000-row batch boundary, null-`ip` record types,
    and a mixed create/update/delete changeset asserting exact row counts
    (proves updates remain delete-then-insert with no dropped/duplicate rows).

  ## Notes

  - No schema or config changes.
  - Complements #1515 (which chunks the sync write transaction but does not
    change this per-row insert path); best landed on top of it so the cache
    write participates in #1515's batch transaction.

### DIFF
--- a/modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlRecordSetCacheRepositoryIntegrationSpec.scala
+++ b/modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlRecordSetCacheRepositoryIntegrationSpec.scala
@@ -71,9 +71,7 @@ class MySqlRecordSetCacheRepositoryIntegrationSpec
     val pendingChanges = generateInserts(zone, count, word)
     val bigPendingChangeSet = ChangeSet(pendingChanges)
     executeWithinTransaction { db: DB =>
-      recordSetCacheRepo.save(db, bigPendingChangeSet)
-      recordSetRepo.apply(db, bigPendingChangeSet)
-
+      recordSetCacheRepo.save(db, bigPendingChangeSet).flatMap(_ => recordSetRepo.apply(db, bigPendingChangeSet))
     }.unsafeRunSync()
     pendingChanges
   }
@@ -81,9 +79,7 @@ class MySqlRecordSetCacheRepositoryIntegrationSpec
   def insert(changes: List[RecordSetChange]): Unit = {
     val bigPendingChangeSet = ChangeSet(changes)
     executeWithinTransaction { db: DB =>
-      recordSetCacheRepo.save(db, bigPendingChangeSet)
-      recordSetRepo.apply(db, bigPendingChangeSet)
-
+      recordSetCacheRepo.save(db, bigPendingChangeSet).flatMap(_ => recordSetRepo.apply(db, bigPendingChangeSet))
     }.unsafeRunSync()
     ()
   }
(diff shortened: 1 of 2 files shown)


modules/mysql/src/main/scala/vinyldns/mysql/repository/MySqlRecordSetCacheRepository.scala
```diff
@@ -64,68 +64,57 @@ class MySqlRecordSetCacheRepository
...
+    // the existing rows for anything being (re)written so updates are delete-then-insert.
+    val deleteParams: Seq[Seq[Any]] =
+      (completeDeletes.map(_.recordSet.id) ++
+        failedCreates.map(_.recordSet.id) ++
+        completeUpdates.map(_.recordSet.id) ++
+        failedUpdatesOrDeletes.flatMap(_.updates.map(_.id))).map(id => Seq[Any](id))
+
+    // Rows to insert: creates/pending and updated records use the new record set; failed
+    // update/delete reversions restore the previous record set.
+    val insertParams: Seq[Seq[(Symbol, Any)]] =
```
JoshSEdwards, line 92: This is definitely an improvement, but currently expands the entire changeset into insertParams before entering the IO. For the 300k-record workload this change targets, `.grouped(1000)` limits JDBC statement size but as written it still keeps all of the expanded parameter rows in memory.

Could be worthwhile to build an iterator of rows and consume its 1,000-row groups inside IO/withinTx so each group is prepared only when it's ready to be written, instead of materializing the expanded params for the entire changeset up front.

Conceptually, the current flow is:
```
Create all 300k parameter rows
        ↓
Return IO
        ↓
Begin transaction
        ↓
Send rows 1–1000
Send rows 1001–2000
...
```

Whereas a genuinely bounded flow could look like:
```
Return IO
        ↓
Begin transaction
        ↓
Create rows 1–1000 → send them
Create rows 1001–2000 → send them
...
```

modules/mysql/src/it/scala/vinyldns/mysql/repository/MySqlRecordSetCacheRepositoryIntegrationSpec.scala
```diff
@@ -260,6 +280,49 @@ class MySqlRecordSetCacheRepositoryIntegrationSpec
       recordSetCacheRepo.getRecordSetData(inserts(1).recordSet.id).unsafeRunSync().map(_.name) shouldBe
         Some(inserts(1).recordSet.name)
     }
+    "apply a mixed create/update/delete changeset atomically with correct row counts" in {
```
JoshSEdwards, line 283: This verifies that a successful mixed changeset produces the correct final rows, but it doesn’t exercise rollback, so “atomically” isn’t fully established here. I think we should keep this test for the happy path and add a separate rollback regression test.

The new test should seed the cache, call `recordSetCacheRepo.save(db, cs)`, then return `IO.raiseError` from the same `executeWithinTransaction`. Afterward, verify the deleted rows still exist with their original values, updated rows retain their original names, newly created rows are absent, and the original row count is unchanged. That would specifically catch cache writes escaping the caller’s transaction.

JoshSEdwards (changes requested):
Overall, looks pretty solid. Just added a few comments with suggestions to make the flow even more bounded than it currently is and to add a new regression test to prove atomicity in the new sequence.

PR #1529, opened 2026-07-24
Fix "Group with ID null was not found" popups on shared zones

nspadaccino:
VDNS-368

 ## Problem

  When viewing certain shared zones, the portal fires a cascade of error popups:

  > HTTP 404 (Not Found): Group with ID null was not found

  One popup per affected record. In the reported zone, 9 CNAME records were
  affected — all with valid `ownerGroupId` values but
  `recordSetGroupChange.requestedOwnerGroupId` persisted as the literal string
  `"null"` inside the protobuf blob.

  ## Root cause

  Two parts:

  1. **Weak portal guards.** The ownership-status code path only guarded with
     truthiness / `!= undefined`, which does not catch the string `"null"` (it is
     truthy and `!= undefined`). On zone load, for each owned record Angular built
     `GET /api/groups/null/members`; the API echoed the segment back as
     `Group with ID null was not found` and it rendered as a popup. A sibling
     `getGroup` helper was already hardened for exactly this case, but the guard
     was never applied to the ownership-status path.
  2. **Bad data on the write path.** Affected records had `requestedOwnerGroupId`
     stored as the string `"null"` rather than an absent/`None` value.

  ## Changes

  ### Portal — stops the popups
  `modules/portal/public/lib/controllers/controller.records.js`
  - Added `isValidGroupId(groupId)` helper that treats `undefined`, `null`, and
    the string `"null"` as absent.
  - Applied it to all three ownership-status call sites:
    `isNotRequestedGroupMemberResponse`, `recordSetGroupOwnershipStatus` (early
    return before the member lookup), and the `updateRecordDisplay` per-record loop.
  - Refactored `getGroup` to use the shared helper.

  ### API — prevents recurrence
  `modules/api/src/main/scala/vinyldns/api/route/DnsJsonProtocol.scala`
  - Sanitized `requestedOwnerGroupId` at the parse boundary in
    `ownershipTransferSerializer.fromJson`, so blank and `"null"` values map to
    `None` before they can be written to the protobuf blob — regardless of which
    client sends them.

  ### Tests
  - Portal (`controller.records.spec.js`): no member lookup for a `"null"` group id;
    a `"null"` requested owner group is skipped while a valid owner group is still
    checked; valid ids still fetch.
  - API (`VinylDNSJsonProtocolSpec.scala`): `"null"` and blank `requestedOwnerGroupId`
    both parse to `None`.

  ## Acceptance criteria
  - [x] Viewing a shared zone with affected records produces no "Group with ID null" popups.
  - [x] Newly created/updated records never persist `requestedOwnerGroupId` as the string `"null"`.
  - [x] The get-group-members call is not issued when the group id is absent, `null`, or `"null"`.

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/route/DnsJsonProtocol.scala
+++ b/modules/api/src/main/scala/vinyldns/api/route/DnsJsonProtocol.scala
@@ -267,10 +267,15 @@ trait DnsJsonProtocol extends JsonValidation {
 
 
   case object ownershipTransferSerializer extends ValidationSerializer[OwnershipTransfer] {
+    // Treat blank and the literal string "null" as an absent group id so bad
+    // client payloads never persist requestedOwnerGroupId as the string "null".
+    private def sanitizeRequestedOwnerGroupId(id: Option[String]): Option[String] =
+      id.map(_.trim).filter(v => v.nonEmpty && !v.equalsIgnoreCase("null"))
+
     override def fromJson(js: JValue): ValidatedNel[String, OwnershipTransfer] =
       (
         (js \ "ownershipTransferStatus").required[OwnershipTransferStatus]("Missing ownershipTransfer.ownershipTransferStatus"),
-        (js \ "requestedOwnerGroupId").optional[String],
+        (js \ "requestedOwnerGroupId").optional[String].map(sanitizeRequestedOwnerGroupId),
         ).mapN(OwnershipTransfer.apply)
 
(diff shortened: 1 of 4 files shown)


JoshSEdwards (approved):
LGTM. Basically the perfect PR -- structurally sound, tightly scoped, and the PR description explains everything very clearly. Love it!

arpit4ever (approved):
LGTM

PR #1553, opened 2026-09-07
Fix Multiple deleteRecordSet Batch Failure

### LINKED ISSUE #1542: Batch Change Failure When Multiple DeleteRecordSet Entries Target the Same RecordSet With Invalid Record Data
A batch DNS change request fails when it contains multiple DeleteRecordSet operations for the same DNS record and one or more delete entries reference record data that does not exist in the target RecordSet.

Although initially observed with IPv6 PTR records, the issue appears to affect all record types that use the common DeleteRecordSet processing path.

Impact

When a batch contains multiple delete operations against the same RecordSet, VinylDNS combines them into a single RecordSet change. If one of the requested deletions references non-existent record data, the combined update can result in an empty record set (records=List()), causing the DNS backend to reject the update.

Example Failure

Batch Request

Record 1:
Type: PTR
Name: FD69:27cc:fe92:2000::25
Record Data: vinyldns.io.

Record 2:
Type: PTR
Name: FD69:27cc:fe92:2000::25
Record Data: test.vinyldns.io. (does not exist)

Result
Batch change status: Failed

Sokitha-21:
Fixes #1542.
JIRA: VDNS-442

**Root Cause:**
When multiple `DeleteRecordSet` operations target the same RecordSet, VinylDNS combines them into a single RecordSet change. If one delete request contains record data that does not exist the combined change can result in an empty RecordSet (`records=List()`), which is rejected by the DNS backend.

**Fix**
Updated the `DeleteRecordSet` handling to intersect the requested record data with the existing records before generating the RecordSet change. This prevents non-existent record data from producing an invalid empty RecordSet update.
Non-existent PTR deletion requests are now treated as no-ops (record does not exist) instead of being included in the shared RecordSet change.

### DIFF
--- a/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
+++ b/modules/api/src/main/scala/vinyldns/api/domain/batch/BatchChangeValidations.scala
@@ -357,7 +357,11 @@ class BatchChangeValidations(
     }
 
     val addInBatch = groupedChanges.getProposedAdds(change.recordKey)
-    val isSameRecordUpdateInBatch = recordData.nonEmpty && addInBatch.contains(RecordData.fromString(recordData, change.inputChange.typ).get)
+    val requestedDeletes = groupedChanges.getRequestedDeleteRecordData(change.recordKey)
+    val isSameRecordUpdateInBatch =
+      recordData.nonEmpty &&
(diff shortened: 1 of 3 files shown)


Sokitha-21:
image
Attached image for reference.

PR #1557, opened 2026-09-10
Fix: Create service log files before API functional/Integration tests startup

### LINKED ISSUE #1556: API Functional/Integration Tests Fail Because LocalStack Log File Is Missing
VDNS-461

The API functional/Integration tests fail during service startup when the test bootstrap attempts to tail the LocalStack log file:

{{tail: cannot open '/localstack/localstack.log' for reading: No such file or directory
tail: no files remaining
make: *** [Makefile:40: run] Error 1}}

The bootstrap starts tail before LocalStack has created /localstack/localstack.log. This creates a first-run race condition where tail fails because the log file does not yet exist, causing the tests to terminate with exit code 1.

Sokitha-21:
VDNS-461

Fixes #1556.

**Problem**
API functional/integration tests were failing during service startup because the bootstrap script tails the LocalStack log file before LocalStack creates it:

tail: cannot open '/localstack/localstack.log' for reading: No such file or directory

This caused the bootstrap to exit and the tests to fail.

**Root Cause**

The LocalStack and VinylDNS log files were created by their respective services during startup. However the test bootstrap starts tail before the log producers have created the files resulting in a first-run race condition.

**Fix**

Created the required log directory and log files during the Docker image build.
This ensures the log files exist before the bootstrap starts tailing them, avoiding the missing-file race condition and allowing the API functional/integration tests to proceed successfully.

### DIFF
--- a/test/api/integration/Dockerfile
+++ b/test/api/integration/Dockerfile
@@ -31,4 +31,7 @@ COPY quickstart/bind9/etc/named.conf.* /etc/bind/
 COPY quickstart/bind9/zones/ /var/bind/
 RUN named-checkconf
 
+# Create log file before startup to avoid tail race.
+RUN touch /opt/vinyldns/vinyldns.log
+
 ENV RUN_SERVICES="all"
--- a/test/api/functional/Dockerfile
+++ b/test/api/functional/Dockerfile
@@ -24,6 +24,9 @@ COPY quickstart/bind9/etc/named.conf.* /etc/bind/
 COPY quickstart/bind9/zones/ /var/bind/
 RUN named-checkconf
 
+# Create log files before startup to avoid tail race.
+RUN mkdir -p /localstack && touch /localstack/localstack.log /opt/vinyldns/vinyldns.log
+
 # Copy over the functional tests
 COPY  modules/api/src/test/functional /functional_test
 


test/api/integration/Dockerfile
```diff
@@ -31,4 +31,7 @@ COPY quickstart/bind9/etc/named.conf.* /etc/bind/
 COPY quickstart/bind9/zones/ /var/bind/
 RUN named-checkconf
 
+# Create log file before startup to avoid tail race.
+RUN touch /opt/vinyldns/vinyldns.log
```
nspadaccino, line 35: This is missing the LocalStack log. Integration runs `/initialize.sh all (RUN_SERVICES=all)`, which starts LocalStack via `start_localstack.sh`, which is the same script that backgrounds `localstack … &> /localstack/localstack.log` and then immediately `tail -fs` it. 

I think we want this line to match the functional test file:
`RUN mkdir -p /localstack && touch /localstack/localstack.log /opt/vinyldns/vinyldns.log`

arpit4ever (approved):
LGTM

test/api/integration/Dockerfile
arpit4ever: Fixed on [6e6a22f](https://github.com/vinyldns/vinyldns/pull/1557/commits/6e6a22f562dc610f53562eac80357427297f5eb7)

JoshSEdwards (approved):
LGTM

PR #1558, opened 2026-09-10
Don't lock disabled service accounts during Graph API user sync

nspadaccino:
VDNS-463

  ## Problem

  The Graph API user sync task locks any active user whose account is
  `accountEnabled=false` in Microsoft Entra. Service accounts are commonly
  disabled for interactive sign-in while still in active use (they
  authenticate non-interactively), so the task can flag live service
  accounts to be locked.

  ## Change

  Add an **optional** setting:

  user-sync.graph-api.service-account-employee-type = "S"

  When set, disabled accounts whose Graph `employeeType` equals the value are
  treated as service accounts and are **not** locked. `employeeType` is added
  to the Graph `$select` only when the setting is present.

  When **unset** (default), behavior and the Graph query are unchanged, so
  existing deployments are unaffected.

  ## Why `employeeType` rather than username patterns

  The `employeeType` attribute distinguishes service accounts from departed
  employees reliably, whereas username conventions do not. Some service
  accounts follow no naming convention, and some non-service accounts share a
  service-account-style prefix. Keying off the attribute avoids both
  false negatives and false positives that a name-based heuristic would hit.

  ## Behavior

  Only disabled accounts whose `employeeType` matches the configured marker
  are skipped. Accounts absent from the directory, and disabled accounts with
  a non-matching or missing `employeeType`, are still locked as before.

  ## Testing

  Added unit tests covering:
  - matching `employeeType` marker → not locked
  - non-matching / missing `employeeType` → locked
  - account absent from directory → locked regardless of the setting
  - no marker configured → prior behavior (any disabled account locked)

  ## Config docs

  `config-portal.md` documents the new optional setting.

### DIFF
--- a/modules/docs/src/main/mdoc/operator/config-portal.md
+++ b/modules/docs/src/main/mdoc/operator/config-portal.md
@@ -215,6 +215,12 @@ user-sync {
     # The Graph API user attribute to match against VinylDNS usernames
     # Default: "onPremisesSamAccountName"
     username-attribute = "onPremisesSamAccountName"
+
+    # Optional. When set, disabled accounts whose "employeeType" equals this value
+    # are treated as service accounts and are NOT locked, since service accounts are
+    # commonly disabled for interactive sign-in while still in active use. Leave unset
+    # to lock every disabled account (default behavior).
+    # service-account-employee-type = "S"
   }
 }
 ```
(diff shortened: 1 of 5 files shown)


JoshSEdwards (approved):
LGTM

arpit4ever (approved):
LGTM

